Issue #9, opened 2018-11-18
[Code health] Refactor remaining logic out of Fragments into ViewModels


gino-m:
This will facilitate testing and improve consistency across UI implementations.

gino-m:
This has been mostly done iiuc, so unassigning for now. I'll keep this issue open to track cleanup of the last holdouts.

gino-m:
Rather than keeping this hanging open, @scolsen, @shobhitagarwal1612, @dturner let's file specific issues as we find them.

PR #38, opened 2019-02-22
Move progress bar logic out of the RecordDetailsFragment.

scolsen:
This changes moves the progress bar visibility logic out of the RecordDetailsFragment and into its associated viewModel. See issue #9 

Since this is my first go at this, there may be some flubs. I followed the example of 8f4719ef. Please leave comments and push back as needed!

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
+++ b/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
@@ -121,13 +121,13 @@ public class DataRepository {
                     .orElse(Single.error(new DocumentNotFoundException())));
   }
 
-  public Observable<Resource<Record>> getRecordDetails(
+  public Flowable<Resource<Record>> getRecordDetails(
       String projectId, String featureId, String recordId) {
     return getFeature(projectId, featureId)
         .flatMap(feature -> remoteDataService.loadRecordDetails(feature, recordId))
         .map(Resource::loaded)
         .onErrorReturn(Resource::error)
-        .toObservable();
+        .toFlowable();
   }
 
   public Single<Resource<Record>> getRecordSnapshot(
--- a/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
@@ -52,9 +52,6 @@ public class RecordDetailsFragment extends AbstractFragment {
   @BindView(R.id.record_details_toolbar)
   TwoLineToolbar toolbar;
 
-  @BindView(R.id.form_name)
-  TextView formNameView;
-
   @BindView(R.id.record_details_layout)
   LinearLayout recordDetailsLayout;
 
@@ -63,13 +60,21 @@ public class RecordDetailsFragment extends AbstractFragment {
   @Override
   public void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
+    RecordDetailsFragmentArgs args = getRecordDetailFragmentArgs();
     viewModel = getViewModel(RecordDetailsViewModel.class);
+    viewModel.getRecordDetailsRequest(args);
+    //TODO: Move toolbar setting logic into the ViewModel once we have
+    // determined the fate of the toolbar.
+    viewModel.toolbarTitle.observe(this, this::setToolbarTitle);
+    viewModel.toolbarSubtitle.observe(this, this::setToolbarSubtitle);
+    viewModel.records.observe(this, this::onUpdate);
   }
 
   @Override
   public View onCreateView(
       LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
-    RecordDetailsFragBinding binding = 
+    super.onCreateView(inflater, container, savedInstanceState);
+    RecordDetailsFragBinding binding =
       RecordDetailsFragBinding.inflate(inflater, container, false);
     binding.setViewModel(viewModel);
     binding.setLifecycleOwner(this);
@@ -91,21 +96,23 @@ public class RecordDetailsFragment extends AbstractFragment {
   public void onActivityCreated(@Nullable Bundle savedInstanceState) {
     super.onActivityCreated(savedInstanceState);
     setHasOptionsMenu(true);
-    viewModel.getRecord().observe(this, this::onUpdate);
   }
 
   @Override
   public void onStart() {
     super.onStart();
-    RecordDetailsFragmentArgs args = getRecordDetailFragmentArgs();
-    viewModel.loadRecordDetails(args.getProjectId(), args.getFeatureId(), args.getRecordId());
+  }
+
+  private void setToolbarTitle(String title) {
+    toolbar.setTitle(title);
+  }
+
+  private void setToolbarSubtitle(String subtitle) {
+    toolbar.setSubtitle(subtitle);
   }
 
   private void onUpdate(Resource<Record> record) {
     switch (record.operationState().get()) {
-      case LOADING:
-        showProgressBar();
-        break;
       case LOADED:
         record.ifPresent(this::showRecord);
         break;
@@ -118,18 +125,7 @@ public class RecordDetailsFragment extends AbstractFragment {
     }
   }
 
-  private void showProgressBar() {
-    toolbar.setTitle("");
-    toolbar.setSubtitle("");
-    formNameView.setText("");
-    recordDetailsLayout.setVisibility(View.GONE);
-  }
-
-  // TODO: Move into separate ViewHolder class.
   private void showRecord(Record record) {
-    toolbar.setTitle(record.getFeature().getTitle());
-    toolbar.setSubtitle(record.getFeature().getSubtitle());
-    formNameView.setText(record.getForm().getTitle());
     recordDetailsLayout.removeAllViews();
     for (Form.Element element : record.getForm().getElements()) {
       switch (element.getType()) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
@@ -27,10 +27,10 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.LinearLayout;
-import android.widget.ProgressBar;
 import android.widget.TextView;
 import butterknife.BindView;
 import butterknife.ButterKnife;
+import com.google.android.gnd.databinding.RecordDetailsFragBinding;
 import com.google.android.gnd.MainActivity;
 import com.google.android.gnd.R;
 import com.google.android.gnd.inject.ActivityScoped;
@@ -55,9 +55,6 @@ public class RecordDetailsFragment extends AbstractFragment {
   @BindView(R.id.form_name)
   TextView formNameView;
 
-  @BindView(R.id.record_details_progress_bar)
-  ProgressBar progressBar;
-
   @BindView(R.id.record_details_layout)
   LinearLayout recordDetailsLayout;
 
@@ -72,7 +69,10 @@ public class RecordDetailsFragment extends AbstractFragment {
   @Override
   public View onCreateView(
       LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
-    return inflater.inflate(R.layout.record_details_frag, container, false);
+    RecordDetailsFragBinding binding = 
+      RecordDetailsFragBinding.inflate(inflater, container, false);
+    binding.setViewModel(viewModel);
+    return binding.getRoot();
   }
 
   @Override
@@ -122,12 +122,10 @@ public class RecordDetailsFragment extends AbstractFragment {
     toolbar.setSubtitle("");
     formNameView.setText("");
     recordDetailsLayout.setVisibility(View.GONE);
-    progressBar.setVisibility(View.VISIBLE);
   }
 
   // TODO: Move into separate ViewHolder class.
   private void showRecord(Record record) {
-    progressBar.setVisibility(View.GONE);
     toolbar.setTitle(record.getFeature().getTitle());
     toolbar.setSubtitle(record.getFeature().getSubtitle());
     formNameView.setText(record.getForm().getTitle());
--- a/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
@@ -62,12 +62,12 @@ public class RecordDetailsFragment extends AbstractFragment {
     super.onCreate(savedInstanceState);
     RecordDetailsFragmentArgs args = getRecordDetailFragmentArgs();
     viewModel = getViewModel(RecordDetailsViewModel.class);
-    viewModel.loadRecordDetails(args);
     // TODO: Move toolbar setting logic into the ViewModel once we have
     // determined the fate of the toolbar.
     viewModel.toolbarTitle.observe(this, this::setToolbarTitle);
     viewModel.toolbarSubtitle.observe(this, this::setToolbarSubtitle);
     viewModel.records.observe(this, this::onUpdate);
+    viewModel.loadRecordDetails(args);
   }
 
   @Override
@@ -103,11 +103,15 @@ public class RecordDetailsFragment extends AbstractFragment {
   }
 
   private void setToolbarTitle(String title) {
-    toolbar.setTitle(title);
+    if (toolbar != null) {
+      toolbar.setTitle(title);
+    }
   }
 
   private void setToolbarSubtitle(String subtitle) {
-    toolbar.setSubtitle(subtitle);
+    if (toolbar != null) {
+      toolbar.setSubtitle(subtitle);
+    }
   }
 
   private void onUpdate(Resource<Record> record) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
@@ -72,6 +72,7 @@ public class RecordDetailsFragment extends AbstractFragment {
     RecordDetailsFragBinding binding = 
       RecordDetailsFragBinding.inflate(inflater, container, false);
     binding.setViewModel(viewModel);
+    binding.setLifecycleOwner(this);
     return binding.getRoot();
   }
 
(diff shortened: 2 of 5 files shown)


gnd/src/main/res/layout/record_details_frag.xml
```diff
@@ -14,79 +14,88 @@
...
-
-    </ScrollView>
-
-  </RelativeLayout>
-
-</android.support.design.widget.CoordinatorLayout>
\ No newline at end of file
+        android:layout_gravity="center"
+        android:padding="48dp"
+          android:visibility="@{viewModel.progressBarVisibility}"/>
```
gino-m: Extra indent?

gnd/src/main/res/layout/record_details_frag.xml
```diff
@@ -14,79 +14,88 @@
...
+            android:textAllCaps="true"
+            android:textColor="@color/colorForeground"
+            android:textSize="12sp"
+            />
+  
+        </LinearLayout>
+  
+      </android.support.v7.widget.CardView>
+  
+      <ProgressBar
```
gino-m: Does it make more sense for this to live at the top level of the layout? Does it still work there?

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -43,6 +47,21 @@ public void loadRecordDetails(String projectId, String featureId, String recordI
     disposeOnClear(
         dataRepository
             .getRecordDetails(projectId, featureId, recordId)
-            .subscribe(record::setValue));
+            .subscribe(this::onRecordUpdate));
```
gino-m: Rather than subscribing, while we're at it can we transform this into a LiveData<Integer> instead?

I think you just need to set lifecycle of the binding in the view with:
binding.setLifecycleOwner(this);

And in the constructor you would do something like:

```java
progressBarVisibility = 
    dataRepository
            .getRecordDetails(projectId, featureId, recordId)
            .map(this::toProgressBarVisibility)
            .toFlowable(BackpressureStrategy.LATEST));
```

Then the view would fully manage the subscription based on its own lifecycle!

gino-m: Oh, and you'll also need to handle errors since LiveData doesn't:
```java
progressBarVisibility = 
    dataRepository
            .getRecordDetails(projectId, featureId, recordId)
            .map(this::toProgressBarVisibility)
            .onErrorReturn(err -> Visibility.HIDDEN)
            .toFlowable(BackpressureStrategy.LATEST));
```
In the future we'll also want to report these errors, so perhaps a TODO would be in order.

gnd/src/main/res/layout/record_details_frag.xml
scolsen: fixed in e678555f2

gnd/src/main/res/layout/record_details_frag.xml
scolsen: Think that should be fine!

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
scolsen: We'll see how this goes. Gave it a first pass: e678555f2 It also required me to use safeUnbox() in the layout to unbox the data--not sure whether or not that's something we should be doing. lmk your thoughts!

scolsen:
Alrighty, updated the implementation to use Livedata+databinding+flowables. @gino-m

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
   @Inject
   RecordDetailsViewModel(DataRepository dataRepository) {
     this.dataRepository = dataRepository;
-    this.record = new MutableLiveData<>();
+
+    this.recordDetailsRequest = BehaviorSubject.create();
+
+    this.recordStream =
+        recordDetailsRequest
+          .switchMap(a ->
```
gino-m: Nit: s/a/args/

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+
 import javax.inject.Inject;
 
 public class RecordDetailsViewModel extends AbstractViewModel {
 
+  private static final String TAG = RecordDetailsViewModel.class.getSimpleName();
+
   private final DataRepository dataRepository;
-  private final MutableLiveData<Resource<Record>> record;
+  private final Subject<RecordDetailsFragmentArgs> recordDetailsRequest;
```
gino-m: Ah, I found it:
s/Subject/Processor/

Processors are already Flowable!

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+  public final LiveData<String> toolbarTitle;
+  public final LiveData<String> toolbarSubtitle;
+  public final LiveData<String> formNameView;
 
   @Inject
   RecordDetailsViewModel(DataRepository dataRepository) {
     this.dataRepository = dataRepository;
-    this.record = new MutableLiveData<>();
+
+    this.recordDetailsRequest = BehaviorSubject.create();
```
gino-m: s/BehaviorSubject/BehaviorProcessor/

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+
+    this.recordStream =
+        recordDetailsRequest
+          .switchMap(a ->
+              this.dataRepository.getRecordDetails(a.getProjectId(), a.getFeatureId(), a.getRecordId()))
+          .toFlowable(BackpressureStrategy.LATEST);
```
gino-m: After switching to Processor you can remove this line.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+
+    //TODO: Refactor to expose the fetched record directly.
```
gino-m: Nit: Space after "//"

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+    );
+
+    this.formNameView = LiveDataReactiveStreams.fromPublisher(
+        recordStream.map(this::toFormNameView)
+        .onErrorReturn(x -> "")
+    );
+
+    recordDetailsRequest.subscribe(args -> {
+      Log.d(TAG, "" + args);
+    });
```
gino-m: Leftover from debugging?

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+  }
+
+  public void getRecordDetailsRequest(RecordDetailsFragmentArgs args) {
```
gino-m: "loadRecordDetails" (see comments in Fragment)

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+    this.recordDetailsRequest.onNext(args);
+  }
+
+  private Integer toProgressBarVisibility(Resource<Record> record) {
```
gino-m: I guess we need to return Integer instead of int because autoboxing doesn't work on lambdas (i.e. recordStream.map(this::toProgressBarVisibility)?

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -17,32 +17,124 @@
...
+  private String toFormNameView(Resource<Record> record) {
+    switch (record.operationState().get()) {
+      case LOADED:
+        return
+          record.data()
+            .map(r -> r.getForm().getTitle())
+            .orElse("");
+      default:
+        return "";
+    }
```
gino-m: Resource has a method data() which returns Optional.empty() if no data is loaded, allowing you to simplify the above "to" methods. For example, for toFormNameView:
`record.data().map(Record::getForm).map(Record::getTitle).orElse("");`

toProgressBarVisibility could be simplified as:
`record.data().isPresent() ? View.VISIBLE : View.GONE`

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -66,13 +60,25 @@
   @Override
   public void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
+    RecordDetailsFragmentArgs args = getRecordDetailFragmentArgs();
     viewModel = getViewModel(RecordDetailsViewModel.class);
+    viewModel.getRecordDetailsRequest(args);
```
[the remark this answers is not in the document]
scolsen: Totally. loadRecordDetails sounds straightforward and accurate to me! I'll make the change.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -66,13 +60,25 @@
...
+    //TODO: Move toolbar setting logic into the ViewModel once we have
```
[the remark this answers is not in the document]
scolsen: done!

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -66,13 +60,25 @@
...
+    // determined the fate of the toolbar.
+    viewModel.toolbarTitle.observe(this, this::setToolbarTitle);
```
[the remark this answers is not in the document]
scolsen: Thanks for pointing this out! Yeah those methods were pretty extraneous huh

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -66,13 +60,25 @@
...
+    viewModel.toolbarSubtitle.observe(this, this::setToolbarSubtitle);
```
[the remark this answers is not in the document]
scolsen: 👍

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -90,21 +96,23 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
-    viewModel.loadRecordDetails(args.getProjectId(), args.getFeatureId(), args.getRecordId());
+  }
+
+  private void setToolbarTitle(String title) {
+    toolbar.setTitle(title);
+  }
+
+  private void setToolbarSubtitle(String subtitle) {
+    toolbar.setSubtitle(subtitle);
   }
```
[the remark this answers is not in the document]
scolsen: got it!

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
scolsen: Sure thing, on this subject, is there a preferred guideline for naming arguments? I usually default to longer, descriptive names, *except* in lambdas--are we just following the google style guide? Maybe we should note this in CONTRIBUTING.md?

@gino-m

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
scolsen: done

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
scolsen: renamed appropriately.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
scolsen: Sweet! Simplification complete :)

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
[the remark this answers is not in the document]
scolsen: @gino-m 

So it looks like we may need these after all. I guess the toolbar isn't available at create time? I get a null object reference when trying to call toolbar::setX. Unfortunately, I believe we're supposed to observe in onCreate to avoid redundant calls because of the way the lifecycle works. https://developer.android.com/topic/libraries/architecture/livedata#observe_livedata_objects. 

For now I've just added back the wrapper methods.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
```diff
@@ -52,89 +57,47 @@
...
+        LiveDataReactiveStreams.fromPublisher(
+            recordProcessor.map(this::toFormNameView).onErrorReturn(x -> ""));
   }
 
-  public void getRecordDetailsRequest(RecordDetailsFragmentArgs args) {
-    this.recordDetailsRequest.onNext(args);
+  public void loadRecordDetails(RecordDetailsFragmentArgs args) {
+    this.dataRepository
+        .getRecordDetails(args.getProjectId(), args.getFeatureId(), args.getRecordId())
+        .subscribe(record -> this.recordProcessor.onNext(record));
```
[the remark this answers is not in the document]
gino-m: Any time we subscribe we also need to keep track of the subscription and dispose of it at the proper point. Doing this would actually put us back where we started, with subscriptions being created each time loadRecordDetails() is called. You probably get an Android Studio warning here that we're not doing anything with the return value of subscribe().

Why did you change this back to subscribe() from the switchMapped solution?

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
```diff
@@ -66,13 +60,24 @@
   @Override
   public void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
+    RecordDetailsFragmentArgs args = getRecordDetailFragmentArgs();
     viewModel = getViewModel(RecordDetailsViewModel.class);
+    viewModel.loadRecordDetails(args);
```
gino-m: This should be the last thing we do here to avoid race condition w/observe. Technically if the LiveData are backed by BehaviorSubjects it won't matter, but we don't know that's what's used here, so better to be safe.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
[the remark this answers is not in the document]
gino-m: I see. In that case we'll also need to check toolbar != null in each method in case load operation finished before onCreateView has completed.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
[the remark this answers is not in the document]
scolsen: Ahh, I see, duh. I got rid of it just because it allowed me to drop the `RecordDetailsRequest` and `RecordStream` properties, which I felt simplified the class since they're really just temporary conversions/type negotiations. Obviously not worthwhile if it makes the solution incorrect 😵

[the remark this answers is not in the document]
gino-m: Yeah, the reason we added them was to not need to manage Disposable subscriptions here.

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsFragment.java
scolsen: got it!

gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java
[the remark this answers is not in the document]
scolsen: Sounds good, I've switched back to the old, new approach.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -121,13 +121,13 @@ private void onRemoteFeatureVectorChange(DatastoreEvent<Feature> event) {
                     .orElse(Single.error(new DocumentNotFoundException())));
   }
 
-  public Observable<Resource<Record>> getRecordDetails(
+  public Flowable<Resource<Record>> getRecordDetails(
```
scolsen: As of now, the record details viewmodel is the only place this method is called, so we may as well just have it return a Flowable.

gino-m: I think this makes sense semantically; iiuc Flowables are for cases where control the rate of emissions (e.g., database queries), whereas Observables are for cases where we don't (e.g., button clicks). We should eventually try to be consistent with what we use when, but I'm not yet sure of the right thing to do in the DataRepository. For now this seems ok!

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -121,13 +121,13 @@ private void onRemoteFeatureVectorChange(DatastoreEvent<Feature> event) {
...
       String projectId, String featureId, String recordId) {
     return getFeature(projectId, featureId)
         .flatMap(feature -> remoteDataService.loadRecordDetails(feature, recordId))
         .map(Resource::loaded)
         .onErrorReturn(Resource::error)
-        .toObservable();
+        .toFlowable();
```
gino-m: Does this compile? I thought BackpressureStrategy was mandatory?

scolsen: @gino-m (should I be @ing you in these comments? or are you already notified?)
It does compile! I thought it was mandatory too, weirdly enough, [Single's toFlowable](http://reactivex.io/RxJava/javadoc/io/reactivex/Single.html#toFlowable) takes no argument. It simply "honors the backpressure of the downstream consumer" which I don't *think* is an issue for us, but I could be wrong.

gino-m: Oooh it's a Single. Doh. Ok LGTM! :)

Issue #19, opened 2018-11-19
[Offline support] Store Feature and Record snapshots and mutations locally


gino-m:
Will need to use Room 2.1.0 to allow Room annotations on AutoValue value objects. This may require updating to Android X:
https://stackoverflow.com/questions/44869545/android-room-persistence-library-entity-with-autovalue

Blocking #15, #16, #17, #18.

gino-m:
Filed https://issuetracker.google.com/issues/129183837 to avoid needing to add create() methods to all @Entity classes, for which we already have AutoValue Builders.

gino-m:
@navinsivakumar @scolsen, for the CREATE, UPDATE, and DELETE queue stored in the local database, do you prefer the -Edit suffix (FeatureEdit, RecordEdit, etc.), -Operation (FeatureOperation, RecordOperation), or something else? Thoughts?
 
_There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors._ ~Leon Bambrick

gino-m:
FYI, committed first working test to my fork:
https://github.com/gino-m/ground-android/commit/24740a2041abc8f239b90c6fcf590e6cd91a46e5

@navinsivakumar, I'll send a PR once db entity classes are in so you can start referencing them in the remote sync impl. Like the *Doc objects I defined for Firebase, these are defined independently of value objects (`.vo` package) to allow us to use the best suited model for each layer and to avoid data layer abstractions leaking into the value objects.

scolsen:
> @navinsivakumar @scolsen, for the CREATE, UPDATE, and DELETE queue stored in the local database, do you prefer the -Edit suffix (FeatureEdit, RecordEdit, etc.), -Operation (FeatureOperation, RecordOperation), or something else? Thoughts?

Edit is nice because it's shorter, but slightly susceptible to misinterpretation since edit ~= modify in the stricter sense of alter the properties of an existing entity (i.e. what "update" means). My natural interpretation of the word 'Edit' doesn't include "create" or "delete", but it's a small adjustment. If we're only talking CRUD operations, I don't think it's that confusing once you encounter its use. I like '-Operation' because it's technically more accurate. Edit really only has the benefit of being shorter, which isn't a big benefit given autocomplete and the other conveniences of IDEs. Here are three other alternatives I can think of:

-Op (as an abbreviation for 'operation' pro: short and sweet, con: it isn't necessarily obvious what this is an abbreviation for.)
[...]

PR #39, opened 2019-03-24
Upgrade to AndroidX

gino-m:
This is required to use the latest Room db needed for local data storage (#19)


gnd/src/main/java/com/google/android/gnd/repository/local/Converters.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.repository.local;
+
+import static java8.util.J8Arrays.stream;
+
+import androidx.room.TypeConverter;
+
+public abstract class Converters {
```
navinsivakumar: If this is meant to be a non-instantiable utility class, I suggest making it final (not abstract) and making the constructor private to prevent instantiation and subclassing.

gnd/src/main/java/com/google/android/gnd/repository/local/EditType.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.google.android.gnd.repository.local;
+
+public enum EditType {
+  UNKNOWN(0),
+  CREATE(1),
+  UPDATE(2),
+  DELETE(3);
+
+  public final int intValue;
```
navinsivakumar: Can this be package-private? That way we enforce using the Converters class to convert (and avoid some of the brittleness risk that comes with public fields).

(I personally would prefer a getter instead of ever exposing the field directly so we have an abstraction that gives flexibility to change how the value is implemented, but those risks would also be lower if this is package-private instead of public.)

gnd/src/main/java/com/google/android/gnd/repository/local/SnapshotType.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.google.android.gnd.repository.local;
+
+/* Types may only be renamed or added, never removed. */
+public enum SnapshotType {
+  UNKNOWN(0),
+  PRIMARY(1),
+  ROLLBACK(2);
+
+  public final int intValue;
```
navinsivakumar: Same about making this package-private.

gnd/build.gradle
```diff
@@ -128,10 +134,10 @@ dependencies {
     implementation "com.jakewharton.rxbinding2:rxbinding-design:$project.rxBindingVersion"
 
     // Android Arch
-    implementation "android.arch.lifecycle:runtime:$project.lifecycleVersion"
-    implementation "android.arch.lifecycle:extensions:$project.lifecycleVersion"
-    implementation "android.arch.lifecycle:reactivestreams:$project.lifecycleVersion"
-    annotationProcessor "android.arch.lifecycle:compiler:$project.lifecycleVersion"
+    implementation "androidx.lifecycle:lifecycle-runtime:2.0.0-beta01"
```
navinsivakumar: How about a variable (like lifecycleVersion) for the androidx.lifecycle version? That will make it easier to keep these consistent.

gino-m:
I didn't realize PRs include all commits in a specific branch, even after the PR is created. I'll resubmit this PR from the appropriate branch, sorry for the confusion. I'll also incorporate your comments into the current WIP implementation. Thanks!

PR #40, opened 2019-03-28
Migrate to AndroidX and update build config

gino-m:
Required to set up local datastore using Room (#19)


gino-m:
Thanks, Navin! 🎉🎊

PR #41, opened 2019-03-29
Implement Feature and Record local db entities and DAOs (#19)

gino-m:
Sharing an early draft for discussion based on reversable edits instead of rollback snapshots. Once we've agreed on the final sync algorithm design I'll update this to reflect our discussions.


gino-m:
Sorry the PR is already getting pretty big. Could you kindly review and pull into `offline-sync` branch?

The main TODO left from this PR is to implement Record values serialization, which I'll implement using JSON for easy debuggability.

Note that this PR only includes DAOs for Feature and Record entities and edits, and not the local cache of Projects. For now the user will need a connection to join the project; the ability to join/leave projects offline shouldn't block testing offline sync for Features and Records though.

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -0,0 +1,72 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.room.ColumnInfo;
+import androidx.room.Ignore;
+import androidx.room.TypeConverter;
+import com.google.android.gnd.vo.Record.Value;
+
+/** Fields describing operation type, old, and new values, common to Feature and Record edits. */
+public class Edit {
```
navinsivakumar: `Edit` should be immutable (and probably implemented using AutoValue, since it is included as a property in other AutoValue classes so we would like to have generated `hashCode`, `equals`, etc. with similar value class semantics).

gnd/src/main/java/com/google/android/gnd/repository/local/FeatureEntity.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.auto.value.AutoValue.CopyAnnotations;
+
+/** Representation of a {@link com.google.android.gnd.vo.Feature} in local db. */
+@AutoValue
+@Entity(
+    tableName = "feature",
+    indices = {@Index("id")})
+public abstract class FeatureEntity {
+  /**
+   * Key used in {@link com.google.android.gnd.repository.local.FeatureEditEntity} when edit
```
navinsivakumar: Where in the `FeatureEditEntity` would this appear?

gnd/src/main/java/com/google/android/gnd/repository/local/IntEnum.java
```diff
@@ -0,0 +1,39 @@
...
+public interface IntEnum {
+  int intValue();
+
+  static <E extends IntEnum> int fromIntEnum(E enumValue, @NonNull E defaultValue) {
+    return enumValue == null ? defaultValue.intValue() : enumValue.intValue() ;
+  }
+
+  @NonNull
+  static <E extends Enum<E> & IntEnum> E toIntEnum(
+      @NonNull Stream<E> values, int intValue, @NonNull E defaultValue) {
```
navinsivakumar: What's the use case for this? In particular, why does it take a `Stream` as an argument instead of, say, `Iterable`? That seems unusual to me (and depending on the use case, there might be a more idiomatic way to express this). Do you have a simple code snippet to show a typical usage pattern?

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -0,0 +1,72 @@
...
+  @NonNull
+  public Type type;
+
+  /** When null the edit refers to the entity itself. */
+  @ColumnInfo(name = "key")
+  @Nullable
+  public String key;
+
+  // TODO: Add Converter to convert to/from JSON.
+  @Ignore @Nullable public Value oldValue;
```
navinsivakumar: Aren't the `Value` fields specific to record edits? Why put them on the shared `Edit` instead of `RecordEditEntity`?

gnd/src/main/java/com/google/android/gnd/repository/local/Coordinates.java
```diff
@@ -0,0 +1,61 @@
...
+  public abstract double getLatitude();
+
+  @CopyAnnotations
+  @NonNull
+  @ColumnInfo(name = "longitude")
+  public abstract double getLongitude();
+
+  // Auto-generated boilerplate:
+
+  public static Coordinates create(double latitude, double longitude) {
```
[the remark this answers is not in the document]
gino-m: Unfortunately yes it's needed by Room. Filed #46 and related IssueTracker bug to use builder instead, but it hasn't been accepted yet.

gnd/src/main/java/com/google/android/gnd/repository/local/EntityState.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static java8.util.J8Arrays.stream;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.room.TypeConverter;
+
+/** Mutually exclusive entity states shared by Features and Records. */
+public enum EntityState implements IntEnum {
+  DELETED(-1),
```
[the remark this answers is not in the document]
gino-m: Good point. It was really just to make the db more intuitive to debug, but I agree probably wiser to not mix semantics with arbitrary ids.

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
gino-m: I completely missed `Edit` when converting to @AutoValue. Good catch!

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -0,0 +1,72 @@
...
+    @NonNull
+    @TypeConverter
+    public static Edit.Type fromInt(int intValue) {
+      return IntEnum.toIntEnum(stream(values()), intValue, UNKNOWN);
+    }
+  }
+
+  @ColumnInfo(name = "type")
+  @NonNull
+  public Type type;
```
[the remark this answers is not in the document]
gino-m: Nope, this was accidentally leaked from my original prototype. I've now also migrated Edit to AutoValue.

gnd/src/main/java/com/google/android/gnd/repository/local/FeatureEntity.java
gino-m: This will be actually defined in the LocalDataService class that will map between JSONObjects fetched from the DB and actual columns in Feature (i.e., location). Deleting for now.

gnd/src/main/java/com/google/android/gnd/repository/local/IntEnum.java
gino-m: TypeConverters were using this like so:
```java
  @NonNull
  @TypeConverter
  public static EntityState fromInt(int intValue) {
    return IntEnum.fromInt(stream(values()), intValue, UNKNOWN);
  }
```

Where `values()` is the array of const values defined for the `Enum`. This allows us to iterate (or stream, rather) over the array without the intermediate step of converting to an `Iterable`.  Unfortunately polymorphism doesn't work on Arrays in Java, so we can't pass `IntEnum[]` as a generic argument either.

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
gino-m: The idea was to use the same mechanism to represent edits on `Feature`s and `Record`s to a) allow code sharing and b) to facilitate adding custom attributes to Features and/or pre-defined attributes to Record in the future.

I've removed Value from the db persistence as per our other discussions and using a generic representation analogous to Firestore (JSON) for key-value pairs of changed values. Wdyt?

gino-m:
Responses and six new commits. PTAL!

gino-m:
@navinsivakumar Also, in what packages should these classes go? The remote data service interface is in c.g.a.gnd.service.RemoteDataService, with implementation in c.g.a.gnd.service.firestore. Shall we create interface c.g.a.gnd.service.LocalDataService and put impl in c.g.a.service.room as well?

gnd/src/main/java/com/google/android/gnd/repository/local/IntEnum.java
navinsivakumar: I would suggest just accepting an `Iterable` instead of a `Stream` (and convert to a stream to perform the operation internally). It doesn't feel like that much different at the call site to call `Arrays.asList(a)` instead of `Arrays.stream(a)`, and eliminates the possibility for weird behavior when a stream is passed across different method calls (e.g. if the stream is accessed concurrently and closed somewhere else while this method is still trying to read it). I think it's easier to understand how iterable behave in those situations than streams.

If operating on streams instead of iterables, this would be more idiomatic as a `Collector` -- I think this is roughly the correct signature:
```java
static <E> Collector<E, ?, E> matchingInt(int intValue, E defaultValue);
``` 
and then the caller can do something like
```java
return stream(values()).collect((IntEnum.matchingInt(intValue, UNKNOWN));
```

But implementing a `Collector` seems like more complexity than is really needed here.

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
navinsivakumar: Hmm, using the JSON means that each edit would store the entire state of the feature or record, right? Instead of just storing the values for the particular key that is to be updated. That could make merging logic trickier if we ever need to support concurrent modifications -- it isn't as clear to see which keys were explicitly modified in this edit, so we might need to overwrite the entire feature state (including keys that the user didn't explicitly set when creating the edit).

We might have discussed this issue before, but I don't remember what we decided :).

gnd/src/main/java/com/google/android/gnd/repository/local/IntEnum.java
gino-m: Good point about exposing streams directly. Even though this package private we should establishing precedent for potentially dangerous patterns. I actually could use generics here to pass the array, which seems to be the simplest option. Doing that instead.

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @CopyAnnotations
+  @ColumnInfo(name = "type")
+  @NonNull
+  public abstract Type getType();
+
+  @CopyAnnotations
+  @ColumnInfo(name = "old_values")
+  @Nullable
+  public abstract JSONObject getOldValues();
```
gino-m: @navinsivakumar re your comment on previous version, if we're using JSON for both data in Record and changes in Edits, it would be more consistent to let key:value pairs be stored inside the JSON in both rather than the key being stored separately in the Edit.

In that case, why not store multiple changes in a single Edit, especially since they tend to get created together from a single Save action? If we only store the modified pairs, merging is no different than previously discussed ; we'd simply treat the presence of a key:value pair in an Edit UPDATE operation as if they were individual operations. Some illustrative examples:

**User populates Record for the first time:**
*oldValues*
```json
{}

```
*newValues*
```json
{"name": "Brutus", "breed": "Yorkshire Terrier"}
```

**User updates name:**
*oldValues*
```json
{"name": "Brutus"}
```

*newValues*
```json
{"name": "Fido"}
```

Wdyt? If you agree I'll make sure to document this in Javadoc.

gnd/src/main/java/com/google/android/gnd/repository/local/IntEnum.java
navinsivakumar: Did you make a change here?

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
navinsivakumar: Sure, that works. And (should we support it) you would express adding a kv pair by including the key in newValues but not old (and similar for delete). That works for me.

PR #44, opened 2019-03-31
Rename Record "Value" to "Response" for consistency

gino-m:
Also, the meaning of "Value" is broad and conflicts with other symbols. This became apparent when implementing local db support because edits can refer to user responses but also changes to featute coordinates, for example.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
+++ b/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
@@ -146,9 +146,9 @@ public class FirestoreDataService implements RemoteDataService {
           responseUpdate
               .getResponse()
               .ifPresent(
-                  value ->
+                  response ->
                       updatedResponses.put(
-                          responseUpdate.getElementId(), RecordDoc.toObject(value)));
+                          responseUpdate.getElementId(), RecordDoc.toObject(response)));
           break;
         case DELETE:
           // FieldValue.delete() is not working in nested objects; if it doesn't work in the future
--- a/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
+++ b/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
@@ -22,7 +22,7 @@ import com.google.android.gnd.service.DatastoreEvent;
 import com.google.android.gnd.service.RemoteDataService;
 import com.google.android.gnd.system.AuthenticationManager.User;
 import com.google.android.gnd.vo.Feature;
-import com.google.android.gnd.vo.FeatureUpdate.RecordUpdate.ValueUpdate;
+import com.google.android.gnd.vo.FeatureUpdate.RecordUpdate.ResponseUpdate;
 import com.google.android.gnd.vo.Project;
 import com.google.android.gnd.vo.Record;
 import com.google.android.gnd.vo.Timestamps;
@@ -105,7 +105,7 @@ public class FirestoreDataService implements RemoteDataService {
 
   // TODO: Move relevant Record fields and updates into "RecordUpdate" object.
   @Override
-  public Single<Record> saveChanges(Record record, ImmutableList<ValueUpdate> updates) {
+  public Single<Record> saveChanges(Record record, ImmutableList<ResponseUpdate> updates) {
     GndFirestore.RecordsCollectionReference records =
         db.projects().project(record.getProject().getId()).records();
 
@@ -119,7 +119,7 @@ public class FirestoreDataService implements RemoteDataService {
   }
 
   private Single<Record> saveChanges(
-      DocumentReference recordDocRef, Record record, ImmutableList<ValueUpdate> updates) {
+      DocumentReference recordDocRef, Record record, ImmutableList<ResponseUpdate> updates) {
     return RxTask.toCompletable(
             () ->
                 db.batch()
@@ -134,22 +134,22 @@ public class FirestoreDataService implements RemoteDataService {
     return db.projects().project(projectId).features().add(feature);
   }
 
-  private Map<String, Object> updatedValues(ImmutableList<ValueUpdate> updates) {
+  private Map<String, Object> updatedValues(ImmutableList<ResponseUpdate> updates) {
     Map<String, Object> updatedValues = new HashMap<>();
-    for (ValueUpdate valueUpdate : updates) {
-      switch (valueUpdate.getOperation()) {
+    for (ResponseUpdate responseUpdate : updates) {
+      switch (responseUpdate.getOperation()) {
         case CREATE:
         case UPDATE:
-          valueUpdate
-              .getValue()
+          responseUpdate
+              .getResponse()
               .ifPresent(
                   value ->
-                      updatedValues.put(valueUpdate.getElementId(), RecordDoc.toObject(value)));
+                      updatedValues.put(responseUpdate.getElementId(), RecordDoc.toObject(value)));
           break;
         case DELETE:
           // FieldValue.delete() is not working in nested objects; if it doesn't work in the future
           // we can remove them using dot notation ("responses.{elementId}").
-          updatedValues.put(valueUpdate.getElementId(), "");
+          updatedValues.put(responseUpdate.getElementId(), "");
           break;
       }
     }
--- a/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
+++ b/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
@@ -123,7 +123,10 @@ public class FirestoreDataService implements RemoteDataService {
     return RxTask.toCompletable(
             () ->
                 db.batch()
-                    .set(recordDocRef, RecordDoc.forUpdates(record, updatedValues(updates)), MERGE)
+                    .set(
+                        recordDocRef,
+                        RecordDoc.forUpdates(record, updatedResponses(updates)),
+                        MERGE)
                     .commit())
         .andThen(Single.just(record));
   }
@@ -134,8 +137,8 @@ public class FirestoreDataService implements RemoteDataService {
     return db.projects().project(projectId).features().add(feature);
   }
 
-  private Map<String, Object> updatedValues(ImmutableList<ResponseUpdate> updates) {
-    Map<String, Object> updatedValues = new HashMap<>();
+  private Map<String, Object> updatedResponses(ImmutableList<ResponseUpdate> updates) {
+    Map<String, Object> updatedResponses = new HashMap<>();
     for (ResponseUpdate responseUpdate : updates) {
       switch (responseUpdate.getOperation()) {
         case CREATE:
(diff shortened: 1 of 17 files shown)


gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -62,7 +62,7 @@
   private final SingleLiveEvent<Void> showUnsavedChangesDialogEvents;
   private final SingleLiveEvent<Void> showErrorDialogEvents;
   private final Resources resources;
-  private final ObservableMap<String, Value> values = new ObservableArrayMap<>();
+  private final ObservableMap<String, Response> values = new ObservableArrayMap<>();
```
scolsen: Should this be called `responses`?

It might be simpler than maintaining the distinction that value==response in this ViewModel. I might also be missing something. 

The corresponding getters would also need to be changed.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordFragment.java
```diff
@@ -207,7 +207,7 @@ public void addMultipleChoiceField(Field field) {
 
   public void onShowDialog(Field field) {
     Cardinality cardinality = field.getMultipleChoice().getCardinality();
-    Optional<Value> currentValue = viewModel.getValue(field.getId());
+    Optional<Response> currentValue = viewModel.getValue(field.getId());
```
scolsen: Is the response/value distinction something we're strictly adhering to at the type level only? To me, it's easier if I can think of "answer to form field" as a single thing called a response in general (at both the type and term level).

In other words, this would be called `currentResponse`

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
```diff
@@ -134,22 +134,22 @@ static Timestamps toTimestamps(@Nullable Date created, @Nullable Date modified)
     return db.projects().project(projectId).features().add(feature);
   }
 
-  private Map<String, Object> updatedValues(ImmutableList<ValueUpdate> updates) {
+  private Map<String, Object> updatedValues(ImmutableList<ResponseUpdate> updates) {
```
scolsen: s%:/`updatedValues`/`updatedResponses`?

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
```diff
@@ -134,22 +134,22 @@ static Timestamps toTimestamps(@Nullable Date created, @Nullable Date modified)
...
+    for (ResponseUpdate responseUpdate : updates) {
+      switch (responseUpdate.getOperation()) {
         case CREATE:
         case UPDATE:
-          valueUpdate
-              .getValue()
+          responseUpdate
+              .getResponse()
               .ifPresent(
                   value ->
```
scolsen: s%:/`value`/`response`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -216,35 +216,35 @@ private void saveChanges(Record r) {
...
+  private ImmutableList<ResponseUpdate> getChangeList(Record r) {
     return getChanges(r).collect(toImmutableList());
   }
 
-  private Stream<ValueUpdate> getChanges(Record r, Field field) {
+  private Stream<ResponseUpdate> getChanges(Record r, Field field) {
     String fieldId = field.getId();
-    Optional<Value> originalValue = r.getValue(fieldId);
-    Optional<Value> currentValue = getValue(fieldId).filter(v -> !v.isEmpty());
+    Optional<Response> originalValue = r.getValue(fieldId);
```
scolsen: s/`originalValue`/`originalResponse`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -216,35 +216,35 @@ private void saveChanges(Record r) {
...
+    Optional<Response> currentValue = getValue(fieldId).filter(v -> !v.isEmpty());
```
scolsen: s/`currentValue`/`currentResponse`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -266,7 +266,7 @@ private void updateError(Field field) {
     updateError(field, getValue(field.getId()));
   }
 
-  private void updateError(Field field, Optional<Value> value) {
+  private void updateError(Field field, Optional<Response> value) {
```
scolsen: /`value`/`response`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/MultiSelectDialogFactory.java
```diff
@@ -41,7 +41,7 @@
   }
 
   AlertDialog create(
-      Field field, Optional<Value> initialValue, Consumer<Optional<Value>> valueChangeCallback) {
+      Field field, Optional<Response> initialValue, Consumer<Optional<Response>> valueChangeCallback) {
```
scolsen: `value`/`response`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/MultiSelectDialogFactory.java
```diff
@@ -66,25 +66,25 @@ AlertDialog create(
 
     private boolean[] checkedItems;
 
-    public DialogState(MultipleChoice multipleChoice, Optional<Value> initialValue) {
+    public DialogState(MultipleChoice multipleChoice, Optional<Response> initialValue) {
```
scolsen: `initialValue`/`initialResponse`

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
gino-m: These were overlooked due to search & replace fatigue, thanks for catching them!

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordFragment.java
gino-m: Right, Response is only defined in the context of Records, as the user's response to a particular question.

gino-m:
Pushed an update. PTAL!

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
```diff
@@ -144,15 +147,16 @@ static Timestamps toTimestamps(@Nullable Date created, @Nullable Date modified)
               .getResponse()
               .ifPresent(
                   value ->
```
scolsen: /value/response

gnd/src/main/java/com/google/android/gnd/ui/editrecord/MultiSelectDialogFactory.java
```diff
@@ -66,15 +68,17 @@ AlertDialog create(
...
           v ->
               IntStreams.range(0, options.size())
                   .forEach(
-                      i -> checkedItems[i] = ((MultipleChoiceResponse) v).isSelected(options.get(i))));
+                      i ->
+                          checkedItems[i] =
+                              ((MultipleChoiceResponse) v).isSelected(options.get(i))));
     }
 
     private Optional<Response> getSelectedValues(List<Option> options) {
```
scolsen: /Values/Responses/

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java
gino-m: Eye like an eagle, done.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/MultiSelectDialogFactory.java
gino-m: Good catch. Done!

gino-m:
Caught a few more. Using "r" in lambdas, and "response" for variables. Also, renamed ambiguous and incorrect usages of "v ->" and "value ->".

gino-m:
So... 🙄

Been thinking about this a bit more, and have an uncomfortable question to ask late in the review:

We will want a way to represent changed values in the edit queue both at Feature level (currently only for location, in the future probably custom id, and maybe even custom attributes), and Record level (currently only for form responses, but perhaps eventually for fixed attributes as well). In that case, rather than make Value more specific (record), perhaps we should be making it more generic by moving it out of Record, and letting it represent the allowed user-provided data types (String, Float, Array of Strings, Coordinates) rather than the allowed form element types (Text, Number, Multiple Choice)? We could then even use the same JSON representation in the local db's edit queue as in remote Firestore.

@navinsivakumar in case you have any amount of cents to add.

scolsen:
I just took a cursory look at the current definitions, so lmk if I'm misunderstanding here.

Sounds like a reasonable proposal. Just to make sure I'm getting it, essentially our Queue would take operations for any object that implements the value interface, which would enable us to stick pretty much whatever edits we want (feature, record, etc.) into the edit queue? That sounds like a good idea--I don't see any particular reason for needing distinct queues at the moment if we're trying to capture a sequence of operations on a set of entities, editBoo, editBah, etc.

I think that makes sense. So we'd have a rather general interface Value that defines the methods each editable object in the queue must implement, with further specifications for particular value types (sort of how it's setup in the record over form elements today). So Value, StringValue, FloatValue, etc. and Feature and Record can implement these as needed, and anything that implements them can be put into the edit queue?

As for the existing value interface, can we just call this FormElement, or FormField, or something equivalent?

navinsivakumar:
If we're looking at the data abstractions that are presented to application logic (which I think is what we have in this PR), I don't really see a case for trying to unify the Feature and Record value representations -- business logic of, say, a ViewModel, would never use them interchangeably, would it?

We might want to do something clever about unifying the representations further down the stack in the data layer, but I don't think that needs to be decided right now (or if it *does* need to be decided right now, then we're not enforcing the abstraction layers properly). My two cents is that it might be easier to treat them as separate at that level as well, although I am not familiar with how Firestore plays into this.

gino-m:
Thanks for the replies. @scolsen, your understanding is correct, that's what I was proposing. I also  share @navinsivakumar's concerns about persisting shared objects from the canonical model in the local db. Let me try to break down the different data models we have today, and how the Db layer might be scoped:

1. [Value objects](https://github.com/google/ground-android/tree/offline-sync/gnd/src/main/java/com/google/android/gnd/vo): Canonical representation of data used by UI and core business logic. Not serialized directly into any persistent data store.
1. [Doc objects](https://github.com/google/ground-android/tree/offline-sync/gnd/src/main/java/com/google/android/gnd/service/firestore): Intermediate representation used for data in remote db, designed specifically designed for working Ground schema via Firebase API. Responses are represented as `Map<String, Object>` in this layer, since there's no easy way to customize Firebase APIs deserialization of an arbitrary, potentially nested structure. These are only visible inside `com.google.gnd.service.firestore`; [FirestoreDataService](https://github.com/google/ground-android/blob/master/gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataService.java) exposes only Value objects in its API.
1. Db objects (#41): Objects mapped to local db tables by Room. Following the preceding patterns, we should also have a LocalDatabaseService that coordinates the various DAOs and DB objects, exposing Value objects externally to the DataRespository layer. This would allow us to use whatever representation is convenient for local db persistence internally (e.g., `JSONObject`, `String`, or `Map<String, Object>`), as long as won't expose those details, and map to the shared value object.

Does this make more sense?

gino-m:
@navinsivkumar as per our other discussions, at the moment we won't use the Response for Features, so applying YAGNI here and leaving as is; we can refactor later if necessary.

Re reusing Value/Response in the local datastore, also agreed in the other thread that this would be bad practice for several reasons mentioned there.

@scottolsen, sorry for the detour. Would you proceed with merge if you think it still looks ok?

scolsen:
You got it, merging!

PR #48, opened 2019-04-12
Add Javadoc, formId field, and move db impl (#19)

gino-m:


gino-m:
Move didn't keep diffs. I'll split this into two PRs.

PR #49, opened 2019-04-12
[db-setup] Add formId field and related Javadoc

gino-m:
Issue #19


gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -60,11 +60,19 @@ public static int toInt(@Nullable Edit.Type value) {
   @NonNull
   public abstract Type getType();
 
+  /**
+   * For edits of type UPDATE, returns a JSON object with the original value of modified attributes
```
navinsivakumar: Nit: maybe use code font for UPDATE (since it's capitalized to represent the code symbol).

navinsivakumar: This might be as good a place as any to document how added and deleted attributes work. My understanding is:

* if a key is present in both `old_values` and `new_values`, the edit represents that the attribute has changed
* if a key is present in `old_values` but not `new_values`, the edit represents that the attribute is deleted
* if a key is present in `new_values` but not `old_values, the edit represents that the attribute is added

gino-m: Added link instead

gino-m: Yes, that makes sense. Adding to Javadoc now.

gino-m:
Updated. PTAL!

gnd/src/main/java/com/google/android/gnd/repository/local/Edit.java
```diff
@@ -61,17 +65,30 @@ public static int toInt(@Nullable Edit.Type value) {
   public abstract Type getType();
 
   /**
-   * For edits of type UPDATE, returns a JSON object with the original value of modified attributes
-   * before this edit. For all other edit types this will return null.
+   * For edits of type {@link Edit.Type.UPDATE}, returns a JSON object with the original values of
```
navinsivakumar: I *think* the correct way to specify the `@link` target here is:
```java
{@link Type#UPDATE}
```
since `UPDATE` is a (static) field in a different class, and `Edit` isn't necessary since we are already in the same class. Check the link syntax and let me know if you come to the same conclusion: https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html#JSSOR654

Also `old_values` will be set on type `DELETE` edits as well, correct?

gino-m: Done. Thanks!

Also, I believe we would in fact need to keep old values on DELETE in order to allow rollback. Good catch.

gino-m:
There was some delay in the diff updating between when I merged with latest upstream offline-sync but looks ok now. PTAL.

PR #53, opened 2019-04-14
Update copyright year to 2019

gino-m:


gino-m:
Note that Android Studio also added copyright notices on a view files where there were missing as well.

gino-m:
Apparently we're not supposed to update copyrights on older files. Learn something new every day! Abandoning this PR.

PR #58, opened 2019-04-18
[local-db - PART 1] Store new features in local db

gino-m:


gnd/src/main/java/com/google/android/gnd/repository/local/LocalDataStore.java
```diff
@@ -0,0 +1,37 @@
...
+ * Provides access to local persistent data store. This is the app's canonical store for latest
+ * state; local changes are saved here before being sync'ed with remote stores, and remote updates
+ * are persistent here before being made visible to the user.
+ */
+public interface LocalDataStore {
+
+  /**
+   * Adds a new Feature to the local db and queues the Feature to be added to the remote datastore.
+   *
+   * @param feature the new feature to be added. If set, the feature's ids will be ignore.
```
navinsivakumar: Nit: "The Feature's IDs, if set, will be ignored." So it's clear what "if set" refers to.

gnd/src/main/java/com/google/android/gnd/repository/local/LocalDataStore.java
```diff
@@ -0,0 +1,37 @@
...
+   * @return a Single that completes once the Feature has been saved locally. The emitted instance
+   *     be identical to the one provided except with the new id set.
```
navinsivakumar: Missing words? ("The emitted instance *will* be"?)

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
```diff
@@ -0,0 +1,82 @@
...
+        Room.databaseBuilder(app.getApplicationContext(), GndRoomDatabase.class, DB_NAME).build();
+  }
+
+  @Override
+  @Transaction
+  public Single<Feature> addNewFeature(Feature feature) {
+    return db.featureDao()
+        .insert(toFeatureEntity(feature))
+        .map(newId -> feature.toBuilder().setLocalId(newId).build())
+        .flatMap(f -> db.featureEditDao().insert(toFeatureEditEntity(f)).map(__ -> f));
```
navinsivakumar: Let's avoid double underscore for unused variable names, as per https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names (and because there are rumblings of single underscore becoming a special character in future Java versions, and it seems best to avoid something that will be visually almost the same).

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private static FeatureEntity toFeatureEntity(Feature feature) {
+    return FeatureEntity.builder()
+        .setState(EntityState.DEFAULT)
+        .setProjectId(feature.getProject().getId())
+        .setLocation(toCoordinates(feature.getPoint()))
+        .build();
+  }
+
+  private static FeatureEditEntity toFeatureEditEntity(Feature feature) throws JSONException {
```
navinsivakumar: WDYT about `toCreateFeatureEditEntity`? I imagine we'll end up with similar converters to make update edits, etc., and we wouldn't want to use overloads.

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
```diff
@@ -0,0 +1,82 @@
...
+        .setLatitude(point.getLatitude())
+        .setLongitude(point.getLongitude())
+        .build();
+  }
+
+  private static JSONObject toJSONObject(Feature feature) throws JSONException {
+    return new JSONObject().put(LOCATION_JSON_KEY, toJSONArray(feature.getPoint()));
+  }
+
+  private static JSONArray toJSONArray(Point point) throws JSONException {
```
navinsivakumar: It don't know if this is something you've already committed to in the Firestore data model, but a JSON object might be better for representing locations (with keys for latitude vs. longitude, instead of depending on a specific order in the array).

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
gino-m: It appears Java banned them because they're on a path to using them specifically for this purpose: http://openjdk.java.net/jeps/302

Once added to the language, it would be a natural and easy change from `__` to `_`. This is also probably why many online examples of RxJava and other functional APIs use this syntax.

Wdyt?

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
gino-m: Firestore has a special data type for this (GeoPoint). The thought here was that coordinates are commonly written as (lat, lng, [z]), so it was a natural representation. It also avoid redundant bytes in every feature in the local db for the keys.

gnd/src/main/java/com/google/android/gnd/repository/local/LocalDataStore.java
gino-m: Done.

gnd/src/main/java/com/google/android/gnd/repository/local/LocalDataStore.java
gino-m: Done. Updated to draft API in https://github.com/google/ground-android/pull/60/files.

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
gino-m: Just double checked, and seems Google architecture examples also rely on this convention, for ex.:
https://github.com/googlesamples/android-architecture/blob/dev-todo-mvvm-rxjava/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailFragment.java

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
gino-m: Done. Tbh after creating PR #60, I'm thinking it might make more sense to accept value-object `FeatureEdit` from the UI ViewModels/Interactions for `CREATE`, `UPDATE`, and `DELETE`, imo the end impl might differ a bit here.

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
scolsen: Just chiming in on this one since it applies to general practices throughout the codebase. `_` as a non-bindable name/"don't care" pattern match is a widely adopted convention in FP languages and libraries. Typically I'd agree with Navin and drop `__` to avoid ambiguity once `_` is adopted into Java, but, since we're talking about lambdas, the scope of `__`'s will be very limited anyway, plus we're not actually using the binding, so even if we were to miss a couple of conversions to `_` when it hits it wouldn't be a huge deal. 

I'm for adopting `__`. To me it's a clear signal that the binding isn't going to be used in the lambda body. It's better than using an actual name, e.g. ('x' or 'notused' or whatever) which would for often lead me to wonder if someone forgot to use the binding in the function body. 

That said, I think Java's method reference form, `(something::function)`, should be preferred where possible. That's not possible, of course, for cases in which we do more than call a single function in the lambda.

For the particular code we're talking about here, I'd wonder whether there's a better idiom than a lambda since you're (I think? just skimmed, so I might be wrong) making direct replacements with map. For example, in something like pseudo Haskell you could replace the lambda here with `map((const f))` Const is just `const:: a -> b -> a` but this particular use of it relies on higher order functions and currying.

Guava Functions does have [constant](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Functions.html#constant(E)) function so something like `.map(Guava.Functions.constant(f))` should work? And eliminate the lambda in this case. No guarantees. I'm still not super familiar with Java's approach to FP and a lot of intuitive solutions in FP style languages are stymied in Java from time to time.

All that said, I have no qualms with the lambda since it's plenty clear to me.

gino-m: I initially used `__` because in addition to being a common pattern found throughout Google and RxJava code samples, and similar to `_` in other languages, the convention makes it explicit that the lambda param is intentionally being ignored. If there's a better identifier that makes this fact clear I'd be happy to replace it globally.

Re this specific case, I want to condition the result of the `Single<Feature>` on the outcome of the inner `insert()` call. There's almost certainly a better way of doing this in Rx. Looking into it now.

scolsen: Totally—I might’ve been unclear with my comment, I’m onboard with __ and I agree that it makes the lack of usage of the binding explicit. 

Back to the single, I’m dealing with a similar thing right now for the leaky subscriptions refactor—finding a way to conditionally manipulate flowables and the like. 

Thus far switchIfEmpty, concat, merge, ambArray, and some others seem potentially useful depending on the situation. 

Curious to see what you find out!

gino-m: See latest commit; this was the closest I could find to using pure Rx semantics to achieve the same thing.

gnd/src/main/java/com/google/android/gnd/repository/local/RoomDataStore.java
```diff
@@ -0,0 +1,87 @@
...
+  public Single<Feature> createNewFeature(Feature feature) {
+    return db.featureDao()
+        .insert(toFeatureEntity(feature))
+        .map(newId -> feature.toBuilder().setLocalId(newId).build())
+        .flatMap(
+            f ->
+                db.featureEditDao()
+                    .insert(toCreateFeatureEditEntity(f))
+                    .toCompletable()
+                    .toSingle(() -> f));
```
gino-m: @navinsivakumar @scolsen FYI

gino-m:
Gentle ping. Would like to close the PR soon since I have a few more I'd like to stack. Thanks!

gino-m:
@navinsivakumar, I think I'll abandon this PR in favor of one that incorporates our latest discussion.

PR #60, opened 2019-04-19
Draft LocalDataStore interface and Javadoc

gino-m:
For issue #19.

Sending for early comment, PTAL!


gino-m:
New version on its way...

Issue #24, opened 2018-11-19
[Code health] Update usage of Rx streams to stop accumulating subscriptions 


gino-m:
In many cases we subscribe to a Completable, Single, or Maybe, and keeping around the Disposable for the lifetime of the component (ViewModel, View), leaking memory.

Example:
https://github.com/google/gnd-android/blob/5d87d022dd6e08a308309c93a34dc2ad97a5e195/gnd/src/main/java/com/google/android/gnd/ui/recorddetails/RecordDetailsViewModel.java#L42-L45

Rather than subscribing to the returned Maybe returned by getRecordDetails() and calling setValue on the MutableLiveData, is it possible to merge the Maybe into a existing stream in a such way that subscription is properly managed? If not, how should short-lived Single/Maybe/Completable subscriptions be disposed? I've seen conflicting suggestions online, from keeping them around in a CompositeDisposable (for the lifetime of the View?) to assuming Rx automatically disposes of them.

gino-m:
Fixed for location lock in commit d7031ea5ac697a5b9abf56391ca5f201b7b592f9.

gino-m:
Assigning to you as well since it seems you're already working on this.

PR #63, opened 2019-04-24
Add simple wrapper for streaming results

gino-m:
Blocking #24 

Wraps/unwraps values and errors in a stream to allow us to propagate error states without breaking chained streams.

Proposed usage (for `HomeScreenViewModel`):
```java
    disposeOnClear(
        featureClicks
            .switchMapSingle(wrapErrors(dataRepository::addFeature))
            .subscribe(unwrapErrors(this::showFeatureSheet, this::onAddFeatureError)));
```

Where `[un]wrapErrors` are statically imported.

Wdyt?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/rx/Result.java
+++ b/gnd/src/main/java/com/google/android/gnd/rx/Result.java
@@ -16,9 +16,10 @@
 
 package com.google.android.gnd.rx;
 
-import androidx.annotation.NonNull;
 import io.reactivex.Single;
+import io.reactivex.annotations.NonNull;
 import io.reactivex.functions.Consumer;
+import io.reactivex.functions.Function;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


gino-m:
Note: The benefits of the Result wrapper and helpers for converting to/from are probably more clear with longer and more complex Rx chains than th one in the example. In those cases, it:

* Allows us to keep Single in chain, without converting to Observable. This is more semantically precise.
* Allows us to propagate errors to then end of a chain to be handled there, rather than handling errors at each point in a chain.
* Provides an explicit representation of error vs success states, making logic more readable and debuggable.

scolsen:
I like this quite a lot! I suspect we'll get a lot of mileage out of the Single [un]wrap helpers, and it's straightforward enough to add them for the other types if we wind up needing them.

This really simplifies a lot of the tinkering we have to do with interior streams, thank you!

PR #65, opened 2019-04-26
[WIP DRAFT] Define local data store API

gino-m:


gino-m:
@navinsivakumar @scolsen PTAL

gino-m:
Note: In deciding how to model local changes, tried to walk the line between strong typing (one interface for each type of mutation x each type of entity) and no static typing (String-Object maps where the key would be some const for fixed fields like feature location). This middle ground design assumes some compromise between compile time and runtime type checking inside data store impls.

gino-m:
I've simplified the LocalChange model further by not relying on polymorphism for entity type for some change types and not others. It also makes the local change model similar to the remote one, hopefully reducing cognitive dissidence.

PR #69, opened 2019-05-01
Plumb leaky subscriptions in the RecordListViewModel

scolsen:
Fixed leaky subscriptions in the record list view model. 

Note: I've cherry picked in the Result wrapper--what I should have done, was push those changes to issue-24/master separately, but I started down the wrong path by cherry-picking them into the home screen fragment/view model fixes branch, thus leaving me to pick the same cherries over and over in subsequent branches *sigh*. Oh well, lesson learned. 

Hopefully the approach to naming here is an improvement over the initial PRs!

Contributes toward #24

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/rx/Result.java
+++ b/gnd/src/main/java/com/google/android/gnd/rx/Result.java
@@ -16,9 +16,10 @@
 
 package com.google.android.gnd.rx;
 
-import androidx.annotation.NonNull;
 import io.reactivex.Single;
+import io.reactivex.annotations.NonNull;
 import io.reactivex.functions.Consumer;
+import io.reactivex.functions.Function;
 import javax.annotation.Nullable;
 
 public class Result<T> {
@@ -61,11 +62,12 @@ public class Result<T> {
     return error;
   }
 
-  public static <T> Single<Result<T>> wrap(Single<T> single) {
-    return single.map(Result::success).onErrorReturn(Result::error);
+  public static <T, R> Function<T, Single<Result<R>>> wrapErrors(
+      @NonNull Function<T, Single<R>> fn) {
+    return (T value) -> fn.apply(value).map(Result::success).onErrorReturn(Result::error);
   }
 
-  public static <T> Consumer<? super Result<T>> unwrap(
+  public static <T> Consumer<? super Result<T>> unwrapErrors(
       Consumer<T> onSuccess, Consumer<Throwable> onError) {
     return result -> {
       switch (result.getState()) {
--- a/gnd/src/main/java/com/google/android/gnd/rx/Result.java
+++ b/gnd/src/main/java/com/google/android/gnd/rx/Result.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.rx;
+
+import androidx.annotation.NonNull;
+import io.reactivex.Single;
+import io.reactivex.functions.Consumer;
+import javax.annotation.Nullable;
+
+public class Result<T> {
+  public enum State {
+    SUCCESS,
+    ERROR
+  }
+
+  private final State state;
+
+  @Nullable private final T value;
+  @Nullable private final Throwable error;
+
+  private Result(State state, @Nullable T value, @Nullable Throwable error) {
+    this.state = state;
+    this.value = value;
+    this.error = error;
+  }
+
+  public static <T> Result<T> success(@NonNull T result) {
+    return new Result<>(State.SUCCESS, result, null);
+  }
+
+  public static <T> Result<T> error(@NonNull Throwable t) {
+    return new Result<>(State.ERROR, null, t);
+  }
+
+  @Nullable
+  public State getState() {
+    return state;
+  }
+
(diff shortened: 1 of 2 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
```diff
@@ -57,23 +71,48 @@ public void loadRecordSummaries(Feature feature, Form form) {
...
-                        stream(records)
-                            .filter(record -> record.getForm().getId().equals(formId))
-                            .collect(Collectors.toList()))));
+    recordSummaryRequests.onNext(new RecordSummaryRequest(project, featureId, formId));
+  }
+
+  class RecordSummaryRequest {
+    public Project project;
+    public String featureId;
+    public String formId;
```
gino-m: Please make these final so objects can be treated as immutable.

gino-m:
Sorry you had to do the manual work on cherry-picking!

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
scolsen: Done!

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
```diff
@@ -57,23 +71,48 @@ public void loadRecordSummaries(Feature feature, Form form) {
         feature.getProject(), feature.getFeatureType().getId(), form.getId(), feature.getId());
   }
 
+  /**
+   * Attempts to fetch a list of records based on the contents of a {@link RecordSummaryRequest}.
+   *
+   * @param request A record summary request. A triple of project, featureId, and formId.
+   * @return A list containing fetched records with forms that satisfy the formId
+   *     provided in the request.
+   */
```
gino-m: We really only need Javadoc on public methods, but ok to keep this if you think it's useful.

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
```diff
@@ -57,23 +71,48 @@ public void loadRecordSummaries(Feature feature, Form form) {
...
+  private Single<List<Record>> fetchRecordSummaries(RecordSummaryRequest request) {
+    // TODO: Only fetch records with current formId.
+    return dataRepository
+        .getRecordSummaries(request.project.getId(), request.featureId)
+        .map(
+            records ->
+                stream(records)
+                    .filter(record -> record.getForm().getId().equals(request.formId))
+                    .collect(Collectors.toList()));
+  }
```
gino-m: While we're here can we extract this lambda into a helper for better readability and less nesting?

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -0,0 +1,83 @@
...
+        case SUCCESS:
+          onSuccess.accept(result.get());
+          break;
+        case ERROR:
+          onError.accept(result.getError());
+          break;
+      }
+    };
+  }
+}
```
gino-m: Please resolve merge conflicts with issue master.

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
```diff
@@ -16,32 +16,46 @@
...
 import java8.util.stream.Collectors;
 import javax.inject.Inject;
 
 // TODO: Roll up into parent viewmodel. Simplify VMs overall.
 public class RecordListViewModel extends AbstractViewModel {
+
   private static final String TAG = RecordListViewModel.class.getSimpleName();
   private final DataRepository dataRepository;
   private MutableLiveData<List<Record>> recordSummaries;
+  private PublishSubject<RecordSummaryRequest> recordSummaryRequests;
```
gino-m: I feel like the right thing to do here would be for the VM to have access to the active project, feature, and form, either from a shared VM, or from separate streams in this VM. Wdyt? 

If you think it makes sense could you file a TODO against an Issue to fix this in the future?

gino-m:
Please flag issue number in PR description. Thanks for the PR!

gnd/src/main/java/com/google/android/gnd/rx/Result.java
scolsen: done!

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
scolsen: That makes sense to me, added!

scolsen: Though note that the change in commit 0b6ff51 is an alternative. I'm still open to making these items split streams if you think it simplifies things, but RX does provide operators we can use to express the necessary transforms concisely over a single "requests" stream.

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
scolsen: Gotcha. I also discovered there's a Bifunction version of Flatmap that succinctly expresses what we're doing here: using the initial request emissions to produce a record list, then using some other component of the initial emissions (requests) to make a 'selection' over the record lists (in this case filtering by the form ID contained in the request).

This eliminates the need for nesting to capture previous scope and allows us to use method references instead of lambdas and function calls.

gino-m:
Approved! Please merge when ready.

[Avoiding merging myself so I don't mess up diffs in your other PR that uses the same page. This way you'll have a change to merge changes back into the other PR. Letting reviewers merge their own in our setup is probably a good general practice, maybe... let's see.]

scolsen:
> Approved! Please merge when ready.
> 
> [Avoiding merging myself so I don't mess up diffs in your other PR that uses the same page. This way you'll have a change to merge changes back into the other PR. Letting reviewers merge their own in our setup is probably a good general practice, maybe... let's see.]

Thanks!

PR #70, opened 2019-05-01
Merge Result Wrapper into issue-24/master

scolsen:
So I can stop pulling these changes into every issue-24 branch :)

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/rx/Result.java
+++ b/gnd/src/main/java/com/google/android/gnd/rx/Result.java
@@ -63,26 +63,86 @@ public class Result<T> {
     return error;
   }
 
-  public static <T, R> Function<T, Single<Result<R>>> wrapSingle(
+  /**
+   * Transforms a Single, returned by applying fn to an argument of type T, into a Single that emits
+   * Results, returning Result::error on error.
+   *
+   * <p>This function is useful for handling nested RX streams, for example, in situations in which
+   * a function passed to the RX {@code map} operator returns a Single. {@code mapSingle} ensures we
+   * handle errors produced by such 'interior' Singles without breaking the 'exterior' stream.
+   *
+   * <pre>
+   *     public Single<Project> getProject(String projectId) {
+   *        return dataRepository.getProject(projectId);
+   *     }
+   *
+   *     somePublishSubject.onNext("foobar");
+   *
+   *     somePublishSubject.map(Result.mapSingle(getSingleProject))
+   *                       .subscribe(Result.unwrap(onProjectSuccess, onProjectError))
+   * </pre>
+   *
+   * @param fn A function that takes an argument of type T and returns a Single with emissions of
+   *     type R.
+   * @param <T> The initial type of values passed to fn.
+   * @param <R> The type of emissions produced by the Single returned by fn.
+   * @return
+   */
+  public static <T, R> Function<T, Single<Result<R>>> mapSingle(
       @NonNull Function<T, Single<R>> fn) {
     return (T value) -> fn.apply(value).map(Result::success).onErrorReturn(Result::error);
   }
 
-  public static <T, R> Function<T, Observable<Result<R>>> wrapObservable(
+  /**
+   * Wraps single emissions in a Result without applying any other transformations. Use this
+   * function to wrap and handle potential errors at a given point in a chain of RX transformations.
+   *
+   * @param single
+   * @param <T> The type of the Single's emissions.
+   * @return A Single that emits Results.
+   */
+  public static <T> Single<Result<T>> wrapSingle(Single<T> single) {
+    return single.map(Result::success).onErrorReturn(Result::error);
+  }
+
+  /**
+   * Transforms an Observable, returned by applying fn to an argument of type T, into an Observable
+   * that emits Results, returning Result::error on error.
+   *
+   *<p>This function is useful for handling nested RX streams, for example, in situations in which
+   * a function passed to the RX {@code map} operator returns a Observable. {@code mapObservable} ensures we
+   * handle errors produced by such 'interior' Observables without breaking the 'exterior' stream.
+   *
+   * <pre>
+   *     public Observable<Project> getProject(String projectId) {
+   *        return dataRepository.getProject(projectId);
+   *     }
+   *
+   *     somePublishSubject.onNext("foobar");
+   *
+   *     somePublishSubject.map(Result.mapObservable(getProject))
+   *                       .subscribe(Result.unwrap(onProjectSuccess, onProjectError))
+   * </pre>
+   *
+   *
+   * @param fn A function that takes an argument of type T and returns an Observable with emissions
+   *     of type R.
+   * @param <T> The initial type of values passed to fn.
+   * @param <R> The type of emissions produced by the Observable returned by fn.
+   * @return
+   */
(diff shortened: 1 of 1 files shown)


scolsen:
I've added an additional update to the wrapper here based on some changes I've found useful while re-implementing leaky subscription fixes.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -62,12 +63,32 @@ public Throwable getError() {
     return error;
   }
 
-  public static <T, R> Function<T, Single<Result<R>>> wrapErrors(
+  public static <T, R> Function<T, Single<Result<R>>> wrapSingle(
```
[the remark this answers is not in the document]
gino-m: Is there some convention for naming higher-order functions like these in Java? Feels a bit confusing that `wrapSingle` actually wraps a function to transform returned values.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -62,12 +63,32 @@ public Throwable getError() {
...
   }
 
-  public static <T> Consumer<? super Result<T>> unwrapErrors(
+  public static <T, R> Function<T, Observable<Result<R>>> wrapObservable(
+      @NonNull Function<T, Observable<R>> fn) {
+    return (T value) -> fn.apply(value).map(Result::success).onErrorReturn(Result::error);
+  }
+
+  /**
+   * Wraps observables in a Result without applying any operation. Equivalent to mapping "id" on the
```
gino-m: What does mapping "id" mean here?

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -62,12 +63,32 @@ public Throwable getError() {
...
+   * observable or "casting" or "lifting" a given observable into a Result.
```
gino-m: Casting has a specific meaning in Java, never heard of lifting - is that a term? Would "wrap" be more accurate?

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -62,12 +63,32 @@ public Throwable getError() {
...
+   *
+   * <p>This is mostly needed for situations in which we need to use a different operator aside from
```
gino-m: Nit: s/aside from/besides/

gnd/src/main/java/com/google/android/gnd/rx/Result.java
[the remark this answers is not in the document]
scolsen: I can't find any special convention. We can take a cue from RX and just make these `map<Entity>` (taking inspiration from such functions as `flatMapMaybe`, `switchMapSingle` etc.

A `map` based name would also clearly signify that this variant maps (and expects a function) as opposed to 'wraps' (expecting an object). 

So we could have `Result::mapSingle` or, if we want to be extra explicit `Result::mapSingleToResult`.

I like `mapSingle`, `mapObservable` etc.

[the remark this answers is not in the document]
gino-m: +1, great idea. Unambiguous and consistent.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
scolsen: I ported over a bunch of terminology that isn't really common or applicable in Java :sweat: --sorry just a result of my natural thought process and not double checking myself.

I do think "wrap" is probably the most accurate term for what I'm trying to say here, since, as you note, 'cast' has a rather particular meaning. Basically, I just mean to convey that we "inject" the given Observable into a Result, that we take an `Observable<T>` type and produce a `Result<Observable<T>>` without performing any transformations on the `Observable<T>` in the process. I'll give this doc another go.

Actually, my use of the term "lifting" here is not even correct from the perspective I was writing from -.-, since one lifts functions, not types or values. Basically, `lift id<T>` makes id well defined for whatever type defines `lift` (so, if you have a list of integers and `id` defined on integer, `lift id` allows you to successfully run id on each element of the list 'lifting' the id function into the list context). For lists, lift is the same as `map` `lift id = map id`. There are other variants of lift that allow you to lift n-ary functions, such as a binary function like `+`--the convenience here is the elimination of lambdas `lift2 + = map (x -> (y -> x + y))`. If you're curious, the term comes from the FP realm: https://wiki.haskell.org/Lifting.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
scolsen: I just mean to say that no additional transformation is performed on the object aside from moving it into a result. I'll rewrite this so that it's clearer. 

I used the term id because "wrapping" something like we do here is equivalent to calling `map(x -> x)`. The x -> x part is usually called the identity function, or just id. Java takes care of that part for us implicitly though because of the way Result is defined, so it's kinda silly to mention it!

scolsen:
Added javadocs and updated names!

Writing docs for these methods is tricky :sweat_smile: Should we can nested streams *matryoshkas*? :P

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -0,0 +1,164 @@
...
+   *     somePublishSubject.onNext("foobar");
+   *
+   *     somePublishSubject.map(Result.mapSingle(getSingleProject))
+   *                       .subscribe(Result.unwrap(onProjectSuccess, onProjectError))
+   * </pre>
+   *
+   * @param fn A function that takes an argument of type T and returns a Single with emissions of
+   *     type R.
+   * @param <T> The initial type of values passed to fn.
+   * @param <R> The type of emissions produced by the Single returned by fn.
```
gino-m: Very thorough! I think these are probably more details than we need for our purposes, but don't mind keeping these docs since they're complete and correct.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @param observable
+   * @param <T> The type of the observable's emissions.
+   * @return An Observable that emits Results.
+   */
+  public static <T> Observable<Result<T>> wrapObservable(Observable<T> observable) {
+    return observable.map(Result::success).onErrorReturn(Result::error);
+  }
+
+  public static <T> Consumer<? super Result<T>> unwrap(
```
gino-m: Missing Javadoc here, but I'd like to unblock this PR so we can add it in a follow-up.

gino-m:
Thanks for the PR!

gnd/src/main/java/com/google/android/gnd/rx/Result.java
scolsen: Doh! will do.

gino-m: Going forward I'd add only essential Javadoc to public methods; inferred types probably don't need to be described since the caller doesn't explicitly declare them, and when params and return values are already described by the main text we can omit them as well.

gino-m:
Really excited to see these design issues and potentially bugs resolved. Thanks for your patience iterating to set up a good precedent!

PR #72, opened 2019-05-03
Bug: Fix calls to Result methods in HomeScreenVM

scolsen:
I forgot to update some calls to the Result API in the HomeScreenViewModel after intaking changes -.-

This fixes that mistake.


gino-m:
Please be sure build succeeds locally until we have commit checks working in GitHub. Thanks for the fix!

gino-m:
Also I still feel the way we wrap and unwrap Result is somewhat counterintuitive but don't want to block or bikeshed; that said, reserving the right to bring it up again at some point :)

PR #73, opened 2019-05-03
Fix Leaky Subscriptions in Edit Record 

scolsen:
This change fixes leaky subscriptions in the Edit Record Fragment and View Model.

**Note:** At the moment this includes a quick fix for the homescreen--this change will disappear from this PR once #72 is merged--I just pulled it in here to make sure the edit record interactions were working as expected (since the bug addressed in #72 prevents builds).

Another brick on the road to #24.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
@@ -100,16 +100,18 @@ public class EditRecordViewModel extends AbstractViewModel {
     Flowable<Resource<Record>> editRecordStream =
         argumentProcessor.flatMap(
             args -> {
+              Single<Resource<Record>> value;
               if (args.getRecordId().equals(NEW_RECORD_ID_ARG_PLACEHOLDER)) {
-                return this.dataRepository
+                value = this.dataRepository
                     .createRecord(args.getProjectId(), args.getFeatureId(), args.getFormId())
                     .map(Resource::loaded)
-                    .doOnSuccess(__ -> onNewRecordLoaded()).toFlowable();
+                    .doOnSuccess(__ -> onNewRecordLoaded());
               } else {
-                return this.dataRepository
+                value = this.dataRepository
                     .getRecordSnapshot(args.getProjectId(), args.getFeatureId(), args.getRecordId())
-                    .doOnSuccess(r -> r.data().ifPresent(this::update)).toFlowable();
+                    .doOnSuccess(r -> r.data().ifPresent(this::update));
               }
+              return value.toFlowable();
             });
 
     disposeOnClear(saveRecordStream.subscribe(record::setValue));
--- a/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
@@ -60,6 +60,7 @@ import javax.inject.Inject;
 // TODO: Save draft to local db on each change.
 public class EditRecordViewModel extends AbstractViewModel {
   private static final String TAG = EditRecordViewModel.class.getSimpleName();
+  private static final String NEW_RECORD_ID_ARG_PLACEHOLDER = "NEW_RECORD";
 
   private final DataRepository dataRepository;
   private final AuthenticationManager authManager;
@@ -88,7 +89,7 @@ public class EditRecordViewModel extends AbstractViewModel {
     this.argumentProcessor = BehaviorProcessor.create();
     this.recordProcessor = BehaviorProcessor.create();
 
-    Flowable<Resource<Record>> recordStream =
+    Flowable<Resource<Record>> saveRecordStream =
         recordProcessor.flatMap(
             r ->
                 this.authManager
@@ -96,9 +97,23 @@ public class EditRecordViewModel extends AbstractViewModel {
                     .flatMap(user -> this.dataRepository.saveChanges(r, getChangeList(r), user))
                     .toFlowable(BackpressureStrategy.LATEST));
 
-    Single<Resource<Record>> newRecord = argumentProcessor.map()
+    Flowable<Resource<Record>> editRecordStream =
+        argumentProcessor.flatMap(
+            args -> {
+              if (args.getRecordId().equals(NEW_RECORD_ID_ARG_PLACEHOLDER)) {
+                return this.dataRepository
+                    .createRecord(args.getProjectId(), args.getFeatureId(), args.getFormId())
+                    .map(Resource::loaded)
+                    .doOnSuccess(__ -> onNewRecordLoaded()).toFlowable();
+              } else {
+                return this.dataRepository
+                    .getRecordSnapshot(args.getProjectId(), args.getFeatureId(), args.getRecordId())
+                    .doOnSuccess(r -> r.data().ifPresent(this::update)).toFlowable();
+              }
+            });
 
-    disposeOnClear(recordStream.subscribe(record::setValue));
+    disposeOnClear(saveRecordStream.subscribe(record::setValue));
+    disposeOnClear(editRecordStream.subscribe(this::onRecordSnapshot));
   }
 
   public ObservableMap<String, Response> getResponses() {
@@ -145,19 +160,6 @@ public class EditRecordViewModel extends AbstractViewModel {
     return showErrorDialogEvents;
   }
 
-  void editNewRecord(String projectId, String featureId, String formId) {
-    // TODO(#24): Fix leaky subscriptions!
-
-    argumentProcessor.onNext(args);
-
-    disposeOnClear(
-        dataRepository
-            .createRecord(projectId, featureId, formId)
-            .map(Resource::loaded)
-            .doOnSuccess(__ -> onNewRecordLoaded())
-            .subscribe(this::onRecordSnapshot));
-  }
-
   @NonNull
   private Optional<Record> getCurrentRecord() {
     return Resource.getData(record);
@@ -177,14 +179,8 @@ public class EditRecordViewModel extends AbstractViewModel {
             r.getForm().getField(k).ifPresent(field -> onResponseChanged(field, Optional.of(v))));
   }
 
-  void editExistingRecord(String projectId, String featureId, String recordId) {
-    // TODO: Store and retrieve latest edits from cache and/or db.
-    // TODO(#24): Fix leaky subscriptions!
-    disposeOnClear(
-        dataRepository
-            .getRecordSnapshot(projectId, featureId, recordId)
-            .doOnSuccess(r -> r.data().ifPresent(this::update))
-            .subscribe(this::onRecordSnapshot));
+  void editRecord(EditRecordFragmentArgs args) {
+    argumentProcessor.onNext(args);
   }
 
   private void onRecordSnapshot(Resource<Record> r) {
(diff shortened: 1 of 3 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -65,8 +65,8 @@
 
     disposeOnClear(
         addFeatureClicks
-            .switchMapSingle(Result.wrapErrors(dataRepository::addFeature))
-            .subscribe(Result.unwrapErrors(this::showFeatureSheet, this::onAddFeatureError)));
+            .switchMapSingle(Result.mapSingle(dataRepository::addFeature))
+            .subscribe(Result.unwrap(this::showFeatureSheet, this::onAddFeatureError)));
```
gino-m: Please use stacked PRs when there's overlap to avoid redundant diffs.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +86,47 @@
     this.showUnsavedChangesDialogEvents = new SingleLiveEvent<>();
     this.showErrorDialogEvents = new SingleLiveEvent<>();
     this.authManager = authenticationManager;
+    this.editRecordRequests = PublishSubject.create();
+    this.recordSaveRequests = PublishSubject.create();
+
+    disposeOnClear(
+            recordSaveRequests.zipWith(authManager.getUser(), this::saveRecord)
```
gino-m: Doesn't this try to save the record as soon as the viewmodel is instantiated? Doesn't sound like what we want to do here.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +86,47 @@
...
+            .switchMap(Result::wrapObservable)
```
gino-m: What is this line trying to do? If there's no substream from which to catch errors, why, wrap in Result?

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +86,47 @@
...
+
+  private Single<Resource<Record>> createOrUpdateRecord(EditRecordRequest request) {
+    return request.getIsNew() ? createRecord(request) : updateRecord(request);
+  }
+
+  private Single<Resource<Record>> createRecord(EditRecordRequest request) {
+    return this.dataRepository
+        .createRecord(request.getProjectId(), request.getFeatureId(), request.getFormId())
+        .map(Resource::loaded)
+        .doFinally(this::onNewRecordLoaded);
```
gino-m: `doFinally` is called both on error and on success. Did you mean `doOnSuccess()`?

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +86,47 @@
...
+  }
+
+  private Single<Resource<Record>> updateRecord(EditRecordRequest request) {
+    return this.dataRepository
+        .getRecordSnapshot(request.getProjectId(), request.getFeatureId(), request.getRecordId())
+        .doOnSuccess(r -> r.data().ifPresent(this::update));
```
gino-m: I assume this is out of scope of this PR, but eventually would be nice to stop relying on side-effects, e.g. can we call these in `onRecordSnapshot()` instead. We can get back to this later.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -286,4 +314,36 @@ private boolean hasUnsavedChanges() {
   private boolean hasErrors() {
     return !errors.isEmpty();
   }
+
+  @AutoValue
+  public abstract static class EditRecordRequest {
```
gino-m: This seems like an unfortunate amount of boilerplate.. Can we just use `EditRecordFragmentArgs`? Note that these generated classes are just "structs" w/o and reference to the view.

In future iterations, the view model can already have the required state and then we can just pass the min required state in the stream that triggers the edit.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
scolsen: (long reply ahead)

In this case it actually only calls `saveRecord` on a click, but I will admit that while this is clear on the surface and I think(?) the best transformation for the job as far as RX semantics are concerned, if I attempt to reason about what's actually going on here it's quite difficult. To make certain, I've gone ahead and double checked this manually by running this and logging (via `doOnNext`) on the zipped stream. The subscription here does indeed only fire on save button clicks.

There is a better approach here, but it's mostly due to the fact that we have to deal with streams that cross contextual boundaries, it's not an issue inherent to the use of `zipWith`.

One thing that's important to keep in mind about zips:

>The resulting {@code ObservableSource<R>} returned from {@code zip} will invoke {@link Observer#onNext onNext}
     * as many times as the number of {@code onNext} invocations of the source ObservableSource that emits the fewest
     * items.

So, completions of the respective observables *do matter*. Basically, whichever one of the sources completes first will also cause the resulting zipped stream to complete.

The first piece, `recordSaveRequests` is easy enough to comprehend. It's defined in the same file, and it's `onNext` is called on saves. This means that the zipped stream will propagate emissions from the first observable whenever `saveChanges()` is called. That makes sense. `zipWith` subscribes to each stream in order, so first, it will subscribe to `recordSaveRequests`, which means that this stream will effectively control the flow of the zipped stream. `recordSaveRequests` is only called in this file, and we're confident it does not call `onComplete` anywhere (just like the zipped stream, it just lives as long as the VM)  which means it won't ever shortcircuit emissions from the other observable.

It's more difficult to reason about the `authManager` part since the source observable is coming from elsewhere. It turns out that the observable being returned is actually a  `BehaviorSubject`, but several of the methods that return a reference to it are typed generically as `Observables`. Because it's a behavior subject, we can actually be confident that we'll always have a value to pass to our binop, since the behavior will replay the previous value upon initial subscription--other than this initial value however, I don't believe any other value comes from this source in this particular case. There's actually a major problem in our reference to this stream since we *have no idea when or if it will complete* from the context of this VM, so it could potentially kill our zipped stream by completing it at any time--we need to step outside and peek, potentially at the whole code base, to determine when or if this stream will ever kill or complete our zipper. Luckily, at the moment it is fine, but this is pretty bad situation since it's not guaranteed.  

The problems related to `authManager` actually hold, so far as I can tell, for *any* observable that's coming from a different scope/context/class. Thus far, I can think of two ways of handling potential completions or errors from external contexts, we do a bit of both of these things at present, more frequently relying on (1.)

1. Handle all errors and completions in the original context, and wrap results in a Resource or Result, giving other contexts a means of handling completions or errors through value checking. 
2. Use RX semantics to handle this, e.g. use `doOnComplete` etc. in situations in which the completion of a stream brought in from another context matters, e.g. when using `zipWith`.

The problem with approach (1.) is that this doesn't stop certain RX mechanisms from firing. For example, if we took  this approach for the situation described here, even though we're returning a wrapper object, e.g. `Resource<User>` which the `saveRecrod` function can us to get a sense of whether or not the origin stream completed, the original stream's `onComplete` *can still be called* resulting in the termination of the zipped stream regardless. Contrarily, `onErrorReturn` actually *does* prevent `onError` from being called, letting the stream continue indefinitely or continue to completion--but this is somewhat problematic as it alters the semantics of the observable significantly. That said, this approach eliminates the need for `zipWith`, but the unknown call to `onComplete` can still be problematic.

As for (2.) it's more verbose in most cases, and we don't (?) yet have good markers for things like "Observable that never completes." or "Observable that never errors." or "Observable that completes and errors normally" which would serve as huge hints in handling observers coming from other contexts appropriately. Another option is to simply handle each case always, ensuring that we use robust subscribers (defining onNext, onError, and onComplete handlers) and handling completions and errors appropriately in stream transformations as well (e.g. using `doOnDispose` when using `zipWith`. 

In general, this scenario suggests it's optimal to keep RX entities as self contained and bounded as possible. We're able to make very confident assertions about the `recordSaveRequest` subject because it's entirely defined and manipulated in a single scope/context. If we need to reference such a stream from outside this context, we have no idea whether errors are being handled, whether the stream completes, etc. It also suggests to me that it's best to define all three callbacks on subscribe(onNext, onError, onComplete).

Sooooo, we need `Result` for 'interior' streams that subvert `onError` to continue emitting items. But, to handle errors on 'exterior' streams we should probably just use subscribe callbacks?

This is still a problematic setup, I think, because as an implementer, there's no way for me to know whether a given stream calls `onErrorReturn` unless I look at the implementation, or unless I'm calling it myself on an interior stream using `Result.mapX` -- but I don't need to call this function if the original stream is already calling `onErrorReturn` which I can't know unless, again, I check the implementation.

For many of the functions that return observables that are fetching some data from the db, we can probably ditch `onErrorReturn` in favor of forms of `retry` and error handling in `subscribe()` which would remove the need for some wrappers and disambiguate some streams (i.e. does this stream potentially error or not).

lets chat about this since my hesd's muddle enough at this point haha

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
scolsen: `saveRecord` returns an Observable, so we need to call `switchMap` to flatten the streams. We essentially have two options, call `switchMap` with the identity function `r -> r` or call it with `Result::wrapObservable`. I chose the latter because it gives us a standardized way to handle errors. If the observable returned by `saveRecord` throws an error, it will instead return `Result::error` and continue emitting. I don't think this adds too much in this case, though, so I've switched to the more direct approach of using the identity function.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
scolsen: Shoot, you're right, error on my part. I was trying to make the signatures more "correct". Really, I was just being a bit pedantic.

`onNewRecordLoaded` doesn't do anything to the record, so it's not necessary to pass it one. `doOnSuccess` passes the success value to the method provided as an argument, while `doFinally` does not--`doOnSuccess` suggests the method is dependent on the success value, the `doFinally` that it isn't. This is just a semantic gap in RX since there's no `doOnSuccess(void)` so there's no notion of "only do this action if this succeeds, but do it independently of the value"--actually, this probably is expressible in RX terms but entails superfluous transformations.

I was just trying to avoid it because a function that takes an argument and does nothing with it is sometimes suspicious to me.

scolsen: Some investigation suggests `Single` is an outlier of sorts as it's apparently only implemented by RX Java and its derivatives. 

`doOnSuccess` has to satisfy both the role of `doOnNext` and `doOnCompleted`--that's why it cares about the value :/

In any case, we can use an alternative here.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
scolsen: Sounds good, #78

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
scolsen: Sure thing

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
scolsen: gotcha

PR #74, opened 2019-05-03
Fix leaky subscriptions in the Project Selector

scolsen:
Another step toward finishing #24.

**Note**: This currently contains the quickfix for the homescreen. This change can be ignored and will go away once #72 is merged into issue-24/master.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
@@ -35,6 +35,8 @@ import io.reactivex.subjects.Subject;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+
 import javax.inject.Inject;
 
 public class ProjectSelectorViewModel extends AbstractViewModel {
@@ -42,16 +44,28 @@ public class ProjectSelectorViewModel extends AbstractViewModel {
 
   private final DataRepository dataRepository;
   private final MutableLiveData<Resource<List<Project>>> projectSummaries;
+  private final BehaviorProcessor<Integer> indexProcessor;
+  public final Completable activeProjectStream;
 
   @Inject
   ProjectSelectorViewModel(DataRepository dataRepository, AuthenticationManager authManager) {
     this.dataRepository = dataRepository;
     this.projectSummaries = new MutableLiveData<>();
+    this.indexProcessor = BehaviorProcessor.create();
+
     Observable<AuthenticationManager.User> user = authManager.getUser();
 
     Observable<Resource<List<Project>>> summaryStream =
-        user
-            .flatMap(this.dataRepository::getProjectSummaries);
+        user.flatMap(this.dataRepository::getProjectSummaries);
+
+    this.activeProjectStream =
+        indexProcessor.flatMapCompletable(
+            idx ->
+                this.dataRepository.activateProject(
+                    Resource.getData(this.projectSummaries)
+                        .orElse(Collections.emptyList())
+                        .get(idx)
+                        .getId()));
 
     disposeOnClear(summaryStream.subscribe(projectSummaries::setValue));
   }
@@ -60,8 +74,7 @@ public class ProjectSelectorViewModel extends AbstractViewModel {
     return projectSummaries;
   }
 
-  Completable activateProject(int idx) {
-    return dataRepository.activateProject(
-        Resource.getData(this.projectSummaries).orElse(Collections.emptyList()).get(idx).getId());
+  void activateProject(int idx) {
+    indexProcessor.onNext(idx);
   }
 }
--- a/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
@@ -18,24 +18,17 @@ package com.google.android.gnd.ui.projectselector;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
 
-import com.google.android.gms.auth.api.Auth;
 import com.google.android.gnd.repository.DataRepository;
 import com.google.android.gnd.repository.Resource;
 import com.google.android.gnd.system.AuthenticationManager;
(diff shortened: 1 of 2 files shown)


gino-m:
> **Note**: This currently contains the quickfix for the homescreen. This change can be ignored and will go away once #72 is merged into issue-24/master.

If you set "base" to the PR with the homescreen change it'll disappear from the diff (click Edit next to PR desc).

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -15,48 +15,88 @@
...
   ProjectSelectorViewModel(DataRepository dataRepository, AuthenticationManager authManager) {
     this.dataRepository = dataRepository;
     this.projectSummaries = new MutableLiveData<>();
-    this.authManager = authManager;
-  }
+    this.activeProject = new MutableLiveData<>();
+    this.activateProjectErrors = new MutableLiveData<>();
+    this.projectSelections = PublishSubject.create();
+
+    Observable<AuthenticationManager.User> user = authManager.getUser();
```
gino-m: Are we sure it's safe to do all this work in the constructor? Or should it be run in a separate method and invoked from the appropriate lifecycle?

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -15,48 +15,88 @@
...
     disposeOnClear(
-        authManager
-            .getUser()
-            .flatMap(user -> dataRepository.getProjectSummaries(user))
-            .subscribe(summaries -> projectSummaries.setValue(summaries)));
+        activeProjectStream.subscribe(
+            Result.unwrap(activeProject::setValue, this::onActiveProjectError)));
+    disposeOnClear(
+        availableProjects.subscribe(
+            Result.unwrap(projectSummaries::setValue, this::onProjectSummariesError)));
```
gino-m: Since we're only going half-way to Rx town at this stage, I'm finding it hard to follow what's actually happening now. Let's put our heads together to make these a bit more readable once the fixes are merged.

Also, could you confirm that you've run and tested these changes locally and that the app is WAI?

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
scolsen: For this particular case, I think it makes sense--the list of available projects needs to be fetched as soon as the VM is instantiated, as these are passed as an original list of options to the Fragment. Thereafter, we wait for selections over that list using the PublishSubject defined here.

Instead of being a stream, I've changed this to refer to only the first user returned by the authManager, to prevent inadvertent actions as other users (not that this is likely).

This also enabled me to be rid of those calls to `Result`.

Open to a better approach if you have one!

scolsen:
> > **Note**: This currently contains the quickfix for the homescreen. This change can be ignored and will go away once #72 is merged into issue-24/master.
> 
> If you set "base" to the PR with the homescreen change it'll disappear from the diff (click Edit next to PR desc).

Good to know, thanks!

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -15,48 +15,89 @@
...
     this.dataRepository = dataRepository;
     this.projectSummaries = new MutableLiveData<>();
-    this.authManager = authManager;
-  }
+    this.activeProject = new MutableLiveData<>();
+    this.activateProjectErrors = new MutableLiveData<>();
+    this.projectSelections = PublishSubject.create();
+
+    AuthenticationManager.User user =
+        authManager.getUser().blockingFirst(AuthenticationManager.User.ANONYMOUS);
```
scolsen: Are we okay with using [`blockingFirst`](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#blockingFirst-T-)? It's the most concise way, I think, of doing what we want here (getting and keeping a reference only to the first user returned by the user stream, that way the user referenced throughout the interaction is the same).

There are alternatives, but they involve a bit more manipulation since `getUser` returns an Observable.

gino-m: In the future would like to expose a synchronous `currentUser()` method and create the backing Behavior with default anon user, but for the purposes of this fix it looks like the right thing.

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -15,48 +15,89 @@
...
 
 public class ProjectSelectorViewModel extends AbstractViewModel {
   private static final String TAG = ProjectSelectorViewModel.class.getSimpleName();
 
   private final DataRepository dataRepository;
   private final MutableLiveData<Resource<List<Project>>> projectSummaries;
-  private final AuthenticationManager authManager;
+  private final PublishSubject<Integer> projectSelections;
+  public final MutableLiveData<Project> activeProject;
+  public final MutableLiveData<Throwable> activateProjectErrors;
```
gino-m: I think I need to eat my words about making these public final. Since they're mutable, any client can accidentally update them. We really should expose them as `LiveData` via methods like `projectActivationLiveData()` and `activeProjectLiveData()`. Sorry that I overlooked this!

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
scolsen: Sounds good! #86!

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
scolsen: Cool, I've changed these to private and added accessors. I've also updated the fragment appropriately.

I stuck with the existing naming pattern for live data accessors `get<data>()`. We can take on a future effort to rename these with the LiveData suffix, if you want.

Issue #57, opened 2019-04-17
[Offline support] Support two way offline data sync


gino-m:
Blocked by #19, #132, #133.

gino-m:
This task is too broad so closing in favor of more specific steps.

gino-m:
Lots left to do here. Rough plan:

- [ ] Manage projects
  - [ ] Join project - store in local db
  - [ ] Leave project - delete from local db
  - [ ] Activate project - switch projects in local db
  - [ ] Reload project - show message when project config changes w/option to reload or postpone
  - [ ] Reload project - show reload icon next to projects w/pending updates
  - [ ] On activation, store last active project, restore when app opens
- [ ] Offline functionality for features and records
  - [ ] Sync remote feature changes in active projects to local db when app is open
  - [ ] If opening feature while offline, show offline warning
  - [ ] If opening feature when online, download latest records
  - [ ] On remote change reload records for open feature
- [ ] Manage users
  - [ ] Make all data in local db per-user
  - [ ] Allow switching among available accounts in drawer ([docs](https://developer.android.com/training/id-auth)). Fetch and store Firestore auth token on switch.
  - [ ] Add link to manage accounts / add account to system
- [ ] Cleanup / Code Health
  - [ ] Replace *Doc with *Document Maps

gino-m:
Either marked above TODOs as done, or forked into separate issues.

Branch `issue-57/master` has been merged into `master` and deleted. 🚀

PR #76, opened 2019-05-05
[LOCAL DB - PART 2b] Implement offline uuid generator

gino-m:
Local id generation needed to avoid maintaining separate local and remote ids, and step of reassigning new remote id on sync. Firestore ids are "statistically unique", meaning they are unique for all practical purposes.

Blocking #57


gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataStore.java
```diff
@@ -41,21 +42,23 @@
 import javax.inject.Singleton;
 
 @Singleton
-public class FirestoreDataStore implements RemoteDataStore {
+public class FirestoreDataStore implements RemoteDataStore, OfflineUuidGenerator {
```
navinsivakumar: Instead of implementing the interface in `FirestoreDataStore` itself, how about writing a separate `FirestoreUuidGenerator` class (that also depends on the Firestore instance)? That would separate functionality and make the object dependency graph cleaner for the classes that only need to generate IDs (e.g. where a local entity is created and to be stored locally only) and don't otherwise interact with Firestore (and so don't need to pull in the whole dependency structure of the data store).

gino-m: Ideally yes. To that we'd also need to refactor the Firestore api client instance setup being done in the constructor. Wanted to keep these PRs small, so decided to postpone that for now. Ok for you? Note that this is all part of the implementation details of our Firestore impl. If someone were to implement their own backend data store, they'd be free to split the implementation into two classes or provide a single impl that implements both interfaces.

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataStore.java
```diff
@@ -41,21 +42,23 @@
...
   private static final String TAG = FirestoreDataStore.class.getSimpleName();
+  private static final String ID_COLLECTION = "/ids";
   private final GndFirestore db;
+  private final FirebaseFirestore firestore;
 
   @Inject
   FirestoreDataStore() {
     // TODO: Run on I/O thread, return asynchronously.
-    final FirebaseFirestore firestore = FirebaseFirestore.getInstance();
+    this.firestore = FirebaseFirestore.getInstance();
```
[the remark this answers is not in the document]
gino-m: +1. Done.

gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataStore.java
navinsivakumar: Sure. Maybe add a TODO to split this up.

gino-m: Done!

PR #77, opened 2019-05-05
[LOCAL DB - PART 2c] Generate feature ids immediately when creating value object

gino-m:
This is an interim step before storing data directly to the local db and work queue (#57).


gnd/src/main/java/com/google/android/gnd/service/firestore/FirestoreDataStore.java
```diff
@@ -135,9 +136,9 @@ static Timestamps toTimestamps(@Nullable Date created, @Nullable Date modified)
   }
 
   @Override
-  public Single<Feature> addFeature(Feature feature) {
+  public Completable saveFeature(Feature feature) {
     String projectId = feature.getProject().getId();
-    return db.projects().project(projectId).features().add(feature);
+    return db.projects().project(projectId).features().feature(feature.getId()).set(feature);
```
navinsivakumar: To confirm: does this work when the db does not contain a feature with the given feature ID?

gino-m: That's correct. `set()` creates or overwrites the feature. Since this is a thin wrapper around Firestore I used the same terminology, but could replace with `saveOrUpdate()` if you think that's preferable.

PR #91, opened 2019-05-14
Fix Leaky Subscriptions

### LINKED ISSUE #24: [Code health] Update usage of Rx streams to stop accumulating subscriptions

scolsen:
Whew, so we:

* Created a Result for wrapping errors in nested streams
* Had some fruitful discussions about our handling of streams
* Fixed the leaky subscriptions.

Hooray!

Closes #24 

@navinsivakumar FYI

### DIFF
--- a/gnd/build.gradle
+++ b/gnd/build.gradle
@@ -17,7 +17,7 @@
 apply plugin: 'com.android.application'
 
 project.ext {
-    autoDisposeVersion = "0.8.0"
+    autoDisposeVersion = "1.2.0"
     autoValueVersion = "1.6"
     butterKnifeVersion = "10.1.0"
     constraintLayoutVersion = "1.0.2"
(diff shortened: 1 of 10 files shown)


scolsen:
@gino-m looks like  I can't delete this branch  since it's protected. 

It's safe to delete since we've merged it.

Issue #84, opened 2019-05-07
[Code health] Verify exceptions are either wrapped or handled in all Rx streams


gino-m:
We can rename `Result` to something like `ValueOrError`, and wrap the value at the source, obviating the need to wrap/unwrap at arbitrary steps in the stream.

@scolsen FYI.

gino-m:
Should we really though? Posted [this question](https://groups.google.com/forum/#!topic/rxjava/sIFNQng5yg4) asking about an alternative approach, a là:
```
    disposeOnClear(
        addFeatureClicks
            .switchMap(
                f ->
                    dataRepository
                        .saveFeature(f)
                        .doOnComplete(() -> showFeatureSheet(f))
                        .doOnError(t -> onAddFeatureError(t))
                        .onErrorComplete()
                        .toObservable())
            .subscribe());
```

There's not need for a wrapper unless we want to propagate errors across long-lived streams. Since Rx streams obtained imperatively will return Completable, Single, or Maybe, this will usually only happen when declaratively setting up and subscribing in ViewModels.

gino-m:
@scolsen, I discussed this with ex-colleagues at FAO (@cdanielw, @lpaolini) who work with RxJS extensively on [SEPAL](https://github.com/openforis/sepal), and consensus is that it's better to wrap and unwrap in some intermediate object to represent errors as a value rather than at stream level (as we have now), rather than to rely on side-effects on the inner stream, e.g.:

https://github.com/google/ground-android/blob/fd1324aa8e59159c9941c5dfab6c7c082cff0b6f/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java#L62-L67

My guest is that with some additional syntactic sugar and clearer naming we can make this more self-documenting / obvious without sacrificing FRP "purity".

scolsen:
Sounds good! Let's stick with our basic idea then and iterate on improving the clarity of the semantics. 

`ValueOrError` is fairly straightforward. 

Would it be clearer to the majority of Java/Android devs if we emulated or took inspiration from an existing interface like `Optional`? i.e. instead of doing something like `unwrap` doing something like `orError` -- this particular example doesn't sound like a great idea to me, just speculating here haha, but maybe it's an approach we can consider.

gino-m:
Yes, I was thinking the same thing. Afaict, the closest use of Optional is
Java 9's "ifPresentOrElse()". What would the equivalent be here?
ifSuccessOrElse?

[...]

PR #94, opened 2019-06-04
Simplify edit record an add feature Rx error handling

gino-m:
https://github.com/google/ground-android/issues/84#issuecomment-498366352

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
@@ -36,7 +36,6 @@ import com.google.android.gnd.GndApplication;
 import com.google.android.gnd.R;
 import com.google.android.gnd.repository.DataRepository;
 import com.google.android.gnd.repository.Resource;
-import com.google.android.gnd.rx.Result;
 import com.google.android.gnd.system.AuthenticationManager;
 import com.google.android.gnd.ui.common.AbstractViewModel;
 import com.google.android.gnd.ui.common.SingleLiveEvent;
@@ -47,12 +46,11 @@ import com.google.android.gnd.vo.Record;
 import com.google.android.gnd.vo.Record.Response;
 import com.google.android.gnd.vo.Record.TextResponse;
 import com.google.common.collect.ImmutableList;
-import java.util.Arrays;
-import java.util.Collections;
-
 import io.reactivex.Observable;
 import io.reactivex.Single;
 import io.reactivex.subjects.PublishSubject;
+import java.util.Arrays;
+import java.util.Collections;
 import java8.util.Optional;
 import java8.util.stream.Stream;
 import javax.inject.Inject;
@@ -96,8 +94,10 @@ public class EditRecordViewModel extends AbstractViewModel {
 
     disposeOnClear(
         editRecordRequests
-            .switchMapSingle(Result.mapSingle(this::createOrUpdateRecord))
-            .subscribe(Result.unwrap(this::onRecordSnapshot, this::onEditRecordError)));
+            .switchMapSingle(
+                record ->
+                    createOrUpdateRecord(record).onErrorResumeNext(t -> __ -> onEditRecordError(t)))
+            .subscribe(this::onRecordSnapshot));
   }
 
   private Single<Resource<Record>> createOrUpdateRecord(EditRecordRequest request) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
@@ -96,7 +96,9 @@ public class EditRecordViewModel extends AbstractViewModel {
         editRecordRequests
             .switchMapSingle(
                 record ->
-                    createOrUpdateRecord(record).onErrorResumeNext(t -> __ -> onEditRecordError(t)))
+                    createOrUpdateRecord(record)
+                        .doOnError(this::onEditRecordError)
+                        .onErrorResumeNext(Single.never()))  // Prevent from breaking upstream.
             .subscribe(this::onRecordSnapshot));
   }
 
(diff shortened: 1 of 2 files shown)


scolsen:
oh... just noticed this is still a draft PR :sweat_smile: I'll take a second look once you mark it ready for review.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -96,8 +94,10 @@
 
     disposeOnClear(
         editRecordRequests
-            .switchMapSingle(Result.mapSingle(this::createOrUpdateRecord))
-            .subscribe(Result.unwrap(this::onRecordSnapshot, this::onEditRecordError)));
+            .switchMapSingle(
+                record ->
+                    createOrUpdateRecord(record).onErrorResumeNext(t -> __ -> onEditRecordError(t)))
```
scolsen: I think the lambda lambda is a bit obscure. Is it possible to achieve this in a style similar to the one you proposed in #84 ?

```java
.addFeature(feature)
    .doOnError(this::onAddFeatureError)  // Handle errors.
    .onErrorResumeNext(Single.never()))
```

gino-m:
> oh... just noticed this is still a draft PR 😅 I'll take a second look once you mark it ready for review.

No worries, I realized in hindsight these changes are easily separable into multiple PRs, so don't mind merging this separately.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
gino-m: My thinking here was that `doOnError ` + `onErrorResumeNext` are really a verbose way of saying "do something and recover", and that since the new stream will be a no-op, it was clearer to just not have the extra boilerplate, especially since this is going to be a common pattern. @navinsivakumar any thoughts on this?

scolsen: Ah I see what you mean. I agree, though I am still having some trouble wrapping my brain around this lambda.

What does `t -> __ -> onEditRecordError(t)` actually amount to in Java? My reading is that it's a function that returns an anonymous function (or can you not do that in Java?), but that doesn't seem to square with the type of `OnErrorResumeNext` (`Function<? super Throwable,? extends ObservableSource<? extends T>>` or `ObservableSource<? extends T>`) unless the definition of `onEditRecordError` is changed to return an Observable--but even in that case wouldn't the additional lambda `__ ->` be unnecessary and throw the type off? Wouldn't the type being passed to `onErrorResumeNext` here amount to `Function<Throwable, Function<???, ObservableSource<T>>>`

Would plain old `t -> onEditRecordError(t)` work assuming `onEditRecordError` returns an Observable/Single/whatever's required? 

Or am I missing something about the types? (a definite possibility; I find Java's approach to generics to be ungainly and mystifying)

scolsen: I suppose there's also some potential misunderstanding on my part due to type erasure. I frequently forget that Java type sigs are subject to erasure, which could mean types of `Function<>` are erased and replaced with their return types...which would presumably make your lambda type check just fine?

gino-m: > What does t -> __ -> onEditRecordError(t) actually amount to in Java?

It's a function that returns a SingleSource, defined by `__ -> onEditRecordError(t)`, that calls the error handler and never emits any items or errors. This syntax works because SingleSource is a [functional interface](https://www.geeksforgeeks.org/functional-interfaces-java/). My reasoning was that since the SingleSource never does anything, and exists only to "eat" the error, it would be better to not add additional boilerplate. I could be convince otherwise though.

> Would plain old `t -> onEditRecordError(t)` work assuming `onEditRecordError` returns an Observable/Single/whatever's required?

This would work as well, but would require onEditRecordError(t) to return a SingleSource. I preferred not to do this because it mixes Rx stream semantics into the error handler, which shouldn't need to know anything about where it's being called from.

scolsen: Ah, I see, thanks for the explanation! 

This makes sense--so this is effectively equivalent to `Single.never()`.

While this is more economical (and clever!), I actually find the boilerplate a lot easier to understand and read. I also think it's more flexible if we need to suddenly do more than just toss out a never emitting Single. For example, if we decided that we instead want to return a backup single on error it's easier for me to modify this pattern:

```
//before 
.doOnError(this::onEditRecordError)
.onErrorResumeNext(Single.never())
//after
.doOnError(this::onEditRecordError)
.onErrorResumeNext(Single.just("Foo"))
```

than this pattern

```
//before
.onErrorResumeNext(t -> __ -> onEditRecordError(t))
//after ??
.onErrorResumeNext(t -> __ -> {onEditRecordError(t) return Single.just("Foo")})
```

There may also be situations in which we decide we need to switch from `onErrorResumeNext` to another error handling operator, such as `retry`. I think that's another situation in which keeping the notions of "handle error with side effect (that doesn't effect the stream)" and "alter stream on error" distinct at the stream operator level (instead of mixed in the function passed to onErrorResume) is beneficial:

```
//before 
.doOnError(this::onEditRecordError)
.onErrorResumeNext(Single.never())
//after
.doOnError(this::onEditRecordError)
.retry(2)
```

```
//before
.onErrorResumeNext(t -> __ -> onEditRecordError(t))
//after
.doOnError(this::onEditRecordError) // wind up having to break out the pattern anyway.
.retry(1)
```

In fact, I'm not even confident the change to the lambda in the first example is the correct/best way to implement what I'm attempting to implement. I feel much more comfortable modifying the verbose pattern.

I think sticking with the RX operators, even if its a bit tedious in the majority of cases, keeps things more cleanly separated. 

I think the first is also less difficult to understand at first glance. Utilizing the fact that `SingleSource` is a functional interface was not immediately obvious to me.

That all said, I *also* have a fairly shallow comprehension of Java in general, and find its type system and implementation of functional programming patterns complicated and unclear, so take this all with a grain of salt as it may just be a personal issue on my part.

gino-m: I think that makes sense. If you the details we're obscuring are potentially relevant, that we shouldn't hide them. I'll update accordingly!

gino-m:
Updates. PTAL!

PR #95, opened 2019-06-05
Simplify loading of record summaries and related error handling

gino-m:
Addressing conclusions in https://github.com/google/ground-android/issues/84#issuecomment-498366352

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
+++ b/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
@@ -16,6 +16,8 @@
 
 package com.google.android.gnd.repository;
 
+import static java8.util.stream.StreamSupport.stream;
+
 import android.util.Log;
 import com.google.android.gnd.service.DatastoreEvent;
 import com.google.android.gnd.service.RemoteDataService;
@@ -35,6 +37,7 @@ import io.reactivex.Single;
 import io.reactivex.subjects.BehaviorSubject;
 import io.reactivex.subjects.Subject;
 import java.util.List;
+import java8.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
@@ -102,12 +105,19 @@ public class DataRepository {
     event.getEntity().ifPresentOrElse(cache::putFeature, () -> cache.removeFeature(event.getId()));
   }
 
-  // TODO: Return Resource.
-  public Single<List<Record>> getRecordSummaries(String projectId, String featureId) {
+  public Single<List<Record>> getRecordSummaries(
+      String projectId, String featureId, String formId) {
     // TODO: Only fetch first n fields.
     // TODO: Also load from db.
     return getFeature(projectId, featureId)
-        .flatMap(feature -> remoteDataService.loadRecordSummaries(feature));
+        .flatMap(feature -> remoteDataService.loadRecordSummaries(feature))
+        .map(summaries -> filterSummariesByFormId(summaries, formId));
+  }
+
+  private List<Record> filterSummariesByFormId(List<Record> summaries, String formId) {
+    return stream(summaries)
+        .filter(record -> record.getForm().getId().equals(formId))
+        .collect(Collectors.toList());
   }
 
   private Single<Feature> getFeature(String projectId, String featureId) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/RecordListViewModel.java
@@ -17,25 +17,19 @@
 package com.google.android.gnd.ui.home.featuresheet;
 
 import android.util.Log;
-
-import static java8.util.stream.StreamSupport.stream;
-
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
 import com.google.android.gnd.repository.DataRepository;
-import com.google.android.gnd.rx.Result;
 import com.google.android.gnd.ui.common.AbstractViewModel;
 import com.google.android.gnd.vo.Feature;
 import com.google.android.gnd.vo.Form;
 import com.google.android.gnd.vo.Project;
 import com.google.android.gnd.vo.Record;
+import io.reactivex.Single;
+import io.reactivex.subjects.PublishSubject;
 import java.util.Collections;
 import java.util.List;
-
-import io.reactivex.Observable;
-import io.reactivex.subjects.PublishSubject;
 import java8.util.Optional;
-import java8.util.stream.Collectors;
 import javax.inject.Inject;
 
 // TODO: Roll up into parent viewmodel. Simplify VMs overall.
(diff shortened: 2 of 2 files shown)


gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -102,12 +105,19 @@ private void onRemoteFeatureVectorChange(DatastoreEvent<Feature> event) {
...
-  // TODO: Return Resource.
-  public Single<List<Record>> getRecordSummaries(String projectId, String featureId) {
+  public Single<List<Record>> getRecordSummaries(
+      String projectId, String featureId, String formId) {
     // TODO: Only fetch first n fields.
     // TODO: Also load from db.
     return getFeature(projectId, featureId)
-        .flatMap(feature -> remoteDataService.loadRecordSummaries(feature));
+        .flatMap(feature -> remoteDataService.loadRecordSummaries(feature))
+        .map(summaries -> filterSummariesByFormId(summaries, formId));
```
scolsen: I'm asking this more out of curiosity than as an operable suggestion--Would this be even simpler if we removed the interior collection structure? If we changed `GndFirestore` to handle the translation from non-RX structures into RX structures and had it return an Observable.from(Collection<Record>) we could just stick with one layer of structure here and call `filter` on the observable directly, instead of having to dip into the value and additional collection structure. This would create a stronger boundary between non-RX world and RX world--only the lowest level (things returned from the firestore directly would be non-RX as well as some outside view values, everything in between is RX. 

Is that possible? Or is there some implementation details that prevent this approach? I think there's a fairly straightforward way to render the entire list at once using only an `Observable<Record>`. Basically we'd have an `Observable<Record>` all the way up to the view model, which would map the contents emitted by `Observable<Record>` into a collection maintained by the VM, then set a mutable data to the value of this collection once the Observable<Record> stream calls `.onComplete`, letting us render the entire list at once. In other words, we'd be working with an `Observable<Record>` that "completes" once all the records have been emitted instead of a `Single<List<Record>>`. This would also presumably allow us to indicate precisely what summaries failed in some meaningful way in the UI instead of just filtering out the bad summaries in the firestore as we do currently (basically we'd emit those errors downstream all the way to the VM).

Sorry if this winds up being a silly question--I've only peeked at the Firestore impl.

gino-m: There's a semantic difference between a `Single<List<Record>>` and a `Observable<Record>`;  in the former, we expect them to arrive all together, and in the latter we expect them to be streamed. The latter would likely add some complexity in the UI layer, so it may be better to only do it that way if that's our intention (which it's not here). Does that make sense?

gino-m: Resolving for now so I can merge, but let's continue the discussion.

scolsen: That makes perfect sense!

What I'm wondering is whether there's any reason we shouldn't move that complexity into the UI? 

It seems like right now we're mixing the data with its presentation because the only reason we expose it as a list is so that the UI has it as a list (*correct me if I'm wrong about this and there's a more significant reason its a list!*). While this simplifies the code for one particular view/VM it also:

* Complicates the datarepository/data-transferring/pipeline code
* mixes the data repo/pipeline with constraints required by a one particular View Model/UI (the record list)
* limits the flexibility we have using this data source for other potential future UIs: A highly hypothetical example, let's say we want to create a widget that displays longest record summary.  If this stream were just a stream of `Observable<RecordSummary> (or w/e)` We could just reach for this stream and filter, map, subscribe, doOnNext, etc. appropriately, but as the current impl stands we'd have to do extra finagling with the list structure, because we only want one of the record summaries. So even though we've made the one UI simpler by tailoring the datarepo code to its particular needs, we potentially complicate other UIs or wind up implementing other specialized streams in the repo when we could use a more less-specialized single stream (e.g. of records).

If my intuition serves me okay, I don't think building the list in the View Model would be that bad, and I think it may actually be easier to grok than our current approach. In psuedojava:

```java
List<Record> recordSummaries;
MutableLiveData<List<Record>> recordSummaryLiveData

// here we'd have some publishsubject to fire off the record summary request.

//Assume getRecordSummaries is Observable<Record>
// we'd hook up the subject to the repo in the usual fashion.
// for simplicity i'm just subscribing directly to the repo here.
// in place of .add() we do whatever we need to do to populate the list.
// ignoring the fact that its an immutable DS(?) for now. May need a diff structure.
disposeOnClear(
   dataRepository.getRecordSummaries(projectId, featureId))
     .subscribe(summary -> recordSummaries.add(), 
                       __ -> return, 
                       completed -> recordSummaryLiveData.set(recordSummaries))
)
```

So, we subscribe to:

* Add a record summary to the list of record summaries `onNext`
* do nothing `onError` since we'll be handling errors in the way we settled on.
* set the mutable live data of summaries to the list we built when `onComplete` fires. Thus, once we've gotten all the summaries from the stream, the view will get a piece of immutable live data that contains the entire list for rendering simultaneously.

This approach should also make our rendering options more flexible. For example, if we're getting the summaries directly though the stream we can easily switch to rendering them progressively instead of simultaneously, as they come down the stream, with minimal hassle. Contrarily it'd take more work to switch to that style if they're already grouped in the stream--the VM has less flexibility for modelling the data. At present the datarepo is already doing some amount of modelling work for the VM. I think having the data repo pass along "raw" data to the VMs and letting the VMs handle the modelling (building raw streamed information into lists, arrays, foobars, what have you) will ultimately make the datarepo and VM code more flexible and thus easier to scale and modify down the road.

That said I'm only coming from the perspective of the VM/UI layer--don't want to mess with things too much if it  catastrophically breaks expectations on the data storage side.

Sorry for the long, blathering comments :sweat:

scolsen: In other words, it seems like if we can keep DataRepo streams free of as much additional structure as possible it increases their flexibility and will help ensure we're doing all our structuring/modelling of data in the viewmodels and not before hand. 

In general any `Single<List<> or other Iterable structure>` seems fishy to me--it's essentially  equatable to `Observable<Item>` at which point, why embed it in a Single when operating on two layers of structure is more complex? Using `Observable<Item>` instead lets you work on emissions (aka iterable members) at the stream level--you can just use the RX operators directly, you don't need to additionally iterate over the interior structure; instead of iterating over two structures (stream + list) you're just iterating over the stream.  

For example, if we returned an `Observable<Record>` instead of `Single<List<Record>>` this code would become: 

```java
  // of course we need to change getByFeature to return an Observable.
  // We'd just call Observable.from on the contents of the querySnapshot if present.
 // If not present Observable.from(emptyCollection)
 
 public Observable<Record> loadRecordSummaries(Feature feature) {
    return db.projects().project(feature.getProject().getId()).records().getByFeature(feature);
  }

public Observable<Record> getRecordSummaries(
      String projectId, String featureId, String formId) {
    return getFeature(projectId, featureId)
        .flatMap(feature -> remoteDataService.loadRecordSummaries(feature))
        .filter(record -> record.getForm().getId().equals(formId));
```

No need for the `filterSummariesByFormId` helper method anymore.

Note that `Observable<List>` is different--that would be a stream of lists-- so that's semantically distinct at the level of emissions. But the only semantic distinction between `List<Item>` and `Single<List<Item>>` are the RX operation semantics (onNext, onComplete, etc.) which suggests the policies of the library enforced that implementation, not the actual semantics of the code and that the additional structure is extraneous.

I think this is a situation where we can take advantage of RX powers to make the code easier to maintain and modify. We can totally eliminate the list from the equation.

Again I'm saying all this with limited knowledge so I could just be flat out wrong. :exclamation:

gino-m: I think this might make sense, though it probably doesn't reduce complexity, but rather moves it into the ViewModel, where we actually use the stream (i.e., the current model will either need to wait for the whole Observable to complete and update the UI, or accumulate them as they arrive, in which case the UI would necessarily need to be able to handle incremental loading.

Give our current tight timeline for G4G, I think I'd rather punt on changing existing code for now, but feel free move these comments into a new issue, and to apply this technique to your current work on the offline basemaps.

Thanks for digging into this in detail!

PR #96, opened 2019-06-05
Clean up handling of project selection streams

gino-m:
Addressing conclusions in https://github.com/google/ground-android/issues/84#issuecomment-498366352

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
@@ -16,78 +16,67 @@
 package com.google.android.gnd.ui.projectselector;
 
 import android.util.Log;
-
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.MutableLiveData;
-
 import com.google.android.gnd.repository.DataRepository;
 import com.google.android.gnd.repository.Resource;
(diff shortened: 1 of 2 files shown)


gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -16,78 +16,67 @@
...
     this.projectSelections = PublishSubject.create();
 
+    disposeOnClear(
+        projectSelections
+            .switchMapSingle(
+                idx ->
+                    dataRepository
+                        .activateProject(getProjectSummary(idx).getId())
+                        .doOnError(this::onActiveProjectError)
+                        .onErrorResumeNext(Single.never()))
```
scolsen: Not for this change, but is this potentially a place where we'll want to retry on error in the future?

If I'm envisioning this correctly, at present if this errors we'll log (and eventually show) an error then we'll wait idly until the user issues another project selection click to try again. Retrying once or twice, and showing some message to that effect, would be a small UX enhancement (at least when the user has a connection).

gino-m: Good idea in practice, but probably a "nice to have" vs lots of other "must haves" at the moment. Let's keep it in the back of our minds for future improvements!

PR #98, opened 2019-06-09
Merge with latest from upstream master

gino-m:
Work towards #57.


scolsen:
Sorry, I merged #97 which I think needs to be brought in here before you can merge :/

Since this is a big change, I'll probably diff the modified files against master to make sure we're in sync.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +84,56 @@
     this.showUnsavedChangesDialogEvents = new SingleLiveEvent<>();
     this.showErrorDialogEvents = new SingleLiveEvent<>();
     this.authManager = authenticationManager;
+    this.editRecordRequests = PublishSubject.create();
+    this.recordSaveRequests = PublishSubject.create();
+
+    disposeOnClear(
+        recordSaveRequests
+            .switchMap(this::saveRecord)
```
scolsen: Don't worry about it here, but we should change this to use the `doOnError` handling instead of handling the error in the subscription.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
```diff
@@ -78,6 +84,56 @@
...
+            .subscribe(record::setValue, this::onSaveRecordError));
```
scolsen: Don't worry about it here, but we should change this to use the `doOnError` handling instead of handling the error in the subscription.

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -15,48 +15,80 @@
...
+            .switchMapSingle(
+                idx ->
+                    dataRepository
+                        .activateProject(getProjectSummary(idx).getId())
+                        .doOnError(this::onActiveProjectError)
+                        .onErrorResumeNext(Single.never()))
+            .subscribe(activeProject::setValue));
+
+    AuthenticationManager.User user =
+        authManager.getUser().blockingFirst(AuthenticationManager.User.ANONYMOUS);
```
scolsen: Note to self: We should add a TODO for this and the other instance of `blockingFirst` to re-implement this in the `AuthenticationManager`.

gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordViewModel.java
gino-m: I'm pretty sure this is wrong, actually; either the Resource should wrap the error, or it should be handled on the inner stream. I'll add a TODO.

gino-m:
Done! PTAL.

gino-m:
Hmm. Author consent check failed. @scolsen, can you check it both emails you used for commits are all registered with GitHub?

scolsen:
Oh, I think I just manually give my consent and you can merge it by applying a label?

> It's up to you to confirm consent of all the commit author(s), set the cla label to yes (if enabled on your project), and then merge this pull request when appropriate.

At any rate I can register my other address if that helps.

gino-m:
> (Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.)

All commits in question are by @scolsen, who has signed CLAs and is reviewer on this thread.

scolsen:
Looks like we still have no luck here? Let me see if making another account does the trick.

gino-m:
Iiuc you started a review but never actually approved/closed it? It says there's one pending review. I'll try to dismiss it now.

gino-m:
Ah ok, we still need a review with your approval. Can you please restamp?

PR #149, opened 2019-07-11
[Offline Imagery Support] Add Gmaps-Addons Dependency, Fix marker removal bug

scolsen:
This PR adds the cocoahero/android-gmaps-addons lib. We'll use this to *provide* an mbtiles provider.

Additionally, it fixes a small bug in removeMarkers that prevents rendering overlays.


settings.gradle
```diff
@@ -1 +1,4 @@
-include ':gnd'
+include ':gnd', ':android-gmaps-addons'
+
+project(':android-gmaps-addons').projectDir = new File('third_party/android-gmaps-addons')
```
gino-m: Can we nest this isn't another directory w/the repo/user name so that it's clearer that it's a copy of that repo?
i.e. `third_party/cocoahero/android-gmaps-addons`

gino-m:
Great, thanks for consolidating this into a new PR!

third_party/android-gmaps-addons/src/com/cocoahero/android/gmaps/addons/mapbox/MapBoxOfflineTileProvider.java
```diff
@@ -0,0 +1,209 @@
...
+
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.LatLngBounds;
+import com.google.android.gms.maps.model.Tile;
+import com.google.android.gms.maps.model.TileProvider;
+
+public class MapBoxOfflineTileProvider implements TileProvider, Closeable {
```
gino-m: Can we avoid importing the whole third party source, and only include this class plus the LICENSE.md? Iirc it doesn't have any other dependencies.

settings.gradle
scolsen: Sure thing!

third_party/android-gmaps-addons/src/com/cocoahero/android/gmaps/addons/mapbox/MapBoxOfflineTileProvider.java
scolsen: Yep, we can do that, I'll make the change.

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -145,7 +145,9 @@ private void addMarker(MapMarker mapMarker, boolean hasPendingWrites, boolean is
   }
 
   private void removeAllMarkers() {
-    map.clear();
+    for (Marker marker : markers.values()) {
+      marker.remove();
+    }
```
gino-m: Good catch!

third_party/android-gmaps-addons/src/com/cocoahero/android/gmaps/addons/mapbox/MapBoxOfflineTileProvider.java
scolsen: So, I've reduced the source as much as I possibly can in ebec4e4. Unfortunately, there are some files we need to make gradle work with the library, and I don't know enough about gradle to know if there's some other way to set this up to remove the need for things like the manifest file, etc. (it was already a bit of a pain to try to figure out this set up). 

At the very least, we now only have the two classes we need!

gino-m:
Lgtm! Feel free to remove Draft status and I'll approve right away so your next PR can be smaller. Thanks!

scolsen:
> Lgtm! Feel free to remove Draft status and I'll approve right away so your next PR can be smaller. Thanks!

Done! Updated the description and title for clearer history.

PR #153, opened 2019-07-15
Add local db indexes and foreign key relations

gino-m:


gnd/src/main/java/com/google/android/gnd/persistence/local/room/RecordMutationEntity.java
```diff
@@ -39,13 +39,25 @@
...
+          parentColumns = "id",
+          childColumns = "feature_id",
+          onDelete = CASCADE),
+      @ForeignKey(
+          entity = RecordEntity.class,
+          parentColumns = "id",
+          childColumns = "record_id",
+          onDelete = CASCADE)
+    },
+    indices = {
```
scolsen: Are we going to need all these indices!? 0.o

(a facetious comment)

gino-m: No. Good question actually.

PK contraints are implemented as indexes in Sqlite, so @Index isn't needed.

Also, we shouldn't add stuff until we use it (YAGNI), and composite indexes should be used for fields that are queried together.

Updated all the things.

Issue #154, opened 2019-07-16
[Offline support] App doesn't sync data after max retries of sync errors


gino-m:
When retry count mutations exceeds some limit, show notification or other affordance to allow user to revert changes or to try again. Follow-up to #147.

gino-m:
@sufyanAbbasi Can you increase MAX_RETRIES to some large number as a workaround?

sufyanAbbasi:
I found two interesting things...

1. `incrementRetryCount()` is only called in the `MediaUploadWorker` for media upload, but not anywhere else. The implication is that retry count doesn't change therefore, we're never not retrying failed tasks (task failed successfully?)!
    1. Fixing it may not be the correct thing to do here.
1. We don't mark mutations as `COMPLETE` anywhere except for the `MediaUploadWorker` after the media mutation upload is complete. I'm not sure if this is intentional or not.

sufyanAbbasi:
From what I understand, we mark all mutations as having failed when one mutation fails, since the FireStore mutations are done in bulk: in `LocalMutationSyncWorker:121`: `remoteDataStore.applyMutations(mutations, user)`. Since if any fail, none of them will be committed, we want to mark all of these mutations as having failed.

Therefore, if we're retrying uploads, we should consider breaking up the failed mutations into their own worker pool to prevent overloading the network.

sufyanAbbasi:
Unassigning myself for now, since I don't have a full understanding of the mutation/worker/sync model and I don't want to mess it up. But I'm happy to work on it if I can peer program with someone on it!

gino-m:
I'll grab time to discuss tomorrow, thanks!

PR #155, opened 2019-07-16
Retry sync on failure

### LINKED ISSUE #147: [Offline support] Retry failed mutation sync attempts
- [x] Respect backoff if successive edits enqueue updates to same feature
- [x] Increment retry counter
- [ ] Report error if retry counter exceeds some threshold

gino-m:
Closes #147. Future work in #154.

For now retry indefinitely.

Could one you kindly give a quick look? Please reply here if you're looking so the other knows. Thanks in advance!

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/shared/FeatureMutation.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/shared/FeatureMutation.java
@@ -29,7 +29,7 @@ import java8.util.Optional;
  * the UI, and are dequeued and sent to the remote data store by the background data sync service.
  */
 @AutoValue
-public abstract class FeatureMutation extends Mutation {
+public abstract class FeatureMutation extends Mutation<FeatureMutation.Builder> {
 
   /** Returns the UUID of the feature type being modified. */
   public abstract String getFeatureTypeId();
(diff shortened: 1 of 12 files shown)


gino-m:
Note: Contains diffs that will disappear after merging #153.

scolsen:
I'll take a look.

gnd/src/main/java/com/google/android/gnd/persistence/shared/FeatureMutation.java
```diff
@@ -40,6 +40,17 @@
...
   public abstract Optional<Point> getNewLocation();
 
+  @Override
+  public abstract Builder toBuilder();
+
+  /** Returns the mutations of type {@link FeatureMutation} contained in the specified list. */
+  public static ImmutableList<FeatureMutation> filter(ImmutableList<Mutation> mutations) {
+    return stream(mutations)
+        .filter(FeatureMutation.class::isInstance)
+        .map(FeatureMutation.class::cast)
```
scolsen: Why do we need to cast here? Does the compiler require this--is it unable to figure out the type in this case without explicit casting? Wouldn't filter(FeatureMutation.class::isInstance) already concertize to a type: ImmutableList<FeatureMutation>?

gino-m: We can treat `FeatureMutation` as the more generic type `Mutation`, but here we're down casting the items, which in Java is always explicit.

gino-m:
> Looks good, thanks!

Thanks for reviewing!

PR #157, opened 2019-07-17
Offline-Imagery: Add Tile classes to the local db.

scolsen:
In order to represent what tiles users have available for offline viewing, we'll make use of the local data store. We store tile state along with the id of the json feature the tile is associated with and the path at which the tile data is stored. This should allow us to keep track of what tiles are on the device, as well as resume failed or interrupted downloads.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/model/tile/Tile.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/tile/Tile.java
@@ -1,0 +1,29 @@
+package com.google.android.gnd.model.tile;
+
+import com.google.auto.value.AutoValue;
+
+import java.net.URL;
+
+@AutoValue
+public abstract class Tile {
+    public enum State {
+        PENDING,
+        DOWNLOADED,
+        FAILED
+    }
+
+    public abstract String getId();
+    public abstract String getPath();
+    public abstract State getState();
+
+    public static Builder newBuilder() {return new AutoValue_Tile.Builder();}
+
+    @AutoValue.Builder
+    public abstract static class Builder {
+        public abstract Builder setId(String id);
+        public abstract Builder setPath(String path);
+        public abstract Builder setState(State state);
+
+        public abstract Tile build();
+    }
+}
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
@@ -1,0 +1,93 @@
+package com.google.android.gnd.persistence.local.room;
+
+import androidx.annotation.NonNull;
+import androidx.room.ColumnInfo;
+import androidx.room.Entity;
+import androidx.room.PrimaryKey;
+
+import com.google.android.gnd.model.tile.Tile;
+import com.google.auto.value.AutoValue;
+import com.google.auto.value.AutoValue.CopyAnnotations;
+
+@AutoValue
+@Entity(tableName = "tile")
+public abstract class TileEntity {
+    @CopyAnnotations
+    @NonNull
+    @PrimaryKey
+    @ColumnInfo(name = "id")
+    public abstract String getId();
+
+    @CopyAnnotations
+    @NonNull
+    @ColumnInfo(name = "path")
+    public abstract String getPath();
+
+    @CopyAnnotations
+    @NonNull
+    @ColumnInfo(name = "state")
+    public abstract TileEntityState getState();
+
+    public static Tile toTile(TileEntity tileEntity) {
+        Tile.Builder tile =
+                Tile.newBuilder()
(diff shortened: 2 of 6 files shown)


gnd/src/main/java/com/google/android/gnd/model/tile/Tile.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.auto.value.AutoValue;
+
+import java.net.URL;
+
+@AutoValue
+public abstract class Tile {
+    public enum State {
+        PENDING,
+        DOWNLOADED,
+        FAILED
```
gino-m: We may also need STARTED/IN_PROGRESS?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
```diff
@@ -0,0 +1,93 @@
...
+    public static Tile toTile(TileEntity tileEntity) {
+        Tile.Builder tile =
+                Tile.newBuilder()
+                        .setId(tileEntity.getId())
+                        .setPath(tileEntity.getPath())
+                        .setState(toTileState(tileEntity.getState()));
+        return tile.build();
+    }
+
+    private static Tile.State toTileState(TileEntityState state) {
```
gino-m: I really wish there was a better way to define and map between these representations. Maybe there's something we can do in the future with https://kotlinlang.org/docs/reference/data-classes.html

[just sharing, no action required]

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
```diff
@@ -0,0 +1,93 @@
...
+        switch (state) {
+            case PENDING:
+                return Tile.State.PENDING;
+            case DOWNLOADED:
+                return Tile.State.DOWNLOADED;
+            case FAILED:
+                return Tile.State.FAILED;
+            default:
+                return Tile.State.PENDING;
```
gino-m: throw new IllegalArgumentException("Unknown tile state: " + state);

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntityState.java
```diff
@@ -0,0 +1,33 @@
...
+    public static int toInt(@Nullable TileEntityState value) {
+        return IntEnum.toInt(value, UNKNOWN);
+    }
+
+    @NonNull
+    @TypeConverter
+    public static TileEntityState fromInt(int intValue) {
+        return IntEnum.fromInt(values(), intValue, UNKNOWN);
+    }
+}
```
gino-m: Nit: Add CR

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
```diff
@@ -0,0 +1,93 @@
...
+    public abstract static class Builder {
+        public abstract Builder setId(String newId);
+
+        public abstract Builder setPath(String newPath);
+
+        public abstract Builder setState(TileEntityState newState);
+
+        public abstract TileEntity build();
+    }
+}
```
gino-m: Nit: Add CR

google-java-format doesn't do this for us?

gnd/src/main/java/com/google/android/gnd/model/tile/Tile.java
```diff
@@ -0,0 +1,29 @@
+package com.google.android.gnd.model.tile;
```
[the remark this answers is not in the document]
scolsen: Good call, thanks!

gnd/src/main/java/com/google/android/gnd/model/tile/Tile.java
scolsen: Right. Added!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
scolsen: Agreed! Are we planning on rolling our own support for that, or are we going to switch to Kotlin :D?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
scolsen: I just ran it and nope :/

gnd/src/main/java/com/google/android/gnd/persistence/local/room/TileEntity.java
scolsen: Done, makes much more sense :)

scolsen:
@gino-m Looks like my small fixes invalidated your first review and approval :/

Mind re-approving when you get a chance?

PR #158, opened 2019-08-02
Offline Imagery: Add JSON Layer Rendering

scolsen:
This PR introduces JSON layer rendering on the map. Users can select a new primary nave menu option, "Download Tiles" which, when selected, activates a JSON layer on the map that gives users a visual field to use to select tiles. 

Once this looks okay, we can move on to passing the tile selections off to the appropriate VM and subsequently the download worker, providing the user with an indication of available storage space, highlight which tiles are already downloaded on the device, etc.

This change also includes some preliminary setup for passing on a set of selected tiles from the MapAdapter, but we don't do anything with theses as of yet.

I'm not 100% confident about how I've hooked up the UI menu option to the rendering logic, but it seemed to be the cleanest way to do it considering the way the VMs and reactive streams were already set up.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
@@ -335,11 +335,19 @@ public class HomeScreenFragment extends AbstractFragment
         showProjectSelector();
         closeDrawer();
         break;
+      case R.id.nav_download_tiles:
+        showTileDownloadLayer();
+        closeDrawer();
+        break;
       case R.id.nav_sign_out:
         authenticationManager.signOut();
         break;
     }
     return false;
+  }
+
+  private void showTileDownloadLayer() {
+    this.viewModel.getDownloadTileClicks().onNext(new Object());
   }
 
   private class BottomSheetCallback extends BottomSheetBehavior.BottomSheetCallback {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
@@ -47,6 +47,7 @@ public class HomeScreenViewModel extends AbstractViewModel {
   private final LiveData<Persistable<Project>> activeProject;
 
   private final PublishSubject<Feature> addFeatureClicks;
+  private final PublishSubject<Object> downloadTileClicks;
 
   // TODO: Move into MapContainersViewModel
   private final SingleLiveEvent<Point> addFeatureDialogRequests;
@@ -66,6 +67,7 @@ public class HomeScreenViewModel extends AbstractViewModel {
         LiveDataReactiveStreams.fromPublisher(dataRepository.getActiveProjectOnceAndStream());
     this.navigator = navigator;
     this.addFeatureClicks = PublishSubject.create();
+    this.downloadTileClicks = PublishSubject.create();
 
     disposeOnClear(
         addFeatureClicks
@@ -92,6 +94,10 @@ public class HomeScreenViewModel extends AbstractViewModel {
 
   public void openNavDrawer() {
     openDrawerRequests.setValue(null);
+  }
+
+  public PublishSubject<Object> getDownloadTileClicks() {
+    return downloadTileClicks;
   }
 
   public LiveData<Persistable<Project>> getActiveProject() {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -138,6 +138,10 @@ public class MapContainerFragment extends AbstractFragment {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCenter()));
     enableLocationLockBtn();
+    homeScreenViewModel
+        .getDownloadTileClicks()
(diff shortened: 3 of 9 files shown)


scolsen:
Sorry, snuck one more quick change in! Added basic tile coloring to indicated selected tiles.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -335,13 +335,21 @@ public boolean onNavigationItemSelected(@NonNull MenuItem item) {
         showProjectSelector();
         closeDrawer();
         break;
+      case R.id.nav_download_tiles:
+        showTileDownloadLayer();
```
gino-m: Naming nit: This isn't a layer from the user's perspective, it's a UI for selecting and deselecting offline satellite imagery. Also, the fact that users select tiles is for the most part an implementation detail; we could in the future allow people to choose an area instead of individual tiles. In that case, I'd probably call it something like "offline imagery selector" or similar.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -66,6 +67,7 @@
         LiveDataReactiveStreams.fromPublisher(dataRepository.getActiveProjectOnceAndStream());
     this.navigator = navigator;
     this.addFeatureClicks = PublishSubject.create();
+    this.downloadTileClicks = PublishSubject.create();
```
gino-m: When I first read this I thought it was referring to "toggle offline tile clicks", but it appears it's the click to show the offline imagery UI. It might be more obvious if we called it something like "showOfflineImagerySelectorClicks()". 

Ideally I think we may want to actually have the selector be a separate View and ViewModel so we don't need to worry about interactions with other features on the map UI (click feature, location lock, etc). Any thoughts on this?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -138,6 +138,10 @@ private void onMapReady(MapAdapter map) {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCenter()));
     enableLocationLockBtn();
+    homeScreenViewModel
+        .getDownloadTileClicks()
+        .as(autoDisposable(this))
+        .subscribe(__ -> map.renderJsonLayer());
```
gino-m: Similarly, here we're "showing the offline basemap/imagery selector". Technically rendering of this GeoJSON layer is actually done by Maps API, and is an implementation detail.

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -132,6 +161,56 @@ public void moveCamera(Point point, float zoomLevel) {
...
+          Log.d(TAG, "Unable to read JSON.", e);
+          e.printStackTrace();
+        }
+      } catch (IOException e) {
+        Log.d(TAG, "geojson file not found.", e);
+      }
+    }
+  }
+
+  private void onJsonFeatureClick(com.google.maps.android.data.Feature feature) {
```
gino-m: Knowledge of the Ground-specific VOs don't really belong here since this class is lower level; it just exposes functionality of Maps API or whatever implementation we use in a generic way, so it covers only mapping concepts like Points, Lines, Polygons, etc. GeoJSON may fix in this context, but we'd ideally find a way to generalize it to not scope-creep the interface needed to be implemented for new map providers.

gnd/src/main/res/values/strings.xml
```diff
@@ -16,128 +16,129 @@
...
-    VIEW FEATURE DATA
-  -->
-
-  <!-- Shown in record overflow menu to open popup dialog to edit feature metadata. -->
-  <string name="edit_record">Edit record</string>
-  <!-- Shown in record overflow menu to delete a record. -->
-  <string name="delete_record">Delete record</string>
-  <!-- Confirmation shown when deleting a record (TBD). -->
-  <string name="delete_record_confirmation">
+    <!-- Application name. Do not translate. -->
```
gino-m: Any idea why whitespace changed here? Was it committed w/the wrong white space? (we really need to add auto lint checks to PRs!)

scolsen: Ah, I ran format on the file after modifying it--that's the reason it changed, I guess it had "incorrect" formatting at some point.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
scolsen: That makes sense, I'll rename it!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
scolsen: I think keeping them separate would definitely be a better design. I can give this a shot.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
scolsen: Good point. I'll excise the JSON from my names.

gino-m:
Are you planning to abandon this in favor of the separate PRs we discussed? If so kindly mark WIP or close. Thanks!

PR #161, opened 2019-08-22
Offline Imagery: Add tile streams to data store

scolsen:
This change adds methods for accessing all tiles from the local data store.
Additionally, I've updated the tile DAO so that we can fetch all tiles
from the DB.


scolsen:
Small change to enable us to access db tiles from the local data store.

PR #164, opened 2019-09-06
Offline Imagery: Update File Download Worker

scolsen:
I've made necessary updates to ensure the file download worker can manage the entirety of the tile downloading and db writing/mgmt flow. 

After this change is in, we only need to pass tile IDs to the worker from the view model, rather than have to worry about managing tiles.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
@@ -24,6 +24,9 @@ public abstract class Tile {
   public static String pathFromId(String tileId) {
     // Tile ids are stored as x-y-z. Paths must be z-x-y.mbtiles.
     // TODO: Convert tile ids to paths in a less restrictive and less hacky manner.
+    // TODO: Move this method to a more appropriate home? We need to perform (and possibly will no
+    // matter where the tiles are stored) translation between the tile ID and the file path of the
+    // corresponding tile source in remote storage/wherever we pull the source tile from.
     String[] fields = tileId.replaceAll("[()]", "").split(", ");
     String filename = fields[2] + "-" + fields[0] + "-" + fields[1];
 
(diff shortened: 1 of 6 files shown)


gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
```diff
@@ -4,25 +4,42 @@
...
+
+  public static Builder newBuilder() {
+    return new AutoValue_Tile.Builder();
+  }
+
+  public static String pathFromId(String tileId) {
+    // Tile ids are stored as x-y-z. Paths must be z-x-y.mbtiles.
+    // TODO: Convert tile ids to paths in a less restrictive and less hacky manner.
+    String[] fields = tileId.replaceAll("[()]", "").split(", ");
+    String filename = fields[2] + "-" + fields[0] + "-" + fields[1];
```
gino-m: Aren't we keeping a mapping between ID and filename in the DB? Or is that still TBD?

scolsen: We are, but at the moment we never actually set the path for a given tile anywhere--this may not be the right place for this method, I wasn't sure where we should be doing this sort of work.

Our Ids are in X-Y-Z format, while our source tile files in the storage api are stored as Z-X-Y.mbtiles, so we need to do this transformation somewhere--would it be more appropriate to have it in the download worker, keeping the filename independent of the source file name? (at the moment, I'm just saving the files with the same name as they have in storage).

gino-m: I don't have the whole flow paged into memory anymore, but I assume this is because the TileProvider needs an easy way to resolve the filename of the tiles without necessarily going to the DB? If so I think this is fine, maybe just add a TODO to think about a better place for the method in the future (not sure where is best).

scolsen: Sounds good! Yep, once things settle down a bit and we get to cleaning some of this stuff up we can sit down and go over the flow and where precisely things should live. Made the todo more specfi, thanks!

PR #165, opened 2019-09-06
Offline Imagery: All the Rest

scolsen:
This PR includes pretty much all the remaining functionality we need to implement to get an acceptable demo of the offline imagery feature up and running. 

**Please note this PR relies on #163 and #164, once those are in, the set of changes will shrink significantly.**

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
@@ -114,4 +114,7 @@ public interface LocalDataStore {
 
   /** Returns the tile with the specified id from the local data store, if found. */
   Maybe<Tile> getTile(String tileId);
+
+  /** Deletes the tile with the specified id from the local data store, if found.*/
+  Maybe<Integer> deleteTile(String tileId);
 }
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -225,4 +225,9 @@ public class RoomLocalDataStore implements LocalDataStore {
   public Maybe<Tile> getTile(String tileId) {
     return db.tileDao().findById(tileId).map(TileEntity::toTile);
   }
+
+  @Override
+  public Maybe<Integer> deleteTile(String tileId) {
+    return db.tileDao().findById(tileId).flatMap(tileEntity -> db.tileDao().deleteTile(tileEntity));
+  }
 }
--- a/gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
@@ -19,6 +19,7 @@ import com.google.android.gnd.ui.map.Extent;
 import com.google.android.gnd.ui.map.ExtentSelector;
 import com.google.android.gnd.ui.map.MapProvider;
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
+import com.google.common.collect.ImmutableSet;
 
 import javax.inject.Inject;
 
@@ -53,7 +54,6 @@ public class BasemapSelectorFragment extends AbstractFragment {
   @BindView(R.id.basemap_controls_layout)
   ViewGroup basemapControls;
 
-
   @Override
   public void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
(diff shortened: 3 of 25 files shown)


scolsen:
Alrighty, after this change, everything should more or less work! (with a rather ugly UI). 

Two things to note, for now:

- Haven't implemented storage space checking yet.
- The current overlays seem to be pretty lossy--there's a lot of black space when rendering them. I think this is due to zoom level support. The storage I'm pulling from only supports up to level 7 (max). We can either: 
1. restrict zoom levels at which the overlays are rendered for a smoother experience. 
2. Pull from a source that has higher zoom levels (do we have 13s available?)

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -114,4 +114,7 @@
 
   /** Returns the tile with the specified id from the local data store, if found. */
   Maybe<Tile> getTile(String tileId);
+
+  /** Deletes the tile with the specified id from the local data store, if found.*/
+  Maybe<Integer> deleteTile(String tileId);
```
scolsen: Remove this. I originally deleted tiles from the DB upon file removal, but it's actually much simpler to just retain them and update their state.

gino-m: Did you mean to add this as a TODO?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -225,4 +225,9 @@ public Completable insertOrUpdateTile(Tile tile) {
   public Maybe<Tile> getTile(String tileId) {
     return db.tileDao().findById(tileId).map(TileEntity::toTile);
   }
+
+  @Override
+  public Maybe<Integer> deleteTile(String tileId) {
+    return db.tileDao().findById(tileId).flatMap(tileEntity -> db.tileDao().deleteTile(tileEntity));
+  }
```
gino-m: Do we care about the number of tiles deleted? If so, why `Maybe`? And if not, we should just return a Completable?

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
```diff
@@ -50,6 +66,63 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+        .getTiles()
+        .observe(
+            this,
+            tiles -> {
+              map.updateExtentSelections(
+                  stream(tiles).map(Extent::fromTile).collect(toImmutableSet()));
+            });
+
+    viewModel
+        .getDownloadedExtents()
```
gino-m: Why separate the different extents by state? I'd think the ViewModel would just observe all the extends, and sync the states whenever any of them changes?

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
```diff
@@ -50,6 +66,63 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+    viewModel.getExtentsPendingRemoval().observe(this, extents -> {
+      if (extents.isEmpty()) {
+        removeButton.setBackgroundTintList(ColorStateList.valueOf(getResources().getColor(R.color.colorGrey600)));
+      } else {
+        removeButton.setBackgroundTintList(ColorStateList.valueOf(getResources().getColor(R.color.colorAlert)));
+      }
+    });
+
+    downloadButton.setOnClickListener(__ -> viewModel.downloadExtents());
+    removeButton.setOnClickListener(__ -> viewModel.removeExtents());
```
gino-m: I understood we'd have one button "Apply" that would apply any changes, including tiles that were toggled off (remove) and tiles that were toggled on (download). Is that not how it works here?

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorViewModel.java
```diff
@@ -12,8 +28,108 @@
...
+  private static final String TAG = BasemapSelectorViewModel.class.getName();
+
+  private final Flowable<ImmutableSet<Tile>> tilesStream;
+  private final LiveData<ImmutableSet<Tile>> tiles;
+  private HashSet<String> extentsPendingDownload = new HashSet<>();
+  private PublishSubject<HashSet<String>> extentsPendingDownloadStream = PublishSubject.create();
+  private HashSet<String> extentsPendingRemoval = new HashSet<>();
+  private PublishSubject<HashSet<String>> extentsPendingRemovalStream = PublishSubject.create();
+  private LiveData<String> downloadedExtents;
+  private PublishSubject<String> downloadedExtentsSubject = PublishSubject.create();
```
gino-m: Same comment here. I would expect the view model to expose all extents as a single stream.

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
scolsen: Yep much simpler--haphazard midnight impl when scrambling for time. I've updated this according to your suggestion thanks!

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorFragment.java
scolsen: Done!

scolsen: And yes, originally I had two buttons, one for each op.

gnd/src/main/java/com/google/android/gnd/ui/basemapselector/BasemapSelectorViewModel.java
scolsen: Done!

scolsen:
I've updated the approach here, it's a good bit cleaner--though can still use some cleaning up to make it easily grokkable. @gino-m , please take another look when you can!

I need to delete the tile DB deletion method--but let's tackle that later

If you pull and build this you can check out how the whole experience works.

PR #176, opened 2019-09-12
Use icon color in new feature type default style 

### LINKED ISSUE #171: [Map] Respect default style (color) for markers
Thanks @daoyuT for adding this to the web UI.

gino-m:
@DaoyuT, FYI. Fixes #171.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/model/layer/Style.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/layer/Style.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


gnd/src/main/java/com/google/android/gnd/model/layer/Style.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.model.layer;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Style {
```
scolsen: Very nice, leaving a comment here to capture my thoughts for posterity--It'd be nice if we can make this generic enough to support the other cases where we'll need styling too (e.g. tile selections)

PR #183, opened 2019-09-12
Offline Imagery: Remove offline maps from menu

scolsen:
Temporarily removing offline maps from the nav menu, since the
functionality is disabled.

Not sure what your plans are for this, but in case you just want to completely nix the functionality from this build, this PR temporarily removes the offline maps menu option!


gino-m:
Thanks for the fix.

Issue #189, opened 2019-10-03
[Code health] Use data binding in all views


gino-m:
TODO:
- [x] EditRecordFragment
- [x] RecordListAdapter
- [x] FeatureSheetFragment
- [ ] TwoLineHeader

gino-m:
@shobhitagarwal1612 Could you kindly add/update any I missed in the description here?

gino-m:
> @shobhitagarwal1612 Could you kindly add/update any I missed in the description here?

Are these all the ones that are pending?

shobhitagarwal1612:
Sorry for the late reply. 

> @shobhitagarwal1612 Are these all the ones that are pending?

Yes

gino-m:
> Sorry for the late reply.
> 
> > @shobhitagarwal1612 Are these all the ones that are pending?
> 
> Yes

Thanks for verifying!

PR #190, opened 2019-10-15
Use DataBinding in EditRecordFragment

shobhitagarwal1612:
Related to #189


gnd/src/main/java/com/google/android/gnd/ui/editrecord/EditRecordFragment.java
```diff
@@ -127,7 +120,7 @@ public void onStart() {
   private void onRecordChange(Persistable<Record> record) {
     switch (record.state()) {
       case LOADING:
-        saveRecordButton.setVisibility(View.GONE);
+        // do nothing
```
gino-m: Nit: Capitalize first letter, finish with "."

Also, please mention why we do nothing. I assume it's because we're handling this in the ViewModel, and we only handle recordChange in both places because some logic still remains in the Fragment?

PR #197, opened 2019-11-07
Remove/Revert BasemapSelectorMapAdapter

scolsen:
This PR removes the BasemapSelectorMapAdapter and its corresponding extent selection interface since maintaining two adapters proved problematic. 

Note: Redux, forgot to set my email on my new machine 😅


gnd/src/main/java/com/google/android/gnd/ui/map/MapProvider.java
```diff
@@ -62,5 +60,7 @@
     void enableCurrentLocationIndicator();
 
     void updateMarkers(ImmutableSet<Feature> features);
+
+    void renderJsonLayer();
```
gino-m: Now that we're switching to "productionize" phase, please be sure to add Javadoc for all new public methods like this one.

gino-m: Actually did you mean to add this now? I think for the adapter we'd want to have something that would allow the caller to add/remove/show/hide JSON layers, but the usage will probably only be clear once we implement the UI. Can we remove this interface method for now?

scolsen: You're totally right! We don't need it. I just did a plain old revert to undo the changes, so I guess this method was already in prior to the commit I reverted. I've removed it.

scolsen:
@gino-m should be good to go now...except GCB is failing. The summary only says `NOT_FOUND`--any ideas?

PR #207, opened 2019-11-17
[Code health] Lint checks

### LINKED ISSUE #206: [Feature] Android lint checks
The lint tool checks for potential bugs and optimization improvements for correctness, security, performance, usability, accessibility, and internationalization.

- [ ] Add gradle plugin

- [ ] Integrate into GCB

- [ ] Generate reports

shobhitagarwal1612:
Closes #206

### Future plan
Convert warnings to errors (one by one) for enforcing them with the cloud build


shobhitagarwal1612:
> Will this also run in Android Studio? 

Yes. Command added in comments,
`./gradlew lintDebug`

> Is there a way to get it trigger on Gradle build?

I have already added this to cloud build. Is that what you are referring to?

gino-m:
> I have already added this to cloud build step. Is that what you are referring to?

I was wondering if we would be able to see these lint warnings directly in Android Studio, either in real time while editing, or automatically on build.

gino-m:
P.S. Merged! Thanks @shobhitagarwal1612! One question pending above for follow-up.

shobhitagarwal1612:
There is a way to automatically run the lint check with every `assembleDebug` task, i.e. building apk (by adding dependency in build.gradle). But that will significantly increase the build time and effect development time. Same goes for rest of the code health plugins too.

But this can be executed manually in IDE or command-line apart from running automatically on cloud build.

What are your thoughts?

gino-m:
In that case I think it's ok if they need to be run manually. Would you mind adding a line or two to the README.md on how to run these manually from the IDE?

shobhitagarwal1612:
Sure. I'll update the README

Issue #185, opened 2019-09-15
[Add observation] Observations sometimes do not appear in list after saving


gino-m:
This occurred online and offline. The record appeared in the list after closing and reopening the feature sheet.

The bug is intermittent - after reloading the app it no longer occurs.

gino-m:
`maybeSyncFirst()` in `DataRepository` is timing out because it's actually implemented incorrectly; we wait for the first Record event before syncing local changes, which never occurs if there are no records, and occurs multiple times if there's more than one. More likely a `List` of Records should be returned, or we should only sync once when the sheet is opened and drop real-time updates.

gino-m:
Verified at HEAD, both when online and offline.

PR #215, opened 2019-11-19
Fix broken record list, Rx schedulers, and cleanup

### LINKED ISSUE #213: [Code health] Always subscribeOn() and observeOn() at point of subscribe()

gino-m:
Work towards #185 and some cleanup (fixes #213) along the way.

Relevant commits:
- [x] Fix broken record list flow: 4df6ca6df5f99516ec19bc8da88fba70e63e6e46
- [x] Clean up Rx scheduling: 05d71910aded45eff6dfcba94af02f502a4a1d8b, 459421bd57ea15b397310bb2496af749bab05297, a686b73167ba9f2f5da31ea8fc67ac2d9432bc87, c0777748c199861b0452a6c1226b7bf6d689c53d, bd4e783671fc84e5f5f900cdbe162956a9f94f15
- [x] Rename "record summary: a7b5d97bb7798fcd1c3254949d62bdf5c5219032

A few observations re Rx streams:
* `subscribeOn()` applies to entire stream, regardless of where it's declared, except..
* `flatMap()` doesn't propagate the thread to the inner observable(s), so we need to declare it ourselves. For this reason I moved all instances of `subscribeOn` closer to background work would be done (upstream in datastore classes).
* `LiveData` always observes on the main thread, so `observeOn(AndroidThreads.main())` is not needed when wrapping `Flowable` in LiveData with `LiveDataReactiveStreams.fromPublisher()`


gino-m:
@scolsen Tried to write up some guidelines here. Please lmk what you think:
https://github.com/google/ground-android/wiki/RxJava-Usage-Guidelines#threading

gino-m:
@shobhitagarwal1612 How can I see actual root cause of the PMD failure? https://pantheon.corp.google.com/cloud-build/builds/09dd240f-98e0-40ef-a4e9-50a85772af43;step=3?project=ground-android-gcb

gino-m:
@shobhitagarwal1612 How can I see actual root cause of the PMD failure? https://console.cloud.google.com/cloud-build/builds/09dd240f-98e0-40ef-a4e9-50a85772af43;step=3?project=ground-android-gcb

shobhitagarwal1612:
The problem is, GCB doesn't collect artifacts (which includes the html reports) if the build **fails**. 
Treating every error as warning will do the trick but then the build will pass even with code styling or other code health issues. I was hoping it would at least show the error in the stacktrace, but it didn't.

I'll try to come up with a better solution. Till then can you please run the check locally and fix the PMD issues?
Command: `./gradlew pmd`
Reports are saved under: `/gnd/build/reports/pmd/pmd.html`

gino-m:
> I'll try to come up with a better solution. Till then can you please run the check locally and fix the PMD issues?
> Command: `./gradlew pmd`
> Reports are saved under: `/gnd/build/reports/pmd/pmd.html`

If it's difficult to get more diagnostic info into output, running locally is sufficient provided it's mentioned somewhere in the docs. Thanks for taking a look!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -83,51 +84,65 @@ public Completable applyAndEnqueue(FeatureMutation mutation) {
...
   @Transaction
   @Override
   public Completable updateMutations(ImmutableList<Mutation> mutations) {
     return db.featureMutationDao()
         .updateAll(toFeatureMutationEntities(mutations))
-        .andThen(db.recordMutationDao().updateAll(toRecordMutationEntities(mutations)));
+        .andThen(
+            db.recordMutationDao()
+                .updateAll(toRecordMutationEntities(mutations))
+                .subscribeOn(Schedulers.io()))
```
scolsen: Do we know if this inner call to `subscribeOn` is necessary? Since `andThen` passes subscription to its argument, I'd assume it honors the scheduler too.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -158,82 +153,37 @@ private void onProjectLoaded(Project project) {
...
-          // TODO: Propagate this to the user so we can show "network unavailable" message also
-          // when network became unavailable mid-fetch .
-          .doOnError(t -> Log.d(TAG, "Record sync timed out"))
-          .onErrorComplete();
-    } else {
-      Log.d(TAG, "No network; skipping remote sync");
-      return Completable.complete();
-    }
+  private Single<ImmutableList<Record>> getRecords(Feature feature, String formId) {
+    Completable remoteSync =
```
scolsen: No more network check? Or did we move it elsewhere?

Seems like it's not strictly necessary from a functional perspective, but I wonder if we want to differentiate between timeouts and network absence in case we want to propagate that info upwards?

gino-m: Good catch. When we disable local caching via Firebase, it will fail fast and this request will fall back to the local db. We can add def propagation of this state once needed.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
gino-m: Unfortunately it doesn't; the inner stream is created on a separate thread as with flatMap (I tested this).

scolsen: Dang! gtk thanks!

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
scolsen: sounds good!

PR #224, opened 2019-11-20
[Bug] Fix a regression caused during code health refactor

shobhitagarwal1612:
Revert changes made in https://github.com/google/ground-android/pull/221/commits/658eeed47895537c00e1ea61c1309b06d40a24eb

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/common/BottomSheetBehavior.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/common/BottomSheetBehavior.java
@@ -406,7 +406,7 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
       return;
     }
     View scrollingChild = nestedScrollingChildRef.get();
-    if (target.equals(scrollingChild)) {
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/ui/common/BottomSheetBehavior.java
```diff
@@ -410,7 +410,7 @@ public void onNestedPreScroll(
       return;
     }
     View scrollingChild = nestedScrollingChildRef.get();
-    if (target.equals(scrollingChild)) {
+    if (!target.equals(scrollingChild)) {
```
gino-m: Is `equals()` the same as `==` for views? We really wanted to ensure it was the same instance. Is this correct?

shobhitagarwal1612: Yes, it has the same effect.

gino-m: Sgtm!

PR #226, opened 2019-11-20
[Code health] Spotbugs plugin

shobhitagarwal1612:
Add [spotbugs](https://github.com/spotbugs/spotbugs) plugin for finding common bugs in Java classes. This can also be run manually like other plugins in the IDE or command-line.

- [x] Added to cloud build

- [x] Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,12 +62,12 @@ to the base repository using a pull request.
 1. Run code health checks locally and fix any errors.
 
    1. Using Command-line:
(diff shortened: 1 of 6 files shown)


shobhitagarwal1612:
@gino-m This is the last plugin that I had in mind. You might notice that there is an overlap between some of the tools. But each one of them have some unique feature. So, the best configuration is to use all 4 plugins.

PR #232, opened 2019-11-23
[Code health] Enable check for unused imports

shobhitagarwal1612:
Adds PMD check for detecting unused imports in classes.


gnd/src/main/java/com/google/android/gnd/ui/common/SharedViewModel.java
```diff
@@ -16,7 +16,7 @@
 
 package com.google.android.gnd.ui.common;
 
-import androidx.fragment.app.Fragment;
+import androidx.fragment.app.Fragment; // NOPMD
```
shobhitagarwal1612: This import is used in the JavaDoc for code navigation in Android Studio.
(Suppressing this warning)

Alternatively, @SuppressWarnings() can also be used for disabling 1 or more checks for an entire class or method.

PR #241, opened 2019-11-26
[Offline areas] Area data model

scolsen:
This PR adds a model class and DB classes for representing areas. Areas are sets of downloadable tiles of offline imagery that form a contiguous geodesic rectangle. They are uniquely identifiable by their Northwestern and southeastern bounds. 

Areas may also have differing zoom levels, but we don't account for this yet.


gnd/src/main/java/com/google/android/gnd/model/basemap/Area.java
```diff
@@ -0,0 +1,64 @@
...
+        public abstract Builder setBounds(LatLngBounds bounds);
+
+        public abstract Builder setState(State state);
+
+        abstract Builder setId(String id);
+
+        abstract Area autoBuild();
+
+        public Area build() {
+            setId(getBounds().toString());
```
gino-m: Rather than hard code this into the builder, it'd be simpler and more consistent to let the caller call setId(), generating a UUID using our UUidGenerator class.

scolsen: A much better approach, thanks! 

We'll expose setId and handle generating a uuid in the datarepo. Since the bounds are no longer used as an identifier, I've added an additional field to the entity to persist them.

scolsen:
@gino-m switched over to using sweet, sweet uuids 😎 (though we don't use them yet in this PR).

Mind taking another look?

gnd/src/main/java/com/google/android/gnd/model/basemap/Area.java
```diff
@@ -19,46 +19,39 @@
 import com.google.auto.value.AutoValue;
 
 /**
- * An area is a contiguous set of tiles that form a geodesic rectangle.
- * They are uniquely identifiable by their NE and SW bounds (or NW and SE).
+ * An area is a contiguous set of tiles that form a geodesic rectangle. They are uniquely
```
gino-m: Since we won't (and shouldn't?) rely on this as the PK maybe it's not worth mentioning here?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
```diff
@@ -20,22 +20,23 @@
...
 @AutoValue
 @Entity(tableName = "area")
 public abstract class AreaEntity {
   public static Area toArea(AreaEntity areaEntity) {
     Area.Builder area =
         Area.newBuilder()
             .setBounds(boundsFromString(areaEntity.getId()))
-            .setState(toAreaState(areaEntity.getState()));
+            .setState(toAreaState(areaEntity.getState()))
+            .setId(areaEntity.getId());
```
gino-m: Nit: Doesn't really matter, but I'd expect setId() to be called first since it's the PK.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
```diff
@@ -54,16 +55,15 @@ public static Area toArea(AreaEntity areaEntity) {
     }
   }
 
-  private static LatLngBounds boundsFromString(String id) {
-    String[] coords = id.split(",");
-    LatLng NE = new LatLng(Double.parseDouble(coords[0]), Double.parseDouble(coords[1]));
-    LatLng SW = new LatLng(Double.parseDouble(coords[2]), Double.parseDouble(coords[3]));
-    return new LatLngBounds(NE, SW);
-  }
+  // The Bounds conversion methods bellow double as type converters so that we can query areas
```
gino-m: This comment can be removed

scolsen:
Time for another round! 🥊 

I've removed the latlngbounds converter and switched to using explicit doubles as we discussed. Tbh, I actually prefer the other approach. I feel like this approach presents more opportunity for things to get messed up in the future. 

Using the converter, we can more tightly couple the Area representation to the DB representation--here, there's more that we need to explicitly remember. Of course, it may just be that there's a better way to use this approach that I haven't implemented!

gnd/src/main/java/com/google/android/gnd/model/basemap/Area.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.google.android.gms.maps.model.LatLngBounds;
+import com.google.auto.value.AutoValue;
+
+/**
+ * An area is a contiguous set of tiles that form a geodesic rectangle.
+ */
+@AutoValue
+public abstract class Area {
+  public static Builder newBuilder() {
```
gino-m: Nit: We've been putting `newBuilder` above  the Builder class definition before.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaDao.java
```diff
@@ -0,0 +1,38 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import java.util.List;
+
+/** Provides read/write operations for writing {@link AreaEntity} to the local db. */
+@Dao
+public interface AreaDao {
+  @Insert(onConflict = OnConflictStrategy.ABORT)
+  Completable insertOrUpdate(AreaEntity areaEntity);
```
gino-m: This technically only inserts (iiuc it will abore if the record exists based on the conflict stategy)

Either rename the method to insert() if that's what you really want,, or extend `BaseDao<AreaEntity>` to inherit insertOrUpdate().

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
```diff
@@ -0,0 +1,149 @@
...
+import androidx.room.Entity;
+import androidx.room.PrimaryKey;
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.LatLngBounds;
+import com.google.android.gnd.model.basemap.Area;
+import com.google.auto.value.AutoValue;
+
+/** Represents a {@link Area} in the local data store. */
+@AutoValue
+@Entity(tableName = "area")
```
gino-m: Naming nit: Looking at the db and the class name in isolation "Area" may be too general - should we call them `OfflineAreaEntity` and `offline_area` instead for clarity?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
```diff
@@ -0,0 +1,149 @@
...
+public abstract class AreaEntity {
+  public static Area toArea(AreaEntity areaEntity) {
+    LatLng northEast = new LatLng(areaEntity.getNorth(), areaEntity.getEast());
+    LatLng southWest = new LatLng(areaEntity.getSouth(), areaEntity.getWest());
+    LatLngBounds bounds = new LatLngBounds(northEast, southWest);
+
+    Area.Builder area =
```
gino-m: Nit: We don't need the intermittent variable `area` here.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @AutoValue.CopyAnnotations
+  @NonNull
+  @ColumnInfo(name = "state")
+  public abstract AreaEntityState getState();
+
+  @AutoValue.CopyAnnotations
+  @NonNull
+  @ColumnInfo(name = "north")
+  public abstract Double getNorth();
```
gino-m: Since these are not null, can we remove `@NonNull` and change them to `double` here and in the builder below?

gino-m:
> I've removed the latlngbounds converter and switched to using explicit doubles as we discussed. Tbh, I actually prefer the other approach. I feel like this approach presents more opportunity for things to get messed up in the future.

You may be right, I'm just having trouble imagining what that might be at this late-ish hour - let's discuss tomorrow, curious to hear your perspective.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
shobhitagarwal1612: Also, can we move these static methods after the variables of the entity are declared?

gnd/src/main/java/com/google/android/gnd/model/basemap/Area.java
scolsen: Gotcha--whatever formatting settings I have want to push this up to the top. 

I feel like I may still not have the right .idea files or something--are those checked into git?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaDao.java
scolsen: Yep this should've been REPLACE -- Thanks! I've inherited the baseDAO, much nicer :)

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
scolsen: sg!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
scolsen: Yes! My formatter keeps moving stuff around whenever I run it :/

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AreaEntity.java
scolsen: 👍

scolsen:
> LGTM! There are lint errors that need to be fixed before we can merge. You can see them here:
> https://console.cloud.google.com/cloud-build/builds/c64c5af9-efe7-4867-bce4-6f4ae7c885c6;step=7?project=ground-android-gcb
> 
> Are you using google-java-format?

Just pushed some fixes for this...there may be a few indentation errors left over but hopefully this eliminated most if not all of the other lint issues.

PR #238, opened 2019-11-25
[Add/edit observation] Major cleanup of view model and view

gino-m:
Fixes #242.

PR #246, opened 2019-12-01
[Offline Support] Persist projects in local db

### LINKED ISSUE #148: [Offline support] Persist projects in local db
- [ ] Join project - store in local db
- [ ] Leave project - delete from local db
- [ ] Activate project - load project from local db

To do this, we'll need to add a model and related DAOs for project entities (Form, Observation, etc) in the local db:
https://github.com/google/ground-android/tree/master/gnd/src/main/java/com/google/android/gnd/persistence/local/room

We may want to tackle #143 before introducing these new classes.

The methods to load/save/delete the project data would then be added to:
https://github.com/google/ground-android/blob/master/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java

We can then update the data repository class to load from local db when offline when activating the project (consider #25 before this step):
https://github.com/google/ground-android/blob/master/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java

Once this is done, we can switch off local persistence in Firebase API, allowing us full control and transparency over sync lifecycle.

shobhitagarwal1612:
Closes #148 

- [x]  Join project - store in local db

- [x]  Leave project - delete from local db

- [x]  Activate project - load project from local db

- [x] Offline mode switch in navigation drawer

P.S. I also integrated [Stetho](http://facebook.github.io/stetho/) for debugging DB. Let me know what you guys think of it and whether it should be removed or not from this PR?


gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
```diff
@@ -61,7 +61,7 @@
     disposeOnClear(
         dataRepository
             .getProjectSummaries(user)
-            .subscribe(projectSummaries::setValue, this::onProjectSummariesError));
+            .subscribe(projectSummaries::postValue, this::onProjectSummariesError));
```
shobhitagarwal1612: This had to be modified because the db calls are scheduled in background thread. `postValue` allows async calls while `setValue` doesn't.

gnd/src/main/java/com/google/android/gnd/GndApplication.java
```diff
@@ -23,6 +23,7 @@
 import androidx.work.Configuration;
 import androidx.work.WorkManager;
 import com.akaita.java.rxjava2debug.RxJava2Debug;
+import com.facebook.stetho.Stetho;
```
scolsen: Sorry to be a pain, but is it possible to split out the Stetho changes into a separate PR?

Just keeps the history slightly cleaner; if we need to revert our use of Stetho or our project persistence we can do so separately.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
```diff
@@ -46,17 +47,19 @@ public String getLastActiveProjectId() {
...
   /** Removes the last active project id in the local value store. */
   public void clearLastActiveProjectId() {
-    preferences
-      .edit()
-      .remove(ACTIVE_PROJECT_ID_KEY)
-      .apply();
+    preferences.edit().remove(ACTIVE_PROJECT_ID_KEY).apply();
+  }
+
+  public boolean isOfflineModeEnabled() {
```
scolsen: Is enabling offline mode explicitly part of the intended design? On the offline imagery side of things, I believe our decision was to render offline imagery automatically, without the user having to specify any sort of offline mode as off or on. I might be mistaken about this, though.

It seems odd to me to maintain to separate approaches to our handling of the offline modality. 

@gino-m

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -268,5 +285,17 @@ public void clearActiveProject() {
     cache.clearActiveProject();
     localValueStore.clearLastActiveProjectId();
     activeProject.onNext(Persistable.notLoaded());
+    localDataStore
+        .getLastActiveProject()
+        .doOnSuccess(project -> localDataStore.removeProject(project).subscribe())
```
scolsen: Instead of using doOnSuccesss here, I believe you should be able to use flat map keeping one level of streams instead of nesting (since the Maybe emits a project):

```java
.getLastActiveProject()
.flatMapCompleteable(localDataStore::removeProject)
.subscribe();
```

Or roughly that http://reactivex.io/RxJava/javadoc/io/reactivex/Maybe.html#flatMapCompletable-io.reactivex.functions.Function-

Also, just subscribing here seems fishy to me, but it's been a while since I've used Rx so it may be totally fine.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -130,12 +131,20 @@ private void onProjectLoaded(Project project) {
     cache.setActiveProject(project);
     activeProject.onNext(Persistable.loaded(project));
     localValueStore.setLastActiveProjectId(project.getId());
+    localDataStore.activateProject(project).subscribeOn(Schedulers.io()).subscribe();
+  }
+
+  private Single<List<Project>> loadProjects(User user) {
+    if (isOfflineModeEnabled()) {
```
scolsen: Related to the previous comment about explicit offline checking. 

I believe our approach was to try to connect to remotes by default and if that doesn't succeed, fallback to local stores, without really making this reliant on the user (while still informing them)--in other words, offline enabled by default and by design.

This might have changed though! Just a heads up based on my prior understanding. I defer to @gino-m

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
shobhitagarwal1612: So sorry for the confusion.

No, it's not a part of the design. I added it as a debug-only feature while working on the issue. It's easier to test the offline behavior for a specific target without cutting-off the network connectivity for all other elements (which still rely on internet!)

My plan is to remove this added UI feature and replace the usages of `ifOfflineModeEnabled()` with `isOffline()` later, i.e. when all modules support offline mode.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
shobhitagarwal1612: Please see https://github.com/google/ground-android/pull/246#discussion_r353000972

gnd/src/main/java/com/google/android/gnd/GndApplication.java
```diff
@@ -45,6 +46,7 @@ public void onCreate() {
     if (BuildConfig.DEBUG) {
       Log.d(TAG, "DEBUG build config active; enabling debug tooling");
       setStrictMode();
+      Stetho.initializeWithDefaults(this);
```
gino-m: Please add brief comment before this explaining what this is.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
gino-m: Correct. The user is offline implicitly when they don't have a network connection. Airplane mode can be a proxy for this, so we don't need a UI to force them offline.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ProjectDao.java
```diff
@@ -0,0 +1,41 @@
...
+import androidx.room.Insert;
+import androidx.room.OnConflictStrategy;
+import androidx.room.Query;
+import io.reactivex.Completable;
+import io.reactivex.Maybe;
+import java.util.List;
+
+@Dao
+public interface ProjectDao {
+  @Insert(onConflict = OnConflictStrategy.REPLACE)
```
gino-m: I recently discovered that one replace the record is deleted then inserted, triggering cascading deletes of all related data! Please copy and paste this instead with a TODO to refactor once I merge the base class:

https://github.com/google/ground-android/pull/238/files#diff-1c91753f181f2553b29d35ecba14c670R30-R38

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ProjectDao.java
```diff
@@ -0,0 +1,41 @@
...
+  Completable insertOrUpdate(ProjectEntity project);
+
+  @Delete
+  Completable deleteProject(ProjectEntity projectEntity);
+
+  @Query("SELECT * FROM project")
+  Maybe<List<ProjectEntity>> findAll();
```
gino-m: Can we use `Single` here instead? If there are no records we should get an empty list.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ProjectEntity.java
```diff
@@ -0,0 +1,99 @@
...
+  public abstract String getTitle();
+
+  @CopyAnnotations
+  @Nullable
+  @ColumnInfo(name = "description")
+  public abstract String getDescription();
+
+  @CopyAnnotations
+  @ColumnInfo(name = "is_active")
+  public abstract boolean isActive();
```
gino-m: We already store the active project in the Preference object, so it would be redundant here and would require extra logic to update the Project on activation. Or is there another reason to have it here?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -65,6 +66,35 @@ public RoomLocalDataStore(GndApplication app) {
...
+    return db.projectDao()
+        .findAll()
+        .map(list -> stream(list).map(ProjectEntity::toProject).collect(toList()))
+        .toSingle()
+        .subscribeOn(Schedulers.io());
+  }
+
+  @Override
+  public Completable activateProject(Project project) {
+    return db.projectDao().insertOrUpdate(ProjectEntity.fromProject(project, true));
```
gino-m: Please add `.subscribeOn(Schedulers.io())`

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: Nope, still the same!

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: +1.  The call the subscribe() would generate a new subscription which would need to be kept and properly disposed of. By chaining streams this is handled for you.

However, since this is an async operation, you'd then need to return the stream to the caller so that it can subscribe() and dispose properly as part of the project activation stream.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
gino-m: Afaik, Project and related entities like Layer and Form are the last ones that aren't yet serialized locally for offline use. In that case would it make sense to remove it now?

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -249,13 +258,21 @@ public Completable saveFeature(Feature feature) {
         .andThen(dataSyncWorkManager.enqueueSyncWorker(feature.getId()));
   }
 
+  private Maybe<Project> loadLastActiveProject() {
+    if (isOfflineModeEnabled()) {
```
gino-m: This and `loadProjects()` would need to be updated to be lazy, i.e., logic only evaluate on subscribe. This can be achieved by returning a single stream as we do for `loadRecords`:

https://github.com/google/ground-android/blob/55b3ebcb358b6e51d7b2741a11033d4f3c213ed4/gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java#L170-L179

In that case we rely on the remote store returning an error immediately when not available. In that case we fall back to the local data store. In this case, the logic might be:

1. Try to sync remote project config to local db
2. Load config from local db.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -249,13 +258,21 @@ public Completable saveFeature(Feature feature) {
...
+      return localDataStore.getLastActiveProject();
```
gino-m: As mentioned, can we just use the id in the localValueStore and fetch project by id?

gnd/src/main/java/com/google/android/gnd/ui/projectselector/ProjectSelectorViewModel.java
gino-m: Good call, thanks!

gnd/src/main/res/drawable/ic_signal_cellular_off_grey_700_24dp.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
gino-m: Please remove temp code and assets used for local debugging.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ProjectEntity.java
shobhitagarwal1612: I added this property so that we can save all available projects and activate the current project. Having this property would make querying easy for last active project. But like you said, we can do that with the id saved in prefs. So, removing it now.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
shobhitagarwal1612: Thanks!

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
shobhitagarwal1612: Done

gnd/src/main/res/drawable/ic_signal_cellular_off_grey_700_24dp.xml
shobhitagarwal1612: Done

gino-m:
@shobhitagarwal1612 I see you did some more cleanup. Please request review when ready for another look.

shobhitagarwal1612:
@gino-m I've squashed and re-ordered some commits to clean git history. This is now ready for review!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ProjectEntity.java
```diff
@@ -0,0 +1,91 @@
...
+  @CopyAnnotations
+  @Nullable
+  @ColumnInfo(name = "description")
+  public abstract String getDescription();
+
+  public static ProjectEntity fromProject(Project project) {
+    return fromProject(project, false);
+  }
+
+  public static ProjectEntity fromProject(Project project, boolean active) {
```
gino-m: Please remove unused helper with `active` arg and merge into related methods.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -248,13 +260,21 @@ public Completable saveFeature(Feature feature) {
         .andThen(dataSyncWorkManager.enqueueSyncWorker(feature.getId()));
   }
 
+  private Maybe<Project> loadLastActiveProject() {
+    String projectId = localValueStore.getLastActiveProjectId();
+    if (isOffline()) {
```
gino-m: A safer and potentially cleaner way of doing this is just falling back to local db on the error case, and setting an explicit timeout as I do for loading records. This way we also catch the case when the OS thinks the network is available, but in practice it isn't (common in environments w/intermittent connectivity).

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -264,8 +284,16 @@ public Completable saveFeature(Feature feature) {
...
+    localDataStore
+        .getProjectById(localValueStore.getLastActiveProjectId())
+        .flatMapCompletable(localDataStore::removeProject)
+        .subscribe();
     cache.clearActiveProject();
     localValueStore.clearLastActiveProjectId();
     activeProject.onNext(Persistable.notLoaded());
   }
+
+  private boolean isOffline() {
```
gino-m: If we rely on error state we can remove this. Note that Firestore API fails immediately if network connection is unavailable.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: (in either case we'll need to handle the error case)

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: See this comment for example of the logic for falling back to local db mentioned in other comments.

shobhitagarwal1612:
Need help!!!
I am loading from database if Firestore throws an exception. But the app is crashing with the below stacktrace. It appears like the exception is happening while loading data from room. Any idea how to tackle this?

```
019-12-11 20:57:41.504 19072-19072/com.google.android.gnd E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.google.android.gnd, PID: 19072
    java.lang.RuntimeException: LiveData does not handle errors. Errors from publishers should be handled upstream and propagated as state
        at androidx.lifecycle.LiveDataReactiveStreams$PublisherLiveData$LiveDataSubscriber$1.run(LiveDataReactiveStreams.java:264)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.util.NoSuchElementException: No value present
        at java8.util.Optional.orElseThrow(Optional.java:383)
        at java8.util.Optional.get(Optional.java:150)
        at com.google.android.gnd.persistence.local.room.FeatureEntity.toFeature(FeatureEntity.java:95)
        at com.google.android.gnd.persistence.local.room.RoomLocalDataStore.lambda$null$1(RoomLocalDataStore.java:117)
        at com.google.android.gnd.persistence.local.room.-$$Lambda$RoomLocalDataStore$NjFg4GviXLQ9cXRB7cVB2x1I6P0.apply(Unknown Source:4)
        at java8.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:196)
        at java8.util.DelegatingSpliterator$ConsumerDelegate.accept(DelegatingSpliterator.java:93)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1390)
        at java8.util.DelegatingSpliterator.forEachRemaining(DelegatingSpliterator.java:57)
        at java8.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
        at java8.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
        at java8.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:1269)
        at java8.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java8.util.stream.ReferencePipeline.collect(ReferencePipeline.java:602)
        at com.google.android.gnd.persistence.local.room.RoomLocalDataStore.lambda$getFeaturesOnceAndStream$2(RoomLocalDataStore.java:118)
        at com.google.android.gnd.persistence.local.room.-$$Lambda$RoomLocalDataStore$MyOWjBnAt-A-hSU5HuyLgqGl0f0.apply(Unknown Source:4)
        at io.reactivex.internal.operators.flowable.FlowableMap$MapSubscriber.onNext(FlowableMap.java:63)
        at com.akaita.java.rxjava2debug.extensions.FlowableOnAssembly$OnAssemblySubscriber.onNext(FlowableOnAssembly.java:61)
        at io.reactivex.internal.operators.flowable.FlowableFlatMapMaybe$FlatMapMaybeSubscriber.innerSuccess(FlowableFlatMapMaybe.java:175)
        at io.reactivex.internal.operators.flowable.FlowableFlatMapMaybe$FlatMapMaybeSubscriber$InnerObserver.onSuccess(FlowableFlatMapMaybe.java:397)
        at com.akaita.java.rxjava2debug.extensions.MaybeOnAssembly$OnAssemblyMaybeObserver.onSuccess(MaybeOnAssembly.java:68)
        at io.reactivex.internal.operators.maybe.MaybeFromCallable.subscribeActual(MaybeFromCallable.java:61)
        at io.reactivex.Maybe.subscribe(Maybe.java:4290)
        at com.akaita.java.rxjava2debug.extensions.MaybeOnAssemblyCallable.subscribeActual(MaybeOnAssemblyCallable.java:43)
        at io.reactivex.Maybe.subscribe(Maybe.java:4290)
        at io.reactivex.internal.operators.flowable.FlowableFlatMapMaybe$FlatMapMaybeSubscriber.onNext(FlowableFlatMapMaybe.java:132)
        at com.akaita.java.rxjava2debug.extensions.FlowableOnAssembly$OnAssemblySubscriber.onNext(FlowableOnAssembly.java:61)
        at io.reactivex.internal.operators.flowable.FlowableObserveOn$ObserveOnSubscriber.runAsync(FlowableObserveOn.java:407)
        at io.reactivex.internal.operators.flowable.FlowableObserveOn$BaseObserveOnSubscriber.run(FlowableObserveOn.java:176)
        at io.reactivex.internal.schedulers.ExecutorScheduler$ExecutorWorker$BooleanRunnable.run(ExecutorScheduler.java:288)
        at io.reactivex.internal.schedulers.ExecutorScheduler$ExecutorWorker.run(ExecutorScheduler.java:253)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
```

gino-m:
The stack track points to this line as the culprit:
https://github.com/google/ground-android/blob/55b3ebcb358b6e51d7b2741a11033d4f3c213ed4/gnd/src/main/java/com/google/android/gnd/persistence/local/room/FeatureEntity.java#L95

It appears the feature doesn't actually have its featureTypeId set in your local database. We should never call `Optional.get()` directly unless we've checked `isPresent()` to avoid RTEs like these. Normally if the field is Optional we'd use functional predicates like `.ifPresent(lambda)` or `.orElse()`. In this case, it seems featureTypeId shouldn't be `Optional` at all in FeatureEntity?  Sorry I missed this in my original review.

shobhitagarwal1612:
We are not saving `Layers` in the local database yet. So, `getLayers()` returns a empty list. Hence `featureTypeId` isn't found. Also, layer depends upon `Form`. 
We need to save all layers and form for each layer while saving the project locally. Then, create a new class and it's dao which contains relations between the entities
https://developer.android.com/reference/androidx/room/Relation.html

shobhitagarwal1612:
Considering to use TypeConverter instead of creating separate entities for each of the remaining models

gino-m:
> Currently, I have created entity and dao for 3 models: project, layer and form as well as saved their data locally (4 more to go).

I think this is the right approach. I don't think TypeConverters are intended for serializing entire graphs, only for implementing serialization of new field types. We would also paint ourselves into a corner that way, since with TypeConverters we wouldn't have an entity exposed that we can query and update independently of the Project.

shobhitagarwal1612:
@gino-m @scolsen This is ready for another review!

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -44,6 +51,37 @@
  * deltas to allow changes to not rely on prior UI state (i.e., emissions are idempotent).
  */
 public interface LocalDataStore {
+
+  /** Add an option for a multiple choice field. */
+  Completable insertOrUpdateOption(String fieldId, Option option);
```
gino-m: Why does the LocalDataStore class need to expose these? Don't we only need the granular insert or update at the DAO level?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/CardinalityTypeConverter.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.persistence.local.room;
+
+import androidx.annotation.NonNull;
+import androidx.room.TypeConverter;
+import com.google.android.gnd.model.form.MultipleChoice.Cardinality;
+
+public class CardinalityTypeConverter {
```
gino-m: Consider updating the Enum to implement [IntEnum](https://github.com/google/ground-android/search?q=IntEnum&unscoped_q=IntEnum) instead of defining a custom converter.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ElementEntity.java
```diff
@@ -0,0 +1,94 @@
...
+        @ForeignKey(
+            entity = FormEntity.class,
+            parentColumns = "id",
+            childColumns = "form_id",
+            onDelete = ForeignKey.CASCADE),
+    indices = {@Index("form_id"), @Index("field_id")})
+public abstract class ElementEntity {
+
+  @PrimaryKey(autoGenerate = true)
+  public int id;
```
gino-m: We're using String uuids for elements, so you'd want to remove autoGenerate and change this to a string here.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ElementTypeConverter.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.persistence.local.room;
+
+import androidx.room.TypeConverter;
+import com.google.android.gnd.model.form.Element.Type;
+
+public class ElementTypeConverter {
```
gino-m: Same here; use IntEnum instead.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldData.java
```diff
@@ -0,0 +1,36 @@
...
+package com.google.android.gnd.persistence.local.room;
+
+import androidx.room.Embedded;
+import androidx.room.Relation;
+import java.util.List;
+
+/**
+ * Represents relationship among FieldEntity, MultipleChoiceEntity and OptionEntity.
+ *
+ * <p>Querying any of the below data class automatically loads the field annotated as @Relation.
```
gino-m: Not sure I follow what this means. Is there another way to implement this that doesn't require a class with public mutable members and lazy loading of foreign key relations (iiuc)?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldEntity.java
```diff
@@ -0,0 +1,134 @@
...
+        Field.newBuilder()
+            .setId(fieldEntity.getId())
+            .setLabel(fieldEntity.getLabel())
+            .setRequired(fieldEntity.isRequired())
+            .setType(fieldEntity.getType());
+
+    List<MultipleChoiceEntity> multipleChoiceEntities = fieldData.multipleChoiceEntities;
+
+    // TODO: Re-enable this check.
+    //    if (multipleChoiceEntities.size() != 1) {
```
gino-m: We should recover gracefully on unexpected data in the db (Log.e, and ignore extra values for example), since this may happen with schema changes, etc.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldTypeConverter.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.persistence.local.room;
+
+import androidx.room.TypeConverter;
+import com.google.android.gnd.model.form.Field.Type;
+
+public class FieldTypeConverter {
```
gino-m: Convert to FieldEntity.Type IntEnum.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormData.java
```diff
@@ -0,0 +1,37 @@
...
+import androidx.room.Embedded;
+import androidx.room.Relation;
+import java.util.List;
+
+/**
+ * Represents relationship among FormEntity, ElementEntity and FieldEntity.
+ *
+ * <p>Querying any of the below data class automatically loads the field annotated as @Relation.
+ */
+public class FormData {
```
gino-m: Same here. Is this class needed? If so, can we make fields private and access via accessors?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormData.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Embedded public FormEntity formEntity;
+
+  @Relation(parentColumn = "id", entityColumn = "form_id", entity = ElementEntity.class)
+  public List<ElementEntity> elementEntities;
+
+  @Relation(parentColumn = "id", entityColumn = "form_id", entity = FieldEntity.class)
+  public List<FieldData> fields;
```
gino-m: Field is a type of Element (currently the only type), so there should be one list, ElementEntity, whose members can be Fields.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormEntity.java
```diff
@@ -0,0 +1,104 @@
...
+
+  public static FormEntity create(String id, String title, String layerId) {
+    return builder().setId(id).setTitle(title).setLayerId(layerId).build();
+  }
+
+  public static Builder builder() {
+    return new AutoValue_FormEntity.Builder();
+  }
+
+  @AutoValue.Builder
```
gino-m: Random thought: Should we just get rid of the Builders at some point and use the `create()` method? Android Studio shows us param names inline, so from a readability perspective there's little difference, and we need the create() method for Room. I'm undecided, but it might reduce some boilerplate.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/LayerData.java
```diff
@@ -0,0 +1,33 @@
...
+import androidx.room.Embedded;
+import androidx.room.Relation;
+import java.util.List;
+
+/**
+ * Represents relationship between LayerEntity and FormEntity.
+ *
+ * <p>Querying any of the below data class automatically loads the field annotated as @Relation.
+ */
+public class LayerData {
```
gino-m: Same here. Do you mean these are being lazily loading from the db when you iterate or access the list?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -65,6 +73,113 @@ public RoomLocalDataStore(GndApplication app) {
             .build();
   }
 
+  @Override
+  public Completable insertOrUpdateOption(String fieldId, Option option) {
```
gino-m: Same comment as before; only the top level insertOrUpdateProject needs to be public.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ElementEntity.java
shobhitagarwal1612: Elements didn't have a separate id of it's own.

Ideally, it would be cleaner if I add FieldEntity as `@Embedded` in this entity. But since we are keeping the options open regarding element types, I chose not to.

But having an autoGenerate primary key isn't useful and is causing duplicate entries. I need to fix this

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldData.java
shobhitagarwal1612: The comment is quite confusing, I'll try to simplify.

Our goal is to fetch a Project object from the local database which is exactly similar to the one fetched from the remote db. That means, the project should contains all layers and the layers should then have the respective forms and so on.

When we do a fetch query for `ProjectEntity` in room, it just returns a list of projects and nothing else. So, to recursively fetch data for all the relations, I had to introduce another class. This was the recommended solution I could find in the official docs. https://developer.android.com/reference/androidx/room/Relation.html
Please suggest if you have other ideas.

>  Is there another way to implement this that doesn't require a class with public mutable members and lazy loading of foreign key relations

We later convert the list into a `ImmutableList` while converting the `*Data` objects into their respective model objects.

Any suggestions regarding naming this class or maybe packaging them separately so that it's less confusing?

shobhitagarwal1612: Alternatively, we could also add these relations to the Entity class directly so that these `*Data` classes can be removed. But that means all of the data is always loaded regardless of whether we need it or not. This way, we have an option to fetch from either a `*Data` class to load all data or fetch from an `*Entity` class to load only specific data.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormEntity.java
shobhitagarwal1612: Sure, let's do this discussion separately in a refactor issue.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/LayerData.java
shobhitagarwal1612: It doesn't load the data lazily. The point of having a separate class is to load LayerEntity and all FormData (which is basically FormEntity with similar relations)

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldData.java
shobhitagarwal1612: I tried to put the `@Relation` in the entity class too, but it doesn't work

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
shobhitagarwal1612: Nice catch! Fixed

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ElementEntity.java
gino-m: > Elements didn't have a separate id of their own.

They should.. where did you see that they don't? New Elements are never created on the Android side, their uuid is set by the form designer when they're added there. The schema is defined by example here:
https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation#document-structure

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldData.java
gino-m: Thanks for doing the research on this one. I see the `Data` classes are wrapping both the Entity, and the `@Relation` Entities. Would it be possible to eliminate these classes and move the `@Relation` into the Entity class itself? That way the whole project would be loaded at once (which matches the current use), and we can eliminate the extra layer of indirection introduced by the `*Data` classes. Wdyt?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/LayerData.java
gino-m: Gotcha. Thanks for bearing with me! See previous comment.

gino-m:
Hmm, didn't see this comment in the Conversation thread, only the email.
This is what I was suggesting in my comment. These only foreseen use case
where would only load part of the project is for project summaries, where
we'd only want the project title (and eventually thumbnail), but for that
case I assume we'd create a separate Entity with only those fields,
pointing to the same table. In that case I'd apply YAGNI here and embed
the @Relation inside the Entity.

gino-m:
>
> I tried to put the @Relation in the entity class too, but it doesn't work
>
> Ah:
"Note that @Relation annotation can be used only in POJO classes, an Entity
<https://developer.android.com/reference/androidx/room/Entity.html> class
cannot have relations. This is a design decision to avoid common pitfalls
in Entity
<https://developer.android.com/reference/androidx/room/Entity.html> setups.
You can read more about it in the main Room documentation. When loading
data, you can simply work around this limitation by creating POJO classes
that extend the Entity
<https://developer.android.com/reference/androidx/room/Entity.html>."
https://developer.android.com/reference/androidx/room/Relation.html

Oops. Ok, in that case I would keep them and call them something like
*EntityAndRelations (ProjectEntityAndRelations, etc).

Thanks again for your patience!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldData.java
shobhitagarwal1612: > I tried to put the @Relation in the entity class too, but it doesn't work

Pointing to my previous comment.

My first attempt to put the relations int the Entity class didn't work. But I'll try to dig deeper and try to make it work. Removing the extra layer does sound tempting enough for this extra work.

shobhitagarwal1612: Saw your comment on the response later https://github.com/google/ground-android/pull/246#issuecomment-565855870

On it.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/ElementEntity.java
shobhitagarwal1612: I got confused because the Element model didn't have an `id` field. I'll add that to both places

shobhitagarwal1612: I think we can remove `ElementEntity` and store `FieldEntity` directly for each element. It does not contain any unique information anyways.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldEntity.java
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldEntity.java
```diff
@@ -0,0 +1,158 @@
...
+            .setId(fieldEntity.getId())
+            .setLabel(fieldEntity.getLabel())
+            .setRequired(fieldEntity.isRequired())
+            .setType(fieldEntity.getFieldType().toFieldType());
+
+    List<MultipleChoiceEntity> multipleChoiceEntities =
+        fieldEntityAndRelations.multipleChoiceEntities;
+
+    if (!multipleChoiceEntities.isEmpty()) {
+      if (multipleChoiceEntities.size() > 1) {
```
gino-m: We may want to use @Embed for these since the relationship is 1:[0|1], but we can keep this for now.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldEntityAndRelations.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Represents relationship among FieldEntity, MultipleChoiceEntity and OptionEntity.
+ *
+ * <p>Querying any of the below data class automatically loads the field annotated as @Relation.
+ */
+public class FieldEntityAndRelations {
+  @Embedded public FieldEntity fieldEntity;
+
+  @Relation(parentColumn = "id", entityColumn = "field_id", entity = MultipleChoiceEntity.class)
+  public List<MultipleChoiceEntity> multipleChoiceEntities;
```
gino-m: For example, this could be embedded in `FieldEntity` and OptionEntity moved under `MultipleChoiceEntityAndRelations`, since it only applies there.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormEntity.java
```diff
@@ -0,0 +1,101 @@
...
+        .build();
+  }
+
+  public static Form toForm(FormEntityAndRelations formEntityAndRelations) {
+    FormEntity formEntity = formEntityAndRelations.formEntity;
+    Form.Builder formBuilder =
+        Form.newBuilder().setId(formEntity.getId()).setTitle(formEntity.getTitle());
+
+    ImmutableList.Builder<Element> listBuilder = ImmutableList.builder();
+    for (FieldEntityAndRelations fieldEntityAndRelations :
```
gino-m: Functional form might be a little more readable here, but not urgent.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
```diff
@@ -49,7 +49,7 @@
   // TODO(#57): Set to false to disable Firebase-managed offline persistence once local db sync
   // is implemented for project config and users.
   private static final FirebaseFirestoreSettings FIRESTORE_SETTINGS =
-      new FirebaseFirestoreSettings.Builder().setPersistenceEnabled(true).build();
+      new FirebaseFirestoreSettings.Builder().setPersistenceEnabled(false).build();
```
gino-m: W00t! Congratulations! 🚀

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -121,7 +122,17 @@ private Completable updateLocalFeature(RemoteDataEvent<Feature> event) {
     Log.d(TAG, " Activating project " + projectId);
     return remoteDataStore
         .loadProject(projectId)
-        .doOnError(e -> Log.e(TAG, "Project not found", e))
+        .flatMap(project -> localDataStore.insertOrUpdateProject(project).toSingleDefault(project))
+        .onErrorResumeNext(
+            throwable -> {
```
gino-m: Pretty sure this can be simplified, let's revisit in a follow-up.

shobhitagarwal1612: Yes. Let's do a bit cleanup sometime soon.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
shobhitagarwal1612: I forgot to enable it back after testing. But I don't think we need it anymore. Also, keeping the persistence disabled will help us test the local db with daily usage.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FormEntity.java
shobhitagarwal1612: Are you referring to renaming it or something else?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/FieldEntityAndRelations.java
shobhitagarwal1612: Sounds right!

shobhitagarwal1612:
Thanks for the useful comments! We can use them for the next iteration of refactor/cleanup.

PR #247, opened 2019-12-02
[Bug] Fix navigation flow

### LINKED ISSUE #245: [Bug] Blank screen after pressing back from home screen
**Describe the bug**
Blank screen appears after pressing back button in home screen

**To Reproduce**
1. Go to home screen
2. Press 'back' button
3. White blank screen appears

**Expected behavior**
App should either close or show confirmation to exit

shobhitagarwal1612:
Closes #245 

`MainActivity` handles the flow of the app (navigation) irrespective of the current displayed fragment.  The problem with the current approach was we that we were not respecting the actual nav_graph intended for this purpose.

There are multiple paths for the flow, we need to use different `NavDirections` based on the current visible fragment. To address that, I've used the `NavController` to determine which fragment is currently loaded.

### Nodes (excluded rest of the fragments here since their behavior is fine):
1. Start Fragment
2. Login Screen
3. Home Screen
4. Exit

### Case 1:
Fresh app install
`1 --> 2 --> 3`
Back press from 2 or 3 should exit the app

### Case 2:
Already logged in
`1 --> 3`
Back press from 3 should exit the app

### Case 3:
Logging out
`1 --> 3 --> 2`
Back press from 3 or 2 should exit the app


gino-m:
Is the problem that [Global actions](https://developer.android.com/guide/navigation/navigation-global-action) behave differently than we expect w.r.t. up/back? If so, why doesn't it?

shobhitagarwal1612:
I might be wrong as I couldn't find a similar blog or post but IMO the problem with `Global Actions` is that it doesn't pop the fragments up in the navigation stack while starting the new fragment.

gino-m:
> I might be wrong as I couldn't find a similar blog or post but IMO the problem with `Global Actions` is that it doesn't pop the fragments up in the navigation stack while starting the new fragment.

I believe that's what [this article](https://medium.com/a-problem-like-maria/a-problem-like-navigation-part-2-63e46a565d4b) is saying, actually (skimmed it quickly).

gino-m:
> `MainActivity` handles the flow of the app (navigation) irrespective of the current displayed fragment. The problem with the current approach was we that we were not respecting the actual nav_graph intended for this purpose.

Is this a problem because we need to call the getNavController() on the Fragment we're navigating away from instead of the one provided by the Activity? My understanding is that they're the same instance.

@cketcham in case you have anything to add.

shobhitagarwal1612:
> Is this a problem because we need to call the getNavController() on the Fragment we're navigating away from instead of the one provided by the Activity? My understanding is that they're the same instance.

I've fetched NavDirections based on the current visible fragment. For that purpose only, I used the NavController. The instance is same like you said

PR #260, opened 2019-12-10
[Testing] Add tests for InMemoryCacheTest

shobhitagarwal1612:
I have chosen the most easiest class for adding tests. 

The motivation behind this is to have at least one test class written so that I can start working on running unit tests on cloud build as a first step (next Instrumentation Tests)

### DIFF
--- a/gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
+++ b/gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.android.gnd.repository;
 
 import static org.junit.Assert.assertEquals;
--- a/gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
+++ b/gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
@@ -41,6 +41,7 @@ public class InMemoryCacheTest {
   public void testSaveFeatures() {
     assertEquals(0, inMemoryCache.getFeatures().size());
 
+    // save feature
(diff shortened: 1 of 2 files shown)


gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
```diff
@@ -0,0 +1,107 @@
...
+        Feature.newBuilder()
+            .setId("foo feature id")
+            .setProject(Project.newBuilder().build())
+            .setLayer(Layer.newBuilder().build())
+            .setPoint(Point.newBuilder().setLatitude(0.0).setLongitude(0.0).build())
+            .build();
+  }
+
+  @Test
+  public void testSaveFeatures() {
```
gino-m: Nit: +1 to testing behavior and not methods. When entry points (e.g. putFeature()) match 1:1 with behavior under test, can we call the test methods accordingly?

Also, introducing a new term "save" in this context may be a little confusing, since it implies there's something different going on besides putting and getting.

gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    assertEquals(1, inMemoryCache.getFeatures().size());
+    assertEquals(feature, inMemoryCache.getFeatures().asList().get(0));
+  }
+
+  @Test
+  public void testDeleteFeatures() {
+    inMemoryCache.putFeature(feature);
+
+    assertEquals(1, inMemoryCache.getFeatures().size());
```
gino-m: We already have a test for the behavior implemented by putFeature(), so this check is redundant and can be removed. Ideally each test should test only one behavior (https://testing.googleblog.com/2014/04/testing-on-toilet-test-behaviors-not.html).

shobhitagarwal1612: I added this because features saved using `putFeature` get cleared as a side effect of `clearActiveProject`

gino-m: If these are two "expected behaviors" then they should be expressed in two different tests.

P.S. If possible please avoid force-push so that comment stay associated with related code:
image

shobhitagarwal1612: Done

gnd/src/test/java/com/google/android/gnd/repository/InMemoryCacheTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private InMemoryCache inMemoryCache;
+  private Project project;
+  private Feature feature;
+
+  @Before
+  public void setUp() {
+    inMemoryCache = new InMemoryCache();
+
+    project =
```
gino-m: These are constants so they can be defined as `public static final`s.

PR #264, opened 2019-12-14
[CI] Fix lint and PMD errors breaking build

gino-m:


gino-m:
Note that Android Studio keeps updating `Project.xml`, so it was accidentally committed. Maybe we should allow this commit so it will become stable?

PR #272, opened 2019-12-18
moved android contributing information out of readme

### LINKED ISSUE #268: [Doc] Refactor dev workflow out of README.me into CONTRIBUTING.me
README is visible at https://github.com/google/ground-android and includes instructions on how to contribute. CONTRIBUTING.md also contains guidelines for contributing. Ideally we'd only have these instructions in one place. Would it make sense to move them out of the README.md and reference CONTIBUTING.md there instead?

mmisim:
Fixes #268


scolsen:
Nice! In the future, we may want to split up the set up (what I need to do to build the app and run it) from the developer workflow (what I need to do to hack around)--esp. if we're committed to the OSS idea--we'll need to support people building the app themselves, but the people that want to do that aren't always also looking to contribute, so maybe we should surface that info elsewhere?

PR #274, opened 2019-12-19
[Active project] Fix project spinner bug and streamline activation flow

### LINKED ISSUE #186: [Add record] "Project loading" spinner sometimes appears after saving record while offline
The app becomes unresponsive:
Screenshot_1568512192

gino-m:
- [x] Moves caching of active project from InMemoryCache to stream.
- [x] Renames `Persistable` to `Loadable` now that we've removed saved state.
- [x] Adds Rx composable helpers to `Loadable` and `Result` to wrap and unwrap values in this stream. This is especially useful when propagating error states from streams implicitly subscribed inside `flatMap` and `switchMap`, in which case Rx stream errors are not propagated, but rather break the outer stream.
- [x] Move subscriptions of project activation into initialization of view layer so that disposables are cleaned up on Android lifecycle events.

Fixes #186. Contributes to #84.


gino-m:
@shobhitagarwal1612 @scolsen Some ideas and WIP in e8a02d59cc4fb027dcf020f0a587b01846822c90.

gnd/src/main/java/com/google/android/gnd/rx/Result.java
```diff
@@ -54,4 +57,35 @@ protected Result(@Nullable T value, @Nullable Throwable error) {
...
+
+  /**
+   * Returns a {@link Flowable} that maps values emitted from the source stream as Result instances.
+   * Errors in the provided stream are handled and wrapped in a Result with ERROR state. The
+   * returned stream itself should never fail with an error.
+   *
+   * @param source the stream to be modified.
+   * @param <T> the type of value being emitted.
+   */
+  public static <T> Flowable<Result<T>> wrap(Flowable<T> source) {
```
gino-m: Would changing these to operate on `Single` instead of `Flowable` possibly make more sense semantically?

gino-m: I tried this out and it seems to be more intuitive with Single, but in the end, wrap/unwrap aren't needed for this PR so removing them. Stashing them in [this gist](https://gist.github.com/gino-m/efcfd158a4d6e4461ac17a1a064d8a7e) for use when we want to wrap potentials errors without breaking a `Flowable` or `Observable`.

shobhitagarwal1612: I don't have anything useful to add to the conversation. @scolsen ?

gino-m: There's a long thread on #84, but I think we can leave that for a separate PR since this one is already growing unwieldy (sorry for that!).

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -72,22 +74,65 @@ public DataRepository(
...
+    }
+    String id = projectId.get();
+
+    return syncProjectWithRemote(id)
+        .doOnSubscribe(__ -> Log.i(TAG, "Activating project " + id))
+        .onErrorResumeNext(
+            localDataStore
+                .getProjectById(id)
+                .doOnSubscribe(__ -> Log.i(TAG, "Falling back to local db"))
+                .toSingle())
```
shobhitagarwal1612: This would consume all kinds of exceptions even when it's not `FirebaseFirestoreException`. Do we want that?

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
```diff
@@ -114,51 +159,21 @@ private Completable updateLocalFeature(RemoteDataEvent<Feature> event) {
...
-              if (throwable instanceof FirebaseFirestoreException) {
-                return localDataStore.getProjects();
-              }
-              return Single.error(throwable);
-            })
-        .map(Persistable::loaded)
-        .onErrorReturn(Persistable::error)
-        .toObservable()
-        .startWith(Persistable.loading());
+        .onErrorResumeNext(localDataStore.getProjects())
```
shobhitagarwal1612: Same as above

gnd/src/main/java/com/google/android/gnd/rx/Result.java
shobhitagarwal1612: Thanks! Dealing with it separately sounds ok to me

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: Good catch.. is that the only exception that would be meaningful here? In either case, we would want to abstract this type of failure (`[Local|Remote]DataStoreException`?), since the specific exceptions thrown are a detail of the local and remote data store implementations.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
gino-m: I've updated these to log failures so we can track which exceptions are relevant and update accordingly.

gnd/src/main/java/com/google/android/gnd/repository/DataRepository.java
shobhitagarwal1612: I don't know what other kinds of exceptions are thrown. This was based on my observation while targeting only internet availability as the culprit for the error.

.idea/codeStyles/Project.xml
```diff
@@ -8,7 +8,6 @@
       </value>
     </option>
     <AndroidXmlCodeStyleSettings>
-      <option name="ARRANGEMENT_SETTINGS_MIGRATED_TO_191" value="true" />
```
shobhitagarwal1612: Might be worth keeping this if it the change is harmless.

PR #275, opened 2019-12-19
[Refactor] Move common methods to BaseDao and refactor DAOs

shobhitagarwal1612:
Moves commonly used update and delete methods to BaseDao. Also, all DAOs now extend BaseDao


shobhitagarwal1612:
Fixed!

PR #277, opened 2019-12-21
[Refactor] Inject DAOs directly

shobhitagarwal1612:
This refactor is based on an existing TODO

- Adds new file to store project configurations
- Create database in module
- Inject DAOs directly

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/GndApplicationModule.java
+++ b/gnd/src/main/java/com/google/android/gnd/GndApplicationModule.java
@@ -24,8 +24,20 @@ import androidx.work.WorkManager;
 import com.google.android.gms.common.GoogleApiAvailability;
 import com.google.android.gnd.inject.ActivityScoped;
 import com.google.android.gnd.persistence.local.LocalDataStore;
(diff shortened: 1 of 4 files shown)


gnd/src/main/java/com/google/android/gnd/GndApplicationModule.java
```diff
@@ -83,4 +100,73 @@ static SharedPreferences sharedPreferences(GndApplication application) {
...
+  @Provides
+  @Singleton
+  static LocalDatabase localDatabase(Context context) {
+    return Room.databaseBuilder(context, LocalDatabase.class, Config.DB_NAME)
+        // TODO(#128): Disable before official release.
+        .fallbackToDestructiveMigration()
+        .build();
+  }
+
+  @Provides
```
gino-m: Instead of having separate Providers here, can we instead just use `@Inject` on each Dao?

gino-m: Nvm, I don't see an easy way to get the annotation to be applied to the generated DAO classes.

PR #280, opened 2019-12-26
[Code health] Split up DataRepository

gino-m:
Splits `DataRepository` into multiple classes (improves separation of concerns, readability, testability).


gino-m:
Rename changes were getting bit, so moving into their own PR. PTAL!

gino-m:
Reassigning @scolsen to better load balance.

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -0,0 +1,126 @@
...
+
+    streamFeaturesToLocalDb(remoteDataStore);
+  }
+
+  /**
+   * Mirrors features in the current project from the remote db into the local db when the network
+   * is available. When invoked, will first attempt to resync all features from the remote db,
+   * subsequently syncing only remote changes.
+   */
+  private void streamFeaturesToLocalDb(RemoteDataStore remoteDataStore) {
```
scolsen: N.B. Most of the following doesn't apply if we decouple features from projects (as noted in the comments in this file).

I wonder a bit about having methods that rely internally on the current project and methods that take a project as argument. 

It's more verbose, but I think it simplifies things if we stick to a standard functional signature that expresses the relational constraints between the entities. e.g. all features require a project context so all feature functions are mappings from : project -> feature(s). 

This makes functions like this one slightly more self-explanatory. As this function currently stands, one needs to read the impl or the doc to know that it uses the current project context. 

Taking the project context explicitly in each method also makes these methods composable at the point of use and more flexible e.g. 

```java
projectRepo.getActiveProjectOnceAndStream().switchMap(featureRepo::doSomethingWithFeatures).switchMap(datastore::DosomethingWithDataStore).subscribe()
```

This keeps our APIs functional and usable as component-wise compositions, rather than surfacing preconceived uses/sequences as a single method. Our api methods are the morphemes (smallest possible composable units) and rx is the grammar.

This approach should also give callers control over subscriptions (partially tackling the problem noted in the TODO on this method, from the API side, we have no idea when the disposable might be garbage collected, but if we pass over subscription control to the caller, it's more likely they'll be able to anticipate and account for possible GC calls).

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -0,0 +1,126 @@
...
+
+  // TODO: Only return feature fields needed to render features on map.
+  // TODO(#127): Decouple from Project and accept id instead.
+  public Flowable<ImmutableSet<Feature>> getFeaturesOnceAndStream(Project project) {
+    return localDataStore.getFeaturesOnceAndStream(project);
+  }
+
+  // TODO(#127): Decouple Project from Feature and remove projectId.
+  // TODO: Replace with Single and treat missing feature as error.
+  public Maybe<Feature> getFeature(String projectId, String featureId) {
```
scolsen: If my reading is accurate, I think the impl is slightly misleading here. The signature would suggest that if I know project P exists and feature F exists, getFeature(P, F) would return F, however, this method actually has the additional requirement that P === ActiveProject P* if that doesn't hold, even if P and F exist (P,F) returns nothing.

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Retrieves the records or the specified project, feature, and form.
+   *
+   * <ol>
+   *   <li>Attempt to sync remote observation changes to the local data store. If network is not
+   *       available or operation times out, this step is skipped.
+   *   <li>Relevant records are returned directly from the local data store.
+   * </ol>
+   */
+  public Single<ImmutableList<Observation>> getRecords(
```
scolsen: getObservations?

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
```diff
@@ -0,0 +1,138 @@
...
+    // TODO: Only fetch first n fields.
+    // TODO(#127): Decouple feature from observation so that we don't need to fetch observation
+    // here.
+    return featureRepository
+        .getFeature(projectId, featureId)
+        .switchIfEmpty(Single.error(new DocumentNotFoundException()))
+        .flatMap(feature -> getRecords(feature, formId));
+  }
+
+  private Single<ImmutableList<Observation>> getRecords(Feature feature, String formId) {
```
scolsen: getObservations?

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
```diff
@@ -0,0 +1,138 @@
...
+        remoteDataStore
+            .loadRecords(feature)
+            .timeout(LOAD_REMOTE_RECORDS_TIMEOUT_SECS, TimeUnit.SECONDS)
+            .doOnError(t -> Log.d(TAG, "Observation sync timed out"))
+            .flatMapCompletable(this::mergeRemoteRecords)
+            .onErrorComplete();
+    return remoteSync.andThen(localDataStore.getRecords(feature, formId));
+  }
+
+  private Completable mergeRemoteRecords(ImmutableList<Observation> observations) {
```
scolsen: mergeRemoteObservations?

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -0,0 +1,126 @@
...
+  private final DataSyncWorkManager dataSyncWorkManager;
+
+  @Inject
+  public FeatureRepository(
+      LocalDataStore localDataStore,
+      RemoteDataStore remoteDataStore,
+      ProjectRepository projectRepository,
+      DataSyncWorkManager dataSyncWorkManager) {
+    this.localDataStore = localDataStore;
+    this.projectRepository = projectRepository;
```
scolsen: I forget off the top of my head, but don't features and projects have foreign keys? If that's the case, don't we need to ensure that projectRepo.local/remoteDataStore and FeatureRepo.local/remoteDataStore are equivalent?

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -0,0 +1,159 @@
...
+        .timeout(LOAD_REMOTE_PROJECT_TIMEOUT_SECS, TimeUnit.SECONDS)
+        .doOnSubscribe(__ -> Log.d(TAG, "Activating project " + id))
+        .doOnSuccess(localDataStore::insertOrUpdateProject);
+  }
+
+  private Single<Project> getProjectFromLocal(String id, Throwable error) {
+    return localDataStore
+        .getProjectById(id)
+        .toSingle()
+        .doOnSubscribe(__ -> Log.d(TAG, "Failed to load project from remote", error));
```
scolsen: remote -> local?

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -0,0 +1,159 @@
...
+
+  private Single<List<Project>> loadProjectSummariesFromLocal(Throwable error) {
+    return localDataStore
+        .getProjects()
+        .doOnSubscribe(__ -> Log.d(TAG, "Failed to load project list from remote", error));
+  }
+
+  // TODO: Only return feature fields needed to render features on map.
+  // TODO(#127): Decouple from Project and accept id instead.
+  public Flowable<ImmutableSet<Feature>> getFeaturesOnceAndStream(Project project) {
```
scolsen: Are we exposing this method in two places? Projectrepo and featurerepo? Does this belong in the feature repo only?

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
gino-m: This was only run when the remote fetch failed. That probably wasn't too clear - pushed e54973073436e080ebb73ffba53ad23b678269e2 to make things clearer.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
gino-m: Wow, good catch. Removing from ProjectRepo in c2fd49688e10ec20ed55a3a92df4fcfdec7279d5.

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
gino-m: Rename of existing methods is taking place in #282.

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
gino-m: Yep, this is a known limitation. In practice features will only ever be loaded for the active project so this isn't an issue. In either case fixing it here is out of the scope of this PR. Added a TODO to fix it going forward.

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
gino-m: See above.

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
gino-m: UUIDs are being used as PKs and FKs, and we use the same ones for local and remote. How would we ensure they're equivalent? Not sure I follow..

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
gino-m: See #282.

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
gino-m: Valid observation and good suggestions for improvement! Let's follow up on this when we fix the poorly defined subscription pattern. Thanks!

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
gino-m: P.S. Trying to focus on splitting up the repo now without changing actual functionality for now. If relevant please create an issue with the suggestion and assign to me for follow-up. Thanks for all the input!

gnd/src/main/java/com/google/android/gnd/repository/ObservationRepository.java
scolsen: thanks!

gino-m:
> Looks good!

Thanks for the review!

PR #174, opened 2019-09-11
UI: Rename "record" to "observation"

scolsen:
This change renames UI instances of the string record to observation, in
keeping with our adopted terminology.

Closes #121

Issue #143, opened 2019-07-11
Rename key concepts in code


gino-m:
Based on discussion in https://github.com/google/ground-platform/issues/81:

* FeatureType ➡ Layer
* Record ➡ Observation

Persistence shouldn't be changed in this PR; it needs to be coordinated with web app in #142.

shobhitagarwal1612:
Is this up for grabs or is someone already working on this?

gino-m:
It's all yours! :)

gino-m:
Besides in the data layer, there are a few instances of "record" in constants and resources:
https://github.com/google/ground-android/search?q=record

gino-m:
We can hold off on updating the remaining instances of `featureType` and `record` until the new web console is fully operational. This will make it easier to create and manage projects being used for development.

PR #282, opened 2019-12-26
[Code Health] Rename Record to Observation everywhere except remote db

gino-m:
Work towards #143, #174. Depends on #280.


gino-m:
@scolsen merged #280, so commits and diffs should all be correct now. PTAL!

gino-m:
Yep, that's intentional. We'll change those once we switch over to the
latest schema. thanks for reviewing!

PR #284, opened 2019-12-27
[Code health] Fix feature stream

gino-m:
As discussed in #280, we now subscribe in the MainViewModel instead of hiding the subscription in a Repository. We also expose the fact that sync depends on the current project as suggested. The fact that the sync'ing happens between local and remote data stores is intentional not exposed, following the pattern that the Repositories completely abstract away where data is stored (in memory, local, remote, etc).

Depends on #280 and #282. Relevant commit: 4e3003e67a05e67ec98df24015ddc722af128352


gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -50,29 +51,24 @@ public FeatureRepository(
...
-    projectRepository
-        .getActiveProjectOnceAndStream()
-        .compose(Loadable::values)
-        .switchMap(remoteDataStore::loadFeaturesOnceAndStreamChanges)
+  public Completable syncFeatures(Project project) {
+    return remoteDataStore
+        .loadFeaturesOnceAndStreamChanges(project)
         .switchMap(event -> updateLocalFeature(event).toFlowable())
-        .subscribe();
+        .ignoreElements();
```
scolsen: Neat! Didn't know about this one. It's a nice signal that we only care about side effects here...but do we actually need it? Converting the completable to the flowable I *think* does effectively the same thing? http://reactivex.io/RxJava/javadoc/io/reactivex/Completable.html#toFlowable-- -- since the completable already only emits terminal events.

gino-m: Trying to simplify this... PTAL!

Issue #10, opened 2018-11-18
[Offline imagery] UI for managing offline areas


gino-m:
Allows download of selected mbtiles for offline use. Related to #113.

gino-m:
Thanks for taking this one on @scolsen.

Some mocks for reference:
![image](https://user-images.githubusercontent.com/228050/55904574-2f6ace80-5b9e-11e9-94b5-b14be289eae1.png) ![image](https://user-images.githubusercontent.com/228050/55904591-3560af80-5b9e-11e9-8a19-a997fcb9b768.png) ![image](https://user-images.githubusercontent.com/228050/55904618-47425280-5b9e-11e9-8b50-9f26836d6af0.png)

Please ping me when you get to it and I'll PM you the test data URL.

scolsen:
Remaining work:
- [ ] Add icon next to "Offline maps" menu option
- [ ] Add top app bar with title "Offline maps", subtitle "Tap to toggle offline areas", and back ("up") arrow
- [ ] Show default offline area extends defined in GeoJSON.
- [ ] Tap to toggle areas; selected areas opaque, unselected ones semi-opaque
- [ ] Update work queue when toggling on and off
- [ ]  Listen to work queue and download imagery in background
- [ ] Abort pending download if deselected
- [ ] Remove deselected areas from device

scolsen:
- [ ] Refactor BasemapSelector -> OfflineMaps

gino-m:
Done, and merged into #481.

PR #287, opened 2019-12-27
Offline area download button

scolsen:
This PR adds an download button to the offline areas UI. Currently, the button has no functionality. In a future PR, we'll hook up the button to an area selection UI that users can manipulate to select areas for download. 

I've also made small adjustments to the styling of the offline area UI to make it more harmonious with the rest of the app.

This work counts towards issue #10.

Screen Shot 2019-12-27 at 12 42 47 PM


gnd/src/main/res/drawable-anydpi/ic_area_download.xml
```diff
@@ -0,0 +1,11 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gino-m: I think for this operation the "+" icon would be more appropriate (i.e. add an offline area); I'd expect "download" to trigger/start a download when pressed, whereas here it opens a UI to select an area for download. @jacobmclaws please cmiiw.

gnd/src/main/res/layout/offline_areas_frag.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
gino-m: \n needed between > and <. Formatter did this to me too, but it will fail checkstyle.

gnd/src/main/res/values/strings.xml
```diff
@@ -143,5 +143,6 @@
   <string name="unknown_user">Unknown user</string>
   <string name="add_observation_toolbar_title">Add observation</string>
   <string name="edit_observation_toolbar_title">Edit observation</string>
+  <string name="offline_area_selector_prompt">Select an area for download</string>
```
gino-m: Suggestion: "download" -> "offline use"?

gino-m:
@scolsen Seems there are only a few nits blocking merge.. could you PTAL?

gnd/src/main/res/drawable-anydpi/ic_area_download.xml
scolsen: Sounds reasonable. I've switched it to ic_add_light. 

Technically we should remove the download icon at this point...but do you think we'll use it later? I can sort of envision using it in the area selector, but we could also just use text, in which case we should scrap the download icon.

gnd/src/main/res/layout/offline_areas_frag.xml
scolsen: Done! Thanks!

gnd/src/main/res/values/strings.xml
scolsen: 🙇

gino-m:
We can leave it for now, since we're probably only a few PRs away from
needing it. Thanks!

PR #288, opened 2019-12-27
[BLOCKED] [Code health] Run code health checks on build

gino-m:


gino-m:
@shobhitagarwal1612 @scolsen FYI, WIP - stuck on getting past JAXBException not found when committing through Android Studio.

gino-m:
> @shobhitagarwal1612 @scolsen FYI, WIP - stuck on getting past JAXBException not found when committing through Android Studio.

I got this working by upgrading Gradle to 6.0. I also needed to upgrade Spotbugs for compatibility.

I'm having second thoughts about running checks on commit, however:
* Even when committing from Android Studio, Gradle is run in a different environment with its own cache. This mean whole project will usually needs to be built, which takes a minute or two on my machine.
* Check code errors are note easily visible in the IDE.
image

Running on build is much faster because the code is already compiled. Wdyt?

shobhitagarwal1612:
Thanks for the fix! It seems that the spotbugs plugin is broken after the update. I'll take a look at it.

> Running checkCode on build is much faster because the code is already compiled, and links to analyses are printed to the Build pane in Android Studio. Sending a commit to make this happen, please lmk if you have any thoughts/concerns.

+1 for running the checks on `assembleDebug`. Just tried it out, all checks run parallely too! :rocket:

build.gradle
```diff
@@ -23,6 +23,9 @@ buildscript {
         mavenLocal()
         maven { url "https://maven.google.com"  }
         maven { url "https://github.com/ben-manes/gradle-versions-plugin/raw/mvnrepo" }
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
```
shobhitagarwal1612: This repository can be removed from here. Already included in `spotbugs.gradle`

shobhitagarwal1612:
I slightly prefer the execution of `checkCode` automatically when running `assembleDebug` task.

- I tried to solve the `spotbugs` plugin issue happening after the upgrade. But no solutions so far.
 `Unable to infer the version of SpotBugs from currently specified SpotBugs classpath: [spotbugs-gradle-plugin-3.0.0.jar]`.

- It does introduce one change. `checkCode` gets triggered in CloudBuild during `Build debug APK` step instead of the intended step `Run code quality checks`. This can be overcomed by running a different flavor of `assemble` other than `assembleDebug`.

Final thoughts, if we are leaning towards running the check on build, then we can revert back the changes made for gradle and spotbugs plugins (because of the spotbugs issue)

build.gradle
gino-m: When I remove this I get:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':gnd:spotbugs'.
> Unable to infer the version of SpotBugs from currently specified SpotBugs classpath: [spotbugs-gradle-plugin-3.0.0.jar]
```

I think that's because the spotbugs.gradle is imported into gnd/build.gradle and not ./build.gradle?

gino-m:
> I slightly prefer the execution of `checkCode` automatically when running `assembleDebug` task.

Ack.

> * I tried to solve the `spotbugs` plugin issue happening after the upgrade. But no solutions so far.
>   `Unable to infer the version of SpotBugs from currently specified SpotBugs classpath: [spotbugs-gradle-plugin-3.0.0.jar]`.

I only get that error when removing the repo mentioned above from ./build.gradle. Did you happen to remove it in your local copy?

> * It does introduce one change. `checkCode` gets triggered in CloudBuild during `Build debug APK` step instead of the intended step `Run code quality checks`. This can be overcomed by running a different flavor of `assemble` other than `assembleDebug`.

What's the right way to do this in Gradle? Would we want a flavor, or a variant? I see we can possibly also look for `properties.containsKey('android.injected.invoked.from.ide')`, but this would introduce non-determinism w.r.t. build targets - could lead to surprises later. 

> Final thoughts, if we are leaning towards running the check on build, then we can revert back the changes made for gradle and spotbugs plugins (because of the spotbugs issue). But we'd have to tackle the cloud build issue in that case

See above. If Spotbugs is working for you, let's keep the upgrade of the two versions so we're more up to date?

shobhitagarwal1612:
> I only get that error when removing the repo mentioned above from ./build.gradle. Did you happen to remove it in your local copy?

It happens while running the `spotbugs` check. Please see the cloud build failure

> What's the right way to do this in Gradle? Would we want a flavor, or a variant? I see we can possibly also look for properties.containsKey('android.injected.invoked.from.ide'), but this would introduce non-determinism w.r.t. build targets - could lead to surprises later.

I suppose a variant for CI, like `assembleCI` only for cloud builds

gino-m:
> > I only get that error when removing the repo mentioned above from ./build.gradle. Did you happen to remove it in your local copy?
> 
> It happens while running the `spotbugs` check. Please see the cloud build failure

Hm. It's actually failing for me always, though I could have sworn I had it was working when building in the IDE. I've spent a lot of time debugging this one so I'm going to punt on it for now. That said, we'll need to revisit since we'll eventually want to upgrade to Gradle 6.

shobhitagarwal1612:
Can we close this one?

gino-m:
Abandoned in favor of #320. Thanks again @shobhitagarwal1612! 🚀

PR #289, opened 2019-12-28
[Feature] Save user to local db

shobhitagarwal1612:
Add user entity and save user info on login


gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -137,4 +139,8 @@ public void clearActiveProject() {
     localValueStore.clearLastActiveProjectId();
     activateProjectRequests.onNext(Optional.empty());
   }
+
+  public Completable saveUser(User user) {
```
gino-m: This class is scoped specifically to Project - could you please create a UserRepository and add the method there instead?

shobhitagarwal1612: Done

shobhitagarwal1612:
> Will we also need to serialize some tokens here eventually for OAuth or Firebase? 

Currently, the OAuth is stored in the SDK and we can configure how it is stored there. That behavior might need some tweaking when we want to support multi-user login. Until then, I don't think we need to worry about storing it here as well.

> And are we allow to as per Terms of Service?
Couldn't find anything related to this in the docs

Issue #15, opened 2018-11-18
[Offline support] Allow user to add/remove accounts for offline use


gino-m:
The prototype currently requests sign-in when opening the app. If the user signs out while offline, the app currently becomes unusable, since there's no way to log back in. In order to properly support offline use, account tokens will likely need to be stored locally.

- [ ] Ensure workers connect to remote using user initiating mutation, not current user
- [ ] When user removed, also remove all pending mutations 
- [ ] Verify error handling and race conditions when syncing mutations from deleted users
- [ ] List users UI
- [ ] Add user UI
- [ ] Remove user UI

gino-m:
Hi @chinesco did you mean to close this or another issue?

gino-m:
At the very least we should change the UI to make it harder for the user to accidentally sign out while offline. @ngosue FYI.

scolsen:
@gino-m -- firebase should store these for us, close in favor of making it hard to log out while offline.

Closing this in favor of a bug around offline logout.

scolsen:
Tracking against #769. Closing.

PR #293, opened 2020-01-05
[Auth] Save audit details with features and observations

### LINKED ISSUE #79: [Add/edit feature] Save user id and timestamps on save

gino-m:
* Introduces `AuditInfo*` across all persistence layers. This include the user making feature or observation mutations, the client timestamp, and the server timestamp once synced.
* User id is stored with mutations in local db.
* User is fetched from local store when syncing mutations to server.
* User email, display name, and id are replicated across observations and features for simplicity and to reduce no. of server queries.
* Removes "Anonymous" as a concept in code; user is required for most options, and is generally non-null, except in data objects for data integrity robustness.

Fixes #79.


gino-m:
Remaining TODOs:
- [ ] Write audit info to remote features
- [ ] Write audit info to remote observations

gino-m:
Example of an observation created, then modified:

image

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -298,22 +299,68 @@ public Completable mergeObservation(Observation observation) {
...
+    builder.setLastModified(lastModified);
+    Log.v(TAG, "Merged observation " + builder.build());
+    return builder.build();
+  }
+
+  @Override
+  public Single<User> loadUser(String id) {
+    return userDao
+        .findById(id)
+        .doOnComplete(() -> Log.e(TAG, "User missing local db: " + id))
```
shobhitagarwal1612: Would this trigger even when the user is found, since it is within doOnComplete?

gnd/src/main/java/com/google/android/gnd/model/Mutation.java
```diff
@@ -86,7 +91,9 @@ public String toString() {
 
     public abstract T setProjectId(String newProjectId);
 
-    public abstract T setUserId(String newUserId);
+    public abstract T setUserId(@Nullable String newUserId);
```
shobhitagarwal1612: Instead of passing userId, can we pass `User` object instead? The reason is so that we don't have to load `User` again from the id in `RoomLocalDataStore`
See usages of `loadUser` in `RoomLocalDataStore`

shobhitagarwal1612: But that would require adding `*MutationEntityRelations` to fetch `User` when loading from local db.

gnd/src/main/java/com/google/android/gnd/model/User.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.auto.value.AutoValue;
+
+/** Represents a single application user. */
+@AutoValue
+public abstract class User {
+
+  // TODO: Replace this workaround with Optional<User> or UNKNOWN.
+  public static final User ANONYMOUS =
+      User.builder().setId("").setEmail("").setDisplayName("").build();
```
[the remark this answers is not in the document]
gino-m: Weird, I thought I removed this. Amending my last commit to include its removal.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
gino-m: You have an impeccable eye for detail! It should have been `doOnError` Fixed.

Added some follow-up TODOs in the description of #15.

gnd/src/main/java/com/google/android/gnd/model/Mutation.java
gino-m: That would push the loading of the `User` to the time of loading `Mutations`, which at DB level would mean either loading each `User` individually, or doing a join to fetch them together. Loading them individually isn't a big deal since they're small, there will be few of them (max 2?) and the db is practically in memory. If we go that route, I think doing a query to fill them in the persistence later is more discoverable than using `MutationEntityRelations`. 

Filed #301 to revisit this.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/AuditInfoEntity.java
```diff
@@ -0,0 +1,97 @@
...
+  public static Builder builder() {
+    return new AutoValue_AuditInfoEntity.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    public abstract Builder setUser(UserDetails newUser);
+
+    public abstract Builder setClientTimeMillis(@NonNull long newClientTimeMillis);
```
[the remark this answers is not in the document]
gino-m: We generally use `long` for non-nulls, so wasn't sure which convention to be consistent with :). In doubt, pushed a commit to switch to `Long`.

shobhitagarwal1612:
Looks like the last thing left is to fix the broken unit tests. Then, we are ready to merge!

gino-m:
> Looks like the last thing left is to fix the broken unit tests. Then, we are ready to merge!

Ah, sorry I missed this! Pushed a fix, CI should pass now.

PR #309, opened 2020-01-13
Offline area selector layout

scolsen:
This PR adds a basic offline area selector UI. I've also removed the old BasemapSelector UI (and the corresponding references to it) since it's obsolete.

Here's what the UI looks like:

Screen Shot 2020-01-13 at 2 38 15 PM


scolsen:
@gino-m This UI isn't what we discussed, but it is more idiomatic when accounting for the patterns in the rest of our app--is there an official UI mock I can follow instead?

gnd/src/main/res/layout/offline_area_selector_frag.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright 2019 Google LLC
```
gino-m: Should this be 2020?

gnd/src/main/res/layout/offline_area_selector_frag.xml
```diff
@@ -0,0 +1,61 @@
...
+          android:layout_width="wrap_content"
+          android:layout_height="wrap_content"
+          android:layout_marginBottom="@dimen/map_btn_margin_bottom"
+          android:layout_marginRight="@dimen/chip_button_margin_right"
+          android:layout_marginEnd="@dimen/chip_button_margin_right"
+          android:elevation="6dp"
+          android:text="@string/offline_area_selector_download"
+          app:chipIcon="@drawable/ic_area_download"
+          app:useCompatPadding="true"
+          android:layout_gravity="bottom|end" />
```
gino-m: Would "center|end" would look better for this particular UI?

gino-m:
> @gino-m This UI isn't what we discussed, but it is more idiomatic when accounting for the patterns in the rest of our app--is there an official UI mock I can follow instead?

The UI in this PR is a great start! I wouldn't spend too much time on the UI details for now, since it will very likely evolve as we start using it.

Going forward, we can also draw inspiration from this early mock by @donmccurdy (thank you!):
image

In particular, the "Download" chip would be in the top|end corner to accommodate the Storage card at the bottom. Note that a chip is the correct element to use if we're consistent with the current design of the edit observation dialog.

gnd/src/main/res/layout/offline_area_selector_frag.xml
scolsen: If my calendar's correct, yes! :P

scolsen:
> > @gino-m This UI isn't what we discussed, but it is more idiomatic when accounting for the patterns in the rest of our app--is there an official UI mock I can follow instead?
> 
> The UI in this PR is a great start! I wouldn't spend too much time on the UI details for now, since it will very likely evolve as we start using it.
> 
> Going forward, we can also draw inspiration from this early mock by @donmccurdy (thank you!):
> image
> 
> In particular, the "Download" chip would be in the top|end corner to accommodate the Storage card at the bottom. Note that a chip is the correct element to use if we're consistent with the current design of the edit observation dialog.

Sounds good!

scolsen:
@gino-m Hrm, just re-requested review--but, did you want me to move the download chip in this PR? --Haven't done so yet. I figure we can hold off until the storage indicator is added?

gino-m:
The chip can stay where it is for now, we can fine tune the UI in
subsequent iterations. Thanks!

>

Issue #310, opened 2020-01-14
[Feature] Integrate Firebase Storage for uploading photos


shobhitagarwal1612:
https://firebase.google.com/docs/storage/android/start

Add capability to upload photos to firebase storage.

- [x]  Prototype uploading images to Firebase Storage

- [x] Create new Field type, ‘photo’ - plumb through data model and local and remote persistence

- [x] UI for uploading photos / show thumbnails

- [x] Add storage runtime permissions

- [x] Capture photos / Select from local storage

- [x] Upload selected/captured photo to Firebase Storage

- [x] Fix storage path for the uploaded media

- [x]  Add photo field type to EditObservationFragment that allows photos to be uploaded, store on local device and db

- [x] Update EditObservationFragment to allow user to select existing images (and video if it’s easy), or to trigger camera to capture a new one using StaX design for inspiration. 

- [x] Store image path and uuid locally (in Observation and ObservationMutation).

- [x]  On remote sync of the relevant ObservationMutation, write photo to storage

- [x] Write photo to remote storage in LocalMutationSyncWorker.

- [x]  Update ObservationDetails to show tiny thumbnail of photo (tiny because this will eventually be moved to the observation list)

- [x] Use RxJava to broadcast upload state across the application for UI updates

- [ ]  Allow removing uploaded photo in relevant EditObservationFragment fields

- [ ] Add support for PHOTO field type to the form designer in the web client

- [x] Create mutation table for sync queue and merge with DataSyncWorker

Storage path:
Root > Uploaded Media > Project > Layer > Observation Id > [field no + element no]

gino-m:
@shobhitagarwal1612 FYI:

Select photo UI:
![image](https://user-images.githubusercontent.com/228050/76426508-8b773d80-6381-11ea-90e5-09a89fce82bf.png)
(except link should say "ADD PHOTO", show in application accent color)

Thumbnail:
![image](https://user-images.githubusercontent.com/228050/76426684-c4171700-6381-11ea-8d5a-d41af06148f2.png)
Shown immediately below field label.

gino-m:
@shobhitagarwal1612 I think this is done? Please reopen if not.

PR #328, opened 2020-01-19
[Support Photos] Add photo field to observation

shobhitagarwal1612:
Towards #310 

- [x] UI for bottom modal sheet (options: Camera or Storage)

- [x] Helper class for handling camera permission and capturing photo

- [x] Helper class for handling read permission and choosing photo from storage

- [x] Save photo locally and update form response

Screenshot_1579456054


gnd/src/main/res/layout/bottom_sheet.xml
```diff
@@ -0,0 +1,46 @@
...
+      android:text="@string/action_camera" />
+    <TextView
+      style="@style/AddPhotoActionItem"
+      android:drawableStart="@drawable/ic_sd_storage"
+      android:drawableLeft="@drawable/ic_sd_storage"
+      android:onClick="@{(view) -> fragment.onSelectPhoto()}"
+      android:text="@string/action_storage" />
+  </LinearLayout>
+
+</layout>
```
scolsen: Add newline.

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -0,0 +1,72 @@
...
+  @Inject
+  public StorageManager(PermissionsManager permissionsManager, ActivityStreams activityStreams) {
+    this.permissionsManager = permissionsManager;
+    this.activityStreams = activityStreams;
+  }
+
+  public void imagePicker() {
+    permissionsManager
+        .obtainPermission(permission.READ_EXTERNAL_STORAGE)
+        .andThen(RxCompletable.completeIf(this::sendIntent))
```
scolsen: Instead of using `completeIf`, why not have `sendIntent` return a completable?

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -0,0 +1,72 @@
...
+                      } else if (activityResult.isCanceled()) {
+                        Log.d(TAG, "file picker canceled");
+                      }
+                    }))
+        .subscribe();
+  }
+
+  private boolean sendIntent() {
+    Log.d(TAG, "Sending file picker intent");
+    activityStreams.withActivity(
```
scolsen: Wrap in a call to `Completable.fromObservable` and return the completable? (This way we'll catch any (highly unlikely) errors emitted by the publish subject)

see: http://reactivex.io/RxJava/javadoc/io/reactivex/Completable.html#fromObservable-io.reactivex.ObservableSource-

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -0,0 +1,74 @@
...
+                      } else if (activityResult.isCanceled()) {
+                        Log.d(TAG, "capture photo canceled");
+                      }
+                    }))
+        .subscribe();
+  }
+
+  private boolean sendIntent() {
+    Log.d(TAG, "Sending capture photo intent");
+    activityStreams.withActivity(
```
scolsen: Wrap in a call to Completable.fromObservable and return the completable? (This way we'll catch any (highly unlikely) errors emitted by the publish subject)

see: http://reactivex.io/RxJava/javadoc/io/reactivex/Completable.html#fromObservable-io.reactivex.ObservableSource-

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -0,0 +1,74 @@
...
+                        Intent intent = activityResult.getData();
+                        if (intent != null) {
+                          Bitmap photo = (Bitmap) intent.getExtras().get("data");
+                          Log.d(TAG, photo.toString());
+                        }
+                      } else if (activityResult.isCanceled()) {
+                        Log.d(TAG, "capture photo canceled");
+                      }
+                    }))
+        .subscribe();
```
scolsen: Eventually, we probably don't want to subscribe in these classes themselves, but to expose their streams to callers and subscribe there.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -197,6 +203,12 @@ public void onShowDialog(Field field) {
     }
   }
 
+  public void onShowPhotoSelectorDialog(Field field) {
```
scolsen: Can we somehow use butterknife binding  `butterknife.BindView` here instead instead of calls to show, etc.?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/ActionBottomDialogFragment.java
```diff
@@ -0,0 +1,85 @@
...
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import com.google.android.gnd.databinding.BottomSheetBinding;
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment;
+
+public class ActionBottomDialogFragment extends BottomSheetDialogFragment {
```
scolsen: Instead of making this generic, why not just call it PhotoDialogFragment for now and abstract later once and if we need to?

gnd/src/main/res/layout/bottom_sheet.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
shobhitagarwal1612: Thanks

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
shobhitagarwal1612: I had to use `Completable.fromAction` as `activityStreams.withActivity` doesn't return anything

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
shobhitagarwal1612: I'm okay with either options. Just wanted to be consistent with other places. Should I replace this with `Butterknife` and remove `DataBinding`?

scolsen: Ah, all good--I think my initial comment is irrelevant here since this is dialog frag. 

This should be fine.

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
shobhitagarwal1612: Fixed

shobhitagarwal1612:
@scolsen Thanks for your comments! I've fixed all of them and also did some cleanup. Can you please take a look again?

scolsen:
> @scolsen Thanks for your comments! I've fixed them and also did some cleanup. Can you please take a look again?

Generally speaking looks pretty good--I'll take a detailed look later today. 

Are we supposed to be able to access the dialog in the app at this point? I can't seem to get it to show up.

shobhitagarwal1612:
> Are we supposed to be able to access the dialog in the app at this point? I can't seem to get it to show up.

For that, you need to either tweak the field type in db to 2 or do this in `EditObservationFragment` (easy UI verification)
```
  private void addField(Field field) {
    switch (field.getType()) {
      case TEXT:
        addTextField(field);
        break;
      case MULTIPLE_CHOICE:
//        addMultipleChoiceField(field);
//        break;
//      case PHOTO:
        addPhotoField(field);
        break;
      default:
        Log.w(TAG, "Unimplemented field type: " + field.getType());
        break;
    }
  }
```

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoDialogFragment.java
```diff
@@ -0,0 +1,99 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import com.google.android.gnd.databinding.EditObservationBottomSheetBinding;
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment;
+
+public class PhotoDialogFragment extends BottomSheetDialogFragment {
+
+  public static final String TAG = "ActionBottomDialog";
+  public static final String FIELD_ID = "field_id";
```
gino-m: Javadoc here as well, maybe rename "FIELD_ID_BUNDLE_ARG" for clarity.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -193,6 +212,52 @@ public void onFocusChange(Field field, boolean hasFocus) {
...
+    /*
+     * Didn't subscribe this with Fragment's lifecycle because we need to retain the disposable
+     * after the fragment is destroyed (for activity result)
+     */
+    disposeOnClear(
+        storageManager
+            .launchImagePicker()
+            .andThen(
+                storageManager
+                    .imagePickerResult()
```
gino-m: Can this and the other lambdas blocks below be refactored into methods with clear names to make it easier to follow?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -193,6 +212,52 @@ public void onFocusChange(Field field, boolean hasFocus) {
     }
   }
 
+  void initPhotoSelector(String fieldId) {
+    /*
+     * Didn't subscribe this with Fragment's lifecycle because we need to retain the disposable
+     * after the fragment is destroyed (for activity result)
+     */
+    disposeOnClear(
```
gino-m: disposeOnClear will remove the subscription as soon as the ViewModel is removed, and we can't assume the ViewModel will stick around after the Fragment is destroyed. Do we need this because the photo picker Intent causes the fragment to go away? If so, can we launch image picker synchronously, and only subscribe to image picker results?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -120,12 +134,18 @@
...
-      AuthenticationManager authenticationManager) {
+      AuthenticationManager authenticationManager,
+      StorageManager storageManager,
+      CameraManager cameraManager) {
     this.resources = application.getResources();
     this.observationRepository = observationRepository;
     this.authManager = authenticationManager;
+    this.storageManager = storageManager;
+    this.cameraManager = cameraManager;
+    this.fileUtil = new FileUtil(application);
```
gino-m: Can we inject this?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -91,10 +101,22 @@ public void onViewCreated(
     viewModel.getForm().observe(this, this::rebuildForm);
     viewModel.getToolbarTitle().observe(this, toolbar::setTitle);
     viewModel.getSaveResults().observe(this, e -> e.ifUnhandled(this::handleSaveResult));
+    viewModel.getAddedPhoto().observe(this, this::onPhotoAdded);
     // Initialize view model.
     viewModel.initialize(EditObservationFragmentArgs.fromBundle(getArguments()));
   }
 
+  private void onPhotoAdded(Map<Field, File> fieldFileMap) {
```
gino-m: Naming nit: `fileFieldMap`

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -91,10 +101,22 @@ public void onViewCreated(
...
+    Log.d(TAG, fieldFileMap.toString());
+
+    // TODO: Upload photo to Firestore Storage
+    // TODO: Fetch download url and update response in viewModel
+    for (Field field : fieldFileMap.keySet()) {
+      File file = fieldFileMap.get(field);
+      viewModel.onResponseChanged(field, TextResponse.fromString(file.getPath()));
```
gino-m: What does this do, and why does it do it for all image fields, not just the changed one?

gnd/src/main/java/com/google/android/gnd/ui/util/FileUtil.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Creates a new file from bitmap and saves under internal app directory
+   * /data/data/com.google.android.gnd/files/
+   */
+  public File saveBitmap(Bitmap bitmap, String filename) {
+    File file = new File(context.getFilesDir(), filename);
+    try (FileOutputStream fos = context.openFileOutput(filename, Context.MODE_PRIVATE)) {
+      bitmap.compress(Bitmap.CompressFormat.JPEG, 100, fos);
+    } catch (IOException e) {
+      Log.e(TAG, e.getMessage(), e);
```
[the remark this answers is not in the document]
shobhitagarwal1612: Agreed

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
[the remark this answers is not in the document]
shobhitagarwal1612: Thanks for pointing this out! This makes the code much simpler

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -63,13 +69,18 @@
...
+  private final CameraManager cameraManager;
+  private final FileUtil fileUtil;
 
   // Input events.
 
   /** Arguments passed in from view on initialize(). */
   private final BehaviorProcessor<EditObservationFragmentArgs> viewArgs =
       BehaviorProcessor.create();
 
+  private final BehaviorProcessor<Map<Field, File>> addedPhoto = BehaviorProcessor.create();
```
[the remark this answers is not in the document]
shobhitagarwal1612: Removed. Entire logic for updating responses moved directly into ViewModel

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -78,6 +89,9 @@
   /** Form definition, loaded when view is initialized. */
   private final LiveData<Form> form;
 
+  /** Added image field. */
+  private final LiveData<Map<Field, File>> photo;
```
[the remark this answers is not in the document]
shobhitagarwal1612: Same as above.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -120,12 +134,18 @@
...
+      CameraManager cameraManager) {
     this.resources = application.getResources();
     this.observationRepository = observationRepository;
     this.authManager = authenticationManager;
+    this.storageManager = storageManager;
+    this.cameraManager = cameraManager;
+    this.fileUtil = new FileUtil(application);
     this.form = fromPublisher(viewArgs.switchMapSingle(this::onInitialize));
     this.saveResults = fromPublisher(saveClicks.switchMapSingle(__ -> onSave()));
+    this.photo = fromPublisher(addedPhoto);
```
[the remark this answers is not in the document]
shobhitagarwal1612: Removed

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
shobhitagarwal1612: Cleaned up

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
shobhitagarwal1612: Initially, I tried subscribing the RxChain in Fragment using it's lifecycle. But during launch of `Intent`, the fragment gets destroyed and so does the disposable.

> can we launch image picker synchronously, and only subscribe to image picker results?

If we separate this logic, then how would we pass the requesting `fieldId` to update the response later?
Since one observation can have multiple photo fields, we have to know the requesting fieldId in advance. Thoughts?

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -50,19 +51,17 @@ public Completable launchImageCapture() {
...
-                Bitmap photo = (Bitmap) intent.getExtras().get("data");
-                return Optional.ofNullable(photo);
-              }
-              return Optional.empty();
-            });
+            activityResult ->
+                Optional.ofNullable(activityResult.getData())
+                    .map(Intent::getExtras)
+                    .map(extras -> (Bitmap) extras.get("data")))
+        .filter(Optional::isPresent)
```
gino-m: If we don't need to return `Optional`, then this can be further simplified to remove the need to wrap and unwrap in `Optional`, e.g.:





Sorry!

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -50,19 +51,17 @@ public Completable launchImageCapture() {
...
+        .map(Optional::get);
```
gino-m:
``` 
  .getNextActivityResult(CAPTURE_PHOTO_REQUEST_CODE)
  .flatMap(this::onCaptureImageResult)
```

Then:
```
private static Observable<Bitmap> onCaptureImageResult(ActivityResult result) {
  return Observable.create(em -> {
    if (!result.isOk()) {
      return;
    }
    Intent data = activityResult.getData();
    if (data == null) {
      return;
    }
    // and so on..
    em.onNext((Bitmap) extras.get("data"));
  });
}
```

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -50,19 +51,17 @@ public Completable launchImageCapture() {
         .andThen(sendCaptureImageIntent());
   }
 
-  public Observable<Optional<Bitmap>> captureImageResult() {
+  public Observable<Bitmap> captureImageResult() {
     return activityStreams
         .getNextActivityResult(CAPTURE_PHOTO_REQUEST_CODE)
```
gino-m: Please use IMAGE (vs PHOTO) throughout for consistency.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
gino-m: Can you use the Bundle stored in the Intent's "extras" to pass the field id through to the photo picker and back again?

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
shobhitagarwal1612: This would convert our RxChains into flat code. What's the benefit of that?

gino-m: In this case, do you mean it would replace the functional reactive programming (FRP) paradigm with an imperative one?

Functional reactive constructs are very useful for representing flows of data while managing/limiting side effects in an expressive, consistent way. But since Java is primarily imperative, we'll always be jumping back and forth between functional and imperative paradigms anyway. Where we choose to do that is up to us, and imo, one clear place where it makes sense to switch to the imperative paradigm is synchronous code that would be much less readable if declared using the FRP model. At least this is what I've observed working with RxJava over the past two years or so.

@scolsen, who has done more research on this than I have, what are your thoughts?

shobhitagarwal1612: I agree with your points. Fixed

shobhitagarwal1612:
> Can you use the Bundle stored in the Intent's "extras" to pass the field id through to the photo picker and back again?

I tried that out, but the extras were available in the resulting intent.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
gino-m: This method can be called multiple times, accumulating subscriptions until the view model is destroyed. If we can't decouple request from the activity response, please add a TODO here to trigger this from a switchMap or to find another way to split the stream in two to avoid accumulating subscriptions.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -193,6 +204,44 @@ public void onFocusChange(Field field, boolean hasFocus) {
...
+        .compose(bitmap -> saveBitmapAndUpdateResponse(bitmap, fieldId))
+        .ignoreElements();
+  }
+
+  void initPhotoCapture(String fieldId) {
+    /*
+     * Didn't subscribe this with Fragment's lifecycle because we need to retain the disposable
+     * after the fragment is destroyed (for activity result)
+     */
+    disposeOnClear(
```
gino-m: Same here re TODO.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -193,6 +204,44 @@ public void onFocusChange(Field field, boolean hasFocus) {
...
+  }
+
+  private Completable handlePhotoPickerResult(String fieldId) {
+    return storageManager
+        .photoPickerResult()
+        .compose(bitmap -> saveBitmapAndUpdateResponse(bitmap, fieldId))
+        .ignoreElements();
+  }
+
+  void initPhotoCapture(String fieldId) {
```
gino-m: Same here re name.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -193,6 +204,44 @@ public void onFocusChange(Field field, boolean hasFocus) {
     }
   }
 
+  void initPhotoSelector(String fieldId) {
```
gino-m: "init" may imply it sets up the photo selector, but doesn't actually show it. One clearer name could be "showPhotoSelector()"

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -0,0 +1,95 @@
...
+            activityStreams.withActivity(
+                activity -> {
+                  Intent cameraIntent = new Intent(MediaStore.ACTION_IMAGE_CAPTURE);
+                  activity.startActivityForResult(cameraIntent, CAPTURE_PHOTO_REQUEST_CODE);
+                  Log.d(TAG, "capture photo intent sent");
+                }));
+  }
+
+  /** Observe for the result of request code {@link CameraManager#CAPTURE_PHOTO_REQUEST_CODE}. */
+  public Observable<Bitmap> capturePhotoResult() {
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -0,0 +1,95 @@
...
+    return activityStreams
+        .getNextActivityResult(CAPTURE_PHOTO_REQUEST_CODE)
+        .flatMap(this::onCapturePhotoResult);
+  }
+
+  /** Fetch bitmap from the result, if present. */
+  private Observable<Bitmap> onCapturePhotoResult(ActivityResult result) {
+    return Observable.create(
```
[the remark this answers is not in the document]
shobhitagarwal1612: Added a TODO for the same

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -169,12 +177,31 @@ public void addMultipleChoiceField(Field field) {
...
     binding.setField(field);
+    binding.setFragment(this);
     formLayout.addView(binding.getRoot());
     assignGeneratedId(binding.getRoot().findViewById(R.id.image_thumbnail_preview));
     assignGeneratedId(binding.getRoot().findViewById(R.id.btn_select_photo));
+
+    viewModel
+        .getResponses()
+        .addOnMapChangedCallback(
+            new OnMapChangedCallback<ObservableMap<String, Response>, String, Response>() {
```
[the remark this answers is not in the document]
shobhitagarwal1612: Sadly no. I was looking into converting the Observable into a PublishProcessor for the Fragment to subscribe. But that needs some more refactoring which I'm planning to do later as it touches other codebase as well

shobhitagarwal1612:
@gino-m Please have a look again

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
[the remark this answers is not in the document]
gino-m: Ok, thanks for checking.

PR #332, opened 2020-01-25
[Code health] Enforce check to detect hard-coded strings in layouts

shobhitagarwal1612:
Enable lint check to detect hardcoded texts in xml layouts.

This should help us in future localization tasks.

@gino-m @scolsen I also noticed that some of strings are duplicate (like `add_feature_cancel` and `discard_multiple_choice_changes`). Any particular reason for that? Also, would it be fine to use android's resource for common strings (android.R.string.cancel) ?

### DIFF
--- a/config/lint/lint.xml
+++ b/config/lint/lint.xml
@@ -23,5 +23,6 @@
   <!-- Severity: Error -->
   <issue id="ApplySharedPref" severity="error" />
   <issue id="DuplicateIds" severity="error" />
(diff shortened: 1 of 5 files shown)


gino-m:
> @gino-m @scolsen I also noticed that some of strings are duplicate (like `add_feature_cancel` and `discard_multiple_choice_changes`). Any particular reason for that? Also, would it be fine to use android's resource for common strings (android.R.string.cancel) ?

+1 to consolidating these. Filed #334 to track.

PR #333, opened 2020-01-27
[Code health] Tweak license header check

shobhitagarwal1612:
It is annoying when Android Studio wraps the header's first line in xml when running `auto-format`.

`<?xml version="1.0" encoding="utf-8"?>`

Since it is optional and doesn't add much value to the code since Android Studio is smart enough to handle them anyways, I propose that we remove them.

So, this PR aims at reducing the developer's time wasted at reverting back the changes manually.
Also, enables the license header check for all xml files and not just under `layout` dir


shobhitagarwal1612:
Another solution would be to introduce a blank space between the xml declaration tag and license.
@gino-m @scolsen Penny for a thought?

shobhitagarwal1612:
Now with the updated checkstyle template, it will be notified with the checkCode task

shobhitagarwal1612:
But we'll still need to remove it manually

gino-m:
These headers seem to be recommended, and are used fairly consistently in the Android sources, for example:
https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/res/res/layout/

It seems adding a blank line between the XML declaration and the comment stops to formatter from removing the newline.

shobhitagarwal1612:
SGTM

PR #335, opened 2020-01-28
[Cleanup] Split up GndDirestore

gino-m:
Closes #322, work towards #323.  Depends on #326.

PR #327, opened 2020-01-17
WIP: Experimenting with robust remote Firestore layer

gino-m:
Sharing an early draft of the revised remote persistence layer. @shobhitagarwal1612 @scolsen FYI.

Issue #144, opened 2019-07-11
[Code health] Replace *Doc classes with sparse Maps


gino-m:
RecordDoc and FeatureDoc currently use two representations; static type fields for fetching/storing the entire object, and `Map<String, Object>` for selectively updating fields on remote server. It's worth looking into consolidating these two representations.

gino-m:
This is low priority, and it's questionable whether we'd want to lose static types offered by these classes.

Issue #142, opened 2019-07-11
[Code health] Big bang change to Firebase schema


gino-m:
To be coordinated together with https://github.com/google/ground-platform/issues/87

gino-m:
Missed a few instances of `center` in Feature.

gino-m:
> Missed a few instances of `center` in Feature.

This is not correct; my client was somehow out of date despite pulling from upstream master.

PR #337, opened 2020-01-28
DRAFT: [Code health] Use maps to represent Firestore data

gino-m:
...in place of relying on Firestore to read/write fields from POJOs. Key benefits:

* Removes duplicate representation of field names needed when building Maps for partial data updates and merges (i.e., mutations).
* Allows more control over failure behaviors, e.g., when database types don't correspond to expected ones.

In the process, we also introduce more carefully separated concerns regarding converters, data contains, and database schema definition.

Based on #327, contributes to #144 and sets us up for non-breaking db migration #142. 

Depends on #335.


gino-m:
@scolsen @shobhitagarwal1612 Please take a glance at the proposed cleanup of our Firestore persistence layer. Relevant commits:
https://github.com/google/ground-android/pull/337/files/97548b924ac73f256770c87a1a1558b50c05ffd5..HEAD

If you think it look good in principle please give a thumbs up and I'll proceed with polishing this PR and migrating the rest of the persistence layer.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/AuditInfoDoc.java
```diff
@@ -59,20 +62,20 @@
...
+        .setUser(UserMapConverter.toUser(Data.fromMap(doc.user)))
         .setClientTimeMillis(doc.clientTimeMillis.toDate())
         .setServerTimeMillis(Optional.ofNullable(doc.serverTimeMillis).map(Timestamp::toDate))
         .build();
   }
 
   public static AuditInfoDoc fromMutationAndUser(Mutation mutation, User user) {
     AuditInfoDoc auditInfo = new AuditInfoDoc();
-    auditInfo.user = UserDoc.fromObject(user);
+    auditInfo.user = UserMapConverter.fromUser(user).toMap();
```
gino-m: Note that nesting will become simplified once we migrate the rest of the model to use `Data`.

gino-m:
@scolsen @shobhitagarwal1612 I've only just begun migrating existing classes to this model and I can see there's a lot of boilerplate and additional conversion logic needed; considering that we'll only be mutating specific fields in the db (and therefore don't need parity with read+write of all objects), and since Firestore does a lot of the type conversions for us, wondering if I should abandon this approach. Anything thoughts before I switch tacks?

gino-m:
Here's an example of the alternative: https://github.com/google/ground-android/pull/350/files

gino-m:
Abandoned in favor of recent refactors.

PR #342, opened 2020-01-31
Add a utility class for parsing GeoJson

scolsen:
To download tiles based on Offline Imagery selections, we need to parse geojson files for tile specifications (polygons) and determine which tiles in a given geojson file intersect with the bounds of an offline imagery selection. This utility class handles both requirements.

This PR also contains changes to the Tile model--we need to update our approach to modeling tiles now that we download them by regional intersection instead of direct selection.


gnd/src/main/java/com/google/android/gnd/util/GeoJsonParser.java
```diff
@@ -0,0 +1,224 @@
...
+        sb.append(line).append('\n');
+        line = buf.readLine();
+      }
+
+      JSONObject geoJson = new JSONObject(sb.toString());
+      JSONArray features = geoJson.getJSONArray("features");
+
+      stream(toArray(features))
+          .filter(GeoJsonParser::hasGeometry)
+          .filter(GeoJsonParser::isPolygon)
```
scolsen: This is wrong -.- Have to fix this.

scolsen: isPolygon expects a geometry object, but here we return the entire feature object (and we need to maintain it down stream). 

I should move these checks into the GeoJsonTile.fromJson method anyway.

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
```diff
@@ -37,14 +46,16 @@ public static Builder newBuilder() {
     return new AutoValue_Tile.Builder();
   }
 
-  public static String pathFromId(String tileId) {
+  public static String pathFromCoords(int x, int y, int z) {
```
gino-m: Would `getFilenameFromCoords()` be more appropriate?

gnd/src/main/java/com/google/android/gnd/workers/FileDownloadWorker.java
```diff
@@ -183,24 +150,21 @@ private Result checkDownload(Tile tile) {
   @NonNull
   @Override
   public Result doWork() {
-    Log.d(TAG, "Downloading tile: " + Tile.pathFromId(tileId));
     Tile tile = localDataStore.getTile(tileId).blockingGet();
 
     // When there is no tile in the db, the maybe completes and returns null.
```
gino-m: Nit: maybe -> Maybe

gnd/src/main/java/com/google/android/gnd/workers/FileDownloadWorker.java
```diff
@@ -183,24 +150,21 @@ private Result checkDownload(Tile tile) {
...
+    // We expect tiles to be added to the DB prior to downloading.
+    // If that isn't the case, we fail.
     if (tile == null) {
```
gino-m: What if the tile has been deleted since this job was queued? Maybe we shouldn't fail but just exit?

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
scolsen: I like filenameFromCoords --but I'm dropping `get` because this method is static (though maybe it shouldn't be static)!

gnd/src/main/java/com/google/android/gnd/workers/FileDownloadWorker.java
scolsen: Hmm, it seems like the right solution here is to use the workManager to cancel the jobs as soon as the tile is deleted: https://developer.android.com/topic/libraries/architecture/workmanager/how-to/cancel-stop-work. Probably some listener on the DB that fires off `cancelAllWork` when a user deletes an area and all its corresponding tiles are up for deletion. i.e. whatever method we implement that deletes tiles first checks in w/ the manager to see if any work is queued and stops all the relevant work first. -- we'll probably have to tag jobs by area id to ensure we stop work for just that area (and not e.g. other areas that were queued up immediately after)

I think this would appropriately handle that case, while this check explicitly uncovers whether or not we queued this without adding something to the DB first.

gnd/src/main/java/com/google/android/gnd/util/GeoJsonParser.java
```diff
@@ -0,0 +1,283 @@
+/*
```
gino-m: A few suggestions on code structure:
- [ ] This deals with deserializing JSON files, so it probably in the persistence package. (e.g. `...persistence.geojson?`)
- [ ] Rather than pass JSONObject around to static methods, can we wrap the different instances in typed accessor objects? (e.g., JsonGeometry, JsonTileExtent, etc)? (https://testing.googleblog.com/2017/11/obsessed-with-primitives.html)
- [ ] If this file is  moved to its own package, the inner classes and be promoted to outer classes in their own file to help readability.
- [ ] GeoJsonParser can then provide a custom GeoJson object with the data we're interested in.

Wdyt?

gnd/src/main/java/com/google/android/gnd/workers/FileDownloadWorker.java
gino-m: I don't feel smart enough to consider all the interactions between the workers and managers, so as a defensive maneuver would normally try to make everything idempotent that can be. Besides the user deleting the tile, it could also get remove if the user clears the app cache; I'm not sure if workers get cancelled as well when that happens. Even if they do, it would be nice not to need to know whether that's the case. Perhaps we can just bail out and log a debug log message instead?

gnd/src/main/java/com/google/android/gnd/util/GeoJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+
+        return new LatLng[] {
+          new LatLng(south, west),
+          new LatLng(south, east),
+          new LatLng(north, west),
+          new LatLng(north, east)
+        };
+
+      } catch (JSONException e) {
+        Log.e(TAG, "couldn't parse json tile coordinates", e);
```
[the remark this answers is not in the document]
scolsen: My thinking was to particularize the point of failure by catching and logging in the helper methods + the caller.

So for instance, if we failed to get coordinates, we'd have a log + stack trace of roughly:

GeoJsonParser: Failed to parse geojson
GeoJsonParser: couldn't parse json tile coordinates

In which case we have more info to work with to debug -- wdyt? Is that overkill? Is it better to define a custom exception instead to differentiate and throw those? --that way we can still be specific about the cause of failure and only need to catch in the caller

gnd/src/main/java/com/google/android/gnd/util/GeoJsonParser.java
scolsen: SG! After the initial impl, this component seems complex enough to me to up-level it to its own package, which will make it easier to define accessor objects like you suggest--plus it'll make it easier to extend in the future.

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/Tile.java
```diff
@@ -37,14 +46,16 @@ public static Builder newBuilder() {
...
     // TODO: Convert tile ids to paths in a less restrictive and less hacky manner.
     // TODO: Move this method to a more appropriate home? We need to perform (and possibly will no
     // matter where the tiles are stored) translation between the tile ID and the file path of the
     // corresponding tile source in remote storage/wherever we pull the source tile from.
-    String[] fields = tileId.replaceAll("[()]", "").split(", ");
-    String filename = fields[2] + "-" + fields[0] + "-" + fields[1];
+    String xstr = String.valueOf(x);
+    String ystr = String.valueOf(y);
+    String zstr = String.valueOf(z);
+    String filename = zstr + "-" + xstr + "-" + ystr;
```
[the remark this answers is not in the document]
scolsen: Gotcha, thanks!

gnd/src/main/java/com/google/android/gnd/workers/FileDownloadWorker.java
scolsen: That sounds fair to me--so, I suppose we need to move this nullness check into the manager and avoid scheduling any work in the first place if we can't find the tile in the DB.

gino-m: We will need the nullness check here regardless, since this logic is executed in a different scheduler than the thread that queues the tile. Currently the manager will only queue work after adding the tiles anyway iirc?

gnd/src/main/java/com/google/android/gnd/util/GeoJsonParser.java
gino-m: Thanks for taking a look!

scolsen:
gino-m@ Mind taking another look? This may still not be totally optimal, but we're getting closer!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+   * bounds}.
+   */
+  public ImmutableList<Tile> intersectingTiles(LatLngBounds bounds, File geojson) {
+    try {
+      InputStream is = new FileInputStream(geojson);
+      BufferedReader buf = new BufferedReader(new InputStreamReader(is));
+      String line = buf.readLine();
+      StringBuilder sb = new StringBuilder();
+      while (line != null) {
+        sb.append(line).append('\n');
```
gino-m: Please use `System.lineSeparator()` instead of `\n`.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    return result;
+  }
+
+  /**
+   * Returns the immutable list of tiles specified in {@param geojson} that intersect {@param
+   * bounds}.
+   */
+  public ImmutableList<Tile> intersectingTiles(LatLngBounds bounds, File geojson) {
```
gino-m: The arg name `geojson` collides with bar `geoJson`. Perhaps rename to `file` for clarity?

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+      Log.e(TAG, "Unable to load JSON layer", e);
+    }
+    return ImmutableList.of();
+  }
+
+  /** Returns the {@link Tile} specified by {@param json}. */
+  private Tile jsonToTile(GeoJsonTile json) {
+    return Tile.newBuilder()
+        .setId(uuidGenerator.generateUuid())
+        .setUrl(json.getUrl().get())
```
gino-m: Either use `orElse()` to provide a default here and below, make required (not Optional), or check these fields are present before hand to prevent `get()` from crashing the app.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonTile.java
```diff
@@ -0,0 +1,86 @@
...
+   * <p>GeoJSON Polygons are described using coordinate arrays that form a linear ring. The first
+   * and last value in a linear ring are equivalent. We assume coordinates are ordered, S/W, S/E,
+   * N/E, N/W, (S/W again, closing the ring).
+   *
+   * <p>Interior rings, which describe holes in the polygon, are ignored.
+   */
+  GeoJsonTile(JSONObject jsonObject) {
+    this.extent = new GeoJsonExtent(new GeoJsonGeometry(jsonObject));
+    this.id = Optional.of(jsonObject.optString(ID_KEY));
+    this.url =
```
gino-m: Same here re cached state.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+  public ImmutableList<Tile> intersectingTiles(LatLngBounds bounds, File geojson) {
+    try {
+      InputStream is = new FileInputStream(geojson);
+      BufferedReader buf = new BufferedReader(new InputStreamReader(is));
+      String line = buf.readLine();
+      StringBuilder sb = new StringBuilder();
+      while (line != null) {
+        sb.append(line).append('\n');
+        line = buf.readLine();
+      }
```
gino-m: Can we split the loading of the file into a String into a separate method so that this method will focus on  parsing its contents?

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.LatLngBounds;
+import java8.util.Optional;
+import javax.annotation.Nullable;
+import org.json.JSONArray;
+
+public class GeoJsonExtent {
+
+  private final GeoJsonGeometry geometry;
+  private final Optional<LatLng[]> coordinates;
```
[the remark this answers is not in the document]
scolsen: Oh, nice! Thanks for the tip!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Nullable
+  GeoJsonExtent(GeoJsonGeometry geometry) {
+    this.geometry = geometry;
+    this.coordinates = getPolygonCoordinates();
```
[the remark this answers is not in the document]
scolsen: sg!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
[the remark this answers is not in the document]
scolsen: Done!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  Optional<LatLngBounds> getBounds() {
+    return this.coordinates.map(cs -> new LatLngBounds(cs[0], cs[2]));
```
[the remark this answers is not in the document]
scolsen: Done!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private Optional<LatLng[]> getPolygonCoordinates() {
+    JSONArray sw = this.geometry.getCoordinates().map(j -> j.optJSONArray(0)).get();
+    JSONArray ne = this.geometry.getCoordinates().map(j -> j.optJSONArray(2)).get();
```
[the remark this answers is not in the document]
scolsen: sg, for the first commit addressing these comments I'm going to stick with the optional and just add a check for empties--I think your suggestion to remove the optional entirely makes sense in this case though, so in subsequent commits I may do that.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -0,0 +1,65 @@
...
+    if (sw == null || ne == null) {
+      return Optional.empty();
+    }
+
+    double south = sw.optDouble(0, 0.0);
+    double west = sw.optDouble(1, 0.0);
+    double north = ne.optDouble(0, 0.0);
+    double east = ne.optDouble(1, 0.0);
+
+    return Optional.of(
```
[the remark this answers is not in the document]
scolsen: Gotcha

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonGeometry.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final String GEOMETRY_KEY = "geometry";
+  private static final String GEOMETRY_TYPE_KEY = "type";
+  private static final String COORDINATES_JSON_KEY = "coordinates";
+
+  private final Optional<JSONObject> json;
+  private final Optional<String> type;
+  private final Optional<JSONArray> coordinates;
+
+  GeoJsonGeometry(JSONObject jsonObject) {
+    this.json = Optional.ofNullable(jsonObject.optJSONObject(GEOMETRY_KEY));
```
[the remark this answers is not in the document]
scolsen: 👍

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonGeometry.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.persistence.geojson;
+
+import java8.util.Optional;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+public class GeoJsonGeometry {
```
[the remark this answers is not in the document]
scolsen: ✅

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+  GeoJsonParser(OfflineUuidGenerator uuidGenerator) {
+    this.uuidGenerator = uuidGenerator;
+  }
+
+  /**
+   * Converts a JSONArray to an array of JSONObjects. Provided for compatibility with java8 streams.
+   * JSONArray itself only inherits from Object, and is not convertible to a stream.
+   */
+  private static JSONObject[] toArray(JSONArray arr) {
+    JSONObject[] result = new JSONObject[arr.length()];
```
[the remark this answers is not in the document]
scolsen: That is way better! We can use optJSONObject instead of getJSONObject to workaround exceptions.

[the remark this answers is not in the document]
scolsen: Turns out we can't conveniently use this approach due to lack of `range` functions (the libs we're using for stream support require spliterators)

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: done

scolsen: Changed to reading this char by char

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: done

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Using orElse() for now and setting an empty string--but these should probably be required.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Done!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonTile.java
scolsen: done

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -17,49 +17,44 @@
...
   @Nullable
   GeoJsonExtent(GeoJsonGeometry geometry) {
     this.geometry = geometry;
-    this.coordinates = getPolygonCoordinates();
   }
 
-  Optional<LatLngBounds> getBounds() {
-    return this.coordinates.map(cs -> new LatLngBounds(cs[0], cs[2]));
+  ImmutableList<LatLng> getVertices() {
+    return getPolygonVertices();
```
gino-m: Can we get rid of this method and expose or rename `getPolygonVertices()`

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -46,41 +46,32 @@
...
+        sb.append(res);
+        res = buf.read();
       }
+      return sb.toString();
+    } catch (IOException e) {
+      Log.e(TAG, "Unable to load JSON", e);
     }
-
-    return result;
+    return "";
```
gino-m: Instead of eating the exception and returning "", we can throw it and catch it in `intersectingTiles`, eliminating the need for a "fake" value in the interim.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -89,19 +80,39 @@
...
+  /**
+   * Converts a JSONArray to an array of JSONObjects. Provided for compatibility with java8 streams.
+   * JSONArray itself only inherits from Object, and is not convertible to a stream.
+   */
+  private static JSONObject[] toArray(JSONArray arr) {
+    JSONObject[] result = new JSONObject[arr.length()];
+
+    for (int i = 0; i < arr.length(); i++) {
+      try {
+        JSONObject o = arr.getJSONObject(i);
```
gino-m: These two lines can be combined.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -89,19 +80,39 @@
...
+
   /** Returns the {@link Tile} specified by {@param json}. */
   private Tile jsonToTile(GeoJsonTile json) {
     return Tile.newBuilder()
         .setId(uuidGenerator.generateUuid())
-        .setUrl(json.getUrl().get())
+        .setUrl(json.getUrl().orElse(""))
         .setState(State.PENDING)
-        .setPath(Tile.pathFromId(json.getId().get()))
+        .setPath(Tile.pathFromId(json.getId().orElse("")))
```
gino-m: Ideally we wouldn't keep Tile in an invalidate state if the file is invalid. Could you add a TODO here to throw an Exception instead that gets handled downstream?

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonTile.java
```diff
@@ -53,34 +52,22 @@
...
-
-  public Optional<LatLng[]> getCoordinates() {
-    return this.extent.getCoordinates();
+  private ImmutableList<LatLng> getVertices() {
+    return new GeoJsonExtent(new GeoJsonGeometry(json)).getVertices();
   }
 
   public Optional<String> getId() {
-    return this.id;
+    return Optional.of(json.optString(ID_KEY));
```
gino-m: This will return Optional.of("") if it doesn't exists. We should either model missing as `null`, `Optional.empty()`, or `""`. Any of those alternatives are fine, but all these will make edge cases harder to catch.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
scolsen: Sgtm, done!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Ah, gotcha. ✅

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Done!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Added.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonTile.java
scolsen: Gotcha, returning Optional.empty on...well, empty values.

scolsen:
Mind taking one more look, @gino-m? 👼

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -60,17 +60,17 @@ private String readJsonFile(File file) {
       return sb.toString();
     } catch (IOException e) {
```
gino-m: If you're throwing the exception callers will need to catch it, and they will try to log it as well; in that case no need for the try/catch and `Log`,

gino-m: In fact I see it's logged in the calling function, so ok to remove the try/catch block.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -96,8 +96,7 @@ private String readJsonFile(File file) {
 
     for (int i = 0; i < arr.length(); i++) {
       try {
-        JSONObject o = arr.getJSONObject(i);
-        result[i] = o;
+        result[i] = arr.getJSONObject(i);
       } catch (JSONException e) {
         Log.e(TAG, "couldn't parse json", e);
```
gino-m: This error message is slightly misleading "Ignoring Error in JSON" might be more accurate.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -96,8 +96,7 @@ private String readJsonFile(File file) {
 
     for (int i = 0; i < arr.length(); i++) {
       try {
-        JSONObject o = arr.getJSONObject(i);
-        result[i] = o;
+        result[i] = arr.getJSONObject(i);
```
gino-m: When an exception is thrown here `result[i]` will remain uninitialized (`null`). In this case you'd need to `filter` them out in the caller `intersectingTiles()`. A cleaner/simpler approach would be to add them to an `ArrayList` instead of a primitive array.

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Thanks!

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Gotcha, switched to ArrayList

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: Good call, updated to "Ignoring error in JSON array"

gino-m:
Weird, checks are failing to run with:
```
Finished Step #2 - "Copy .git directory for versioning"
ERROR
ERROR: build step 2 "gcr.io/cloud-builders/git" failed: step exited with non-zero status: 128
```

Are you able to run `checkCode` locally?

scolsen:
> Weird, checks are failing to run with:
> 
> ```
> Finished Step #2 - "Copy .git directory for versioning"
> ERROR
> ERROR: build step 2 "gcr.io/cloud-builders/git" failed: step exited with non-zero status: 128
> ```
> 
> Are you able to run `checkCode` locally?

Oh weird, I'll take a look

scolsen:
> > Weird, checks are failing to run with:
> > ```
> > Finished Step #2 - "Copy .git directory for versioning"
> > ERROR
> > ERROR: build step 2 "gcr.io/cloud-builders/git" failed: step exited with non-zero status: 128
> > ```
> > 
> > 
> > Are you able to run `checkCode` locally?
> 
> Oh weird, I'll take a look

Sure enough, I just cloned this to my home machine and I can't get gradle to sync b/c of the following error:

```
ERROR: android.defaultConfig.versionCode is set to -1, but it should be a positive integer.
```

which in fact sounds like it would result from something like the original failure:

```
Finished Step #2 - "Copy .git directory for versioning"
> > ERROR
```

If it failed to copy the git dir for versioning it'd break out versions.

Not really sure how or why this happened or how to debug this--Idk anything about gitVersioner (that's what we're using) I'll try to figure it out

scolsen:
Sooooo this is proving tricky. I cloned the repo to my home laptop and cannot get this thing to build at all. gitVersioner refuses to identify it as a git repo even though it is one and every git command works as expected. I've even tried stepping back a couple commits and that doesn't solve anything--there doesn't seem to be any corruption in the repo itself.....so I'm fairly stumped.

scolsen:
I do notice that we use the variable `$_HEAD_REPO_URL` to copy the git file (see cloudbuild.yaml) --I'm guessing that's set in our cloud build setup? Could that have been messed up at some point? Though I don't see how that'd result in local issues building....

As a last effort for now I'm going to remove the gitVersioner plugin entirely and see if I can get this to build.

scolsen:
Finally got this to build on my home machine. Had to remove the `gitVersioner` plugin completely which revealed another error that  `android.enableSeparateAnnotationProcessing=true` is experimental and not supported...commenting that out finally did the trick.

Unfortunately, I don't think these local remedies solve any problems on the cloud build side. We probably still have to diagnose this. It'll be interesting to see whether these same issues crop up when I try to build on my other machine tomorrow.

Not pushing any fixes since it seems to me that the fix I've found is less than ideal and there's probably some other root cause--maybe we have to change something about the way we use gitVersioner? Or perhaps there's something problematic in the .git folder copying script?

scolsen:
Did some digging. The last time we touched `gitVersioner` was back in December--and we hadn't encountered this up till now--so I suspect this issue ultimately stems from either a bug in gitVersioner itself or some other strangeness in git on my local fork. Checking things out on my other machine (where I originated these commits) will hopefully be instructive.

@shobhitagarwal1612 FYI

scolsen:
@gino-m woo! Checks passed

PR #343, opened 2020-02-02
[Refactor] Add fieldId to TextResponse

shobhitagarwal1612:
Inspired from #328 

There is always a need to bind the field id to the response whenever we are dealing with it (eg. EditObservationFragment / EditObservationViewModel)

Adding this field to the model would remove the extra need keep a Map<String, Response> at all places.

Next steps:

- Cleanup code


shobhitagarwal1612:
I'll reopen when it's needed.

PR #344, opened 2020-02-03
[Observation list] Remove form tab header

gino-m:


gnd/build.gradle
```diff
@@ -205,7 +205,3 @@ apply plugin: 'com.google.gms.google-services'
 task checkCode(type: GradleBuild) {
     dependsOn 'checkstyle', 'lintDebug', 'pmd', 'spotbugs'
 }
-
-afterEvaluate {
```
scolsen: This seems unrelated to the overall PR--what's the impetus for this change?

Looks like this lets us assemble a debug build w/o enforcing style etc. checks?

LGTM, just calling it out in case it slipped into this PR accidentally.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -148,31 +147,28 @@ public void onBottomSheetHidden() {
...
-      Log.e(TAG, "Missing form");
-      return;
-    }
     Feature feature = state.getFeature();
     if (feature == null) {
       Log.e(TAG, "Missing feature");
       return;
     }
+    ImmutableList<Form> forms = feature.getLayer().getForms();
+    if (forms.isEmpty()) {
```
scolsen: Add TODO to show a message to the user?

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/FeatureSheetFragment.java
```diff
@@ -76,23 +69,13 @@ public View onCreateView(
   @Override
   public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
     super.onViewCreated(view, savedInstanceState);
```
scolsen: This method is redundant now -- delete?

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/ObservationListViewModel.java
```diff
@@ -59,14 +59,22 @@ public ObservationListViewModel(ObservationRepository observationRepository) {
   }
 
   /** Loads a list of observations associated with a given feature. */
-  public void loadObservationList(Feature feature, Form form) {
+  public void loadObservationList(Feature feature) {
+    ImmutableList<Form> forms = feature.getLayer().getForms();
     loadObservations(
-        feature.getProject(), feature.getLayer().getId(), form.getId(), feature.getId());
+        feature.getProject(),
+        forms.isEmpty() ? Optional.empty() : Optional.of(forms.get(0).getId()),
```
scolsen: Are we eventually going to support multiple forms again, or will `forms` eventually become `form`?

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/ObservationListViewModel.java
```diff
@@ -76,22 +84,16 @@ public void loadObservationList(Feature feature, Form form) {
...
   }
 
-  private void loadObservations(Project project, String layerId, String formId, String featureId) {
-    Optional<Form> form = project.getLayer(layerId).flatMap(pt -> pt.getForm(formId));
-    if (!form.isPresent()) {
-      // TODO: Show error.
-      return;
-    }
-    // TODO: Use project id instead of object.
+  private void loadObservations(Project project, Optional<String> formId, String featureId) {
```
scolsen: nit: Parallelize parameter positions? 

loadObservations:: project, formId, featureId
ObservationListRequest: project, featureId, formId

Swap feature and form in one or the other?

gnd/build.gradle
gino-m: Actually it did! I removed it because it was slowly down development significantly, but we can discuss this separately. Added it back.

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/FeatureSheetFragment.java
gino-m: Fixed!

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/ObservationListViewModel.java
gino-m: Good catch. Done!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
gino-m: Added TODO to hide Add Observation button instead, since it shouldn't be shown if form is not defined (i.e., this case should never happen).

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/ObservationListViewModel.java
gino-m: The idea is that we'll keep it in the data store schema, but not in the UIs. This way custom or future clients can use multiple forms without breaking the database.

gino-m:
Thanks for the review! RFAL.

gnd/src/main/java/com/google/android/gnd/ui/home/featuresheet/ObservationListViewModel.java
scolsen: Gotcha, maybe this is worthy of a helper method to make sure we always call `.get(0)` for now to fetch only the first form and don't accidentally call `.get(0±n)`? (Not that I anticipate us needing to call this in many other places in the app atm).

gino-m: I'll go you one better - pushing the cardinality change to the persistence layers, so that the application model doesn't reflect this idiosyncrasy in the data models. Please see latest commit 56cfa3df66cd7312961db3633c0fd06e919750c1.

gino-m:
@scolsen should be ready for approval!

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/LayerConverter.java
```diff
@@ -33,10 +35,12 @@ static Layer toLayer(String id, LayerNestedObject obj) {
     if (obj.getDefaultStyle() != null) {
       layer.setDefaultStyle(StyleConverter.toStyle(obj.getDefaultStyle()));
     }
-    if (obj.getForms() != null) {
-      Maps.forEach(
-          obj.getForms(),
-          (formId, formDoc) -> layer.addForm(FormConverter.toForm(formId, formDoc)));
+    if (obj.getForms() != null && !obj.getForms().isEmpty()) {
```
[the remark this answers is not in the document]
gino-m: That's exactly right.. It would if we controlled how the field was populated, but since we don't know if the Firebase API will set a null or empty we need to check both.

PR #349, opened 2020-02-11
Tolerate null style definitions in project config in local db

gino-m:
We don't have string schema enforcement for remote data, so being more laxed about nullness checking in local db as well.


gino-m:
@scolsen Gentle ping!

PR #368, opened 2020-02-15
[Photos support] Load saved photo in ObservationDetailsFragment

shobhitagarwal1612:
Towards #310 

Also fixes a bug for saving photo field to local/remote db

image


gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -138,6 +138,9 @@
   /** True if the observation is being added, false if editing an existing one. */
   private boolean isNew;
 
+  /** True if the photo field is immediately updated. */
```
gino-m: Wording here is unclear: True if the photo field has been updated?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -267,6 +270,7 @@ private Completable handlePhotoCaptureResult(String fieldId) {
             })
         .doOnNext(
             url -> {
+              isPhotoFieldUpdated = true;
```
gino-m: May be out of scope of this PR, but it would seem that saving the bitmap should be an asynchronous operation, whereas here it's modeled as a synchronous one. 

Also, the Rx chain seems to be relying on side effects to do critical work, which is general not harder to debug and maintain. It might be clearer if this method would accept `Bitmap` instead of an `Observable<Bitmap>`, save and upload asynchronously, and return a `Completable` or `Single<String>` when done. Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -316,7 +320,15 @@ public void onSaveClick() {
 
   private void onObservationLoaded(Observation observation) {
     this.originalObservation = observation;
-    refreshResponseMap(observation);
+
+    // Photo field is updated by launching an external intent. This causes the form to reload.
+    // When that happens, we don't want to lose the unsaved changes.
```
gino-m: Wouldn't this be equivalent to opening another app while editing the form? If so, how is this working today for non-photo fields, and why are photos different?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
shobhitagarwal1612: We're handling the result of the activity before the state of the response is reset. To fix this, we should call the external intent from the activity and handle for results in `onObservationLoaded` after `refreshResponseMap` call.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
shobhitagarwal1612: The reason is we handle the result of the photos before the response map is refreshed. As mentioned in the previous comment, this is a side-effect of an earlier pending refactor.

gino-m: If the `initialize` is getting called whenever we reopen the Fragment, I'm not sure how this is actually working today.  There are many reasons the fragment could get reinitialized; it's hard to name `isPhotoFieldUpdated` appropriately because it's not really well defined - iiuc it's asking "am I being reinitialized because the user left to a photo intent and came back?" In reality I don't think we can know that for sure, and as result there are likely many edge cases where this will fail.

I may be a bit out of my depth on Android fragment/VM lifecycles, but I'll approve for now to unblock current work. @dturner, would you mind taking a look at the `EditObservation*` logic when you get a chance to see how we might improve?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
gino-m: See comment below re `isPhotoFieldUpdated`.

Could you kindly collapse the chained `map` and `doOnNext` into a single function and use a more appropriate Rx method for chaining? Might look something like this:

```java
  private Completable handlePhotoCaptureResult(String fieldId) {
    return cameraManager
        .capturePhotoResult()
        .flatMapCompletable(bitmap -> saveBitmapAndUpdateResponse(bitmap, fieldId));
  }

  private Completable saveBitmapAndUpdateResponse(
      Bitmap bitmap, String fieldId) {
    String file = fileUtil.saveBitmap(bitmap, fieldId + ".jpg");
    String destinationPath = getRemoteImagePath(file.getName());
    String url = firestoreStorageManager.uploadMediaFromFile(file, destinationPath);
    isPhotoFieldUpdated = true;
    onTextChanged(form.getValue().getField(fieldId).get(), url);
    return Complete.complete()
  }
```

I assume we'd want `capturePhotoResult()` to run on the UI thread, and `saveBitmapAndUpdateResponse()` to run in the IO thread. @dturner, what's the right way to do this? I vaguely remember downstream `flatMap` logic runs on the same thread as the upstream subscribers (defined with `subscribeOn`), while `switchMap` does not, but I can't find a reference for it.

PR #370, opened 2020-02-16
[Developer workflow] Fixes "No classes configured for SpotBugs analysis" bug

shobhitagarwal1612:
Run check code right before installing the app for testing rather than build.


shobhitagarwal1612:
> a) we run these on commit instead

I'll revisit git-hooks some time later. 

> b) Android Studio will give realtime warnings for most of these.. but it seems that's not possible yet.

Using android studio plugins might be helpful. But the challenge is to load our custom config.

PR #371, opened 2020-02-16
[Remote sync] Refactor remote persistence layer

gino-m:
- [x] Splits conversion logic out from remote entity classes
- [x] Makes remote entity classes immutable
- [x] Renames and moves all entities and converters into `com.google.android.gnd.persistence.remote.firestore`
- [x] Additional error checking for Features and Observations
- [x] Other minor cleanup around remote persistence

We should also handle null fields in the `Project` document correctly, but saving this for a follow-up PR since this one is already a getting to be a scary beast!


gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
```diff
@@ -91,30 +97,37 @@ public Completable applyMutations(ImmutableCollection<Mutation> mutations, User
   private Task<?> applyMutationsInternal(ImmutableCollection<Mutation> mutations, User user) {
     WriteBatch batch = db.batch();
     for (Mutation mutation : mutations) {
-      addMutationToBatch(mutation, user, batch);
+      try {
+        addMutationToBatch(mutation, user, batch);
+      } catch (DataStoreException e) {
```
scolsen: Should we use exceptions for this case? 

I'm of the opinion that exceptions are more useful for states that are actually unrecoverable/catastrophic--places where crashing w/ a message + trace is a valid course of action, whereas when we do stuff like ignore or log a potential value we don't care about and otherwise continue running some kind of value wrapper type is more appropriate—it doesn't make much of a difference at the end of the day, so I would bother if you have a different stance on this.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.android.gnd.model.User;
+import com.google.firebase.Timestamp;
+import java.util.Date;
+import java8.util.Optional;
+
+/** Converts between Firestore nested objects and {@link AuditInfo} instances. */
+class AuditInfoConverter {
+
+  /**
+   * Converts a POJO representing user and timestamp data in Firebase into an equivalent model
```
scolsen: Is it really a POJO? Doesn't it depend on Firebase?

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
```diff
@@ -0,0 +1,56 @@
...
+   * object. This should never be empty in Firebase, but this method returns a default value when
+   * the input is null to support legacy or corrupt dbs.
+   */
+  @NonNull
+  static AuditInfo toAuditInfo(@Nullable AuditInfoNestedObject doc) {
+    if (doc == null || doc.getClientTimeMillis() == null) {
```
scolsen: If we know the user, shouldn't we ascribe the audit to the user even if the time is somehow unknown?

scolsen: We could also return defaults from the AuditInfoNestedObject getters themselves -- since we're accounting for a peculiarity of firebase and not our model. 

Plus that ensures the work performed by the converter is actual conversion from store values to model values (timestamp->date) and we don't give it the double duty of defining defaults for store values.

scolsen: I don't have a particularly strong opinion on this though, so up to you.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/FormConverter.java
```diff
@@ -0,0 +1,34 @@
...
+import static java8.util.stream.StreamSupport.stream;
+
+import com.google.android.gnd.model.form.Form;
+
+/** Converts between Firestore nested objects and {@link Form} instances. */
+class FormConverter {
+
+  static Form toForm(String formId, FormNestedObject obj) {
+    return Form.newBuilder()
+        .setId(formId)
```
scolsen: Is there a reason we don't store formIds on forms themselves?

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ProjectConverter.java
```diff
@@ -14,36 +14,28 @@
...
-  @Nullable public Map<String, String> description;
-
-  // TODO: Add AuditInfoDoc for created and lastModified.
-
-  @Nullable public Map<String, LayerDoc> featureTypes;
-
-  public static Project toObject(DocumentSnapshot doc) {
-    ProjectDoc pd = doc.toObject(ProjectDoc.class);
+  static Project toProject(DocumentSnapshot doc) throws DataStoreException {
+    ProjectDocument pd = doc.toObject(ProjectDocument.class);
```
scolsen: Instead of doing a conversion here, could we just add a single layer of indirection to the DocSnapshots? -- so that their first interior/nested field is the representation of a Project, instead of having them represent the project itself? --or would that be rather fussy/pointless?

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
gino-m: If exceptions were unrecoverable, then there'd be no need for try/catch. Sun says "An exception is an event that occurs during the execution of a program that disrupts the normal flow of instructions." Issues like these with bad input data prevent the deserialization of remote entities, but don't break the entire app. Where we handle these and how obviously an art and not a science, and I agree sometimes it may make sense to use value wrappers instead of exceptions. But for values that really really need to be there for the data be used (i.e., we're dropping data, the database is corrupt) an exception definitely seems like the right mechanism. 

@dturner in case you have anything to add!

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
gino-m: Good question... I *think* in this case they're still considered POJOs in the general sense in that they implement no special interfaces or external conventions; the annotations only provide hints for Firebase, but they're not required to use or create instances.

> The idea is that if the object (actually class) were a POJO before any annotations were added, and would return to POJO status if the annotations are removed then it can still be considered a POJO. Then the basic object remains a POJO in that it has no special characteristics (such as an implemented interface) that makes it a "Specialized Java Object" (SJO or (sic) SoJO).
https://en.wikipedia.org/wiki/Plain_old_Java_object#Definition

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/FormConverter.java
gino-m: Because they come from the enclosing map, e.g.
```
{
   'id1234': {
     // nested object
   }
}
```

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
scolsen: Cool, sounds good! Thanks for clarifying. 

Might be best to just drop the term in this case? I don't think it adds much atm -- I think just calling it a converter to/from Firebase representation to the model is clear and effective.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
gino-m: > We could also return defaults from the AuditInfoNestedObject getters themselves -- since we're accounting for a peculiarity of firebase and not our model.

This would split out part of the conversion of remove data, making the conversion harder to debug.

> Plus that ensures the work performed by the converter is actual conversion from store values to model values (timestamp->date) and we don't give it the double duty of defining defaults for store values.

I think it makes sense to a canonical representation of what's actually in Firebase separate from the handling of that data. It makes things easier to debug, and consolidates all logic that transforms that data (defaults, fallbacks, or otherwise) into a single place.

> If we know the user, shouldn't we ascribe the audit to the user even if the time is somehow unknown?

Actually now that I think about it, an entity without a time should probably just be considered invalid, since it's one of the only identifying attributes of features and observations. Users, on the other hand, could delete their accounts, meaning we would need to delete the user data.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
scolsen: That sounds reasonable to me. I'm fine with using them--I think at times they compete with alternatives like returning Result(x)--thanks for the explanation!

But i will drop [this chart](https://www.progsbase.com/blog/flow-charts-of-programming-language-constructs/#11.-exception) :)

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
scolsen: Sounds good!

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/FormConverter.java
scolsen: I see, thanks!

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ProjectConverter.java
scolsen: Based on:

> Because they come from the enclosing map, e.g.
> 
> ```
> {
>    'id1234': {
>      // nested object
>    }
> }
> ```

Seems like it'd be a pain to do what I'm suggesting/asking here

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
gino-m: Fixed! Relevant patch: ab1b06fb643e4de3e4d029cedbc93f575535674d

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/AuditInfoConverter.java
gino-m: Done (9f16c32f4670c2d8313c87958891da0063c71c21)! I was going to update the comment, but ended up deleting instead, since other converter methods don't have method-level JavaDoc, and what they're doing is mostly clear from the class JavaDoc.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
gino-m: +1, especially when we need a was to propagate errors in streams.

I haven't seen that before, but at first glance it seems they may be confusing code readability with graph complexity, which I intuitively believe may not be strictly correlated. The abstraction provided by exception propagation could allow you to represent the call stack as a single box, in some sense... but I could be wrong.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ProjectConverter.java
gino-m: Firebase client libraries return you `DocumentSnapshot`s; we'd still need to ask the API to convert them to objects for us someplace.

gino-m: If we wanted to abstract his I suppose we could define the classes to convert to declaratively in some sort of a registry, then define a standard interface for converters, but that would lead to a of boilerplate, and not allow us to provide custom args to converter methods as we do now, forcing us to propose the args to member variables of hypothetical converter instances. Oy. 🙄 Open to other ideas if you had something else in mind though!

gino-m:
> 👏 Is this still a WIP? Or are you ready to merge?

Ready, sir! Was waiting for checks to pass, and for your approval!

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java
scolsen: That's a good point! I guess I just think things are easier to reason about the less we mix styles--but it probably is simpler to go with exceptions in certain cases--I just hope doing things on a case-by-case basis doesn't make it more difficult to alter things in the long run. But I think our choices/boundaries are pretty clear right now, so it probably won't be an issue.

scolsen:
> > 👏 Is this still a WIP? Or are you ready to merge?
> 
> Ready, sir! Was waiting for checks to pass, and for your approval!

Looks good, but I am running into a few performance issues when running the app on this branch. It also looks like we're rendering multiple choice options multiple times:

Screen Shot 2020-02-19 at 11 23 08 AM
Screen Shot 2020-02-19 at 11 23 49 AM

The app in general is chugging a lot on this branch compared to master.

Maybe do a double check on this? --the performance problems could be peculiar to my setup, but the multiple choice duplication is new on this branch (as far as I can tell)

gino-m:
Looking into this. I am also seeing the following:
```
2020-02-19 17:09:22.994 29712-29712/com.google.android.gnd E/RxDebug: Unhandled Rx error
    java.lang.Throwable: No properties to serialize found on class com.google.android.gnd.persistence.remote.firestore.schema.FeatureDocument
        at com.google.firebase.firestore.util.CustomClassMapper$BeanMapper.<init>(com.google.firebase:firebase-firestore@@21.3.1:714)
        at com.google.firebase.firestore.util.CustomClassMapper.loadOrCreateBeanMapperForClass(com.google.firebase:firebase-firestore@@21.3.1:377)
        at com.google.firebase.firestore.util.CustomClassMapper.convertBean(com.google.firebase:firebase-firestore@@21.3.1:540)
        at com.google.firebase.firestore.util.CustomClassMapper.deserializeToClass(com.google.firebase:firebase-firestore@@21.3.1:253)
        at com.google.firebase.firestore.util.CustomClassMapper.convertToCustomClass(com.google.firebase:firebase-firestore@@21.3.1:100)
        at com.google.firebase.firestore.DocumentSnapshot.toObject(com.google.firebase:firebase-firestore@@21.3.1:210)
        at com.google.firebase.firestore.QueryDocumentSnapshot.toObject(com.google.firebase:firebase-firestore@@21.3.1:116)
        at com.google.firebase.firestore.DocumentSnapshot.toObject(com.google.firebase:firebase-firestore@@21.3.1:188)
        at com.google.firebase.firestore.QueryDocumentSnapshot.toObject(com.google.firebase:firebase-firestore@@21.3.1:97)
        at com.google.android.gnd.persistence.remote.firestore.schema.FeatureConverter.toFeature(FeatureConverter.java:37)
        at com.google.android.gnd.persistence.remote.firestore.schema.FeaturesCollectionReference.lambda$toRemoteDataEvents$1(FeaturesCollectionReference.java:46)
        at com.google.android.gnd.persistence.remote.firestore.schema.-$$Lambda$FeaturesCollectionReference$pZM8llkKADx-Iu6OKgJCl6BtjYQ.apply(Unknown Source:4)
        at com.google.android.gnd.persistence.remote.firestore.schema.QuerySnapshotConverter.toEvent(QuerySnapshotConverter.java:55)
        at com.google.android.gnd.persistence.remote.firestore.schema.QuerySnapshotConverter.lambda$toEvents$0(QuerySnapshotConverter.java:42)
        at com.google.android.gnd.persistence.remote.firestore.schema.-$$Lambda$QuerySnapshotConverter$q3Ornk4k4Bx5XBQwy8u5mcLNy_k.apply(Unknown Source:4)
        at java8.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:196)
        at java8.util.DelegatingSpliterator$ConsumerDelegate.accept(DelegatingSpliterator.java:93)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1390)
        at java8.util.DelegatingSpliterator.forEachRemaining(DelegatingSpliterator.java:57)
        at java8.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
        at java8.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
        at java8.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:1269)
        at java8.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java8.util.stream.ReferencePipeline.collect(ReferencePipeline.java:602)
        at com.google.android.gnd.persistence.remote.firestore.schema.QuerySnapshotConverter.toEvents(QuerySnapshotConverter.java:43)
        at com.google.android.gnd.persistence.remote.firestore.schema.FeaturesCollectionReference.toRemoteDataEvents(FeaturesCollectionReference.java:45)
        at com.google.android.gnd.persistence.remote.firestore.schema.FeaturesCollectionReference.lambda$loadOnceAndStreamChanges$0(FeaturesCollectionReference.java:40)
        at com.google.android.gnd.persistence.remote.firestore.schema.-$$Lambda$FeaturesCollectionReference$QNEdO5fu7f_msfYGOUBLEj0nn4Q.apply(Unknown Source:4)
        at io.reactivex.internal.operators.flowable.FlowableFlattenIterable$FlattenIterableSubscriber.drain(FlowableFlattenIterable.java:265)
        at io.reactivex.internal.operators.flowable.FlowableFlattenIterable$FlattenIterableSubscriber.onNext(FlowableFlattenIterable.java:174)
        at com.akaita.java.rxjava2debug.extensions.FlowableOnAssembly$OnAssemblySubscriber.onNext(FlowableOnAssembly.java:61)
        at io.reactivex.internal.operators.flowable.FlowableCreate$NoOverflowBaseAsyncEmitter.onNext(FlowableCreate.java:403)
        at durdinapps.rxfirebase2.RxFirestore$18$1.onEvent(RxFirestore.java:530)
```

Investigating..

gino-m:
``

gino-m:
> Investigating..

Looking at the source, it seems POJO conversion with `DocumentSnapshot.toObject()` requires a default constructor and public accessors:
https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentSnapshot#public-methods_1

gino-m:
@scolsen Fixed! PTAL 6bf722f58ef3fca00f1459b0cfba85b722119a6e in particular. Thanks!

gino-m:
> Awesome! 😎

Whooooohooo!

Issue #364, opened 2020-02-13
[Map] Adjust marker size based on zoom level


gino-m:
Zoom level < 16:
![image](https://user-images.githubusercontent.com/228050/74393992-8f8f5a00-4dd9-11ea-8d6d-0168520b12be.png)

Zoom level >= 16:
![image](https://user-images.githubusercontent.com/228050/74393978-84d4c500-4dd9-11ea-9072-6809f00bb60e.png)

JSunde:
I think I can take a stab at this. Gino, do you have any suggestions for modifying the scale on the fly? It looks like today the marker is created once in the MarkerIconFactory, so changing the size based on the zoom level there won't have any effect since the zoom level will always be the default value when the marker is created.

gino-m:
> I think I can take a stab at this. Gino, do you have any suggestions for modifying the scale on the fly? It looks like today the marker is created once in the MarkerIconFactory, so changing the size based on the zoom level there won't have any effect since the zoom level will always be the default value when the marker is created.

Sure, please go ahead! I think you may need to handle this in the view model when the camera changes, updating the MapMarkers accordingly. Please ping if you get stuck and we can look at it together!

PR #372, opened 2020-02-18
[Map] Use custom Ground marker for point features

### LINKED ISSUE #117: [Map] Update default map marker style

### LINKED ISSUE #171: [Map] Respect default style (color) for markers

gino-m:
User-defined fill color required building the bitmap in memory out of constituent vector parts, since it's not possible to edit a single vector path in Android without extra libraries or workarounds. 

Fixes #117, fixes #171. #364, #373 still pending.

Screenshot_1581998363

Screenshot_1581998397

@hintzemichael w00t! 🎉


shobhitagarwal1612:
It doesn't right to have 3 separate xmls for a single icon. Is it possible to have a single xml and dynamically update color using a custom `attr`

gnd/src/main/java/com/google/android/gnd/ui/MapIcon.java
```diff
@@ -17,80 +17,54 @@
...
-      int resourceId =
-          context.getResources().getIdentifier(resourceName, "drawable", context.getPackageName());
-      if (resourceId > 0) {
-        return resourceId;
-      }
-    } catch (Resources.NotFoundException e) {
-      Log.w(TAG, e);
+      return Color.parseColor(colorHexCode);
+    } catch (IllegalArgumentException e) {
+      return context.getResources().getColor(R.color.colorMapAccent);
```
shobhitagarwal1612: In addition, let's log an error as well to keep track of such events.

gino-m: Good idea! In the process I've done a little more cleanup that I'd probably never get back to otherwise: 50049aaea418d44ff6c3fb78d584947b4acb2174

gino-m:
> It doesn't right to have 3 separate xmls for a single icon. Is it possible to have a single xml and dynamically update color using a custom `attr`
> 
> Update:
> I tried to look for a solution which would selectively allow us to color only a single path.
> 
> * custom attribute can be used but that can't be updated on runtime. So, all color must be defined as separate themes (not much usable for us then)
> 
> Let's keep it as it for now and look for a possible solution in the future!

Exactly, that's how I landed on this solution - it seemed the least bad option. Thanks for looking into it too!

gino-m:
By the way, the three-part vector will also be useful when introduce icon overlays (e.g., preloaded or user-specified symbols that can be overlaid inside the map marker).

gnd/src/main/java/com/google/android/gnd/ui/MapIcon.java
shobhitagarwal1612: Thanks!

gino-m:
> app:srcCompat

Thanks, this appeared after merging with your recent PR. Should be fixed now.

PR #375, opened 2020-02-18
Fix path of includes in base checkstyle config

gino-m:
This was resulting in an error when running checkstyle (and therefore post-build):
```
Unable to create Root Module: config {/Users/gmiceli/GitProjects/gino-m/ground-android/third_party/checkstyle/google_checks.xml}, classpath {null}.
```

Not sure what changed to cause this, but it looks like checkstyle is being run from a different base dir.


gino-m:
Looks like this fix works locally but not on GCB. @shobhitagarwal1612 any ideas?

shobhitagarwal1612:
This can be closed now. Fixed by pushing the commit directly to #372

shobhitagarwal1612:
I don't see any errors on GCB
(latest build on master): https://console.cloud.google.com/gcr/builds/43cb0da6-d56a-47e5-9a6b-8b4929191ac2?project=264658303667

I am not sure why this fix is needed. For locally, we should run the task from root dir.

gino-m:
I was only encountering this locally, but after cleaning project, rebuilding, and rerunning checkCode it passes. Closing!

PR #376, opened 2020-02-19
Stop running checkCode on install

gino-m:
Auto running locally has been causing several issues so disabling for now. @shobhitagarwal1612 FYI. See also conversation on #342 and in https://gitter.im/ground-platform/ground-contributors.


gino-m:
> This looks fine, but I still run into issues about 90% of the time I attempt to run checkCode in isolation.
> 
> Usually having to do with `SpotBugs`

I think what's happening is that there are other issues (compile? PMD?) that cause there to be not classes generated, but spotbugs is being run anyway. Filed #377 to track.

shobhitagarwal1612:
I'll try exploring the android studio plugins if that solves our problem

PR #386, opened 2020-03-12
[Photos Support] Improve upload photo UI

shobhitagarwal1612:
Towards #310 

- Bug fix: Display thumbnail after selecting a photo while in EditObersationFragment

- UI tweaked https://github.com/google/ground-android/issues/310#issuecomment-597661636


gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -196,14 +196,13 @@ public void initialize(EditObservationFragmentArgs args) {
...
-    Log.v(TAG, "onTextChanged: " + field.getId());
+    Timber.v("onTextChanged: %s", field.getId());
 
     onResponseChanged(field, TextResponse.fromString(text));
   }
 
   public void onResponseChanged(Field field, Optional<Response> newResponse) {
-    Log.v(
-        TAG, "onResponseChanged: " + field.getId() + " = '" + Response.toString(newResponse) + "'");
+    Timber.v("onResponseChanged: " + field.getId() + " = '" + Response.toString(newResponse) + "'");
```
gino-m: With Timber can we now just do to following?
```
Timber.v("onResponseChanged: %s = '%s'", field.getId(), Response.toString(newResponse));
```

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -392,12 +391,12 @@ private void refreshResponseMap(Observation obs) {
...
   private ImmutableList<ResponseDelta> getResponseDeltas() {
     if (originalObservation == null) {
-      Log.e(TAG, "Response diff attempted before observation loaded");
+      Timber.e("Response diff attempted before observation loaded");
       return ImmutableList.of();
     }
     ImmutableList.Builder<ResponseDelta> deltas = ImmutableList.builder();
     ResponseMap originalResponses = originalObservation.getResponses();
-    Log.v(TAG, "Responses:\n Before: " + originalResponses + " \nAfter:  " + responses);
+    Timber.v("Responses:\n Before: " + originalResponses + " \nAfter:  " + responses);
```
gino-m: Same here

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
```diff
@@ -118,4 +125,29 @@ public static void setOnFocusChangeListener(
   public static void setOnShowDialogListener(MultipleChoiceFieldLayout view, Runnable listener) {
     view.setOnShowDialogListener(listener);
   }
+
+  @BindingAdapter({"bind:storageManager", "bind:response"})
+  public static void bindPhoto(ImageView view, StorageManager storageManager, Response response) {
```
gino-m: This is a lot of logic to be doing inside a binding adapter, especially the hiding/showing of the parent view. Can we instead define a separate view widget that can be tested and implemented independently? It could have its own ViewModel with LiveData containing the actual image to display. Wdyt?

shobhitagarwal1612: Makes sense. Then we can reuse it at other places as well.

shobhitagarwal1612: I've added a custom view + a view model for it. But this is again having the original problem of image view not getting updated (nor the visibility of it's container). Can you please have a look? I tried debugging it but couldn't find anything useful.

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
```diff
@@ -118,4 +129,13 @@ public static void setOnFocusChangeListener(
   public static void setOnShowDialogListener(MultipleChoiceFieldLayout view, Runnable listener) {
     view.setOnShowDialogListener(listener);
   }
+
+  @BindingAdapter("bind:uri")
+  public static void bindUri(ImageView view, Uri uri) {
+    if (uri == null) {
+      return;
```
gino-m: Do we want to set the placeholder here in case Uri is null?

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoView.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public PhotoView(Context context, AttributeSet attrs, int defStyleAttr) {
+    super(context, attrs, defStyleAttr);
+
+    LayoutInflater inflater =
+        (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+    binding = PhotoViewBinding.inflate(inflater);
+  }
+
+  public void setResponse(@Nullable Response response) {
```
gino-m: Since Views are tightly bound to the Android OS and running VM, in general we're trying keep views as "dumb" as possible. This way we can test all/most business logic using ViewModels and upstream classes only. So ideally the views would delegate interaction with custom model classes or logic to the ViewModel. In most cases, that means the View can go away altogether, leaving us with just a layout and viewmodel. Would that be possible in this case?

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoView.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public void setViewModel(PhotoViewViewModel viewModel) {
+    if (viewModel == null) {
+      return;
+    }
+    viewModel.setResponse(response);
+    binding.setViewModel(viewModel);
+  }
+
+  public interface Callback {
```
gino-m: In a non-Rx world we'd probably just use `Consumer<Uri>` instead, but in our case we can just use an Observable<Uri>.

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoViewViewModel.java
```diff
@@ -0,0 +1,52 @@
...
+import androidx.databinding.ObservableInt;
+import androidx.lifecycle.LiveData;
+import androidx.lifecycle.MutableLiveData;
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import com.google.android.gnd.system.StorageManager;
+import com.google.android.gnd.ui.common.AbstractViewModel;
+import javax.inject.Inject;
+
+public class PhotoViewViewModel extends AbstractViewModel {
```
gino-m: Naming: PhotoFieldViewModel, to make it clear that this is specific to photo fields in forms, not just any photo widget?

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoViewViewModel.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private final StorageManager storageManager;
+  private final MutableLiveData<Uri> destinationPath = new MutableLiveData<Uri>();
+  private final ObservableInt isDestinationPathAvailable = new ObservableInt(View.GONE);
```
gino-m: This should probably be calling this "visibility" instead given the value it returns and how it's used.

gino-m: Also, we should use `LiveData` since `ObservableInt` isn't Lifecycle aware.

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoView.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public PhotoView(Context context, AttributeSet attrs) {
+    this(context, attrs, 0);
+  }
+
+  public PhotoView(Context context, AttributeSet attrs, int defStyleAttr) {
+    super(context, attrs, defStyleAttr);
+
+    LayoutInflater inflater =
+        (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
```
gino-m: `LayoutInflater inflater = LayoutInflater.from(context);` can be used here instead.

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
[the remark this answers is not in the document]
shobhitagarwal1612: Ideally, when this happens the visibility of MaterialCard should be set to GONE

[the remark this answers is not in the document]
gino-m: When could this actually be null? If it happens, the widget's previous state will remain. At some point, when it's made visible, the old state will show even though Uri is null, which is obviously incorrect. It would seem we'd want the state of the widget to match to current field binding, but I may be missing a piece!

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
gino-m: Great job finding this issue, I looked over the weekend but didn't find it!

Do you think we can let the binding take a Uri instead of a StorageManager and response to prevent having so much logic in the binder?

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -82,7 +78,7 @@ private Completable sendPhotoPickerIntent() {
                   Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
                   intent.setType("image/*");
                   activity.startActivityForResult(intent, PICK_PHOTO_REQUEST_CODE);
-                  Log.d(TAG, "file picker intent sent");
+                  Timber.d("file picker intent sent");
```
gino-m: This method is UI-specific so should probably be defined in the UI layer (e.g., the Fragment) or related helper.

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -120,35 +116,24 @@ private File getLocalFileFromDestinationPath(String destinationPath)
...
-              // Load file locally
-              File file = getLocalFileFromDestinationPath(destinationPath);
-              Picasso.get()
-                  .load(file)
-                  .placeholder(R.drawable.ic_photo_grey_600_24dp)
-                  .into(imageView);
-            })
+        .doOnSuccess(callback::run)
+        .doOnError(throwable -> callback.run(getFileUriFromDestinationPath(destinationPath)))
         .subscribe();
```
gino-m: Sorry if I mentioned it before, but this probably won't work. Since you need to properly manage the returned disposable, we only subscribe to streams in lifecycle-aware contexts like Fragments. The right thing to do here I believe, would be to return a Single<Uri>, which can be created using RxTask.

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoView.java
```diff
@@ -0,0 +1,58 @@
...
+import android.view.LayoutInflater;
+import androidx.fragment.app.Fragment;
+import com.google.android.gnd.databinding.PhotoViewBinding;
+import com.google.android.gnd.model.observation.Response;
+import com.google.android.gnd.system.StorageManager;
+import com.google.android.material.card.MaterialCardView;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+
+public class PhotoView extends MaterialCardView {
```
gino-m: It's probably my fault for being vague in previous comments, but we're generally trying to avoid creating custom Views unless absolutely necessary, since they're hard to test and easy to get wrong (as you discovered)! Instead, could this view be reduced to a layout xml that can be included in the photo_input_field layout, which in turn gets added by the Fragment?

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
shobhitagarwal1612: Added a TODO for future refactoring

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
shobhitagarwal1612: Thanks for catching this! This simplified the code greatly.

gnd/src/main/java/com/google/android/gnd/ui/common/photoview/PhotoView.java
shobhitagarwal1612: Thanks for the idea. It all looks better than the first approach I took.

shobhitagarwal1612:
@gino-m This is ready for your final review. Thank you!

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
```diff
@@ -118,4 +126,9 @@ public static void setOnFocusChangeListener(
   public static void setOnShowDialogListener(MultipleChoiceFieldLayout view, Runnable listener) {
     view.setOnShowDialogListener(listener);
   }
+
+  @BindingAdapter("imageUri")
+  public static void bindUri(ImageView view, Uri uri) {
+    Picasso.get().load(uri).placeholder(R.drawable.ic_photo_grey_600_24dp).into(view);
```
gino-m: Do we need to handle error states here? Will failing to load the URI cause the application to crash?

gnd/src/main/java/com/google/android/gnd/ui/common/ViewModelModule.java
```diff
@@ -96,6 +97,11 @@
   @ViewModelKey(EditObservationViewModel.class)
   abstract ViewModel bindEditObservationViewModel(EditObservationViewModel viewModel);
 
+  @Binds
+  @IntoMap
+  @ViewModelKey(PhotoFieldViewModel.class)
+  abstract ViewModel bindPhotoViewViewModel(PhotoFieldViewModel viewModel);
```
gino-m: Naming nit: Our classes follow the pattern FooView and FooViewModel, which would make this class here "PhotoViewModel". To make it clear this is specifically a photo in a form field, PhotoFieldView and PhotoFieldViewModel might be more appropriate.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+          @Override
+          public void onMapChanged(ObservableMap<String, Response> sender, String key) {
+            if (key.equals(field.getId())) {
+              updateUI(sender.get(key), field);
+            }
+          }
+        });
+  }
+
+  private void updateUI(Response response, Field field) {
```
gino-m: Nit: Please use a more specific method name like `updateField`. In general avoid capital case in Java method names (`updateUi` instead of `updateUI`).

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+    if (response == null) {
+      photoPreviewVisibility.setValue(View.GONE);
+    } else {
+      String value = response.getDetailsText(field);
+      if (value.isEmpty()) {
+        photoPreviewVisibility.setValue(View.GONE);
+      } else {
+        photoPreviewVisibility.setValue(View.VISIBLE);
+        disposeOnClear(
```
gino-m: This will accumulate subscriptions in the viewmodel ad infinitum every time a photo is changed during the viewmodel's lifecycle. That's probably ok in practice since the user will never likely do this so many times we'll run out of memory, but it is a form of memory leak that as a general rule we're designing to avoid.

The "right", safe way to do this would be to create a `PublishSubject<Optional<String>> destinationPath` class member that emits the path or empty when this method is called. Then in the constructor, you would define `LiveData<Optional<Uri>> uri` using something like:
```java
  this.uri = LiveDataReactiveStreams.fromPublisher(
    destinationPath.switchMap(p -> p.map(storageManager::loadUriFromDestinationPath)));
```

Visibility could be defined in a similar way:
```java
  this.uri = LiveDataReactiveStreams.fromPublisher(
    destinationPath.map(p -> p.isPresent() ? View..VISIBLE : View.GONE));
```

This method would then simply call `destinationPath.onNext` with either `Optional.empty()` or the URI.

Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import com.google.android.gnd.system.StorageManager;
+import com.google.android.gnd.ui.common.AbstractViewModel;
+import javax.inject.Inject;
+
+public class PhotoFieldViewModel extends AbstractViewModel {
+
+  private final StorageManager storageManager;
+  private final MutableLiveData<Uri> destinationPath = new MutableLiveData<>();
```
gino-m: I'm a bit confused by the naming here - it appears Uris are loaded from destination paths, therefore they're not equivalent. If that's the case, shouldn't this be called `uri` instead?

gnd/src/main/java/com/google/android/gnd/ui/observationdetails/ObservationDetailsFragment.java
```diff
@@ -159,7 +163,18 @@ private void addField(Field field, Observation observation) {
                 binding.fieldValue.setVisibility(View.GONE);
                 binding.imagePreview.setVisibility(View.VISIBLE);
 
-                storageManager.loadPhotoFromDestinationPath(binding.imagePreview, value);
+                storageManager
```
gino-m: Same here - subscriptions should only be made inside lifecycle methods so they can be properly disposed of in their equivalent end lifecycle methods. To do this safely, you can create a PublishSubject in the view model that gets exposed to the fragment as a LiveData. Let's discuss if it's not clear how to do this here.

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
shobhitagarwal1612: No, that's why we have a placeholder. If anything unexpected happens, the app doesn't crash and the imageview just displays placeholder.

I tested network not-available and null cases

gino-m: Great, thanks for checking! Could you kindly add a comment for posterity since it's not obvious at first glance?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
shobhitagarwal1612: Looks so clean. ✨

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
shobhitagarwal1612: My bad, thanks for pointing it out

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
gino-m: 🌟

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
gino-m: Np! Thanks for all the careful code hygiene overall!

gnd/src/main/java/com/google/android/gnd/ui/editobservation/PhotoFieldViewModel.java
shobhitagarwal1612: I am having a hard time creating the `uri` LiveData object.

For visibility, the below code works
```
    photoPreviewVisibility =
            destinationPath
                .map(path -> path.isPresent() ? View.VISIBLE : View.GONE)
                .toFlowable(BackpressureStrategy.LATEST));
```

But for Uri, I can't get the correct RxChain

`loadUriFromDestinationPath(path)` returns `Single<Uri>`

gino-m: If you replace destinationPath with a `PublishProcessor` you can get rid of the .toFlowable. Looking into the right syntax for your case now.

shobhitagarwal1612: To be precise, I can't convert Single<Uri> to Optional<Uri> in the below chain. The only way I can think of is by subscribing `loadUriFromDestinationPath` inside `map` but that would defeat the entire purpose of this refactor.

```
    uri =
            Flowable.fromPublisher(destinationPath)
                .map(Optional::get)
                .map(storageManager::loadUriFromDestinationPath));
```

gnd/src/main/java/com/google/android/gnd/ui/observationdetails/ObservationDetailsFragment.java
shobhitagarwal1612: Added a TODO

shobhitagarwal1612:
Thank you so much @gino-m! That really helped me understand the problems in my approach.
This is ready for review! PTAL

gino-m:
@shobhitagarwal1612  There were some failing PMD checks, PTAL.

PR #392, opened 2020-03-12
Refactor Tile Download Worker to Download Pending Tiles in DB

scolsen:
This PR updates the Tile Download worker to check the DB for pending tiles -- rather than take a list of tiles to download from the context.﻿

This simplifies the worker--in the future, it should take an *area* as an argument, and only check for pending tiles that intersect the given area bounds. (It will currently check *all* pending tiles).

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -441,8 +441,8 @@ public class RoomLocalDataStore implements LocalDataStore {
   @Override
   public Single<ImmutableList<Tile>> getPendingTiles() {
     return tileDao
-      .findByState(TileEntityState.PENDING.intValue())
-      .map(ts -> stream(ts).map(TileEntity::toTile).collect(toImmutableList()))
+        .findByState(TileEntityState.PENDING.intValue())
+        .map(ts -> stream(ts).map(TileEntity::toTile).collect(toImmutableList()))
         .subscribeOn(schedulers.io());
   }
 }
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -440,8 +440,6 @@ public class RoomLocalDataStore implements LocalDataStore {
 
   @Override
   public Single<ImmutableList<Tile>> getPendingTiles() {
-    // TODO: Only retrieve tiles for a given area.
-    // This currently retrieves all pending tiles in the DB.
     return tileDao
       .findByState(TileEntityState.PENDING.intValue())
       .map(ts -> stream(ts).map(TileEntity::toTile).collect(toImmutableList()))
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -28,7 +28,6 @@ import com.google.android.gnd.model.Mutation;
 import com.google.android.gnd.model.Project;
 import com.google.android.gnd.model.User;
 import com.google.android.gnd.model.basemap.tile.Tile;
-import com.google.android.gnd.model.basemap.tile.Tile.State;
 import com.google.android.gnd.model.feature.Feature;
 import com.google.android.gnd.model.feature.FeatureMutation;
 import com.google.android.gnd.model.form.Element;
@@ -444,13 +443,8 @@ public class RoomLocalDataStore implements LocalDataStore {
     // TODO: Only retrieve tiles for a given area.
     // This currently retrieves all pending tiles in the DB.
     return tileDao
-        .findAll()
-        .map(
-            ts ->
-                stream(ts)
-                    .map(TileEntity::toTile)
-                    .filter(t -> t.getState() == State.PENDING)
-                    .collect(toImmutableList()))
+      .findByState(TileEntityState.PENDING.intValue())
+      .map(ts -> stream(ts).map(TileEntity::toTile).collect(toImmutableList()))
         .subscribeOn(schedulers.io());
   }
 }
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -28,6 +28,7 @@ import com.google.android.gnd.model.Mutation;
 import com.google.android.gnd.model.Project;
 import com.google.android.gnd.model.User;
 import com.google.android.gnd.model.basemap.tile.Tile;
+import com.google.android.gnd.model.basemap.tile.Tile.State;
 import com.google.android.gnd.model.feature.Feature;
 import com.google.android.gnd.model.feature.FeatureMutation;
 import com.google.android.gnd.model.form.Element;
@@ -230,6 +231,7 @@ public class RoomLocalDataStore implements LocalDataStore {
     return tileDao
         .findAll()
         .map(list -> stream(list).map(TileEntity::toTile).collect(toImmutableSet()))
+        .toFlowable()
         .subscribeOn(schedulers.io());
   }
 
@@ -435,5 +437,20 @@ public class RoomLocalDataStore implements LocalDataStore {
   @Override
   public Maybe<Tile> getTile(String tileId) {
     return tileDao.findById(tileId).map(TileEntity::toTile).subscribeOn(schedulers.io());
+  }
+
+  @Override
+  public Single<ImmutableList<Tile>> getPendingTiles() {
+    // TODO: Only retrieve tiles for a given area.
+    // This currently retrieves all pending tiles in the DB.
+    return tileDao
+        .findAll()
+        .map(
+            ts ->
+                stream(ts)
+                    .map(TileEntity::toTile)
+                    .filter(t -> t.getState() == State.PENDING)
+                    .collect(toImmutableList()))
+        .subscribeOn(schedulers.io());
   }
 }
--- a/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
+++ b/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
@@ -17,7 +17,6 @@
 package com.google.android.gnd.workers;
 
 import android.content.Context;
-import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.work.Data;
 import androidx.work.Worker;
@@ -100,7 +99,7 @@ public class TileDownloadWorker extends Worker {
 
   /** Update a tile's state in the database and initiate a download of the tile source file. */
   private Completable downloadTile(Tile tile) {
-    HashMap<String, String> requestProperties = new HashMap<>();
+    Map<String, String> requestProperties = new HashMap<>();
 
     if (tile.getState() == State.IN_PROGRESS) {
       File existingTileFile = new File(context.getFilesDir(), tile.getPath());
@@ -172,13 +171,13 @@ public class TileDownloadWorker extends Worker {
       return Result.failure();
     }
 
-    Log.d(TAG, "Downloading tiles: " + pendingTiles);
+    Timber.d("Downloading tiles: %s", pendingTiles);
 
     try {
       processTiles(pendingTiles).blockingAwait();
       return Result.success();
     } catch (Throwable t) {
-      Log.d(TAG, "Downloads for tiles failed: " + pendingTiles, t);
+      Timber.d(t, "Downloads for tiles failed: %s", pendingTiles);
       return Result.failure();
     }
   }
--- a/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
+++ b/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
@@ -165,10 +165,10 @@ public class TileDownloadWorker extends Worker {
     ImmutableList<Tile> pendingTiles = localDataStore.getPendingTiles().blockingGet();
 
     // When there are no tiles in the db, the blockingGet returns null.
-    // We expect tiles to be added to the DB prior to downloading.
-    // If that isn't the case, we fail.
+    // If that isn't the case, another worker may have already taken care of the work.
+    // In this case, we return a result immediately to stop the worker.
     if (pendingTiles == null) {
-      return Result.failure();
+      return Result.success();
     }
 
     Timber.d("Downloading tiles: %s", pendingTiles);
--- a/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
+++ b/gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
@@ -43,8 +43,6 @@ import timber.log.Timber;
  * connection.
  */
 public class TileDownloadWorker extends Worker {
-  private static final String TAG = TileDownloadWorker.class.getSimpleName();
-
   private static final int BUFFER_SIZE = 4096;
 
   private final Context context;
@@ -69,7 +67,10 @@ public class TileDownloadWorker extends Worker {
    * storage. Optional HTTP request header {@param requestProperties} may be provided.
    */
   private void downloadTileFile(Tile tile, Map<String, String> requestProperties)
-    throws TileDownloadException {
+      throws TileDownloadException {
+
+    int mode = Context.MODE_PRIVATE;
+
     try {
       URL url = new URL(tile.getUrl());
       HttpURLConnection connection = (HttpURLConnection) url.openConnection();
@@ -78,12 +79,13 @@ public class TileDownloadWorker extends Worker {
         for (Map.Entry<String, String> property : requestProperties.entrySet()) {
           connection.setRequestProperty(property.getKey(), property.getValue());
         }
+        mode = Context.MODE_APPEND;
       }
 
       connection.connect();
 
       try (InputStream is = connection.getInputStream();
-        FileOutputStream fos = context.openFileOutput(tile.getPath(), Context.MODE_PRIVATE)) {
+          FileOutputStream fos = context.openFileOutput(tile.getPath(), mode)) {
 
         byte[] byteChunk = new byte[BUFFER_SIZE];
         int n;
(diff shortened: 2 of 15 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -436,4 +438,19 @@ public Completable insertOrUpdateTile(Tile tile) {
   public Maybe<Tile> getTile(String tileId) {
     return tileDao.findById(tileId).map(TileEntity::toTile).subscribeOn(schedulers.io());
   }
+
+  @Override
+  public Single<ImmutableList<Tile>> getPendingTiles() {
+    // TODO: Only retrieve tiles for a given area.
+    // This currently retrieves all pending tiles in the DB.
+    return tileDao
+        .findAll()
```
gino-m: Rather than using findAll and filtering, can we define a method on the DAO that only returns tiles with state == PENDING?

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -42,106 +45,106 @@
...
-
-  /** Update a tile's state in the database and initiate a download of the tile source file. */
-  private Result downloadTile(Tile tile) {
-    localDataStore
-        .insertOrUpdateTile(tile.toBuilder().setState(Tile.State.IN_PROGRESS).build())
-        .blockingAwait();
 
-    return downloadTileFile(tile, Optional.empty());
+    is.close();
+    fos.close();
```
gino-m: Does this need to be in a `finally` block?

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -42,106 +45,106 @@
...
   }
 
-  /** Resumes downloading the source for {@param tile} marked as {@code Tile.State.IN_PROGRESS}. */
-  private Result resumeTileDownload(Tile tile) {
-    File existingTileFile = new File(context.getFilesDir(), tile.getPath());
+  /** Update a tile's state in the database and initiate a download of the tile source file. */
+  private Completable downloadTile(Tile tile) {
```
gino-m: Do I understand correct that "Tile" actually refers to an entire pyramid of tiles, and not a single tile? If so, at some point we may want to rename it something more appropriate to avoid confusing others (and ourselves).

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -42,106 +45,106 @@
...
+    if (tile.getState() == State.IN_PROGRESS) {
+      File existingTileFile = new File(context.getFilesDir(), tile.getPath());
+      requestProperties.put("Range", existingTileFile.length() + "-");
+    }
 
-    return downloadTileFile(tile, Optional.of(requestProperties));
+    return localDataStore
+        .insertOrUpdateTile(tile.toBuilder().setState(Tile.State.IN_PROGRESS).build())
+        .andThen(
+            Completable.fromCallable(
```
gino-m: If you use `Completable.fromRunnable` instead you don't need to return anything and can use an inline lambda instead of a block.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -42,106 +45,106 @@
...
 
-  /** Resumes downloading the source for {@param tile} marked as {@code Tile.State.IN_PROGRESS}. */
-  private Result resumeTileDownload(Tile tile) {
-    File existingTileFile = new File(context.getFilesDir(), tile.getPath());
+  /** Update a tile's state in the database and initiate a download of the tile source file. */
+  private Completable downloadTile(Tile tile) {
     HashMap<String, String> requestProperties = new HashMap<>();
 
-    requestProperties.put("Range", existingTileFile.length() + "-");
+    if (tile.getState() == State.IN_PROGRESS) {
```
gino-m: Please add a comment explaining why we do this (to result downloads)?

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -42,106 +45,106 @@
...
+    is.close();
+    fos.close();
   }
 
-  /** Resumes downloading the source for {@param tile} marked as {@code Tile.State.IN_PROGRESS}. */
-  private Result resumeTileDownload(Tile tile) {
-    File existingTileFile = new File(context.getFilesDir(), tile.getPath());
+  /** Update a tile's state in the database and initiate a download of the tile source file. */
+  private Completable downloadTile(Tile tile) {
     HashMap<String, String> requestProperties = new HashMap<>();
```
gino-m: Please define variable as `Map` instead.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -150,27 +153,23 @@ private Result checkDownload(Tile tile) {
...
   public Result doWork() {
-    Tile tile = localDataStore.getTile(tileId).blockingGet();
+    ImmutableList<Tile> pendingTiles = localDataStore.getPendingTiles().blockingGet();
 
-    // When there is no tile in the db, the Maybe completes and returns null.
+    // When there are no tiles in the db, the blockingGet returns null.
     // We expect tiles to be added to the DB prior to downloading.
     // If that isn't the case, we fail.
-    if (tile == null) {
+    if (pendingTiles == null) {
```
gino-m: We shouldn't care if there are no pending tiles; if so another instance of the worker may have completed the job for us, and we should just exit since everything is in the proper state.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -150,27 +153,23 @@ private Result checkDownload(Tile tile) {
...
-      case DOWNLOADED:
-        return checkDownload(tile);
-      case PENDING:
-      case FAILED:
-        return downloadTile(tile);
-      case IN_PROGRESS:
-        return resumeTileDownload(tile);
-      default:
-        return Result.failure();
+    Log.d(TAG, "Downloading tiles: " + pendingTiles);
```
gino-m: Fyi: After Shobhit's recent PR, you can now just `Timber.d("Downloading tiles %s", pendingTiles)`.

scolsen:
> Awesome refactor, thanks a lot! A few comments/suggestions.
> 
> Also, question - are you sure we'd want to trigger one worker per Area in the future? What would be the advantage of doing that instead of having one sync the entire queue?

There's no real technical advantage, but I do think it would better fit user expectations. By tackling everything at once, there's potential for us to do more than matches the user's perception of what's going on based on the actions they took in the UI. For example if we tackle all pending tiles at once and someone attempts to download area A then quits, leaving it partially downloaded, then attempts to download area B, that will result in area A continuing (worker picks up pending tiles), and eventually being downloaded, even though they really only issued actions in the UI that explicitly communicate "download area B"--I'd rather give the user more control to determine if they want to try to download area A again, or just leave it be, or delete it.

I'm not married to such an approach though. I'm fine with chugging through everything that's still Pending—but I do think it could lead to weird interactions or "hidden features" like the case above (e.g. did you know triggering a new area download will resume your previous downloads?).

I guess the downside to triggering a worker on each area though is that if someone goes download crazy and triggers a bunch of different areas at once, or potentially the same area twice, it could result in a lot of extra work—but since we'll probably just lock the Area selection UI for now until a download fails, completes, or takes too long, I think we can ignore that case.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: I wanted to do this, but since the IO exception here is a checked exception, we'd have to explicitly handle the error in the runnable itself, losing the elegance of its automatic propagation upstream (though I suppose NBD since we're already losing elegance with this approach). see here for more info: https://stackoverflow.com/questions/44179444/how-to-create-observable-out-of-runnable

Our options are what we currently have here, or:

```java
Completable.fromRunnable(() -> {try {downloadTile(tile)} catch (IOException e) { Log.d(TAG, "Failed to download tile: " + tile, e);}})
``` 

I opted for the Callable variant since it's shorter. Though I think it may be more confusing to someone unfamiliar with the checked exception restriction....but then again so would handling the exception in the runnable itself.

I'm good with using whichever style you prefer.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: The range request is supposed to let us [resume the file download from where it stopped](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests), but tbh I haven't confirmed it works in practice, and I don't really know much about HTTP range requests at all.

It might be safer to just start the entire file download again when we resume something in progress and overwrite anything that was already on disk. Even though we're technically doing extra work in that case.

gino-m: Hmm.. perhaps.. we can leave it for now provided we test it out.

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
gino-m: Perhaps `downloadTileFile()` should catch the error and rethrow an unchecked exception?

We need to try {} finally {} in `downloadTileFile()` to close the connection, so adding `catch()` should be easy. Wdyt?

scolsen: > Perhaps `downloadTileFile()` should catch the error and rethrow an unchecked exception?
> 
> We need to try {} finally {} in `downloadTileFile()` to close the connection, so adding `catch()` should be easy. Wdyt?

That's a great idea! Let's do that.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
scolsen: The DAO Query strings have to be constant, which unfortunately means we can't exactly add a Pending tiles method to the DAO directly without using an int literal, which is too brittle for my taste.

However, I have switched to using a `findByState` method instead, which is simpler!

scolsen: see commit #2b867a8

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Used a try-with-resources block instead (since android studio suggested it).

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: sg!

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Done! See 2b81c0b

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Done see 2b81c0b

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Done see 2b81c0b

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Have any ideas on what we should call it?

scolsen: Will probably hold off on this for a later PR

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Now it's done :) c68a03b

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: c68a03b

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Gotcha, we'll just treat it as a success then eec1754

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
gino-m: SGTM. Filed #400 to track!

gino-m:
> > Also, question - are you sure we'd want to trigger one worker per Area in the future? What would be the advantage of doing that instead of having one sync the entire queue?
> 
> There's no real technical advantage, but I do think it would better fit user expectations. By tackling everything at once, there's potential for us to do more than matches the user's perception of what's going on based on the actions they took in the UI. For example if we tackle all pending tiles at once and someone attempts to download area A then quits, leaving it partially downloaded, then attempts to download area B, that will result in area A continuing (worker picks up pending tiles), and eventually being downloaded, even though they really only issued actions in the UI that explicitly communicate "download area B"--I'd rather give the user more control to determine if they want to try to download area A again, or just leave it be, or delete it.

Workers continue to run even when the user quits. If the user adds area B to the database and there's already a worker running, if the worker polls the database on each iteration until empty, it will find the new area and process it. 

Also, what needs to be downloaded is currently represented by the state of the local db. If we leverage the Worker framework to represent this as well, the download queue will essentially be represented in two places (the Worker queue, and the db). Ideally we'd only have to manage one or the other.

> I'm not married to such an approach though. I'm fine with chugging through everything that's still Pending—but I do think it could lead to weird interactions or "hidden features" like the case above (e.g. did you know triggering a new area download will resume your previous downloads?).

There's more room for weird edge cases if we have have essentially two ways of representing the download queue (see previous comment). It shouldn't happen that downloading a new area will resume previous downloads - if the previous downloads failed we won't retry them unless the user asks to.
 
> I guess the downside to triggering a worker on each area though is that if someone goes download crazy and triggers a bunch of different areas at once, or potentially the same area twice, it could result in a lot of extra work—but since we'll probably just lock the Area selection UI for now until a download fails, completes, or takes too long, I think we can ignore that case.

Workers run in the background completely asynchronous, regardless of whether the app is open or not, so it wouldn't make sense to block the Area selection UI on the download queue.

I think there may be some confusion about how Workers work (could be my confusion as well). Please let me know if this makes sense!

scolsen:
> > > Also, question - are you sure we'd want to trigger one worker per Area in the future? What would be the advantage of doing that instead of having one sync the entire queue?
> > 
> > 
> > There's no real technical advantage, but I do think it would better fit user expectations. By tackling everything at once, there's potential for us to do more than matches the user's perception of what's going on based on the actions they took in the UI. For example if we tackle all pending tiles at once and someone attempts to download area A then quits, leaving it partially downloaded, then attempts to download area B, that will result in area A continuing (worker picks up pending tiles), and eventually being downloaded, even though they really only issued actions in the UI that explicitly communicate "download area B"--I'd rather give the user more control to determine if they want to try to download area A again, or just leave it be, or delete it.
> 
> Workers continue to run even when the user quits. If the user adds area B to the database and there's already a worker running, if the worker polls the database on each iteration until empty, it will find the new area and process it.
> 
> Also, what needs to be downloaded is currently represented by the state of the local db. If we leverage the Worker framework to represent this as well, the download queue will essentially be represented in two places (the Worker queue, and the db). Ideally we'd only have to manage one or the other.
> 
> > I'm not married to such an approach though. I'm fine with chugging through everything that's still Pending—but I do think it could lead to weird interactions or "hidden features" like the case above (e.g. did you know triggering a new area download will resume your previous downloads?).
> 
> There's more room for weird edge cases if we have have essentially two ways of representing the download queue (see previous comment). It shouldn't happen that downloading a new area will resume previous downloads - if the previous downloads failed we won't retry them unless the user asks to.
> 
> > I guess the downside to triggering a worker on each area though is that if someone goes download crazy and triggers a bunch of different areas at once, or potentially the same area twice, it could result in a lot of extra work—but since we'll probably just lock the Area selection UI for now until a download fails, completes, or takes too long, I think we can ignore that case.
> 
> Workers run in the background completely asynchronous, regardless of whether the app is open or not, so it wouldn't make sense to block the Area selection UI on the download queue.
> 
> I think there may be some confusion about how Workers work (could be my confusion as well). Please let me know if this makes sense!

Thanks for the clarification! That makes sense—I wasn't accounting for the fact that the workers continue even after app exit. I'll remove the TODO

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Fixed some problems w/ this and added explanatory comments in de5953f

scolsen:
> > > > Also, question - are you sure we'd want to trigger one worker per Area in the future? What would be the advantage of doing that instead of having one sync the entire queue?
>>> ...lots of text
> Thanks for the clarification! That makes sense—I wasn't accounting for the fact that the workers continue even after app exit. I'll remove the TODO

Removed the TODO 096707a

scolsen:
Think it's ready for another look, @gino-m!

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -32,116 +35,140 @@
...
+            localDataStore.insertOrUpdateTile(tile.toBuilder().setState(State.DOWNLOADED).build()));
   }
 
   /**
    * Verifies that {@param tile} marked as {@code Tile.State.DOWNLOADED} in the local database still
    * exists in the app's storage. If the tile's source file isn't present, initiates a download of
    * source file.
    */
-  private Result checkDownload(Tile tile) {
+  private Completable checkDownload(Tile tile) {
```
gino-m: Naming nit: Since this both checking and downloads, maybe `downloadIfNotFound`? This can be done in a follow up though.

gino-m:
Amazing work, Scott, this is going to be a very important feature for Ground. Thanks for your perserverance!

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
scolsen: Done! 6373a71

PR #398, opened 2020-03-17
[Code-health] Add task for dependency check to Cloud Build

shobhitagarwal1612:
- Config task updateDependency
- Set output dir to /gnd/build/reports/ so that the reports are uploaded to GCS
- Add TODO for future task

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ allprojects {
 }
 
 task checkCode(type: GradleBuild) {
-    tasks = ['checkstyle', 'lintDebug', 'pmd', 'spotbugs']
+    tasks = ['checkstyle', 'lintDebug', 'pmd', 'spotbugs', 'dependencyUpdates']
 }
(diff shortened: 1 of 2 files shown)


gino-m:
What does this do in practice? Please describe in PR description how it will look to contributors. (e.g., will checks fail on PRs if dependencies aren't all at the latest version? If so this may be too strict - we can't always use the latest version due to compat issues between different libraries we use, for example)

shobhitagarwal1612:
This only displays a report in the console and also creates a report in json format. The plugin also claims to account for constraints while checking for latest version.

This will never fail any build and only prints the suggestions. My guess is that no one will ever run this check manually. So, I plan to copy the reports file to GCS dir and dynamically create a SVG with dependency health

eg.
The json contains the total number of dependencies with latest versions and obsolete ones too. 
Just a simple % healthy dependencies with background color

PR #399, opened 2020-03-18
[Code health] Add dependency health badge to README

shobhitagarwal1612:
Generates SVG from json report and adds badge to README

**Demo**
https://github.com/shobhitagarwal1612/ground-android/tree/generate_dependency_status

**Screenshot** 
image

(Follow up of https://github.com/google/ground-android/pull/398)


cloud-builder/generate_dependency_health_svg.py
```diff
@@ -0,0 +1,87 @@
...
+COLOR_GREEN = '#43CC11'
+
+
+def get_status_color(health):
+    if health >= 67:
+        return COLOR_GREEN
+    elif 33 < health < 67:
+        return COLOR_ORANGE
+    else:
+        return COLOR_RED
```
shobhitagarwal1612: @gino-m @scolsen @dturner 
Thoughts on the current chosen range?

cloud-builder/generate_dependency_health_svg.py
```diff
@@ -0,0 +1,87 @@
+"""Generates a SVG by parsing json report.
```
gino-m: This should have Copyright header as well.

cloud-builder/generate_dependency_health_svg.py
gino-m: I like the last one (5 ranges), since I think the impact of having older dependencies shouldn't be weighted so harshly - this way we'll have more yellow and green, only slipping into orange if we're really out of date.

cloudbuild.yaml
```diff
@@ -105,14 +105,27 @@ steps:
...
+
+          # Build Status
           if [[ $(< build-status.txt) == "fail" ]]; then
             gsutil cp cloud-builder/failure.svg gs://${_CACHE_BUCKET}/status.svg
           else
             gsutil cp cloud-builder/success.svg gs://${_CACHE_BUCKET}/status.svg
           fi
 
-          # Make file public readable
+          # Make files public readable
```
gino-m: Nit: public -> publicly

PR #401, opened 2020-03-18
[Cleanup] Replace manual subscription handling with ViewModel

shobhitagarwal1612:
Resolves a TODO added in https://github.com/google/ground-android/pull/386

- Factor out Photo Field UI elements into photo_field.xml
- Reuse above layout and PhotoFieldViewModel to render ObservationDetailsFragment

Tested manually:
- ObservationDetailsFragment page (both with and without image)
- EditObservationFragment page (both with and without image)


shobhitagarwal1612:
@gino-m PTAL!

gnd/src/main/java/com/google/android/gnd/ui/observationdetails/ObservationDetailsFragment.java
```diff
@@ -156,35 +149,25 @@ private void addField(Field field, Observation observation) {
...
               }
             });
   }
 
+  private void addPhotoField(ViewGroup container, Field field, Response response) {
+    PhotoFieldBinding photoFieldBinding = PhotoFieldBinding.inflate(getLayoutInflater());
+    PhotoFieldViewModel photoFieldViewModel = viewModelFactory.create(PhotoFieldViewModel.class);
+    photoFieldBinding.setLifecycleOwner(this);
+    photoFieldBinding.setViewModel(photoFieldViewModel);
+    photoFieldViewModel.updateField(response, field);
```
gino-m: It appears this method is called both from `init()` and here. Is this intentional? If so, why?

shobhitagarwal1612: `init()` takes an `ObservableMap` as a parameter and calls `updateField()` whenever the response is changed. Here, we don't call `init()` as this is a one time update.

shobhitagarwal1612: Makes me wonder if `init()` should a part of `PhotoFieldViewModel` at all....

PR #402, opened 2020-03-18
Clean up MapAdapter

scolsen:
This PR contains two minor clean up changes to the MapAdapter:

- Use timber for logging
- Remove the renderJsonLayer -- we don't need it anymore since we now take a viewport based approach to offline area selection﻿

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
@@ -51,7 +51,6 @@ import timber.log.Timber;
  */
 class GoogleMapsMapAdapter implements MapAdapter {
 
-  private static final String TAG = GoogleMapsMapAdapter.class.getSimpleName();
   private final GoogleMap map;
   private final Context context;
   private final MarkerIconFactory markerIconFactory;
@@ -197,7 +196,7 @@ class GoogleMapsMapAdapter implements MapAdapter {
   }
 
   private void removeMarker(Marker marker) {
-    Timber.tag(TAG).v("Removing marker %s", marker.getId());
+    Timber.v("Removing marker %s", marker.getId());
     marker.remove();
   }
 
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -197,15 +197,15 @@ private static LatLng toLatLng(Point point) {
   }
 
   private void removeMarker(Marker marker) {
-    Log.v(TAG, "Removing marker " + marker.getId());
+    Timber.tag(TAG).v("Removing marker %s", marker.getId());
```
shobhitagarwal1612: We don't need to mention a custom tag. Timber uses the filename as the default tag. 

So, the above can be written as `Timber.v("Removing marker %s", marker.getId());`

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -197,15 +197,15 @@ private static LatLng toLatLng(Point point) {
...
     marker.remove();
   }
 
   private int parseColor(@Nullable String colorHexCode) {
     try {
       return Color.parseColor(String.valueOf(colorHexCode));
     } catch (IllegalArgumentException e) {
-      Log.w(TAG, "Invalid color code in layer style: " + colorHexCode);
+      Timber.tag(TAG).w("Invalid color code in layer style: %s", colorHexCode);
```
shobhitagarwal1612: Same as above

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
shobhitagarwal1612: This also makes `TAG` variable unused. So, please remove that as well

scolsen: Gotcha, done! 513ef9a

scolsen:
Thanks @shobhitagarwal1612! It's ready for another check when you get a chance.

PR #406, opened 2020-03-20
[Feature] Add button to switch map type

shobhitagarwal1612:
Demo:

map_type_2


gino-m:
Thanks for the change! Is there any background on why this is needed or preferred? Before starting a change please make sure there's an Issue filed so we can discuss, since we need to consider all the different ongoing feature development and how they'll fit together.

Note that we're using the [] before the PRs and Issues to indicate which feature/subarea of the app is being changes, not the type of PR/issue.

shobhitagarwal1612:
> Thanks for the change! Is there any background on why this is needed or preferred? Before starting a change please make sure there's an Issue filed so we can discuss, since we need to consider all the different ongoing feature development and how they'll fit together.

Sorry for being so impatient. I remember asking for this feature in hangouts some days ago and you replied that it would somewhat similar to Google Maps. Point noted for future issues.

Points regarding use cases:
 - Satellite view requires better internet connection if offline tiles are not available as it consumes for data
 - I prefer using normal view when navigating using maps as it contains less details.

> Note that we're using the [] before the PRs and Issues to indicate which feature/subarea of the app is being changes, not the type of PR/issue.

Done

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapProvider.java
```diff
@@ -73,29 +74,28 @@ public Fragment getFragment() {
...
   @Override
-  public HashMap<Integer, String> getMapTypes() {
-    HashMap<Integer, String> hashMap = new HashMap<>();
-    hashMap.put(GoogleMap.MAP_TYPE_NONE, "None");
-    hashMap.put(GoogleMap.MAP_TYPE_NORMAL, "Normal");
-    hashMap.put(GoogleMap.MAP_TYPE_SATELLITE, "Satellite");
-    hashMap.put(GoogleMap.MAP_TYPE_TERRAIN, "Terrain");
-    hashMap.put(GoogleMap.MAP_TYPE_HYBRID, "Hybrid");
-    return hashMap;
+  public Map<Integer, String> getMapTypes() {
```
gino-m: Please use an ImmutableMap<> defined in a constant instead. The constant can then be returned from this method.

gino-m: Also add `// TODO: i18n.`

gino-m:
> Sorry for being so impatient. I remember asking for this feature in hangouts some days ago and you replied that it would somewhat similar to Google Maps. Point noted for future issues.

No worries. We'd normally want to discuss such a change more in depth with UX Designers (@coreyleamon fyi), but since we know we'll need to have a layer switcher eventually to control offline imagery let's keep this change and iterate on it as we get more clarity. Over to @scolsen for review!

gnd/src/main/res/values/strings.xml
```diff
@@ -140,5 +140,6 @@
   <string name="cancel">Cancel</string>
   <string name="join_project">Join project</string>
   <string name="photo_preview">Photo preview</string>
+  <string name="select_map_type">Select Map Type</string>
```
gino-m: Nit: "Select map type"

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -72,6 +74,26 @@
   private HomeScreenViewModel homeScreenViewModel;
   private MainViewModel mainViewModel;
 
+  @OnClick(R.id.map_type_btn)
```
gino-m: We're trying to move away from Butterknife in favor of Data Binding to ViewModels - could we do that here instead?

shobhitagarwal1612: I agree that the class needs a refactor, but doing that in the same PR would broaden the scope. We can open a new refactor bug to handle that independently. Wdyt?

gino-m: I mainly didn't want to add a new instance of Butterknife binding if we could avoid it, esp. since we already have a MapContainerViewModel. Is it just a question of defining `onMapTypeButtonClick()` in the view model and exposing an `Observable<Event<Nil>>` to trigger the dialog in the view? If so I'd prefer to add it straight away if possible. If you have your hands full atm though please lmk and I can try to send a PR to your branch with the recommended changes!

shobhitagarwal1612: Done

gnd/src/main/res/drawable/map_layers.xml
```diff
@@ -0,0 +1,29 @@
...
+  android:width="24dp"
+  android:height="24dp"
+  android:viewportWidth="24"
+  android:viewportHeight="24">
+
+  <path android:pathData="M0 0h24v24H0V0z" />
+  <path
+    android:fillColor="#000000"
+    android:pathData="M11.99 18.54l-7.37-5.73L3 14.07l9 7 9-7-1.63-1.27zM12 16l7.36-5.73L21 9l-9-7-9 7 1.63 1.27L12 16zm0-11.47L17.74 9 12 13.47 6.26 9 12 4.53z" />
+</vector>
```
scolsen: Missing newline

gnd/src/main/res/layout/map_container_frag.xml
```diff
@@ -86,20 +102,24 @@
...
       android:clickable="true"
+      android:focusable="true"
       android:onClick="@{() -> homeScreenViewModel.openNavDrawer()}"
+      android:padding="16dp"
       android:src="@drawable/ic_menu_black_24dp"
-      android:tint="@color/colorBackground"/>
+      android:tint="@color/colorBackground" />
 
   </FrameLayout>
 </layout>
```
scolsen: Missing newline

gnd/src/main/java/com/google/android/gnd/ui/map/MapAdapter.java
```diff
@@ -70,6 +70,12 @@
   /** Update map pins shown on map. */
   void setMapPins(ImmutableSet<MapPin> pins);
 
+  /** Get current map type. */
+  int getMapType();
+
+  /** Update map type. */
+  void setMapType(int mapType);
```
scolsen: Instead of using `int` here directly should we abstract away the particular map types of a given Map impl and introduce an enum and some conversion functions to allow us to set the types regardless of the underlying adapter?

@gino-m wdyt?

e.g. on the adapter we'd have some FN setMapType(TypeENUM) instead of int.

No need to do this now.

gnd/src/main/res/drawable/map_layers.xml
shobhitagarwal1612: Thanks.
We should probably automate this check as well using pmd or checkstyle.

gnd/src/main/java/com/google/android/gnd/ui/map/MapAdapter.java
gino-m: It may make sense to define an enum at the interface level with a superset of basemap types, yes. Also +1 that it can be done in a follow up (please add a TODO)?

Also, somewhere we should clarify that Map Type refers to the basemap shown below map features and offline satellite imagery - they're called "map styles" in Mapbox and "basemaps" in Leaflet.

shobhitagarwal1612: Thanks @scolsen and @gino-m! Done

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -88,6 +94,9 @@
                 .map(Loadable::value)
                 .switchMap(this::getFeaturesStream)
                 .map(MapContainerViewModel::toMapPins));
+    this.mapLayerUpdateRequests =
```
gino-m: Can we simplify this by eliminating `mapLayerTypeClicks` and just making `mapLayerUpdateRequests` a `MutableLiveData` that gets a new Event on each click (`.setValue(Event.create(..))`)?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -29,34 +28,40 @@
...
 public class MapContainerViewModel extends AbstractViewModel {
 
-  private static final String TAG = MapContainerViewModel.class.getSimpleName();
   private static final float DEFAULT_ZOOM_LEVEL = 20.0f;
   private final LiveData<Loadable<Project>> activeProject;
   private final LiveData<ImmutableSet<MapPin>> mapPins;
   private final LiveData<BooleanOrError> locationLockState;
   private final LiveData<CameraUpdate> cameraUpdateRequests;
   private final MutableLiveData<Point> cameraPosition;
+  private final LiveData<Event<Nil>> mapLayerUpdateRequests;
```
gino-m: `showMapTypeSelectorRequests`?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
shobhitagarwal1612: Thanks! Done

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
shobhitagarwal1612: Done

PR #413, opened 2020-03-26
[Refactor] Categorised room classes into packages

shobhitagarwal1612:
Motivation behind the PR:
Adding more visibility by reducing the number of files in a package

- Moves room classes into packages based on the type of class (dao, entity, model, converter, relations)
- Log -> Timber
- Reduce visibility of (some) methods


gnd/src/main/java/com/google/android/gnd/persistence/local/room/relations/ProjectEntityAndRelations.java
```diff
@@ -14,10 +14,12 @@
  * limitations under the License.
  */
 
-package com.google.android.gnd.persistence.local.room;
+package com.google.android.gnd.persistence.local.room.relations;
```
gino-m: Are relations sufficiently closely related to entities (entity-relational model) that they should be in the entity package?

shobhitagarwal1612: I don't have a good answer. There are only a few relations till now but I expect this package to expand with more feature additions. Please suggest what would be better to do here?

gino-m: Ok with leaving it as-is, though will point out that number of instances should only be a rough proxy or a tie-breaker when scoping; i.e., we should seek to group things conceptually rather than by number.

gino-m:
Thanks for the cleanup!

Issue #110, opened 2019-06-14
[Offline support] Show notifications with data sync progress


gino-m:
Blocked by #17

shobhitagarwal1612:
Is someone working on this?

gino-m:
> Is anyone working on this?

Not yet! Interested?

shobhitagarwal1612:
Yes. This looks interesting.

gino-m:
All yours! Let's discuss design before starting, please ping or set up a VC!

shobhitagarwal1612:
Will do!

PR #414, opened 2020-03-26
[Notifications] Show notifications when uploading photos

shobhitagarwal1612:
Towards #110 

For testing: Use `FROGS` project or create one having a photo field


gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteStorageManager.java
```diff
@@ -31,5 +32,5 @@
   Task<Uri> getDownloadUrl(String remoteDestinationPath);
 
   /** Uploads file to a remote path. */
-  Completable uploadMediaFromFile(File file, String remoteDestinationPath);
+  Flowable<Event<UploadProgress>> uploadMediaFromFile(File file, String remoteDestinationPath);
```
gino-m: Why wrapping with `Event`? It's used as a "Wrapper for events passed through streams that should be handled at most once.". It's primarily used to prevent UIs from showing the same message more than ones. In this case, it seems you'd want to just stream the UploadProgress with `Backpressure.LATEST`.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreStorageManager.java
```diff
@@ -65,27 +68,35 @@ private StorageReference createReference(String path) {
...
+      File file, String remoteDestinationPath) {
+    return Flowable.create(
         emitter ->
             createReference(remoteDestinationPath)
                 .putFile(Uri.fromFile(file))
                 .addOnCompleteListener(uploadTask -> emitter.onComplete())
-                .addOnFailureListener(emitter::onError)
+                .addOnPausedListener(
+                    taskSnapshot -> {
+                      emitter.onNext(Event.create(new UploadProgress(UploadState.PAUSED)));
```
gino-m: Instead of calling `new` directly, consider adding "static constructor" methods like `UploadProgress.paused()` and UploadProgress.inProgress(byteCount, bytesTransferred).

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreStorageManager.java
```diff
@@ -65,27 +68,35 @@ private StorageReference createReference(String path) {
...
-                      Timber.d("Uploading in progress: %s %f", remoteDestinationPath, completed);
-                    }));
+                    taskSnapshot ->
+                        emitter.onNext(
+                            Event.create(
+                                new UploadProgress(
+                                    UploadState.IN_PROGRESS,
+                                    (int) taskSnapshot.getTotalByteCount(),
+                                    (int) taskSnapshot.getBytesTransferred())))),
+        BackpressureStrategy.BUFFER);
```
gino-m: Only use buffer if it's important that downstream subscribers catch every single value. In this case, only the latest value is relevant, so we'd use LATEST.

gnd/src/main/java/com/google/android/gnd/persistence/sync/PhotoSyncWorker.java
```diff
@@ -66,7 +72,8 @@ public Result doWork() {
       try {
         remoteStorageManager
             .uploadMediaFromFile(new File(localSourcePath), remoteDestinationPath)
-            .blockingAwait();
+            .blockingForEach(
```
gino-m: Are we showing a notification for each photo being uploaded? Ideally we'd have a single status bar for sync progress.

shobhitagarwal1612: For now, yes. I plan to add notifications to all workers, then move on to refactor PhotoSyncWorker (merge with DataSyncWorker and store mutations in db)

gino-m: I may be wrong, but isn't this fundamentally different than showing a single always-on service UI notification? Won't we be doing a lot of work that will then need to be deleted?

shobhitagarwal1612: Nope, it's just a matter of changing boolean values to keep it sticky and then reuse the same notification for displaying updates. My idea was to have many eyes look at the code without making too many changes at once. But if you think it's better to do all at once, I can give it a try.

gino-m: If it's an incremental change then SGTM!

PR #419, opened 2020-03-26
Adding some comments and questions

dturner:
As a new user to the codebase, and someone unfamiliar with RxJava I thought it'd be useful to add comments/suggestions/questions at points where it took a while for me to grasp what was going on. 

Not intending to merge in this current state, however, if you could answer any questions I can tidy up and merge when we're happy with the comments.


gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -97,6 +99,12 @@ public ProjectRepository(
         .compose(Loadable::loadingOnceAndWrap);
   }
 
+  /**
+   * Loads the project with the given id from the remote data store and saves it to the local data
+   * store
+   * @param id - id of project to load
+   * @return A Single representing the project (@see http://reactivex.io/documentation/single.html)
```
gino-m: If we consider Single as low-level types like `Optional` and `Consumer`, would it make sense to count on the user referring to the JavaDoc instead of pointing them to additional docs? We use Rx very widely, and it may become burdensome to consistently link to Rx docs.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -64,9 +64,11 @@ public ProjectRepository(
     this.localValueStore = localValueStore;
 
     // BehaviorProcessor re-emits last requested project id to late subscribers.
+    // TODO (DT) - Why is this necessary? Under what conditions would there be late subscribers?
```
gino-m: "Late subscribers" here just means UIs that initialized after the active project was loaded. Since the order of initialization of views in Android as often hard to mentally model, we treat it as non-deterministic (and perhaps sometimes it is?). 

I agree Rx concepts are not sufficiently self-describing; that's been a pain point in onboarding. I'm not sure what the best way to mitigate this is; what level of Rx knowledge do you think out docs should assume to strike the right balance between verbosity/redundancy and clarity?

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -50,6 +50,8 @@ public MainViewModel(
     this.navigator = navigator;
 
     // TODO: Move to background service.
+    // Add the project repository subscription to the collection of subscriptions which will be
+    // disposed when the view model is cleared
```
gino-m: Should this be explained in the declaration `disposeOnClear()` instead? It's used widely throughout.

dturner:
Yep, will do, this is definitely a WIP and your comments are extremely helpful in getting to grips with the codebase.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
dturner: Good point. I will remove the link. A base level of RxJava knowledge is required to work with the codebase and it doesn't make sense for us to teach that to the user.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
dturner: I think a class/architecture diagram would reduce the onboarding barrier significantly. I've been sketching one out and will publish when I have it in a useful digital form. 

It's definitely not necessary to explain Rx concepts, but may be necessary to explain which Rx concepts Ground is using and why it's using them.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
dturner: The confusion for me was that I'm in the constructor and the first method talks about disposing something. I'm thinking "why are we talking about disposing objects when we're creating one?". Digging further it makes sense: the subscription (a `Disposable`) is added to a collection of subscriptions which should be unsubscribed and disposed of when the view model is cleared. 

Perhaps it might read better if split onto 2 lines. Have refactored. WDYT?

dturner:
I'm going to close this out as it's probably not the best method of asking questions/adding comments since the speed of development on master requires virtually continuous rebasing.

gino-m:
> I'm going to close this out as it's probably not the best method of asking questions/adding comments since the speed of development on master requires virtually continuous rebasing.

Sgtm. I'll follow up on a few of the questions here now for posterity though.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -50,11 +51,13 @@ public MainViewModel(
...
-            .getActiveProjectOnceAndStream()
-            .switchMapCompletable(this::syncFeatures)
-            .subscribe());
+    // Subscribe to the project repository
+    Disposable projectSubscription = projectRepository.getActiveProjectOnceAndStream()
+      .switchMapCompletable(this::syncFeatures)
+      .subscribe();
+
+    // Ensure that the subscription is disposed of when the model is cleared
+    disposeOnClear(projectSubscription);
```
gino-m, line 60: This will be a frequent enough usage of RxJava that we may not want to make this explicit on each usage; it's part of understanding that Reactive steam subscriptions must be cleaned up.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
gino-m: +1. @mmisim is this something you might be able to help with, i.e. a basic architecture doc -> entry in our Wiki?

mmisim: Yes, I can help with this. I think starting with an architecture diagram and fleshing out into a wiki article would be a good approach. @dturner — do you want me to take a look a what you have so far?

PR #421, opened 2020-03-31
[Background sync] WIP : Run workers in foreground service

### LINKED ISSUE #110: [Offline support] Show notifications with data sync progress

shobhitagarwal1612:
Closes #110

- [x] Adds foreground service

- [x] Enables notifications for all sync workers

- [x] Auto removes sync notification on complete

**[EDIT]**
Questions:
1. The foreground service should terminate after performing its job. Currently, it never gets killed. This drains battery a lot. What should be the ideal behavior in our case?

2. Should clicking on the  "App is running" notification perform some action?

Verified manually:

- Sync notifications for workers
- Sync happens in background if the app is closed


gnd/src/main/java/com/google/android/gnd/persistence/sync/LocalMutationSyncWorker.java
```diff
@@ -72,10 +78,14 @@ public Result doWork() {
     ImmutableList<Mutation> mutations = localDataStore.getPendingMutations(featureId).blockingGet();
     try {
       Timber.v("Mutations: %s", mutations);
-      processMutations(mutations).blockingAwait();
+      processMutations(mutations)
+          .doOnSubscribe(__ -> sendNotification(UploadProgress.starting()))
```
scolsen, line 82: Should we encapsulate all this setup in a method since it's pretty much the same for all workers? Might be overkill since we only have three workers--just a thought.

gnd/src/main/java/com/google/android/gnd/persistence/sync/LocalMutationSyncWorker.java
```diff
@@ -126,4 +136,12 @@ private Mutation incrementRetryCount(Mutation mutation, Throwable error) {
         .setLastError(error.toString())
         .build();
   }
+
+  private void sendNotification(UploadProgress uploadProgress) {
```
scolsen, line 140: Should we add an interface for this method--or make it static and have it live in the Notification manager (we'd just have to make it take the R.string as a second argument)? 

It's more or less the same in the other worker (tiles).

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
```diff
@@ -155,6 +156,13 @@ private Completable downloadIfNotFound(Tile tile) {
 
   private Completable processTiles(ImmutableList<Tile> pendingTiles) {
     return Observable.fromIterable(pendingTiles)
+        .doOnSubscribe(__ -> sendNotification(UploadProgress.starting()))
```
scolsen, line 159: A little weird that this is called UploadProgress from this context, since we're actually downloading tiles, not uploading them. Of course, for photos we're performing an upload not a download...

Maybe make the name more generic?

dturner: TransferProgress?

shobhitagarwal1612: Thanks!

gnd/src/main/java/com/google/android/gnd/persistence/sync/LocalMutationSyncWorker.java
shobhitagarwal1612: dbf0ce9 PTAL

gnd/src/main/java/com/google/android/gnd/persistence/sync/LocalMutationSyncWorker.java
shobhitagarwal1612: 8d75cd8

gnd/src/main/java/com/google/android/gnd/workers/TileDownloadWorker.java
shobhitagarwal1612: 1b72b05

shobhitagarwal1612:
I've move the foreground service related code into https://github.com/google/ground-android/pull/428 since this PR might take some time to reach consensus around designing background sync. Once that is approved and submitted, I'll clean this PR. PTAL

shobhitagarwal1612:
moved to #430

PR #425, opened 2020-04-05
[Code health] Add unit tests for LocalDataStore

shobhitagarwal1612:
Unit tests for LocalDataStore. Code coverage: 96%


gnd/src/androidTest/java/com/google/android/gnd/persistence/local/room/dao/UserDaoTest.java
```diff
@@ -0,0 +1,95 @@
...
+    userDao.insert(UserEntity.fromUser(testUser)).test().assertNoErrors();
+    userDao
+        .findById(testUser.getId())
+        .test()
+        .assertNoErrors()
+        .assertValue(UserEntity.fromUser(testUser));
+  }
+
+  @Test
+  public void insertingAlreadyPresentUser_shouldRaiseError() {
```
gino-m, line 57: Should we include the name of the entry point in this method as well for consistency? Perhaps we can create a convention like testEntryPoint_condition_result, e.g. testInsertUser_alreadyPresent_raisesError?  If condition and result are missing it can be assumed we're testing the default code path?

shobhitagarwal1612: Done

shobhitagarwal1612:
>  I'd normally say auto-generated classes don't need to be tests (i.e. to avoid testing Room's implementation), but since our annotations and signature are actual implementation details it seems to make sense here

I was aiming to write tests for `LocalDataStore`. But that seemed a bit overwhelming given the total size of the class and the total number of codepaths. Maybe we should target testing (in a separate PR) `*Repository` classes instead since they are defined in a more systematic way. Thoughts?

gino-m:
> > I'd normally say auto-generated classes don't need to be tests (i.e. to avoid testing Room's implementation), but since our annotations and signature are actual implementation details it seems to make sense here
> 
> I was aiming to write tests for `LocalDataStore`. But that seemed a bit overwhelming given the total size of the class and the total number of codepaths. Maybe we should target testing (in a separate PR) `*Repository` classes instead since they are defined in a more systematic way. Thoughts?

Assuming we're aiming to test all behaviors under different conditions, `RoomLocalDataStore` would need to be tested anyway, since it's the internal API for interacting with the local db. If we also test Daos through `RoomLocalDataStore`, we'll basically be saying that the "system under test" (or unit) is the Room abstraction layer, which is probably ok considering the DAOs don't have much logic of their own. The fact that `RoomLocalDataStore` is larger shouldn't deter us;  if we feel it's not manageable, we could split it up into e.g., LocalProjectStore, LocalFeatureStore, etc. (at the cost of needing a separate interface for each).

Wdyt, also @dturner @scolsen?

shobhitagarwal1612:
I might be wrong since many of the methods in the `RoomLocalDataStore` class are chained together and testing the start link should cover the rest. But I'm curious to know what others think about this.

shobhitagarwal1612:
What if we restrict `LocalDataStore` to only contain definitions for CRUD operations and move the business logic to `*Repository` classes? That should make the class simple enough to not be further broken down into subclasses. Then we can test the entire `LocalDataStore` in a single go rather than targeting all DAOs separately.

By business logic, I'm referring to methods like `applyMutations`, `updateMutations`, `mergeObservation`,  etc. which are performing multiple CRUD operations in a single method.

gino-m:
> What if we restrict `LocalDataStore` to only contain definitions for CRUD operations and move the business logic to `*Repository` classes? That should make the class simple enough to not be further broken down into subclasses. Then we can test the entire `LocalDataStore` in a single go rather than targeting all DAOs separately.
> 
> By business logic, I'm referring to methods like `applyMutations`, `updateMutations`, `mergeObservation`, etc. which are performing multiple operations in a single method.

By definition, `LocalDataStore` interface abstracts how data is read/written locally, including what gets done as a single transaction; we can't assume all data stores will break down operations along the same CRUD lines as in Room. The idea is that the layers are providing increasingly higher levels of abstraction; mapping `LocalDataStore` 1:1 to DAOs could expose implementation details across the "local data store" abstraction. 

I would be in favor of testing the `RoomLocalDataStore` in addition to, or instead of, the individual Daos. 

Gentle ping @dturner , @scolsen for your input.

shobhitagarwal1612:
@gino-m Sounds good to me! Once we have a word from everyone then we can merge this PR and I'll get started with testing the `LocalDataStore`

dturner:
Warning: layman opinions ahead...

I'm supportive of writing tests for each DAO. However, I don't think we should test methods which are implemented in `BaseDao` such as `insert`, `update` etc as this will result in a lot of test code with overlapping coverage. 

I'd suggest we have a test for `BaseDao` which tests the basic CRUD operations once using one of the entities and then separate tests for each DAO which tests only the methods in that class e.g. `UserDao::findById`.

I'm interested to see how the tests come together for `RoomLocalDataStore` - that class is quite big (~500 LOC) and it looks like there's a lot of repetition `insertOrUpdate*` which could be abstracted in to smaller more manageable classes.

shobhitagarwal1612:
Thanks for your inputs @dturner @gino-m! 

Let's keep this PR open and I'll try to add tests for `RoomLocalDataStore` into this. (Also, if possible will try to convert these into Robolectric tests so they can be tested wiithout an emulator or real device)

gino-m:
> I'm supportive of writing tests for each DAO. However, I don't think we should test methods which are implemented in `BaseDao` such as `insert`, `update` etc as this will result in a lot of test code with overlapping coverage.

+1.

> I'd suggest we have a test for `BaseDao` which tests the basic CRUD operations once using one of the entities and then separate tests for each DAO which tests only the methods in that class e.g. `UserDao::findById`.

+1 again.

> I'm interested to see how the tests come together for `RoomLocalDataStore` - that class is quite big (~500 LOC) and it looks like there's a lot of repetition `insertOrUpdate*` which could be abstracted in to smaller more manageable classes.

+1 yet again.

I would also be in favor of splitting the LocalDataStore into several interfaces (possibly inner interfaces of LocalDataStore) so that we can split the Room impl into multiple classes that get tested separately..

gino-m:
> Thanks for your inputs @dturner @gino-m!
> 
> Let's keep this PR open and I'll try to add tests for `RoomLocalDataStore` into this. (Also, if possible will try to convert these into Robolectric tests so they can be tested without an emulator or real device)

Sgtm. Once you have a few methods tests in RoomLocalDataStore let's discuss further? IMO it'll be easier to write and read the tests if we don't need to mock out the Daos. If we don't mock the Dao in RoomLocalDataStore, the separate Dao tests may not be necessary. Let's see.

Re Roboelectric - can tests that don't involve Android views or view lifecycles be run without it? Does Room work without simulating the SDK with Roboelectric?

shobhitagarwal1612:
> Sgtm. Once you have a few methods tests in RoomLocalDataStore let's discuss further? IMO it'll be easier to write and read the tests if we don't need to mock out the Daos. If we don't mock the Dao in RoomLocalDataStore, the separate Dao tests may not be necessary. Let's see.

 +1

> Re Roboelectric - can tests that don't involve Android views or view lifecycles be run without it? Does Room work without simulating the SDK with Roboelectric?

I hope it does.

shobhitagarwal1612:
@dturner @gino-m, I've covered 81% of the class with tests. A few more are left, but I could definitely use some feedback.
PTAL

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -49,13 +48,13 @@
 public interface LocalDataStore {
 
   /** Load projects stored in local database. */
-  Single<List<Project>> getProjects();
+  Single<ImmutableList<Project>> getProjects();
```
gino-m, line 51: 👏 👏 

Thanks also for the name fixes, the DAO injection, and all the rest!

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -145,4 +144,7 @@
    * the area into the local data store.
    */
   Completable insertOrUpdateOfflineArea(OfflineArea area);
+
+  /** Returns all saved offline areas from the local data store. */
```
gino-m, line 148: "all saved" -> "all queued, failed, and completed"

gnd/src/main/java/com/google/android/gnd/persistence/local/room/dao/OfflineAreaDao.java
```diff
@@ -19,15 +19,15 @@
...
 import io.reactivex.Maybe;
+import io.reactivex.Single;
 import java.util.List;
 
 /** Provides read/write operations for writing {@link OfflineAreaEntity} to the local db. */
 @Dao
 public interface OfflineAreaDao extends BaseDao<OfflineAreaEntity> {
   @Query("SELECT * FROM offline_area")
-  Flowable<List<OfflineAreaEntity>> findAll();
+  Single<List<OfflineAreaEntity>> findAll();
```
gino-m, line 30: Good catch.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/dao/TileDao.java
```diff
@@ -27,7 +28,7 @@
 public interface TileDao extends BaseDao<TileEntity> {
 
   @Query("SELECT * FROM tile")
-  Single<List<TileEntity>> findAll();
+  Flowable<List<TileEntity>> findAll();
```
gino-m, line 31: Why return a Flowable here? Do we want to subscribe to future tile changes?

gnd/src/test/java/com/google/android/gnd/TestScheduler.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.inject.Inject;
+
+public class TestScheduler implements Schedulers {
+
+  @Inject
+  TestScheduler() {}
+
+  @Override
+  public Scheduler io() {
+    return io.reactivex.schedulers.Schedulers.trampoline();
```
gino-m, line 30: Neat.. could you please add a class description that explains what this does in practice?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,565 @@
...
+  @Rule public InstantTaskExecutorRule instantTaskExecutorRule = new InstantTaskExecutorRule();
+
+  @Inject LocalDataStore localDataStore;
+
+  @Before
+  public void setUp() {
+    DaggerTestComponent.create().inject(this);
+  }
+
+  private User createUser() {
```
gino-m, line 80: Please make helpers static, include word "Test" so it's clear these don't just create a new empty instance (e.g. `createTestUser()`).

shobhitagarwal1612:
I've finished testing the `LocalDataStore` class (96% coverage).

gnd/src/main/java/com/google/android/gnd/persistence/local/room/dao/TileDao.java
shobhitagarwal1612: The corresponding method in `LocalDataStore` was unused. But I think @scolsen would be needing that later since he added them. I've added tests for the same to verify that it indeed works as it says.

gnd/src/test/java/com/google/android/gnd/TestScheduler.java
shobhitagarwal1612: Sure

shobhitagarwal1612:
> Looking at the tests, do you think it makes sense to split up the LocalDataStore? We don't do this for RemoteDataStore, but we could. The downside is that we'd need to implement multiple interfaces to implement a new persistence provider, but that would rarely happen in practice, and could be mitigated by using inner interfaces to consolidate them into a single file.

That depends on how we are splitting up the `LocalDataStore`. If it is by model type (similar to repository pattern), then I'd say it would be an overkill because each of those would have 1 or 2 tests each. Having them together does make sense because of the constraints. If it were separate, I'd have to write the same createTest*() methods for each or (even worse) move them into a common other. But that being said, I'm all up for simplifying the class, wherever possible.

config/spotbugs/spotbugs-filter.xml
```diff
@@ -80,4 +81,7 @@
   <Match>
     <Bug pattern="UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR" />
   </Match>
+  <Match>
+    <Bug pattern="URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD" />
```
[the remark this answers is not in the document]
shobhitagarwal1612: That didn't work. I also tried `SuppressFBWarnings` ([docs](url)) but even that isn't working.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
shobhitagarwal1612: I even found a bug while testing.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,611 @@
...
+    Assert.assertEquals(layer, newFeature.getLayer());
+    Assert.assertNull(newFeature.getCustomId());
+    Assert.assertNull(newFeature.getCaption());
+    Assert.assertEquals(point, newFeature.getPoint());
+    Assert.assertEquals(user, newFeature.getCreated().getUser());
+    Assert.assertEquals(user, newFeature.getLastModified().getUser());
+  }
+
+  @Test
+  public void testApplyAndEnqueue_observationMutation() {
```
gino-m, line 365: This test (and ones like it) look they're testing more than one behavior, which is why they're longer . Can we split these into clear preconditions, execute, and postconditions? See previous comments for ways to simplify.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,611 @@
...
+    Assert.assertNull(feature.getCaption());
+    Assert.assertEquals(mutation.getNewLocation().get(), feature.getPoint());
+    Assert.assertEquals(user, feature.getCreated().getUser());
+    Assert.assertEquals(user, feature.getLastModified().getUser());
+  }
+
+  @Test
+  public void testGetFeaturesOnceAndStream() {
+    User user = createTestUser();
+    localDataStore.insertOrUpdateUser(user).test().assertComplete();
```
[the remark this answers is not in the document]
shobhitagarwal1612: `insertOrUpdateUser` needs to be subscribed to create the user. It's either `test().assertComplete()` or `blockingAwait()` or `subscribe()`. I thought the cleanest would be to use `test()`

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,611 @@
...
+    localDataStore.insertOrUpdateProject(project).test().assertComplete();
+
+    Layer layer = project.getLayers().get(0);
+    FeatureMutation mutation =
+        createTestFeatureMutation(user.getId(), project.getId(), layer.getId());
+    localDataStore.applyAndEnqueue(mutation).test().assertComplete();
+
+    ImmutableList<Mutation> savedMutations =
+        localDataStore.getPendingMutations(mutation.getFeatureId()).blockingGet();
+    Assert.assertEquals(1, savedMutations.size());
```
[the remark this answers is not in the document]
shobhitagarwal1612: Thanks. `Truth` assertions are also very easy to read.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,611 @@
...
+            .setId("id 2")
+            .setTitle("project 2")
+            .setDescription("foo description 2")
+            .build();
+    localDataStore.insertOrUpdateProject(project1).test().assertComplete();
+    localDataStore.insertOrUpdateProject(project2).test().assertComplete();
+    localDataStore
+        .getProjects()
+        .test()
+        .assertValue(ImmutableList.<Project>builder().add(project1, project2).build());
```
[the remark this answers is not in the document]
shobhitagarwal1612: Thanks. I didn't realize that this is also available. Done

shobhitagarwal1612:
> The semantics of Single and Flowable are different in Room generated classes; Single returns a result and completes, whereas Flowable continues emitting values on each change. We convert to Flowable in upstream code because we're merging with a hot observable that doesn't complete as soon as a value is emitted. Here we expect the stream to terminate after the first item is emitted, so Single is appropriate here.

Reverting back `Flowable` to `Single` fails the test `testGetTilesOnceAndStream`. Can you or @scolsen please take a look at the test to verify that I am testing the correct behavior or not?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,611 @@
...
+            .setListHeading("heading title")
+            .setDefaultStyle(Style.builder().setColor("000").build())
+            .setForm(form)
+            .build();
+
+    Project.Builder builder =
+        Project.newBuilder()
+            .setId("project id")
+            .setTitle("project 1")
+            .setDescription("foo description");
```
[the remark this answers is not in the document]
shobhitagarwal1612: Thanks. Done

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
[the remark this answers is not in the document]
shobhitagarwal1612: On second thought, we can use `subscribe` for simplicity and to remove redundancy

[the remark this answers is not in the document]
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/persistence/local/room/dao/TileDao.java
shobhitagarwal1612: > The semantics of Single and Flowable are different in Room generated classes; Single returns a result and completes, whereas Flowable continues emitting values on each change. We convert to Flowable in upstream code because we're merging with a hot observable that doesn't complete as soon as a value is emitted. Here we expect the stream to terminate after the first item is emitted, so Single is appropriate here.

Reverting back Flowable to Single fails the test testGetTilesOnceAndStream. @gino-m Can you or @scolsen please take a look at the test to verify that I am testing the correct behavior or not?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
shobhitagarwal1612: Done. PTAL

shobhitagarwal1612:
@gino-m I've resolved all issues. PTAL again!

shobhitagarwal1612:
@gino-m I've made the final set of requested changes. Should be ready to go now!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -268,7 +268,7 @@ public Completable applyAndEnqueue(FeatureMutation mutation) {
   @Override
   public Flowable<ImmutableSet<Tile>> getTilesOnceAndStream() {
     return tileDao
-        .findAll()
+        .getTilesOnceAndStream()
```
gino-m, line 271: let's stick with the "findAll" prefix for consistency with other DAOs, so "`findAllOnceAndStream`?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -67,97 +66,138 @@
...
+          .setResponseDeltas(
+              ImmutableList.of(
+                  ResponseDelta.builder()
+                      .setFieldId("field id")
+                      .setNewResponse(TextResponse.fromString("response for field id"))
+                      .build()))
+          .setClientTimestamp(new Date())
+          .build();
+
+  private static final Tile TEST_TILE_PENDING =
```
gino-m, line 146: Nit: Please use natural word order for readability: TEST_PENDING_TILE here and below.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -67,97 +66,138 @@
...
-        .setType(Mutation.Type.CREATE)
-        .setUserId(userId)
-        .setProjectId(projectId)
-        .setFeatureId("feature id")
-        .setLayerId(layerId)
-        .setNewLocation(
-            Optional.ofNullable(Point.newBuilder().setLatitude(110.0).setLongitude(-23.1).build()))
-        .setClientTimestamp(new Date())
-        .build();
+  private static void assertObservationMutation(
```
gino-m, line 182: One asserts that something Xs, so here it would be `assertObservationMutationEquals`.

But this also does something extra, so should be reflected in the name, e.g.:
`assertEqualsIgnoreId`.

We can omit "ObservationMutation" from the name for brevity since it will be clear from the arguments what's being compared.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -67,97 +66,138 @@
...
+      ObservationMutation expected, ObservationMutation actual) {
+    // TODO: Id is auto-assigned to ObservationMutation.
+    //  If we try to give it while inserting, then it causes problems. Improve this behavior.
+    //  So, copy the id from actual to expected and then compare the objects.
+    expected = expected.toBuilder().setId(actual.getId()).build();
+    Assert.assertEquals(expected, actual);
```
gino-m, line 188: Let's use Google Truth consistently here and throughout to set a stable precedent and to get more useful error messages. i.e. assertThat(actual).isEqualTo(expected).

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -67,97 +66,138 @@
...
-        .setUserId(userId)
-        .build();
+  private static void assertObservation(ObservationMutation mutation, Observation observation) {
+    Assert.assertEquals(mutation.getObservationId(), observation.getId());
+    Assert.assertEquals(mutation.getFeatureId(), observation.getFeature().getId());
+    Assert.assertEquals(TEST_USER, observation.getCreated().getUser());
+    Assert.assertEquals(TEST_FORM, observation.getForm());
+    Assert.assertEquals(TEST_PROJECT, observation.getProject());
+    Assert.assertEquals(TEST_USER, observation.getLastModified().getUser());
+    MatcherAssert.assertThat(
```
gino-m, line 198: Same here: https://truth.dev/

Please be sure to use static imports to make assertions more fluent.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -167,137 +207,97 @@ public void setUp() {
...
-            .setTitle("project 2")
-            .setDescription("foo description 2")
-            .build();
-    localDataStore.insertOrUpdateProject(project1).test().assertComplete();
-    localDataStore.insertOrUpdateProject(project2).test().assertComplete();
-    localDataStore
-        .getProjects()
-        .test()
-        .assertValue(ImmutableList.<Project>builder().add(project1, project2).build());
+    Project project = TEST_PROJECT;
```
gino-m, line 210: Please use constants directly, here and throughout; extra indirection makes tests [marginally] harder to ready.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -167,137 +207,97 @@ public void setUp() {
...
+    localDataStore.insertOrUpdateProject(project).subscribe();
+    localDataStore.getProjects().test().assertValue(ImmutableList.of(project));
   }
 
   @Test
   public void testGetProjectById() {
-    Project project = createTestProject();
-    localDataStore.insertOrUpdateProject(project).test().assertComplete();
+    Project project = TEST_PROJECT;
+    localDataStore.insertOrUpdateProject(project).subscribe();
```
gino-m, line 218: Please use blocking* instead of `subscribe` here and throughout.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -167,137 +207,97 @@ public void setUp() {
...
-    Assert.assertEquals(0, savedMutation.getRetryCount());
-    Assert.assertNull(savedMutation.getLastError());
+    localDataStore
+        .getPendingMutations(mutation.getFeatureId())
+        .test()
+        .assertValue(ImmutableList.of(mutation));
 
     // assert feature is saved to local database
-    Feature feature = localDataStore.getFeature(project, mutation.getFeatureId()).blockingGet();
+    Feature feature =
```
gino-m, line 252: Can this entire block be replaced with:
```
   localDataStore.getFeature(TEST_PROJECT, mutation.getFeatureId()).test().assertValue(TEST_FEATURE);
```
?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -352,136 +341,75 @@ public void testMergeFeature() {
...
-    Assert.assertEquals(mutation.getResponseDeltas(), savedMutation.getResponseDeltas());
-    Assert.assertEquals(mutation.getType(), savedMutation.getType());
-    Assert.assertEquals(mutation.getUserId(), savedMutation.getUserId());
-    Assert.assertEquals(mutation.getProjectId(), savedMutation.getProjectId());
-    Assert.assertEquals(mutation.getFeatureId(), savedMutation.getFeatureId());
-    Assert.assertEquals(layer.getId(), savedMutation.getLayerId());
-    Assert.assertEquals(mutation.getClientTimestamp(), savedMutation.getClientTimestamp());
-    Assert.assertEquals(0, savedMutation.getRetryCount());
-    Assert.assertNull(savedMutation.getLastError());
+    assertObservationMutation(mutation, (ObservationMutation) savedMutations.get(1));
```
gino-m, line 366: You probably will also want to `assertThat(mutation).isInstanceOf(ObservationMutation.class)`

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -352,136 +341,75 @@ public void testMergeFeature() {
...
-                    .setFieldId("really new field")
-                    .setNewResponse(TextResponse.fromString("value for the really new field"))
-                    .build())
-            .build();
+        ImmutableList.of(
+            ResponseDelta.builder()
+                .setFieldId("really new field")
+                .setNewResponse(TextResponse.fromString("value for the really new field"))
+                .build());
     mutation = mutation.toBuilder().setResponseDeltas(deltas).setType(Mutation.Type.UPDATE).build();
```
gino-m, line 382: Please replace secondary assertions (i.e. those that don't test the behavior in question) with `blocking*().` If those code paths don't already have their own tests, please be sure to add them.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
shobhitagarwal1612: That would require mocking a static method. For that power mock can be used. https://github.com/robolectric/robolectric/wiki/Using-PowerMock

But I didn't have any success with that.

config/spotbugs/spotbugs-filter.xml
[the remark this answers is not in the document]
gino-m: Did you try setting value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD" in the annotation?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+  private static void assertEqualsIgnoreId(
+      ObservationMutation expected, ObservationMutation actual) {
+    // TODO: Id is auto-assigned to ObservationMutation.
+    //  If we try to give it while inserting, then it causes problems. Improve this behavior.
+    //  So, copy the id from actual to expected and then compare the objects.
+    expected = expected.toBuilder().setId(actual.getId()).build();
+    assertThat(expected).isEqualTo(actual);
+  }
+
+  private static void assertObservation(ObservationMutation mutation, Observation observation) {
```
gino-m, line 191: Naming nit: assert that observation is what? assertEquivalent?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    assertThat(mutation.getObservationId()).isEqualTo(observation.getId());
+    assertThat(mutation.getFeatureId()).isEqualTo(observation.getFeature().getId());
+    assertThat(TEST_USER).isEqualTo(observation.getCreated().getUser());
```
gino-m, line 194: Seems a bit surprising that this method would compare to constant values - should we be looking instead at the user, form, project in the mutation instead?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    assertThat(TEST_FORM).isEqualTo(observation.getForm());
+    assertThat(TEST_PROJECT).isEqualTo(observation.getProject());
+    assertThat(TEST_USER).isEqualTo(observation.getLastModified().getUser());
+    MatcherAssert.assertThat(
+        ResponseMap.builder().applyDeltas(mutation.getResponseDeltas()).build(),
+        samePropertyValuesAs(observation.getResponses()));
+  }
+
+  private static void assertFeature(String featureId, Point point, Feature feature) {
```
gino-m, line 203: Same here - the name seems to be missing what is actually being asserted (i.e. or it seems to imply it's just checking that the supplied value is a feature)

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    assertThat(featureId).isEqualTo(feature.getId());
+    assertThat(TEST_PROJECT).isEqualTo(feature.getProject());
```
gino-m, line 205: Same comment here re constants.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+  @Test
+  public void testDeleteProject() {
+    localDataStore.insertOrUpdateProject(TEST_PROJECT).blockingAwait();
+    localDataStore.deleteProject(TEST_PROJECT).test().assertComplete();
+    localDataStore.getProjects().test().assertValue(AbstractCollection::isEmpty);
+  }
+
+  @Test
+  public void testInsertUser() {
+    localDataStore.insertOrUpdateUser(TEST_USER).test().assertComplete();
```
gino-m, line 246: In `testDeleteProject` we checked that the project was actually deleted. Should we do the same here for user insertion? The method could then be called `testInsertAndGetUser` and the next method deleted.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+  @Test
+  public void testGetFeaturesOnceAndStream() {
+    localDataStore.insertOrUpdateUser(TEST_USER).blockingAwait();
+    localDataStore.insertOrUpdateProject(TEST_PROJECT).blockingAwait();
+
+    TestSubscriber<ImmutableSet<Feature>> subscriber =
+        localDataStore.getFeaturesOnceAndStream(TEST_PROJECT).test();
+
+    subscriber.assertValueCount(1);
+    subscriber.assertValueAt(0, AbstractCollection::isEmpty);
```
gino-m, line 282: Since we check this at the end of the method, do we also need to check it here?

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+
+    FeatureMutation mutation = TEST_FEATURE_MUTATION;
```
gino-m, line 284: Please inline this variable since it always contains a constant.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    localDataStore.applyAndEnqueue(mutation).blockingAwait();
+
+    Feature feature =
+        localDataStore.getFeature(TEST_PROJECT, mutation.getFeatureId()).blockingGet();
+
+    subscriber.assertValueCount(2);
+    subscriber.assertValueAt(0, AbstractCollection::isEmpty);
```
gino-m, line 291: We could also use `subscriber.assertValues()` here instead of two `assertValueAt` calls.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    subscriber.assertValueAt(1, ImmutableSet.of(feature));
+  }
+
+  @Test
+  public void testUpdateMutations() {
+    localDataStore.insertOrUpdateUser(TEST_USER).blockingAwait();
+    localDataStore.insertOrUpdateProject(TEST_PROJECT).blockingAwait();
+
+    FeatureMutation mutation = TEST_FEATURE_MUTATION;
```
gino-m, line 300: Please inline.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    localDataStore.applyAndEnqueue(mutation).blockingAwait();
+
+    Point newPoint = Point.newBuilder().setLatitude(51.0).setLongitude(44.0).build();
```
gino-m, line 303: Shall we create a constant, TEST_POINT_2, for example, for use here and throughout instead? Otherwise it may seem like these values might have special meaning.

Also, consider creating a constructor method createTestFeatureMutation(point) that you can use for both cases.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+    Mutation updatedMutation =
+        mutation.toBuilder().setNewLocation(Optional.ofNullable(newPoint)).build();
+
+    localDataStore.updateMutations(ImmutableList.of(updatedMutation)).test().assertComplete();
+
+    ImmutableList<Mutation> savedMutations =
+        localDataStore.getPendingMutations(updatedMutation.getFeatureId()).blockingGet();
+    assertThat(savedMutations).hasSize(1);
```
gino-m, line 311: Please use `assertThat(savedMutations).containsExactly` instead of `hasSize`, `get`, `isEqualTo`.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
```diff
@@ -0,0 +1,480 @@
...
+  public void testApplyAndEnqueue_observationMutation() {
+    localDataStore.insertOrUpdateUser(TEST_USER).blockingAwait();
+    localDataStore.insertOrUpdateProject(TEST_PROJECT).blockingAwait();
+    localDataStore.applyAndEnqueue(TEST_FEATURE_MUTATION).blockingAwait();
+
+    localDataStore.applyAndEnqueue(TEST_OBSERVATION_MUTATION).test().assertComplete();
+
+    ImmutableList<Mutation> savedMutations =
+        localDataStore.getPendingMutations("feature id").blockingGet();
+    assertThat(savedMutations).hasSize(2);
```
gino-m, line 359: Same here re containsExactly. I think it will also check the type for you.

gnd/src/test/java/com/google/android/gnd/persistence/local/LocalDataStoreTest.java
shobhitagarwal1612: makes sense. Removing them

shobhitagarwal1612:
> Did you try setting value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD" in the annotation?

Yes, I don't know why it didn't work.

shobhitagarwal1612:
@gino-m This looks much cleaner and simpler too. Thanks for your keen observations.

gino-m:
Great work as always Shobhit. Really glad to have these in place!

PR #427, opened 2020-04-07
Add a couple of extra bits of info

dturner:
I realise now I should've done this on my own user branch, rather than using `ground-android`. Apologies. Will do so in future.


gino-m:
Thanks for this addition @shobhitagarwal1612 !

PR #436, opened 2020-04-11
Reload observation list when restoring feature sheet

### LINKED ISSUE #185: [Add observation] Observations sometimes do not appear in list after saving

gino-m:
Fixes #185.

When returning back to the observation list from the edit observation screen, the list was not reloaded because the selected Feature was only emitted when the user clicks on the map; observing the `LiveData` once more did not cause the selected Feature to be reemitted. Using a BehaviorProcessor, the last value is always reemitted on new subscriptions, causing the list to be reloaded whenever the view is restored.


gino-m:
Actually.. I think this is the wrong way to do this; as it stands, from the view's perspective there's no difference between the user clicking a feature, and a configuration change. This will make it hard for us to maintain the scroll position on config changes. I'll close this for now and try a different approach.

gino-m:
And on third thought.. 

The alternative approach I implemented came up with () looks at the current state of the `LiveData` in onCreateView to restore the UI state. The drawback of this approach is that we mix side-effects, when a purely functional approach would work just as well. 

Instead, what I think we should do (in a future PR) is to emit the scroll position along with the selected Feature. This way to current state is always communicated from the VM to the V together in the same stream. Please lmk if this makes sesn.

shobhitagarwal1612:
Can we merge this?

gino-m:
> Can we merge this?

Waiting for stamp - would you like give it a look

shobhitagarwal1612:
Thanks Gino!

PR #444, opened 2020-04-13
Fix app crashes on missing default layer style

### LINKED ISSUE #433: [Offline areas] Add bottom sheet to area selector with used/free/avail/required estimates
Potential layout:
image

@shobhitagarwal1612 FYI.

### LINKED ISSUE #443: [Project selection] App crashes with "Animal sightings" test db
Some dbs will cause the app to crash since null styles are not tolerated.

gino-m:
Fixes #443. Older projects didn't have defaultStyle set for all layers, causing the app to crash.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/Config.java
+++ b/gnd/src/main/java/com/google/android/gnd/Config.java
@@ -21,7 +21,7 @@ public final class Config {
 
   // Local db settings.
   // TODO(#128): Reset version to 1 before releasing.
-  public static final int DB_VERSION = 42;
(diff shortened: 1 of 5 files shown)


shobhitagarwal1612:
@gino-m The bug linked in the description doesn't seem related. Can you please fix that?

shobhitagarwal1612:
Changes look good but the InMemoryTests are failing.

gino-m:
> Changes look good but the InMemoryTests are failing.

Fixed! Thanks!

PR #452, opened 2020-04-21
[Sync] Execute worker requests under foreground service

### LINKED ISSUE #430: [Background sync] Use foreground service / workers for sync
Originated from #421 @gino-m. says, 

> Q. Foreground service should be running at all times or periodically?
> A. Two planned features will require the service to be run at all times: GPS tracks recording, and syncing added/changed/removed features from joined remote projects to the local device.
> 
> Q. Clicking the "App is running" notification should perform some action?
> A. I assume clicking the notification will take us back into the app home for now. In the future we could also direct to a more detailed status page.
> 
> More general question - what belongs in the Workers (if anything) vs in the foreground Service? On the one hand, it would simplify implementation and give us greater control to sync all data in the same service. On other, with worker we get job constraints (network availability, for example) and parallelism for free.
> 
> One possible prioritization of tasks could be:
> 
> Upload local changes to observations and features (except photos)
> In parallel (?):
> a. Smaller new/changed features in joined projects
> b. Download satellite imagery
> c. Upload photos
> d. Upload new GPS tracks data [TBD]
> There may be two options in settings to constrain when smaller syncs (1 and 2a) and larger ones (2b, 2c) can be executed. I wonder if all sync should be done in the service because:
> 
> The user can change the sync constraints after a sync task is queued. If the worker is already queued, we'd need to add logic to handle this, possibly by aborting and requeueing with the new constraints. (this adds complexity and carrying cost)
> It's not obvious how to implement the ordering of our idempotent workers wrt our queue and network constraints. Putting them in one service might make it clearer how these various tasks interact.
> Curious to hear your thoughts, incl. @scolsen @dturner? Thanks!

shobhitagarwal1612:
Fixes #430 

Things done:
 - All workers create their separate notifications instead of reusing the same id. This is to prevent any running worker from getting interrupted by another worker.
 - Call `setForegroundAsync` to execute worker request under foreground service
 - Moved worker for tileDownload under `persistence/sync` 

Notifications are canceled automatically by the system once the request completes.

Tested manually:
1. Turn off wifi
2. Open app 
3. Select project `Plastic Frog Finder`
4. Create a new feature by using '+' button. Then create a new observation
5. Fill some random data and also attach a photo.
6. Click `Save`
7. Killed the app from recent apps drawer
8. Turn on wifi
9. Notifications for photo and data sync are visible


gino-m:
Sorry for the delay! A few questions:

> All workers create their separate notifications instead of reusing the same id. This is to prevent any running worker from getting interrupted by another worker.

Why would the notification id interrupt the worker? Do you mean "to prevent existing notifications from getting interrupted" rather than "running worker"?

General question - data sync workers use `ExistingWorkPolicy.APPEND`, so they wouldn't get interrupted. What is our sync policy for other workers? 

> Call setForegroundAsync to execute worker request under foreground service

How many notifications are generated? One for each sync type, one for each event, or a single notification continually updated? The latter would be preferable.

shobhitagarwal1612:
> Why would the notification id interrupt the worker? Do you mean "to prevent existing notifications from getting interrupted" rather than "running worker"?
General question - data sync workers use ExistingWorkPolicy.APPEND, so they wouldn't get interrupted. What is our sync policy for other workers?

Yes, to prevent it from getting interrupted.
Let's assume that we started downloading offline satellite imagery for a large area. So, a notification would start for that. Now, if we also add a new feature or observation, the same notification would be updated. So, we'd see disappearing of download progress which might confuse users.

> How many notifications are generated? One for each sync type, one for each event, or a single notification continually updated? The latter would be preferable.

One for each type. When the task gets finished, then the notification also disappears.

PR #453, opened 2020-04-23
[WORK IN PROGRESS] Split feature sheet header out of bottom sheet scroll view

gino-m:


gnd/src/main/java/com/google/android/gnd/ui/home/BottomSheetChromeBehavior.java
```diff
@@ -39,6 +40,10 @@ public BottomSheetChromeBehavior(Context context, AttributeSet attrs) {
   @Override
   protected void onSheetScrolled(
       CoordinatorLayout parent, ViewGroup layout, SheetSlideMetrics metrics) {
+    // Slide feature sheet header up and down together with bottom sheet.
+    View header = parent.findViewById(R.id.feature_sheet_header);
+    header.setY(metrics.getTop() -  header.getHeight());
```
gino-m, line 45: @dturner this is what I had in mind to simplify obscuring the sheet header behind the toolbar; we can split it out completely into the parent FrameLayout and control its position here.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -153,6 +155,15 @@ public View onCreateView(
 
     HomeScreenFragBinding binding = HomeScreenFragBinding.inflate(inflater, container, false);
     binding.featureSheetChrome.setViewModel(viewModel);
+    binding
+        .featureSheetHeader
+        .getRoot()
+        .setOnTouchListener(
+            (v, event) -> {
+              Log.e("!!!", "Touch: " + event);
+              return binding.bottomSheetScrollView.dispatchTouchEvent(event);
```
gino-m, line 164: @dturner ..but I'm having trouble transferring touch events from the header into the bottom sheet. It works when clicking or dragging on some parts of the header, but when clicking on text or the image the touch listener here isn't called. Any ideas how to capture touch events recursively?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -316,30 +311,25 @@ private void showOfflineAreas() {
...
-    double mapHeight = width / COLLAPSED_MAP_ASPECT_RATIO;
-    double peekHeight = height - mapHeight;
-    bottomSheetBehavior.setPeekHeight((int) peekHeight);
+  private int getBottomSheetPeekHeight(WindowInsetsCompat insets) {
+    double screenWidth = getScreenWidth(getActivity());
+    double screenHeight = getScreenHeight(getActivity());
+    double collapseMapHeight = screenWidth / COLLAPSED_MAP_ASPECT_RATIO;
+    int insetTop = insets.getSystemWindowInsetTop();
+    int headerHeight = getView().findViewById(R.id.feature_sheet_header).getHeight();
+    return (int) (screenHeight - insetTop - collapseMapHeight - headerHeight);
```
gino-m, line 332: @dturner The resulting map thumbnail aspect ratio appears to be wrong but not really sure why; I'd expect it to be 4:3 as defined by `COLLAPSED_MAP_ASPECT_RATIO` but it appears closer to 1:1. I'm probably doing the math wrong here, but not sure where.

Screenshot_1587604134

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -238,17 +233,17 @@ private String getVersionName() {
...
+    if (toolbarWrapper == null || bottomSheetBehavior == null || insets == null) {
       return;
     }
     bottomSheetBehavior.setFitToContents(false);
 
     // When the bottom sheet is expanded, the bottom edge of the header needs to be aligned with
     // the bottom edge of the toolbar (the header slides up under it).
-    bottomSheetBehavior.setExpandedOffset(
-        toolbarWrapper.getHeight() - bottomSheetHeader.getHeight());
+    bottomSheetBehavior.setExpandedOffset(toolbarWrapper.getHeight());
```
gino-m, line 243: @dturner The latest version of BottomSheetBehavior now exposes setExpandedOffset, but it must be non-negative. Breaking the header out allows us to set a positive expandedOffset (the height of the scrim for the top inset).

gino-m:
@dturner a few comments on my work in progress - could you please have a look?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
dturner: If by "map thumbnail" you mean the map area above the bottom sheet it could be the line `screenHeight - insetTop - collapseMapHeight - headerHeight` because presumably what you actually want is `screenHeight  - (insetTop + collapseMapHeight + headerHeight)`. Difficult to debug without step debugging.

dturner:
@gino-m Not really sure where to start with this. I understand that this change splits the bottom sheet header into a separate layout which is tracked independently of the bottom sheet to "simplify" things, however, I'm not seeing much simplification. If anything the code and architecture is more difficult to understand than before. 

Also, the fact that touch events now have to be propagated up from the header, then back down to the bottom sheet feels fairly a bit hacky to me. 

Perhaps this is because I'm still struggling with the UI architecture. Maybe a VC to discuss would be beneficial.

gino-m:
Hi Don, thanks for the comments. I've already started integrating FloatingHeaderBehavior as discussed. Closing in favor of https://github.com/google/ground-android/pull/462 (WIP).

PR #457, opened 2020-04-26
Refactor EditObservationFragment

shobhitagarwal1612:
- [x] Move UI creation for each field type into a separate class 
- [x] Create field view generator class
- [x] Decouple layout xml for each field view from `EditObservationFragment`
- [x] Move logic for updating field view UI from EditObservationViewModel to another VM
- [x] Decouple bottomSheetDialog used in PhotoView from `EditObservationViewModel`


shobhitagarwal1612:
@gino-m PTAlL

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -133,101 +129,11 @@ private void rebuildForm(Form form) {
...
-      bottomSheetDialog.show();
-    }
-  }
-
   @Override
   public void onPause() {
-    if (bottomSheetDialog != null && bottomSheetDialog.isShowing()) {
-      bottomSheetDialog.dismiss();
+    for (FieldView fieldView : fieldViews) {
+      fieldView.onPause();
```
gino-m, line 135: Doesn't pause get propagated to child components automatically? And if not, wouldn't we also need to propagate other events like onStart()?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -354,25 +243,13 @@ private static boolean isAddObservationRequest(EditObservationFragmentArgs args)
...
-    }
-  }
-
   private ImmutableList<ResponseDelta> getResponseDeltas() {
     if (originalObservation == null) {
       Timber.e("Response diff attempted before observation loaded");
       return ImmutableList.of();
     }
-    ImmutableList.Builder<ResponseDelta> deltas = ImmutableList.builder();
+    Builder<ResponseDelta> deltas = ImmutableList.builder();
```
gino-m, line 251: Please qualify `Builder` with it's outer class name.

gnd/src/main/java/com/google/android/gnd/ui/field/MultipleChoiceFieldView.java
```diff
@@ -0,0 +1,78 @@
...
+      binding.setViewModel(getViewModel());
+      binding.setLifecycleOwner(getLifecycleOwner());
+      binding.setField(field);
+    }
+  }
+
+  public void onShowDialog() {
+    Cardinality cardinality = field.getMultipleChoice().getCardinality();
+    Optional<Response> currentResponse = getViewModel().getResponse(field.getId());
+    switch (cardinality) {
```
gino-m, line 52: Line 50 and this switch might be cleaner in their own method (@Nullable getDialog())

gnd/src/main/java/com/google/android/gnd/ui/common/BindingAdapters.java
```diff
@@ -31,7 +31,7 @@
 import com.google.android.gnd.databinding.TextInputFieldBinding;
 import com.google.android.gnd.model.form.Field;
 import com.google.android.gnd.model.observation.Response;
-import com.google.android.gnd.ui.editobservation.MultipleChoiceFieldLayout;
+import com.google.android.gnd.ui.field.MultipleChoiceFieldLayout;
```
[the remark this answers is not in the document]
shobhitagarwal1612: Both read-only and editable fields.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
shobhitagarwal1612: It wasn't needed because we are re-creating the views and only VM is retained. `onPause()` was needed because I had to handle state of AlertDialog when app is paused.

shobhitagarwal1612:
> Hi Shobhit! I've added a few general comments/questions. Is there a way to break this up into a few distinct PRs to make them easier to conceptualize? I'm having a hard time understanding all the changes being done here and how they interact. Wdyt?

Thanks for your suggestions. I'll try to explain the motives behind the changes.

 - A lot of code related to creating field views and storing the state is tangled with `EditObservationFragment` and `EditObservationViewModel`

 - This reduces the code readability and with scale, these classes would become unmaintainable and very hard to test.

 - The main idea is to separate out all the logic for UI and ViewModel related to a field into a separate package. So, if `EditObservationFragment` (or `ObservationDetailsFragment`) needs to ever change the way the fields are arranged in the main UI, then it can be done without any extra effort. Another benefit would be easy addition of new types as the project grows with better testability.

 - `EditObservtionViewModel` would then talk to `FieldViewModel` to get a snapshot of current state for saving form, etc. 


Summarized flow:
image

I'm more than happy to split the code into multiple PRs if you think that the design is going correctly so far. Otherwise, let's discuss the architecture first.

gino-m:
Thanks for the diagram and the explanation! It's a bit clearer now. It's mainly the volume of the change that makes it a bit hard to follow in a single PR.

A few observations:

* Based on examples and docs I've seen, the dependency graph is always View -> ViewModel. ViewModels don't interactive with other ViewModels; since ViewModels represents a View's state, exposing the state of two views directly to each other breaks the ViewModel abstraction. 

Rather than have ViewModels talk to each other, can we pass the shared parent ViewModel to the factory, and have the field View channel events from its ViewModel to the parent ViewModel? Does this make sense?

shobhitagarwal1612:
Thanks for the comments! I'll improve the design in the next few days.

gino-m:
> Thanks for the comments! I'll improve the design in the next few days.

Thanks! Please pingif you have any questions. The easiest way would be to send a small PR refactoring a single element into a View-ViewModel (construction can be hard coded for now for example). Wdyt?

shobhitagarwal1612:
> Thanks! Please ping if you have any questions. An easy way to discuss/share would be to send a small PR refactoring a single element into a View-ViewModel (construction can be hard coded for now for example). Wdyt?

I've thought about it and have some leads in the right direction. Stay tuned!

shobhitagarwal1612:
@gino-m I've fixed the code to isolate the two view models. PTAL

new flowchart

shobhitagarwal1612:
FYI, @gino-m and I had a long discussion about the whole design and we agreed upon reusing some ideas from this proposal but stick to standard design patterns (as suggested by Android team). So, I'm starting it from scratch. Hoping to have something soon for review

gino-m:
> FYI, @gino-m and I had a long discussion about the whole design and we agreed upon reusing some ideas from this proposal but stick to standard design patterns (as suggested by Android team). So, I'm starting it from scratch. Hoping to have something soon for review

Sorry for the design churn. I think it's definitely been a valuable and necessary exercise so that we can come up with a consistent approach to building these more complex UIs. Thanks for taking this on! 🚀

PR #462, opened 2020-04-28
[WIP] [Feature sheet] Refactor bottom sheet using custom FloatingHeaderBehavior

gino-m:
- [x] Split up and simplify BottomSheetBehaviors
- [x] Split up and simplify metrics an transitions
- [x] Let header remain on screen when bottom sheet expanded, simplifying code and obviating need for negative expanded offsets or other workarounds.
- [ ] Fix: Bottom sheet is collapsed and screen remains frozen after returning from add/edit observation screens

Punted additional fixes, refactors, and UI tweaks to follow-up PR(s):
- [ ] Additional simplifications to feature sheet layout file structure
- [ ] Consistent naming of feature/bottom sheet components
- [ ] Fix styles to work with latest  Material components, upgrade version
- [ ] Delete custom BottomSheetBehavior class
- [ ] Add back drop shadow below header / above observation list
- [ ] Fix: Marker not centered in collapsed map
- [ ] Remove toolbar drop shadow when  bottom sheet is expanded
- [ ] Make toolbar background transparent until sheet is expanded

Expanded/collapsed states:

Screenshot_1588718397

Screenshot_1588718395


gino-m:
@dturner Almost there.. only issue remaining is the bottom sheet UI state doesn't seem to get persisted after tapping an item in the list; the sheet is returned to its HIDDEN state, even though the HomeScreenFragment still has the map as locked. This works in `master`. Any idea what in this change could have prevented the BottomSheetBehavior from preserving its state?

gino-m:
> @dturner Almost there.. only issue remaining is the bottom sheet UI state doesn't seem to get persisted after tapping an item in the list; the sheet is returned to its HIDDEN state, even though the HomeScreenFragment still has the map as locked. This works in `master`. Any idea what in this change could have prevented the BottomSheetBehavior from preserving its state?

Please disregard - I've started a new PR (#463) to merge in relevant pieces of this change incrementally.

PR #463, opened 2020-05-06
[Feature sheet] Simplify bottom sheet header design

gino-m:
Move header into bottom sheet and stop requiring negative expanded offset. This unblocks upgrading to latest Material components and upcoming simplifications and cleanup. See commit history for details.


gino-m:
@dturner could you PTAL? I have a few other cleanups and fixes, but I'll send out small PRs to make reviews more manageable.

PR #464, opened 2020-05-07
[Feature sheet] Refactor/rename key bottom sheet elements

gino-m:
See commits for details.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenMetrics.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenMetrics.java
@@ -23,11 +23,11 @@ import androidx.coordinatorlayout.widget.CoordinatorLayout;
 import com.google.android.gnd.R;
 import com.google.android.gnd.ui.common.BottomSheetBehavior;
 
-public class SheetSlideMetrics {
+public class HomeScreenMetrics {
   private final CoordinatorLayout parent;
   private final View view;
 
-  public SheetSlideMetrics(CoordinatorLayout parent, View view) {
+  public HomeScreenMetrics(CoordinatorLayout parent, View view) {
     this.parent = parent;
     this.view = view;
   }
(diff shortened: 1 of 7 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenMetrics.java
```diff
@@ -0,0 +1,101 @@
...
+public class HomeScreenMetrics {
+  private final CoordinatorLayout parent;
+  private final View view;
+
+  public HomeScreenMetrics(CoordinatorLayout parent, View view) {
+    this.parent = parent;
+    this.view = view;
+  }
+
+  public static float scale(float value, float before1, float before2, float after1, float after2) {
```
dturner, line 35: Since these are all public methods it might be useful to future readers to add a comment to each one explaining what it does (even if it's obvious from the method name) and the context in which it's likely to be used

gino-m: I've actually rewritten this class completely to use pixel values instead of ratios for anchor points. That version has JavadDoc.

shobhitagarwal1612:
This is strange! 
The GCB finished 10 mins after it was triggered but it is still showing "in progress" here in the PR.

gino-m:
> This is strange!
> The GCB finished 10 mins after it was triggered but it is still showing "in progress" here in the PR.

I've also had issues recently with TravisCI status being communicated, but I thought it was due to TravisCI quota being exhausted. Could be a deeper issue.

gino-m:
I've restarted the build to see if the PR receives the notification this time.

gino-m:
@shobhitagarwal1612 Hmm.. all checks were successful again, but still progress is not updated. Could be related to this? https://github.community/t5/GitHub-API-Development-and/Github-checks-broken-merge-check-count-vs-check-tab-count/m-p/55929/highlight/false#M4729

shobhitagarwal1612:
No idea.

gino-m:
Forcing a merge as admin for now. If the problem persists on other PRs we can contact support.

gino-m:
@shobhitagarwal1612 Looks like this was resolved by GitHub:
https://github.community/t5/forums/replypage/board-id/api/message-id/4822/load-autosave/true

PR #466, opened 2020-05-07
[WIP] Add instrumented tests

dturner:
WIP. The idea is to add instrumented tests for: 

**End to end tests**
These would test a complete user journey. For example, logging into the app, loading a project and adding an observation. 

**Integration/UI tests**
These test how components interact with one another. For example, the "Add observation" button is rendered and clicking it navigates to the "Add Observation" fragment.

@gino-m @shobhitagarwal1612 It'd be good to get some more examples of the above tests to help me figure out the scope boundaries between e2e, integration and unit tests in this project. 

For now, I'm working on a single integration/UI test:

After successful login the project selection view is shown.


gino-m:
Hi Don! Why do we need an interface for AuthenticationManager? Is it exclusively to allow us to swap out the implementation for testing? If so, can we use something like `Mockito` to achieve the same?

gnd/src/main/java/com/google/android/gnd/system/AuthenticationModule.java
```diff
@@ -0,0 +1,14 @@
+package com.google.android.gnd.system;
```
gino-m, line 1: Since there are several of these classes now, may be worth moving to their own package? (e.g., .gnd.system.auth?)

dturner:
> Is it exclusively to allow us to swap out the implementation for testing?

Yes, that was the idea. 

> can we use something like `Mockito` to achieve the same?

Maybe, although the challenge will be getting Dagger to provide the MockAuthenticationManager during testing. I will investigate.

gnd/src/main/java/com/google/android/gnd/system/AuthenticationModule.java
dturner: Good plan. I'll do that.

gino-m:
> > Is it exclusively to allow us to swap out the implementation for testing?
> 
> Yes, that was the idea.
> 
> > can we use something like `Mockito` to achieve the same?
> 
> Maybe, although the challenge will be getting Dagger to provide the MockAuthenticationManager during testing. I will investigate.

We had a similar issue with Dagger in a previous (pure Java) project I worked on. It would be interesting to see if there's another solution by now to avoid a proliferation of interface + impl pairs only used for testing. Thanks for investigating!

gino-m:
I believe if you set up Mockito you can just declare a private member in the test as `@Mock`. As long as `AuthenticationManager` is not final, it should work (even as a class).

shobhitagarwal1612:
Adding my 2 cents to the discussion. 
For UI unit tests using Robolectric, we can create ["shadow"](http://robolectric.org/extending/) objects or use Mockito (or PowerMockito for static methods) in order to override specific methods in the tests.

Can't say if they'll be useful in integration tests.

dturner:
I don't think Mockito will work for e2e / integration tests because what we need is a _fake_ rather than a _mock_. The fake auth manager needs to offer some basic functionality such as sign in and sign out. 

I'll update this PR with all my testing code so you can see the `FakeAuthenticationManager`

dturner:
I've added a very rough version of how an integration test might work. Basic architecture: 

- Tests are run using `CustomTestRunner` (which is specified in `build.gradle`). 
- This instantiates `TestGndApplication` which is a version of the app specifically tailored for testing
- The most important aspect of `TestGndApplication` is that it has a different dependency graph to the normal app, allowing dependencies to be swapped out for test instances of those classes. 
- The entry point to the dependency graph is `TestGndApplicationComponent` which defines the Dagger modules to be used when creating dependencies.
- I have created a new module for the `MainActivity` dependencies called `TestMainActivityInjectorModule`. It has a corresponding class in the main app called `MainActivityInjectorModule`.  
- In the `Test` version of this module we use the `FakeAuthenticationManager` to provide login features which allows `MainActivity` to avoid using Google Sign In and go straight to the map view / project list. 

The test itself can be found in `MyTests::MainActivity_testLogin`. It currently fails 😃 but that's because there are still dependencies on firebase which need faking (or setting up correctly for live instance testing). 

Hope that makes sense. Any and all feedback welcome.

gino-m:
> I don't think Mockito will work for e2e / integration tests because what we need is a _fake_ rather than a _mock_. The fake auth manager needs to offer some basic functionality such as sign in and sign out.
> 
> I'll update this PR with all my testing code so you can see the `FakeAuthenticationManager`

As mentioned by VC, could Spies be helpful here? In that case you can mock out some methods while using the actual implementation of others.
https://github.com/mockito/mockito/wiki/Using-Spies-(and-Fakes)

I realize spies aren't a silver bullet; there will be cases where creating an interface + impl + testImpl will be cleaner, but for simple cases perhaps it's worth considering?

gnd/src/main/java/com/google/android/gnd/system/auth/AuthenticationManagerImpl.java
```diff
@@ -45,10 +44,9 @@
 import java8.util.Optional;
 import javax.inject.Inject;
 
-@ActivityScoped
-public class AuthenticationManager {
-  private static final String TAG = AuthenticationManager.class.getSimpleName();
-  private static final int SIGN_IN_REQUEST_CODE = AuthenticationManager.class.hashCode() & 0xffff;
+public class AuthenticationManagerImpl implements AuthenticationManager {
```
gino-m, line 47: As mentioned by VC, could Spies be helpful here? In that case you can mock out some methods while using the actual implementation of others.
https://github.com/mockito/mockito/wiki/Using-Spies-(and-Fakes)

I realize spies aren't a silver bullet; there will be cases where creating an interface + impl + testImpl will be cleaner, but for simple cases perhaps it's worth considering?

gnd/src/androidTest/java/com/google/android/gnd/CustomTestRunner.java
```diff
@@ -0,0 +1,14 @@
+package com.google.android.gnd;
+
+import android.app.Application;
+import android.content.Context;
+import androidx.test.runner.AndroidJUnitRunner;
+
+public class CustomTestRunner extends AndroidJUnitRunner {
```
gino-m, line 7: Can we mark the custom module, injector, and test runner somehow (name/package/path) to make it clear that they exist to allow integration testing? It seems it would be easy to conflate integration and unit testing; for unit testing, for example, we probably want to instantiate and supply the dependencies ourselves to constructors, so the custom test graph wouldn't apply.

gnd/src/androidTest/java/com/google/android/gnd/MyTest.java
```diff
@@ -0,0 +1,47 @@
...
+      .factory().create(app);
+    graph.inject(app);
+
+    ActivityScenario.launch(MainActivity.class);
+    onView(withId(R.id.sign_in_button)).perform(click());
+
+    // We probably need to mock/fake more than the user login, almost certainly avoid any connection with firebase
+
+    // The project selection view should be displayed
+    onView(withId(R.id.project_list_view)).check(matches(isDisplayed()));
```
gino-m, line 45: espresso is really readable, nice fluent API.

gnd/src/androidTest/java/com/google/android/gnd/TestGndApplication.java
```diff
@@ -0,0 +1,16 @@
+package com.google.android.gnd;
+
+import com.google.android.gnd.inject.DaggerTestGndApplicationComponent;
+import dagger.android.AndroidInjector;
+import dagger.android.support.DaggerApplication;
+
+public class TestGndApplication extends GndApplication {
```
gino-m, line 7: Just a reminder/placeholder to add JavaDoc to public classes here and throughout once the design settles.

gnd/src/main/java/com/google/android/gnd/GndApplication.java
```diff
@@ -39,6 +39,9 @@
 // without extending MultiDexApplication.
 public class GndApplication extends DaggerApplication {
 
+  /**
+   * ctor is annotated with @Inject, that's how Dagger knows!
```
gino-m, line 43: Please update to clarify or remove.

gino-m:
@dturner will you come back to this one? Excited to see a well-defined pattern defined for testing!

dturner:
Of course, sorry for the delay, "real work" getting in the way. Hopefully have some time this week.

gino-m:
> Of course, sorry for the delay, "real work" getting in the way. Hopefully have some time this week.

No worries, looking forward!

gnd/src/androidTest/java/com/google/android/gnd/TestGndApplication.java
dturner: Thanks for the reminder. Added comment.

gnd/src/androidTest/java/com/google/android/gnd/CustomTestRunner.java
dturner: They're all in the `androidTest` folder which is designed for instrumented tests (i.e. integration and e2e tests) - is that sufficient?

gnd/src/main/java/com/google/android/gnd/system/auth/AuthenticationManagerImpl.java
dturner: I've investigated using spies and have come to the conclusion that coding to an interface (aka the API approach) is still the best way to proceed. Here's why: 

- We're already using it. For example, we have a `RemoteStorageManager` interface which is implemented by `FirestoreStorageManager`, and a `RemoteDataStore` which is implemented by `FirestoreDataStore`. 
- When there is more than one implementation of a given interface (e.g. one for prod, one for test) we should reflect that in the code, rather than relying on a spy to do that work for us.  
- Speaking to various people with more testing experience than me, the general feeling around spies is that they are a code smell. There's also a shift away from using mocks. 
- It makes it easy to add new authentication methods, such as Twitter, Facebook, email address. Potentially we might add some anonymous login which can then be converted to an account if the user wishes later. 

Incidentally, the `Impl` suffix adds noise ([and is an anti-pattern](https://octoperf.com/blog/2016/10/27/impl-classes-are-evil/)) we can avoid this by naming the implementations according to their specialization e.g. `GoogleAuthenticationManager` (I'll commit a change for this shortly). 

WDYT?

gino-m: This makes sense. Int his case, I assume won't need interface/impls for all classes, only on those classes that touch remote or system services so that we can reliably and quickly run UI tests? If so, I think this amount of additional complexity is a reasonable price to pay for fast, reproducible UI tests. Please go ahead!

gino-m:
Hi @dturner ! Is this still WIP? Would it make sense at this point to break this PR into smaller chunks? If so, can we add integration and unit tests separately?

PR #468, opened 2020-05-11
[Refactor] Form UI Generator

shobhitagarwal1612:
Split from #457 

I have taken a couple of new classes from #457 to show the proposed architecture changes.

@gino-m PTAL

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/common/Validator.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/common/Validator.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.ui.common;
+
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import java8.util.Optional;
+
+public final class Validator {
+
+  public static boolean isValid(Field field, String textResponse) {
+    if (!field.isRequired()) {
+      return true;
+    }
+    return textResponse != null && !textResponse.isEmpty();
+  }
+
+  public static boolean isValid(Field field, Optional<Response> response) {
+    if (!field.isRequired()) {
+      return true;
+    }
+    return !response.isEmpty();
+  }
+}
--- a/gnd/src/main/java/com/google/android/gnd/ui/field/FieldFactory.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/field/FieldFactory.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


gnd/src/main/java/com/google/android/gnd/ui/common/Validator.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.ui.common;
+
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import java8.util.Optional;
+
+public final class Validator {
```
gino-m, line 23: ->FieldValidator?

gino-m, line 23: Please add essential JavaDoc (on classes and public members) here and throughout.

gnd/src/main/java/com/google/android/gnd/ui/field/FieldFactory.java
```diff
@@ -0,0 +1,49 @@
...
+package com.google.android.gnd.ui.field;
+
+import androidx.annotation.NonNull;
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import com.google.android.gnd.ui.common.AbstractFragment;
+import com.google.android.gnd.ui.editobservation.EditObservationFragmentArgs;
+import java8.util.Optional;
+
+public class FieldFactory {
```
gino-m, line 26: ->FieldViewFactory?

gnd/src/main/java/com/google/android/gnd/ui/field/FieldFactory.java
```diff
@@ -0,0 +1,49 @@
...
+  private final AbstractFragment fragment;
+  private final EditObservationFragmentArgs args;
+
+  public FieldFactory(AbstractFragment fragment, EditObservationFragmentArgs args) {
+    this.fragment = fragment;
+    this.args = args;
+  }
+
+  @NonNull
+  public FieldView createFieldView(Field field, Optional<Response> response) {
```
gino-m, line 37: if this is a FieldViewFactory, this method can be called `create()` to avoid redundancy.

gnd/src/main/java/com/google/android/gnd/ui/field/FieldModel.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.android.gnd.BR;
+import com.google.android.gnd.R;
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.ui.common.Validator;
+
+/**
+ * {@link FieldModel} is not shared among {@link FieldView}. It stores the current state of the View
+ * it's attached to and updates itself using two-way data binding.
+ */
+public class FieldModel extends BaseObservable {
```
gino-m, line 31: Isn't that the point of ViewModels? i.e. I'd assume we'd have a FieldViewModel with the required state; not sure what value the extra indirection adds here. Also, not we've been tending towards using LiveData instead of Android ObservableFoos since they're lifecycle aware. Otherwise we'll need to think carefully about lifecycles each time we use them.

gnd/src/main/java/com/google/android/gnd/ui/field/FieldModel.java
```diff
@@ -0,0 +1,66 @@
...
+    this.field = field;
+    this.resources = resources;
+  }
+
+  @Bindable
+  public String getResponse() {
+    return response == null ? "" : response;
+  }
+
+  public void setResponse(String response) {
```
gino-m, line 49: Since this does more than sets the response I'd suggest using another name (updateResponse?)

gnd/src/main/java/com/google/android/gnd/ui/field/FieldView.java
```diff
@@ -0,0 +1,101 @@
...
+import android.widget.FrameLayout;
+import androidx.lifecycle.LifecycleOwner;
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import com.google.android.gnd.ui.common.AbstractFragment;
+import com.google.android.gnd.ui.editobservation.EditObservationFragmentArgs;
+import com.google.android.gnd.ui.util.ViewUtil;
+import java8.util.Optional;
+
+public abstract class FieldView extends FrameLayout {
```
gino-m, line 29: Do we actually need a custom class for the View at all? View classes are harder to tests since they're tied to the Android UI APIs and lifecycle. Instead, could the factory be used to inflate the relevant layout, and to create the appropriate ViewModel?

PR #469, opened 2020-05-14
[Refactor] Refactor edit observation package

shobhitagarwal1612:
@gino-m PTAL.

I've tried to keep the commits clean and concise for review purpose.


gnd/src/main/java/com/google/android/gnd/ui/common/ResponseValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.android.gnd.R;
+import com.google.android.gnd.model.form.Field;
+import com.google.android.gnd.model.observation.Response;
+import java8.util.Optional;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class ResponseValidator {
```
gino-m, line 30: View business logic generally belongs in ViewModels. Any reason not to include this in the AbstractFieldViewModel class?

gnd/src/main/java/com/google/android/gnd/ui/common/ResponseValidator.java
```diff
@@ -0,0 +1,47 @@
...
+  private final Resources resources;
+
+  @Inject
+  ResponseValidator(GndApplication application) {
+    this.resources = application.getResources();
+  }
+
+  // TODO: Check valid response values
+  @Nullable
+  public String validate(Field field, Optional<Response> response) {
```
gino-m, line 41: We've been using `Optional` instead of `null` for internal APIs. Can we do the same here for consistency?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/AbstractFieldViewModel.java
```diff
@@ -0,0 +1,77 @@
...
+  private Field field;
+
+  AbstractFieldViewModel() {
+    responseText =
+        LiveDataReactiveStreams.fromPublisher(
+            responseSubject.distinctUntilChanged().switchMapSingle(this::getDetailsText));
+    response = LiveDataReactiveStreams.fromPublisher(responseSubject.distinctUntilChanged());
+  }
+
+  void init(Field field, Optional<Response> response) {
```
gino-m, line 45: Can you add a TODO to add a reference to Field to Response? We have these associations in other parts of our internal model, so would make sense there as well.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -65,8 +70,31 @@
   private SingleSelectDialogFactory singleSelectDialogFactory;
   private MultiSelectDialogFactory multiSelectDialogFactory;
 
-  @Nullable private EditObservationBottomSheetBinding addPhotoBottomSheetBinding;
-  @Nullable private BottomSheetDialog bottomSheetDialog;
+  private static Class<? extends AbstractFieldViewModel> getViewModelClass(Field.Type fieldType) {
```
gino-m, line 73: Rather than have `getViewModelClass` and `getFieldLayoutId`, can we instead have a `FieldViewFactory` class that will inflate the appropriate view and viewmodel? This fragment can coordinate the returned views by getting their viewmodels with `DataBindingUtil`. Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/AbstractFieldViewModel.java
```diff
@@ -0,0 +1,77 @@
...
+public class AbstractFieldViewModel extends AbstractViewModel {
+
+  private final LiveData<Optional<Response>> response;
+  private final LiveData<String> responseText;
+  private final MutableLiveData<String> error = new MutableLiveData<>();
+  private final BehaviorProcessor<Optional<Response>> responseSubject = BehaviorProcessor.create();
+
+  private Field field;
+
+  AbstractFieldViewModel() {
```
gino-m, line 38: Please add Javadoc to classes,, public methods, and especially to streams like the ones below to clarify what they're used for and why.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -119,116 +147,107 @@ private void handleSaveResult(EditObservationViewModel.SaveResult saveResult) {
...
-  private void addTextField(Field field) {
-    TextInputFieldBinding binding =
-        TextInputFieldBinding.inflate(getLayoutInflater(), formLayout, false);
-    binding.setViewModel(viewModel);
-    binding.setLifecycleOwner(this);
-    binding.setField(field);
-    formLayout.addView(binding.getRoot());
-    assignGeneratedId(binding.getRoot().findViewById(R.id.text_input_edit_text));
-  }
+    // subscribe to response updates
```
gino-m, line 163: Style nit: Please use capital case with final period in comments. Example: "Please use capital case with final period in comments." :)

gino-m: Also please make sure the comment describes something more than what's clear from the code itself. In this case, for example, a comment isn't needed.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/AbstractFieldViewModel.java
gino-m: [or if it's an easy change feel free to add it in this PR so that the field UI method signatures can be simplified.

gnd/src/main/java/com/google/android/gnd/ui/common/ResponseValidator.java
shobhitagarwal1612: Thanks! Done 2c82209

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
shobhitagarwal1612: Great idea! d4a40af

gnd/src/main/java/com/google/android/gnd/ui/editobservation/AbstractFieldViewModel.java
shobhitagarwal1612: I checked and it is a bit complicated for `MultipleChoiceResponse`. Added a TODO for now

gnd/src/main/java/com/google/android/gnd/ui/common/ResponseValidator.java
shobhitagarwal1612: There were a couple of reasons to separate it out. 

 1. Right now, we are only considering if response is present or not. In future, we'll have to consider a lot of other validations as well (minLen, instanceType, regex, etc.)

 2. In the beginning, I moved all of this logic within AFVM, but it was also present in EOVM. So, to prevent duplication, I kept the original code and left it as is in favor of future cleanups. Ideally, the error should be generated and set by AFVM and EOVM should subscribe to the changes.

gino-m: > There were a couple of reasons to separate it out.
> 
> 1. Right now, we are only considering if a response is present or not. In future, we'll have to consider a lot of other validations as well (minLen, instanceType, regex, etc.). Having a small class would encourage extensive testing with all possible scenarios.

In that case, won't we end up with parallel Validator and ViewModel hierarchies for each field type?

> 2. In the beginning, I moved all of this logic within AFVM, but it was also present in EOVM. So, to prevent duplication, I kept the original code and left it as is in favor of future cleanups. Ideally, the error should be generated and set by AFVM and EOVM should subscribe to the changes.

I see.. so it sounds like it belongs in the individual AFVM?

shobhitagarwal1612: > In that case, won't we end up with parallel Validator and ViewModel hierarchies for each field type?

I see. So each ViewModel should contain the logic instead of aggregating them in a external class and later mapping them. Sgtm.

> I see.. so it sounds like it belongs in the individual AFVM?

Yes.

shobhitagarwal1612: Done b42aa16

shobhitagarwal1612:
@gino-m Resolved all comments.

PR #472, opened 2020-05-20
Disable saving artifacts to cloud storage

shobhitagarwal1612:
A dummy `google-services.json` is used for building apks on GCB. This is fine for unit tests as we don't install the apk. But it can't be used on a real device.

So, let's disable this until the issue is resolved


gino-m:
Thanks Shobhit. This should save us some GCB storage space.

PR #473, opened 2020-05-21
[Settings] Add basic settings page

### LINKED ISSUE #424: [Settings] Add Settings page
Accessible from left hand side nav menu.

shobhitagarwal1612:
Fixes #424.

- [x] Add a new item "settings" to navigation drawer
- [x] Include some dummy preferences to setup the structure for future use.

### Screenshot:
Screenshot_1590054937

@scolsen  PTAL?


gnd/src/main/java/com/google/android/gnd/ui/settings/SettingsFragment.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
+    super.onViewCreated(view, savedInstanceState);
+
+    // TODO: Set dynamically by calculating the actual height of toolbar
+    int topMargin = 108;
+
+    LayoutParams params = (LayoutParams) getListView().getLayoutParams();
+    params.setMargins(0, topMargin, 0, 0);
+    getListView().setLayoutParams(params);
```
shobhitagarwal1612, line 70: I tried to get the height of the view from binding. But it always returns 0 for the first time. Anyone knows why?

The reason we have to add top margin is to display the toolbar. There is no standard practice around how to add a toolbar because official docs recommend using the `PreferenceFragment` from within an Activity. In our case, the activity doesn't have a toolbar. Any guidelines or suggestions are most welcome!

shobhitagarwal1612: Removed

shobhitagarwal1612:
@gino-m I've added a new activity to launch the `SettingsFragment` and did some cleanup as well. PTAL

gnd/src/main/java/com/google/android/gnd/ui/settings/Keys.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.ui.settings;
+
+/** These keys should match 1:1 in res/xml/preferences.xml. */
```
gino-m, line 19: Can these be represented as String resources? Seems unfortunate we need to duplicate these keys in two places.

gnd/src/main/res/xml/preferences.xml
```diff
@@ -0,0 +1,53 @@
...
+      app:title="Visit website" />
+
+    <Preference
+      app:key="feedback"
+      app:summary="Report technical issues or suggest new features"
+      app:title="Send feedback" />
+
+  </PreferenceCategory>
+
+</PreferenceScreen>
```
gino-m, line 53: Does the autoformat remove final CRs? Maybe they should be there?

gnd/src/main/java/com/google/android/gnd/ui/settings/Keys.java
shobhitagarwal1612: Sadly, no. `preferences.xml` only accepts string constants as inputs, not string resources.

gnd/src/main/res/xml/preferences.xml
shobhitagarwal1612: I always forget about this. Thanks!
Maybe there is a checkstyle or pmd rule for it.

Issue #476, opened 2020-05-24
[Code health] Simplify applying window inset to components


shobhitagarwal1612:
The main activity subscribes to window inset changes and all fragments listen to that and apply padding/margin/translationY to adjust themselves accordingly.

A lot of code, dedicated for this purpose can be replaced with tuning the xmls directly by separating the components that need adjustment from rest and setting `fitsSystemWindows` to `true`

gino-m:
Hi @shobhitagarwal1612 can we close this now?

shobhitagarwal1612:
One or two instances are still remaining. See the linked PR for details

shobhitagarwal1612:
`HomeScreenFragment` and `FeatureDetailsFragment` could also use similar refactoring.

shobhitagarwal1612:
Closing this as obsolete.

PR #477, opened 2020-05-24
[Code refactor] Simplify handling of window insets

shobhitagarwal1612:
Towards #476 

 - Flatten hierarchy using constraint layout
 - Replace java code for handling top and bottom insets with `fitsSystemWindows` in xml layouts
 - Move status bar scrim to MainActivity for applying it to all screens at once

Note:
Handle `HomeScreenFragment` and `FeatureDetailsFragment` separately as they are more complex and need extra care when simplifying.


gnd/src/main/res/layout/map_container_frag.xml
```diff
@@ -32,90 +32,96 @@
...
 
-    <RelativeLayout
-      android:id="@+id/map_layout"
+    <FrameLayout
       android:layout_width="match_parent"
       android:layout_height="match_parent"
-      android:layout_gravity="center"
-      android:gravity="center"
-      app:layout_behavior="com.google.android.gnd.ui.home.mapcontainer.MapLayoutBehavior">
+      android:fitsSystemWindows="true">
```
gino-m: I could never get this to work; nicely done!

gnd/src/main/res/layout/offline_areas_frag.xml
```diff
@@ -20,56 +20,51 @@
...
       name="viewModel"
       type="com.google.android.gnd.ui.offlinearea.OfflineAreasViewModel" />
   </data>
 
-  <LinearLayout
+  <androidx.constraintlayout.widget.ConstraintLayout
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical">
+    android:fitsSystemWindows="true">
```
gino-m, line 31: Not a major issue, esp. if it saves us code complexity, but the system inset areas shows as grey instead of blue in Offline areas now:
image

Before:
image

Ideal would be as in the expanded feature sheet:
image

If it's not easy to fix, this is really a P3 UI tweak that can be punted.

shobhitagarwal1612: It was like that because of hardcoded padding. After fixing that, it is showing the same status bar scrim as the home screen. 

To fix it system-wide, I am thinking about using colors (with alpha value) in the status_bar_scrim of MainActivity. It would be the dark shade of the toolbar. (In a separate PR)

gnd/src/main/res/layout/map_container_frag.xml
shobhitagarwal1612: I have read in Romain Guy's blog that flat hierarchy views are better because they consume less resources and are rendered quickly. So, that was one of my motivation too.

shobhitagarwal1612:
@gino-m Back to you 😃

gnd/src/main/res/layout/offline_areas_frag.xml
gino-m: Great, I think that's sufficient. Thanks for the fix!

shobhitagarwal1612:
> LGTM! Please resolve merge conflicts and I'll merge.

Done

Issue #478, opened 2020-05-24
[Code health] Consolidate and clean up styles


gino-m:
All components should ideally use Material theming; specific instances of color attributes can be removed.

gino-m:
@shobhitagarwal1612 @JSunde Any idea when we should be using `?attr/colorFoo` vs `@color/colorBar`? I don't quite understand yet how the two interact, would be good to establish a common approach.

JSunde:
This article seems to recommend ?attr over @color, although it doesn't seem to go into much detail as to why that is recommended.
https://medium.com/androiddevelopers/android-styling-common-theme-attributes-8f7c50c9eaba

JSunde:
```
Specifically we recommended using theme attributes to provide a point of indirection to resources, so that you can vary them (e.g. in [dark theme](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme))
```

JSunde:
```
This enables us to create fewer layouts or styles, isolating changes within a theme. In practice, you largely want to vary colors by theme and as such you should always* refer to colors via theme attributes.
```

https://medium.com/androiddevelopers/android-styling-prefer-theme-attributes-412caa748774

gino-m:
@shobhitagarwal1612 @JSunde Proposed structure of new consolidated styles, by example:

https://github.com/google/ground-android/blob/09bb8e534e021f8427cec5710cb2d52a917e3c95/ground/src/main/res/values/styles.xml#L218-L240

anandwana001:
@gino-m Is it good to close? 
All the places we use `md_theme_*` colors only.

gino-m:
> [@gino-m[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/gmiceli@google.com)](https://github.com/gino-m) Is it good to close? All the places we use `md_theme_*` colors only.

Themes > colors. We should aim to use named themes for consistency and to simplify styling.

PR #479, opened 2020-05-24
[App theme] Upgrade to Material components 1.1.0

### LINKED ISSUE #458: [Code health] Upgrade to Material 1.1.0
Latest Material requires app to use a Material theme, which will require updating many styles defined in styles.xml.

### LINKED ISSUE #459: [Code health] Eliminate forked BottomSheetBehavior class and simplify impl
We made a copy of the BottomSheetBehavior class to allow us to specify set the `expandedOffset` such that the header gets obscured behind the toolbar. Material 1.1.0 allows us to set the offset, but only to a non-negative number. To avoid needing to copy and modify the implementation again, looking for a simpler solution.

I've tried a dozen or so variants, and the _least_ hacky and most succinct is shared here:
https://github.com/gino-m/slidey-demo/blob/master/app/src/main/java/com/google/android/slidey/MainActivity.java
https://github.com/gino-m/slidey-demo/blob/master/app/src/main/java/com/google/android/slidey/FloatingHeaderBehavior.java

slidey

Blocking #458.

@dturner fyi. @cketcham would love to hear what you think of this sol'n when/if you have a sec!

gino-m:
Also upgrades related deps, removes custom BottomSheetBehavior, and fixes several styles. Some styles still need more work (#478). Closes #459, closes #458.


gino-m:
@shobhitagarwal1612 could you PTAL? @dturner FYI.

PR #489, opened 2020-05-27
Fixed Tile Providers

scolsen:
- [x] Revert problematic MapBox changes
- [ ] Ensure all available tiles are rendered on the map.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -161,7 +161,7 @@ public class MapContainerFragment extends AbstractFragment {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCameraTarget()));
     enableLocationLockBtn();
-    map.renderTileOverlay();
+    mapContainerViewModel.getMbtiles().observe(this, map::renderTileOverlays);
   }
 
   @Override
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -107,7 +107,7 @@ public class MapContainerFragment extends AbstractFragment {
         .toObservable()
         .flatMap(MapAdapter::getTileProviders)
         .as(disposeOnDestroy(this))
-        .subscribe(mapContainerViewModel::onTileProvider);
+        .subscribe(mapContainerViewModel::queueTileProvider);
   }
 
   @Override
@@ -154,7 +154,7 @@ public class MapContainerFragment extends AbstractFragment {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCameraTarget()));
     enableLocationLockBtn();
-    mapContainerViewModel.getMbtiles().observe(this, map::addTileOverlays);
+    mapContainerViewModel.getMbtilesFilePaths().observe(this, map::addTileOverlays);
   }
 
   private void onBottomSheetStateChange(BottomSheetState state, MapAdapter map) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -115,6 +115,11 @@ public class MapContainerFragment extends AbstractFragment {
         .flatMap(MapAdapter::getCameraMoves)
         .as(disposeOnDestroy(this))
         .subscribe(mapContainerViewModel::onCameraMove);
+    mapAdapter
+        .toObservable()
+        .flatMap(MapAdapter::getTileProviders)
+        .as(disposeOnDestroy(this))
+        .subscribe(mapContainerViewModel::onTileProvider);
   }
 
   @Override
@@ -255,5 +260,11 @@ public class MapContainerFragment extends AbstractFragment {
   @Override
   public void onSaveInstanceState(@NonNull Bundle outState) {
     saveChildFragment(outState, mapProvider.getFragment(), MAP_FRAGMENT_KEY);
+  }
+
+  @Override
+  public void onDestroy() {
+    mapContainerViewModel.closeProviders();
+    super.onDestroy();
   }
 }
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -154,7 +154,7 @@ public class MapContainerFragment extends AbstractFragment {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCameraTarget()));
     enableLocationLockBtn();
-    mapContainerViewModel.getMbtiles().observe(this, map::renderTileOverlays);
+    mapContainerViewModel.getMbtiles().observe(this, map::addTileOverlays);
   }
 
   private void onBottomSheetStateChange(BottomSheetState state, MapAdapter map) {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
@@ -64,7 +64,11 @@ public class MapContainerViewModel extends AbstractViewModel {
   private final Subject<Boolean> locationLockChangeRequests;
   private final Subject<CameraUpdate> cameraUpdateSubject;
   private final MutableLiveData<Event<Nil>> showMapTypeSelectorRequests = new MutableLiveData<>();
-  private final LiveData<ImmutableSet<String>> mbtiles;
+  private final LiveData<ImmutableSet<String>> mbtilesFilePaths;
+  // TODO: Create our own wrapper/interface for MbTiles providers
+  // The impl we're using unfortunately requires calling a `close` method explicitly
+  // to clean up provider resources; `close` however, is not defined by the `TileProvider`
+  // interface, preventing us from treating providers generically.
   private final List<MapBoxOfflineTileProvider> tileProviders = new ArrayList<>();
 
   @Inject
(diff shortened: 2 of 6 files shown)


scolsen:
@gino-m I'm having some trouble with this. Maybe you can steer me in the right direction?

I'm getting tons of SQL corruption errors whenever we attempt to render tiles. Something's going wrong, but I'm not sure what yet. It could be any of these issues:

- Tiles are marked as `Downloaded` too early resulting in incomplete data
- `MapBoxOfflineTileProviders` need to be closed in order not to leak resources--I'm currently not doing this since I'm not sure where exactly we'd do it? On map/fragment destruction?
- There's some underlying issue with the source data? (least likely).

Anyway, I'm bashing my head against this one, so let me know if you have any ideas!

gino-m:
Hey Scott! 

Did you basically revert back the MapBoxTileProvider to its original state? I'd recommend starting there.

Re destroying TileProviders, not sure you need to do that, I believe they get removed when the Map is destroyed.

Lastly, we should keep in mind that `Tile` is a misnomer in our model; really their MapBox tile sources; each Tile represents an mbtiles file, which contains many tiles (cmiww).

Please lmk if you're using the default MapBoxTileProvider and we can go from there!

Gino

gino-m:
Also, are you working with an an area we know has completed imagery? Some files may have been corrupt in the current test data source.

scolsen:
> Hey Scott!
> 
> Did you basically revert back the MapBoxTileProvider to its original state [exactly]?

Yep. Should be equivalent to the original source right now except for the try/catches I added for the corruption errors (the original just crashes outright on a DB corruption).

> Re destroying TileProviders, not sure you need to do that, I believe they get removed when the Map is destroyed.

That's correct--what I meant was, according to the `MapBoxOfflineTileProvider` we need to explicitly close the underlying SQLite DB:

```java
/**
     * Closes the provider, cleaning up any background resources.
     * 
     * <p>
     * You must call {@link #close()} when you are finished using an instance of
     * this provider. Failing to do so may leak resources, such as the backing
     * SQLiteDatabase.
     * </p>
     */
    @Override
    public void close() {
        if (this.mDatabase != null) {
            this.mDatabase.close();
            this.mDatabase = null;
        }
    }
```

> Lastly, we should keep in mind that `Tile` is a misnomer in our model; really their MapBox tile sources; each Tile represents an mbtiles file, which contains many tiles (cmiww).

You're right! I've been trying to use names like `mbtiles`--maybe we should rename the `Tile` entity to something like `TileSet` or some such 

> Also, are you working with an an area we know has completed imagery? Some files may have been corrupt in the current test data source.

I'm not sure about the integrity of the data source, unfortunately—but I can say with confidence this is the first time that I've encountered these corruption errors with this source, and we'd had this working in the past.

scolsen:
@gino-m It's ready for a look now. I still run into some DB corruptions, and some mbtiles files that are downloaded are only 16kb in size, which is suspicious, but afaict everything should be correct

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -103,6 +103,11 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         .flatMap(MapAdapter::getCameraMoves)
         .as(disposeOnDestroy(this))
         .subscribe(mapContainerViewModel::onCameraMove);
+    mapAdapter
+        .toObservable()
+        .flatMap(MapAdapter::getTileProviders)
+        .as(disposeOnDestroy(this))
+        .subscribe(mapContainerViewModel::onTileProvider);
```
gino-m, line 110: Naming nit: onTileProvider<event/action>

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -149,7 +154,7 @@ private void onMapReady(MapAdapter map) {
     addFeatureBtn.setOnClickListener(
         __ -> homeScreenViewModel.onAddFeatureBtnClick(map.getCameraTarget()));
     enableLocationLockBtn();
-    map.renderTileOverlay();
+    mapContainerViewModel.getMbtiles().observe(this, map::renderTileOverlays);
```
gino-m, line 157: Naming: render -> addTileOverlays

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -90,6 +98,11 @@
                 .map(Loadable::value)
                 .switchMap(this::getFeaturesStream)
                 .map(MapContainerViewModel::toMapPins));
+    this.mbtiles =
```
gino-m, line 101: mbtilesFilePaths maybe? (since this doesn't actually contain the data)

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -247,10 +257,21 @@ public LatLngBounds getViewport() {
   }
 
   @Override
-  public void renderTileOverlay() {
-    try (MapBoxOfflineTileProvider tileProvider =
-        new MapBoxOfflineTileProvider(context.getFilesDir())) {
-      map.addTileOverlay(new TileOverlayOptions().tileProvider(tileProvider));
-    }
+  public void renderTileOverlays(ImmutableSet<String> mbtilesFiles) {
```
gino-m, line 260: Same here: `add` instead of `render`

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -247,10 +257,21 @@ public LatLngBounds getViewport() {
...
+    stream(mbtilesFiles)
+        .forEach(
+            path -> {
+              File mbtiles = new File(context.getFilesDir(), path);
```
gino-m, line 264: Can this inner lambda be extract to its own method for readability?

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -247,10 +257,21 @@ public LatLngBounds getViewport() {
...
+              if (mbtiles.exists()) {
```
gino-m, line 265: Could you short circuit exit here instead and Timber.info the fact the the file was not found?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -59,12 +64,15 @@
   private final Subject<Boolean> locationLockChangeRequests;
   private final Subject<CameraUpdate> cameraUpdateSubject;
   private final MutableLiveData<Event<Nil>> showMapTypeSelectorRequests = new MutableLiveData<>();
+  private final LiveData<ImmutableSet<String>> mbtiles;
+  private final List<MapBoxOfflineTileProvider> tileProviders = new ArrayList<>();
```
gino-m, line 68: Could you kindly add a TODO and Code health task to create our own wrapper interface/impl for these?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
scolsen: 57cb7dd9d31e68d8c97168a3778b7bcb62c173f7

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
scolsen: 57cb7dd9d31e68d8c97168a3778b7bcb62c173f7

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
scolsen: 57cb7dd9d31e68d8c97168a3778b7bcb62c173f7

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
scolsen: 57cb7dd9d31e68d8c97168a3778b7bcb62c173f7

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
scolsen: I renamed it to `queueTileProvider` since `onTileProviderQueue` is ambiguous (not clear whether queue is a continuation of the noun or if it's a verb).

scolsen: f9a41a2650397ca3f118ba6677bb57e2affb43af

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
scolsen: f9a41a2650397ca3f118ba6677bb57e2affb43af

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
scolsen: f9a41a2650397ca3f118ba6677bb57e2affb43af

scolsen:
@gino-m ready for another look whenever

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -107,7 +107,7 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         .toObservable()
         .flatMap(MapAdapter::getTileProviders)
         .as(disposeOnDestroy(this))
-        .subscribe(mapContainerViewModel::onTileProvider);
+        .subscribe(mapContainerViewModel::queueTileProvider);
```
gino-m, line 110: Not sure `queue` is the right name here, since it implies an actual queue is being used, and not a stream. In this case "add" would be correct, since it's really a deferred add operation, but this can be changed later if we like.

PR #499, opened 2020-06-11
Fix GeoJson coordinate parsing

scolsen:
The GeoJson spec states that positions must be described by a longitude followed by a latitude; we previously assumed the reverse order, leading to errors.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
@@ -16,8 +16,14 @@
 
 package com.google.android.gnd.persistence.geojson;
 
+import static com.google.android.gnd.util.ImmutableListCollector.toImmutableList;
+import static java8.util.stream.StreamSupport.stream;
+
 import com.google.android.gms.maps.model.LatLng;
 import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.List;
+import java8.util.Optional;
 import javax.annotation.Nullable;
 import org.json.JSONArray;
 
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -48,10 +48,10 @@
       return ImmutableList.of();
     }
 
-    double south = sw.optDouble(0, 0.0);
-    double west = sw.optDouble(1, 0.0);
-    double north = ne.optDouble(0, 0.0);
-    double east = ne.optDouble(1, 0.0);
+    double south = sw.optDouble(1, 0.0);
```
gino-m, line 51: This code will work, but the variable names may not always be correct because we can't count on the polygon having the sw coordinate first. Why not just tread the polygon vertices as an arbitrary list of coordinates?

scolsen: fantastic point! I've updated the implementation to better reflect this 7b660a7

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonExtent.java
```diff
@@ -31,32 +37,26 @@
...
-      return ImmutableList.of();
-    }
+    Optional<JSONArray> exteriorRing = geometry.getVertices().map(j -> j.optJSONArray(0));
 
-    JSONArray g = geometry.getVertices().map(j -> j.optJSONArray(0)).orElse(null);
+    return ringCoordinatesToLatLngs(exteriorRing.orElse(null));
+  }
 
-    if (g == null) {
+  private ImmutableList<LatLng> ringCoordinatesToLatLngs(JSONArray exteriorRing) {
```
gino-m, line 45: Nice!

PR #506, opened 2020-06-29
[Offline sync] Allow user to specify connection type for sync

### LINKED ISSUE #408: [Offline sync] Allow user to specify connection types required for syncing
Allow user to enable "Wait for Wifi" for: Photos, Uploading observations, Downloading new features (wording and UX TBD). @coreyleamon FYI.

shobhitagarwal1612:
Fixes #408

@dturner PTAL

Tested manually:

 - Turn the setting to sync `Over Wi-Fi only` on
 - Toggle Wifi off and keep Mobile Data on
 - Add a new observation and also attach a photo
 - Click "save"
 - Observe that the form data is saved but the photo isn't uploaded
 - Turn on wifi
 - Photo gets synced in the background automatically


gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
```diff
@@ -46,17 +47,19 @@ public String getLastActiveProjectId() {
...
   /** Removes the last active project id in the local value store. */
   public void clearLastActiveProjectId() {
-    preferences
-      .edit()
-      .remove(ACTIVE_PROJECT_ID_KEY)
-      .apply();
+    preferences.edit().remove(ACTIVE_PROJECT_ID_KEY).apply();
+  }
+
+  public boolean shouldUploadMediaOverWifiOnly() {
```
dturner, line 58: We might want to change this to `shouldUploadMediaOverUnmeteredConnectionOnly`. Whilst most Wi-Fi connections are unmetered some are not (hotspot tethering, airport Wi-Fi), also with 5G there are some mobile connections which _are_ unmetered. 

My point being: we're presumably trying to avoid the user being charged for data rather than simply letting them choose a specific network protocol.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
```diff
@@ -46,17 +47,19 @@ public String getLastActiveProjectId() {
...
+    return preferences.getBoolean(Keys.UPLOAD_MEDIA, false);
+  }
+
+  public boolean shouldDownloadOfflineAreasOverWifiOnly() {
```
dturner, line 62: As comment above

gnd/src/main/java/com/google/android/gnd/GndApplicationModule.java
```diff
@@ -45,8 +45,7 @@
 import javax.inject.Singleton;
 
 @Module(includes = {AndroidSupportInjectionModule.class, ViewModelModule.class})
-abstract class GndApplicationModule {
-  private static final String SHARED_PREFERENCES_NAME = "shared_prefs";
+public abstract class GndApplicationModule {
```
dturner, line 48: Out of interest, what caused this to become `public`?

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
```diff
@@ -0,0 +1,79 @@
...
+    return DEFAULT_NETWORK_TYPE;
+  }
+
+  abstract Class<? extends BaseWorker> getWorkerClass();
+
+  protected OneTimeWorkRequest buildWorkerRequest(@Nullable Data inputData) {
+    Builder builder =
+        new Builder(getWorkerClass())
+            .setConstraints(getWorkerConstraints())
+            .setBackoffCriteria(BackoffPolicy.LINEAR, SYNC_BACKOFF_MILLIS, TimeUnit.MILLISECONDS);
```
dturner, line 71: Any reason for picking `LINEAR` backoff? I thought the generally accepted best practice was to use `EXPONENTIAL`

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
shobhitagarwal1612: Thanks for pointing that out. So does that mean I should update the label in the settings as well?

gnd/src/main/java/com/google/android/gnd/GndApplicationModule.java
shobhitagarwal1612: At first, I reused the same variable `SHARED_PREFERENCES_NAME` in the SettingsFragment, but then moved it `Config.java`. So this can be reverted back now. Nice catch!

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
shobhitagarwal1612: This is the current backoff criteria being used in `DataSyncWorkManager`. I just pulled it out and moved to `BaseWorkerManager` so we can apply a common template to all of our `*WorkManager` and move boiler-plate code to this base class.

But I do agree `EXPONENTIAL` seems to be the best practice. @gino-m Could you please provide some background why it `LINEAR` was set as the default choice?

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
gino-m: Good catch, @dturner. Google Photos and other apps still prompt the user to "Wait for Wi-Fi" rather than "Wait for Unmetered connection". The latter is harder to explain to beginner users - @sergeydolgov1 do you think we should use "Wi-Fi" as a proxy for "Unmetered Network" in the UIs, even though it's not really 100% correct?

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
gino-m: I don't remember for certain, but it might have been an attempt to compensate for the fact that gRPC can take awhile to wake up after a connection becomes available. 

How does backoff interact with the required network type? i.e. is the worker expected to start ASAP after the required network is available, regardless of backoff?

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
```diff
@@ -0,0 +1,79 @@
...
+import androidx.work.Data;
+import androidx.work.NetworkType;
+import androidx.work.OneTimeWorkRequest;
+import androidx.work.OneTimeWorkRequest.Builder;
+import androidx.work.WorkManager;
+import androidx.work.WorkRequest;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Provider;
+
+public abstract class BaseWorkManager {
```
gino-m, line 31: Please add class JavaDoc with quick summary of what this base class provides.

gnd/src/main/java/com/google/android/gnd/persistence/sync/TileDownloadWorkManager.java
```diff
@@ -16,47 +16,48 @@
...
+   * completable upon enqueueing.
    */
   public Completable enqueueTileDownloadWorker() {
     return Completable.fromRunnable(this::enqueueTileDownloadWorkerInternal);
   }
 
   private void enqueueTileDownloadWorkerInternal() {
-    OneTimeWorkRequest request = buildWorkerRequest();
-
+    OneTimeWorkRequest request = buildWorkerRequest(null);
```
gino-m, line 60: Consider adding a noargs override of `buildWorkerRquest()` instead of passing null here with JavaDoc explaining the default behavior with noargs.

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
shobhitagarwal1612: It will check for the constraint only after the required amount of time has passed based of last backoff delay. Only then it will start.

gino-m: Is there no way to wake up the worker as soon as the connection becomes available? (@dturner?).  If not, perhaps LINEAR is a safer bet? As it's currently defined, retries will only occur at 10 sec, then 100 sec, and 1000 sec?

dturner:
FYI `checkstyle` is causing the CI to fail: 

```
[ant:checkstyle] [ERROR] /workspace/gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java:43: Empty line should be followed by <p> tag on the next line. [JavadocParagraph]
```

shobhitagarwal1612:
Checkstyle issue resolved.

gnd/src/main/java/com/google/android/gnd/persistence/sync/BaseWorkManager.java
shobhitagarwal1612: Reverted back to LINEAR

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
sergeydolgov1: @gino-m Suggest that we use "Wi-Fi" as a proxy for Unmetered Network as Wi-Fi is usually the way people use the Internet free of charge. Also, FYI -- this is Android developer documentation, they also use "Wi-Fi" terminology: https://developer.android.com/training/basics/network-ops/managing

PR #508, opened 2020-06-30
Fix crash on rotating home screen

### LINKED ISSUE #440: App crashes when device is rotated
To reproduce, rotate the device with "auto rotate" enabled in Android settings. Confirmed on map screen and expanded observation list.

In theory, we could disable rotation for the app, but the app still needs to be able to survive configuration changes in general.

@dturner Does this look like one you might be able to help with?

shobhitagarwal1612:
Fixes #440


gino-m:
Thanks for the fix!

PR #509, opened 2020-06-30
[Add feature] Show observation form immediately after adding new feature

### LINKED ISSUE #351: [Add feature] Show observation form immediately after adding feature

shobhitagarwal1612:
Fixes #351

@gino-m PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
@@ -16,7 +16,6 @@
 
 package com.google.android.gnd.ui.home;
 
-import android.util.Log;
 import android.view.View;
 import androidx.lifecycle.LiveData;
 import androidx.lifecycle.LiveDataReactiveStreams;
@@ -40,11 +39,11 @@ import io.reactivex.Single;
 import io.reactivex.subjects.PublishSubject;
 import java8.util.Optional;
 import javax.inject.Inject;
+import timber.log.Timber;
 
 @SharedViewModel
 public class HomeScreenViewModel extends AbstractViewModel {
 
-  private static final String TAG = HomeScreenViewModel.class.getSimpleName();
   private final ProjectRepository projectRepository;
   private final Navigator navigator;
   /** The state and value of the currently active project (loading, loaded, etc.). */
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -87,7 +86,14 @@
                         .doOnError(this::onAddFeatureError)
                         .onErrorResumeNext(Single.never())) // Prevent from breaking upstream.
             .observeOn(schedulers.ui())
-            .subscribe(this::showBottomSheet));
+            .subscribe(this::addNewObservation));
```
gino-m, line 89: Cool, thanks! We really needed this! One caveat, we should only direct to add observation if the layer in which the feature lives has a form defined. Can we check that first?

gino-m:
@sergeydolgov1 fyi.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -85,8 +85,10 @@
                         .toSingleDefault(newFeature)
                         .doOnError(this::onAddFeatureError)
                         .onErrorResumeNext(Single.never())) // Prevent from breaking upstream.
+            .filter(feature -> feature.getLayer().getForm().isPresent())
+            .doOnNext(this::addNewObservation)
```
gino-m, line 89: Relying on side-effects to do actual work is an anti-pattern that's we're actively seeking to avoid; rather than using `filter` + `doOnNext` can we just add a method called from `subscribe`, called `onAddFeature` which checks if the form is present and if so calls, `addNewObservation`?

shobhitagarwal1612: Thanks for the wisdom!

gino-m: My pleasure! Added https://github.com/google/ground-android/wiki/RxJava-Usage-and-Conventions#side-effects to clarify.

gino-m:
Nice work!

PR #510, opened 2020-06-30
[Code check] Update checkstyle plugin

### LINKED ISSUE #480: [Code health] Update checkstyle plugin
Update the checkstyle plugin to 8.32 to catch up with the latest release. Also, update the google_checks.xml from the source.

shobhitagarwal1612:
Fixes #480 

@scolsen PTAL?


gnd/src/main/java/com/google/android/gnd/rx/Action.java
```diff
@@ -28,19 +28,15 @@ protected Action() {
...
-  /**
-   * Invokes the provided handler if the value has not yet been handled.
-   *
-   * @param handler
-   */
+  public static Action create() {
+    return new Action();
+  }
+
+  /** Invokes the provided handler if the value has not yet been handled. */
```
dturner, line 35: Just an observation. It feels like the diff got confused with the moving of `create` to above `ifUnhandled`. Might be better to keep the ordering the same to avoid "noise" in future PRs.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalValueStore.java
```diff
@@ -54,4 +54,12 @@ public void setLastActiveProjectId(@NonNull String id) {
   public void clearLastActiveProjectId() {
     preferences.edit().remove(ACTIVE_PROJECT_ID_KEY).apply();
   }
+
+  public boolean shouldUploadMediaOverUnmeteredConnectionOnly() {
```
dturner: Nit: This change isn't related to checkstyle. Might be better to rebase master rather than merge in future.

PR #511, opened 2020-06-30
[Offline base map] Notify user when "download" clicked

### LINKED ISSUE #537: [Offline base maps] Show toast when user clicks "Download"
e.g. "Download started in background". Currently it seems like nothing happens.

scolsen:
This PR adds a basic toast + navigation change to signal to users that they've initiated an offline area download. After tapping the button, the user is immediately returned to the offline area list, where the newly downloaded area is listed. An informative toast also appears. On errors we still return to the list, but display a different toast message.

Screen Shot 2020-06-30 at 5 32 24 PM

Closes #537.


gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
 import com.google.android.gnd.repository.OfflineAreaRepository;
 import com.google.android.gnd.ui.common.AbstractViewModel;
+import io.reactivex.BackpressureStrategy;
+import io.reactivex.subjects.PublishSubject;
 import javax.inject.Inject;
 import timber.log.Timber;
 
 public class OfflineAreaSelectorViewModel extends AbstractViewModel {
 
+  private final LiveData<DownloadTrigger> downloads;
```
gino-m, line 31: I believe in this case the toast will re-appear every time to view is restored (e.g. after configuration change). To ensure they're single use, you can wrap DownloadTrigger in our custom class  `Event<>`.

gino-m, line 31: downloads -> downloadEvents?

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
+
+  public LiveData<DownloadTrigger> getDownloadTriggers() {
+    return this.downloads;
+  }
+
+  enum DownloadTrigger {
```
gino-m, line 37: DownloadEvent?

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
-      .blockingAwait();
+    offlineAreaRepository
+        .addAreaAndEnqueue(viewport)
+        .doOnError(
+            err -> {
+              Timber.e("Failed to add area and queue downloads: %s", err.getMessage());
+              downloadsPublishSubject.onNext(DownloadTrigger.Failure);
+            })
+        .doOnComplete(() -> downloadsPublishSubject.onNext(DownloadTrigger.Started))
+        .blockingAwait();
```
gino-m, line 64: Please add a TODO to handle this asynchronously instead of blocking the UI.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
-      .onErrorComplete()
-      .blockingAwait();
+    offlineAreaRepository
+        .addAreaAndEnqueue(viewport)
+        .doOnError(
+            err -> {
+              Timber.e("Failed to add area and queue downloads: %s", err.getMessage());
+              downloadsPublishSubject.onNext(DownloadTrigger.Failure);
+            })
+        .doOnComplete(() -> downloadsPublishSubject.onNext(DownloadTrigger.Started))
```
gino-m, line 63: Since you're blocking, you can just move this operation outside of the stream immediately following the blockingAwait().

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
   // TODO: Use an abstraction over LatLngBounds
   public void onDownloadClick(LatLngBounds viewport) {
     Timber.d("viewport:%s", viewport);
-    offlineAreaRepository.addAreaAndEnqueue(viewport)
-      .doOnError(err -> Timber.e("Failed to add area and queue downloads: %s", err.getMessage()))
-      .onErrorComplete()
-      .blockingAwait();
+    offlineAreaRepository
+        .addAreaAndEnqueue(viewport)
+        .doOnError(
```
gino-m, line 58: This will report the error but it won't catch it; I believe blockingAwait() will throw an exception for unhandled errors, so may want to use try{} catch around this call instead so you'd catch both cases.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,51 @@
...
 
+  private final LiveData<DownloadTrigger> downloads;
+
+  public LiveData<DownloadTrigger> getDownloadTriggers() {
+    return this.downloads;
+  }
+
+  enum DownloadTrigger {
+    Started,
+    Failure
```
gino-m, line 39: Enum values are constants so they're general ALL_CAPS in Java.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: Done

scolsen: Done

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: Done

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: I just made it non-blocking by introducing a click stream+subscription

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: Handled via subscribe instead

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: Done

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorFragment.java
```diff
@@ -56,6 +61,24 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+            e -> {
+              e.ifUnhandled(this::onDownloadEvent);
+            });
+  }
+
+  private void onDownloadEvent(DownloadEvent downloadEvent) {
+    switch (downloadEvent) {
+      case FAILURE:
+        EphemeralPopups.showError(getContext(), R.string.offline_area_download_failed);
+        navigator.navigateUp();
```
gino-m, line 77: Add "break;" here and below.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,62 @@
...
 public class OfflineAreaSelectorViewModel extends AbstractViewModel {
 
-  private final OfflineAreaRepository offlineAreaRepository;
+  private final LiveData<Event<DownloadEvent>> downloadEvents;
+
+  public LiveData<Event<DownloadEvent>> getDownloadEvents() {
+    return this.downloadEvents;
+  }
+
+  enum DownloadEvent {
```
gino-m, line 40: Pls move enum declaration to start of class.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,62 @@
...
+  private final PublishSubject<LatLngBounds> downloadClickSubject = PublishSubject.create();
 
   @Inject
-  OfflineAreaSelectorViewModel(OfflineAreaRepository offlineAreaRepository) {
-    this.offlineAreaRepository = offlineAreaRepository;
+  OfflineAreaSelectorViewModel(OfflineAreaRepository offlineAreaRepository, Schedulers schedulers) {
+
+    disposeOnClear(
+        downloadClickSubject
+            .flatMap(
```
gino-m, line 53: I think maybe here you'd just want to use `flatMapCompletable` without the additional andThen?

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
```diff
@@ -16,27 +16,62 @@
...
+                // We need to handle this in the inner lambda instead of using a combinator such as
+                // flatMapCompletable because the PublishSubject will never actually complete.
+                // andThen enables us to fill the resulting stream w/ items to propagate completion.
+                viewport ->
+                    offlineAreaRepository
+                        .addAreaAndEnqueue(viewport)
+                        .andThen(Observable.just(new Object())))
+            .observeOn(schedulers.ui())
+            .doOnError(e -> Timber.e("Failed to add area and queue downloads: %s", e.getMessage()))
+            .subscribe(
```
gino-m, line 63: Following up on the more idiomatic use about, you should then be able to do:
```
  .toSingleDefault(DownloadEvent.STARTED)
  .onErrorReturn(DownloadEvent.FAILURE)
 .subscribe(downloadsPublishSubject::onNext);
```

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorViewModel.java
scolsen: unfortunately, `flatMapCompletable` won't work here unless we call `onComplete` explicitly on the `PublishSubject`  from the docs

> Maps each element of the upstream Observable into CompletableSources, subscribes to them and waits until the upstream and all CompletableSources complete.

So, since a PublishSubject continually listens for clicks, it doesn't complete and the flatmap will never complete either (since it's perpetually waiting for the subject to complete). I added the comment just above the lambda to describe this predicament

Unless there's a subject we should use instead of `PublishSubject`? We could also force it to complete, but that would ensure the user can only ever click `Download` once, which is the current effect anyway. wdyt?

scolsen: I think if we're ok with forcing a single click to occur, the right approach is probably to use a `SingleSubject` for the clicks.

scolsen:
@gino-m just merged in your edits, thanks for that! Just an FYI I'll be gone next week so there won't be any github activity from me during that time (except maybe on Wednesday but the likelihood is low).

scolsen:
@gino-m checks should pass now, just need a re-approval!

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/selector/OfflineAreaSelectorFragment.java
```diff
@@ -55,6 +60,21 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+  }
+
+  private void onDownloadMessage(DownloadMessage message) {
+    switch (message) {
+      case STARTED:
+        EphemeralPopups.showSuccess(getContext(), R.string.offline_area_download_started);
+        navigator.navigateUp();
+        break;
+      case FAILURE:
+      default:
```
shobhitagarwal1612, line 73: Why default? Doesn't this make above line `case FAILURE` redundant?

scolsen: It does, but I believe PMD (or some other check) requires a default clause. We could drop the `FAILURE` clause, but I think it's ok to keep for now in case we decide we want to handle that case differently--this way we're explicit about discriminating over all the enum values.

PR #514, opened 2020-07-01
[Code health] Tweak plugin and update gradle dependencies

shobhitagarwal1612:
* Add a filter to `dependencyUpdates` plugin to prevent notifying update from stable to unstable version. (We hardly ever do that. Except for when testing latest libraries e.g. `hilt`). This shows the code health more accurately in the Github badge

* Bump gradle and several other dependencies

@gino-m PTAL.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,9 +77,23 @@ gitVersioner {
     baseBranch "master"
 }
 
+def isNonStable = { String version ->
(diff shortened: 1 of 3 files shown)


gino-m:
Will this prevent the dependency update plugin from suggesting upgrades to an unstable version of libs?

shobhitagarwal1612:
Yes, only if the current used version is stable. 
Since for `hilt` we are using an alpha release, so this will continue showing latest unstable release for them

Issue #517, opened 2020-07-01
[Local data store] Simplify local persistence layer


gino-m:
Ideas:

* We currently split the project config into many different tables, even though the entire hierarchy is always read and written as a single unit. We could reduce complexity by storing the project as a blob. 
* The same may apply to other entities. 
* Using protos (possible with an autocopy lib to copy between abstraction layers) could be a solution to decrease complexity.

gino-m:
@JSunde Would it be absurd to use Kotlin serialization to persist the entire Survey and contained objects in the db to reduce the number of entity mappings we need?

Issue #516, opened 2020-07-01
[Offline base maps] Allow offline base maps to be configured by Manager


gino-m:
- [ ] Add base map sources to remote data store.
- [ ] Cache base map sources in local data store.
- [ ] Read URL from base map sources (initial only use first one for now if present) and fetch GeoJSON extents file from remote when loading offline area selector UI.

gino-m:
Merged into #481 as part of issue consolidation and cleanup.

PR #518, opened 2020-07-01
[Offline base maps] Add offline base map sources to remote and local databases

gino-m:
Work towards #516.

Adding a since repeated field was much more work than expected (286 lines across 13 files). In large part this is due to how we persist data locally. Filed #517 with a few ideas of how to improve this in the future.


gino-m:
@scolsen Could you PTAL?

gnd/src/main/java/com/google/android/gnd/model/Project.java
```diff
@@ -39,6 +40,9 @@
   @NonNull
   protected abstract ImmutableMap<String, Layer> getLayerMap();
 
+  @NonNull
+  public abstract ImmutableList<OfflineBaseMapSource> getOfflineBaseMapSources();
```
scolsen, line 44: Cool! What's the envisioned flow for selecting between these? 

Offline Areas -> Add Area -> Select Source -> Download... or something different?

gino-m: These will be configured via the web app only - there will likely be an overflow menu with advanced project settings where we can paste the URL.

gino-m:
Thanks for the review!

PR #522, opened 2020-07-04
[Manage observations] Add ability to remove observations

### LINKED ISSUE #107: [Delete observation] Add ability to remove observations
Only accessible to project managers and owner of observation.

shobhitagarwal1612:
Fixes #107

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
@@ -25,6 +25,7 @@ import com.google.android.gnd.model.feature.Feature;
 import com.google.android.gnd.model.feature.FeatureMutation;
 import com.google.android.gnd.model.observation.Observation;
 import com.google.android.gnd.model.observation.ObservationMutation;
+import com.google.android.gnd.persistence.local.room.LocalDataStoreException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.reactivex.Completable;
@@ -79,6 +80,12 @@ public interface LocalDataStore {
    * mutation to the local queue for remote sync.
    */
   Completable applyAndEnqueue(ObservationMutation mutation);
+
+  /** Applies the specified {@link ObservationMutation} to the local data store.. */
+  Completable apply(ObservationMutation mutation) throws LocalDataStoreException;
+
+  /** Appends the specified {@link ObservationMutation} to the local queue for remote sync. */
+  Completable enqueue(ObservationMutation mutation);
 
   /**
    * Returns a long-lived stream that emits the full set of features for a project on subscribe, and
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
@@ -84,12 +84,6 @@ public interface LocalDataStore {
   /** Applies the specified {@link ObservationMutation} to the local data store.. */
   Completable apply(ObservationMutation mutation) throws LocalDataStoreException;
 
-  /** Appends the specified {@link ObservationMutation} to the local queue for remote sync. */
-  Completable enqueue(ObservationMutation mutation);
-
-  /** Deletes local observation entry. */
-  Completable deleteObservation(ObservationMutation mutation);
-
   /**
    * Returns a long-lived stream that emits the full set of features for a project on subscribe, and
    * continues to return the full set each time a feature is added/changed/removed.
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
@@ -134,6 +134,9 @@ public interface LocalDataStore {
    */
   Completable mergeObservation(Observation observation);
 
+  /** Deletes observation from local database. */
+  Completable deleteObservation(String observationId);
+
   /**
    * Attempts to update a tile in the local data store. If the tile doesn't exist, inserts the tile
    * into the local data store.
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
@@ -87,13 +87,19 @@ public interface LocalDataStore {
   /** Appends the specified {@link ObservationMutation} to the local queue for remote sync. */
   Completable enqueue(ObservationMutation mutation);
 
+  /** Deletes local observation entry. */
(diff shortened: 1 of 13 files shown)


dturner:
The deleted observation is still appearing in the list after it's been deleted. Steps to repro: 

- Load a feature
- Add observation
- Tap on observation
- Tap "Delete observation"
- View changes to observation list

Expected result

- Observation is removed from list of observations 

Actual result

- Observation is still present and tapping on it causes the app to crash with `java.util.NoSuchElementException: No value present`

shobhitagarwal1612:
@gino-m @dturner PTAL

dturner:
Failing unit test: 

```
com.google.android.gnd.persistence.local.LocalDataStoreTest > testDeleteObservation FAILED
Step #5 - "Run unit tests":     java.lang.AssertionError at LocalDataStoreTest.java:422
```

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -80,13 +81,25 @@
...
   Completable applyAndEnqueue(ObservationMutation mutation);
 
+  /** Applies the specified {@link ObservationMutation} to the local data store.. */
+  Completable apply(ObservationMutation mutation) throws LocalDataStoreException;
+
+  /** Appends the specified {@link ObservationMutation} to the local queue for remote sync. */
+  Completable enqueue(ObservationMutation mutation);
+
+  /** Deletes local observation entry. */
+  Completable deleteObservation(ObservationMutation mutation);
```
gino-m, line 91: Isn't this just another case of `apply(ObservationMutation)`? Adding a separate method would break this model, and it would mean there are many configurations of ObservationMutation that would be invalid and would need to be checked. Can we instead call the impl of this method via `apply()`?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -260,6 +261,8 @@ public Completable applyAndEnqueue(FeatureMutation mutation) {
         .map(
             list ->
                 stream(list)
+                    .filter(
+                        observationEntity -> observationEntity.getState() != EntityState.DELETED)
```
gino-m, line 265: Can we implement this in the DAO instead?

shobhitagarwal1612: Thanks for pointing out! Dao already contains a where clause for DEFAULT state

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
shobhitagarwal1612: Done 7f6dbff

shobhitagarwal1612:
@gino-m Thanks for your comments! I've added some more tests to exercise all moving parts of the diff. PTAL

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -424,8 +427,25 @@ public Completable apply(ObservationMutation mutation) throws LocalDataStoreExce
...
+            .flatMapCompletable(
+                entity -> {
+                  // Mark the ObservationEntity as DELETED if the current state is DEFAULT.
+                  // Otherwise, delete the observation entity from local db.
+                  //
+                  // TODO: If the remote sync fails, reset the state to DEFAULT.
+                  //
+                  // Note: Observations marked as DELETED are not shown in UI.
+                  //       See {@link ObservationDao}
+                  if (entity.getState() == EntityState.DEFAULT) {
```
gino-m, line 441: This logic is not idempotent, which can lead to issues wrt order and timing of mutations; if you call it once, it will do something different than if you call it twice. 

It's not clear to me when/how the second mutation to delete the entity would occur. Wouldn't it occur at a lower level, once the remote observation is actually removed (i.e. not via the mutation interface)?

shobhitagarwal1612: I may have misinterpreted your previous comment. You mentioned that `deleteObservation` is another case of apply and should be called via it rather than directly. So I merged the two methods `markObservationDeleted` and `deleteObservation` under the same `apply` implementation.

Current flow:
mark observation as deleted -> create mutation -> remote sync (delete) -> delete local observation -> delete mutation

@gino-m Did I make an error in understanding the comments? Please confirm.

gino-m: Sorry, I may have misspoke - specifically "mark deleted" is a special case of `apply`. "Delete permanently" isn't represented as a mutation type, since it's not a user action that we need to sync to/from the remote server.  

The flow as you described it seems correct. My comment was relating to how the "delete local observation" step is being invoked - it currently depends on how many times this method is called, and as mentioned, the cleanup step isn't a synced mutation type. Instead, it can be done directly in the db after the remote observation is deleted.

shobhitagarwal1612: Done 3077b29

gnd/src/main/java/com/google/android/gnd/persistence/sync/LocalMutationSyncWorker.java
```diff
@@ -112,17 +112,17 @@ private Completable processMutations(ImmutableList<Mutation> mutations, String u
   private Completable processMutations(ImmutableList<Mutation> mutations, User user) {
     return remoteDataStore
         .applyMutations(mutations, user)
-        .andThen(removeLocalObservationsIfMutationTypeIsDelete(mutations))
+        .andThen(deleteObservationIfRemovedRemotely(mutations))
```
gino-m, line 115: Why don't we just delete the observation as soon as the remote observation is deleted inside `applyMutations`? Sorry if I'm missing something!

shobhitagarwal1612: https://github.com/google/ground-android/blob/1a1246f0a55404c0fe3ed98ea83c1b35dfc594e4/gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/FirestoreDataStore.java#L112

IIUC, deleting the observation just after adding batch request would not ensure that the remote observation was indeed removed. So, to ensure that we only do it after successful task completion, I chose to do it in after the success call.
Does that seem right to you?

gino-m: Ah, good point; sorry I forgot we were deleting them as part of the batch. I was concerned about possible race conditions, but we're using the same set of mutations in both steps, so I think we're ok here. LGTM!

PR #524, opened 2020-07-06
Refactor ObservationRepository

### LINKED ISSUE #523: [Code health] Move creation of observation mutation to ObservationRepository
`EditObsevationViewModel` has a lot of logic and it would be better if we move the logic for creating observation mutations in the `ObservationRepository` where the same is done for creating observations.

This would help in simplifying the logic class and increasing readability.

shobhitagarwal1612:
Closes #523 

Move boilerplate code for generating mutations to ObservationRepository

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
@@ -34,7 +34,6 @@ import com.google.android.gnd.model.form.Element.Type;
 import com.google.android.gnd.model.form.Field;
 import com.google.android.gnd.model.form.Form;
 import com.google.android.gnd.model.observation.Observation;
-import com.google.android.gnd.model.observation.ObservationMutation;
 import com.google.android.gnd.model.observation.Response;
 import com.google.android.gnd.model.observation.ResponseDelta;
 import com.google.android.gnd.model.observation.ResponseMap;
@@ -42,7 +41,6 @@ import com.google.android.gnd.persistence.uuid.OfflineUuidGenerator;
 import com.google.android.gnd.repository.ObservationRepository;
 import com.google.android.gnd.rx.Event;
 import com.google.android.gnd.rx.Nil;
-import com.google.android.gnd.system.AuthenticationManager;
 import com.google.android.gnd.system.CameraManager;
 import com.google.android.gnd.system.StorageManager;
 import com.google.android.gnd.ui.common.AbstractViewModel;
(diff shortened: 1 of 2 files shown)


gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -303,21 +293,13 @@ private void onObservationLoaded(Observation observation) {
...
-            .setUserId(authManager.getCurrentUser().getId())
-            .build();
+    if (originalObservation == null) {
+      return Single.error(new IllegalStateException("Observation is null"));
+    }
+
     return observationRepository
-        .applyAndEnqueue(observationMutation)
+        .addObservationMutation(originalObservation, getResponseDeltas(), isNew)
+        .doOnSubscribe(disposable -> savingProgressVisibility.setValue(View.VISIBLE))
```
dturner, line 302: Just an observation, the viewmodel's `saveProgressVisibility` property could be refactored to `isSaving` and the visibility logic could then be handled in the layout XML using databinding e.g. `android:visibility="@{viewmodel.isSaving ? View.VISIBLE : View.GONE}"`

shobhitagarwal1612: Thanks for the comment! We are already using databinding for toggling the visibility. The only difference is, currently the binded object is an integer type whose value is equal to one of the visibility values. And in this scenario, we would use a boolean type instead with the suggested refactor. Any particular reason for using boolean over integer type?

dturner: Yeah, it's really only a minor thing. The reason for using boolean over integer is that it represents the 2 saving states: "saving" and "not saving". 

Currently the `savingProgressVisibility` integer maps to the 3 possible visibility states: visible, invisible and gone. We don't seem to use "invisible" so refactoring to `isSaving.setValue(true)` makes the code (albeit only slightly) more readable.

dturner: I did the refactor on `ObservationListViewModel` - see what you think :) https://github.com/shobhitagarwal1612/ground-android/pull/1/files

shobhitagarwal1612: I like this better. Thanks for the suggestion! 
Would you like to open a PR with the proposed changes? If not, then I'd be more than happy to do it on your behalf. 😄

dturner: I'm happy to do it. Will do shortly.

dturner: https://github.com/google/ground-android/pull/529

PR #529, opened 2020-07-06
[Code Health] Move visibility logic from ViewModels to layouts

dturner:
Slightly simplifies the view models by refactoring the "visibility" variables from ints which hold the visibility to booleans which hold the underlying state e.g. `isLoading` `isSaving` and can then be used in the layouts.


shobhitagarwal1612:
Just needs some minor PMD warnings fixed. Good to go then!

PR #542, opened 2020-07-14
List Human Readable Area Names in Offline Area List

### LINKED ISSUE #536: [Offline base maps] Show place name in offline area list
Place names can be fetched with https://developer.android.com/reference/android/location/Geocoder when downloading the areas.

This is low priority but adding here to tract - for now we'll show "Area 1, Area 2.." in the offline base map areas list.

@sergeydolgov1 Fyi, please cmiiw.

scolsen:
This PR uses reverse Geocoding to give downloaded offline areas human readable names. 

Area names are determined by walking up (from bottom up) the addressing hierarchy defined by the [Extensible Address Language](https://www.immagic.com/eLibrary/ARCHIVES/TECH/OASIS/XAL_V2.PDF)--according to the docs, this is the schema that Android's [`Address` implementation roughly follows](https://developer.android.com/reference/android/location/Address).

Explicitly, we try to get a name for the area using the following address components in the following order--if the prior component is null, we move on to the next item in the list:

- Feature name (e.g. Golden Gate Bridge)
- Locality name
- Subadministration name
- Administration name
- Country name

Screen Shot 2020-07-14 at 12 48 23 PM

Closes #536.


gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
```diff
@@ -75,9 +90,42 @@ public void onClick(View v) {
   }
 
   @Override
-  public void onBindViewHolder(ViewHolder viewHolder, int position) {
-    viewHolder.binding.offlineAreaName.setText(offlineAreas.get(position).getId());
-    viewHolder.position = position;
+  public void onBindViewHolder(@NonNull ViewHolder viewHolder, int position) {
+    String areaName =
+        viewHolder.binding.getRoot().getContext().getString(R.string.offline_areas_unknown_area);
```
shobhitagarwal1612, line 95: Can you please move this string creation to `catch` block? Otherwise, it seems wasteful to do this init each time.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
```diff
@@ -75,9 +90,42 @@ public void onClick(View v) {
...
+      Address address = addresses.get(0);
+
+      areaName =
+          Optional.ofNullable(address.getFeatureName())
+              .or(() -> Optional.ofNullable(address.getLocality()))
+              .or(() -> Optional.ofNullable(address.getSubAdminArea()))
+              .or(() -> Optional.ofNullable(address.getAdminArea()))
+              .or(() -> Optional.ofNullable(address.getCountryName()))
+              .orElse(areaName);
+    } catch (Exception e) {
```
shobhitagarwal1612, line 118: Can we please specify exactly which exception we are looking for? Is this catching the errors other than `AddressNotFoundException` raised above?

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
```diff
@@ -37,6 +50,7 @@
     public int position;
     private ImmutableList<OfflineArea> areas;
     private final Navigator navigator;
+    private final Geocoder geocoder;
```
gino-m, line 53: Following our current architecture, remote and system services are all hidden either behind a `Repository` or a `Manager`, respectively. Also,
`ListAdapter` is a purely UI component, so it would be best to abstract services like this one so they can be mocked or replaced without needing to mock the lower level API.

How about creating a `GeocodingService` that exposes only the functionality needed to implement this feature?

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
```diff
@@ -75,9 +90,42 @@ public void onClick(View v) {
...
   @Override
-  public void onBindViewHolder(ViewHolder viewHolder, int position) {
-    viewHolder.binding.offlineAreaName.setText(offlineAreas.get(position).getId());
-    viewHolder.position = position;
+  public void onBindViewHolder(@NonNull ViewHolder viewHolder, int position) {
+    String areaName =
+        viewHolder.binding.getRoot().getContext().getString(R.string.offline_areas_unknown_area);
+
+    OfflineArea area = offlineAreas.get(position);
+    String id = area.getId();
```
gino-m, line 98: In particular, this is business logic that would be best defined in a more tightly scoped class imo.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
gino-m: @shobhitagarwal1612 Could you clarify what would be wasted and when?

gino-m:
Added issue number to description, moved to "In progress" pipeline.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
shobhitagarwal1612: IIUC, this is the default area name "unknown area" which would be set if the try block throws an exception. So, this value would only be used if reverse geocoding fails for a particular point

scolsen: After the refactor in 81c0b3f, the setup is a bit different--we still need to do the allocation early as we can't just hold on to the context in the class (iirc this can lead to leaks)--but it should be a bit cleaner now.

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
scolsen: done! 81c0b3f

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
scolsen: Done! 81c0b3f added a `GeocodingManager`

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
scolsen: 81c0b3f

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
```diff
@@ -60,7 +47,7 @@ public AddressNotFoundException(String message) {
       this.binding = binding;
       this.areas = areas;
       this.navigator = navigator;
-      this.geocoder = new Geocoder(binding.getRoot().getContext());
+      this.geocoder = new GeocodingManager(binding.getRoot().getContext());
```
gino-m, line 50: This seems like logic that would normally be called from the ViewModel. Also, it would be preferable to inject the `GeocodingManager` (in the VM) to make it easier to swap out in tests. Would this be possible?

gnd/src/main/java/com/google/android/gnd/system/GeocodingManager.java
```diff
@@ -0,0 +1,85 @@
...
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gnd.R;
+import com.google.android.gnd.model.basemap.OfflineArea;
+import java.io.IOException;
+import java.util.List;
+import java8.util.Optional;
+import timber.log.Timber;
+
+/**
+ * GeocodingManger abstracts over native geocoding facilities, and provides convenience methods for
```
gino-m, line 31: s/abstracts over/abstracts/

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreaListAdapter.java
scolsen: 77aaed7

gnd/src/main/java/com/google/android/gnd/system/GeocodingManager.java
scolsen: 656adc8

scolsen:
Thanks, @gino-m, ready for another look!

gnd/src/main/java/com/google/android/gnd/ui/offlinearea/OfflineAreasViewModel.java
```diff
@@ -16,39 +16,57 @@
...
+  private final GeocodingManager geocoder;
 
   @Inject
-  OfflineAreasViewModel(Navigator navigator, OfflineAreaRepository offlineAreaRepository) {
+  OfflineAreasViewModel(
+      Navigator navigator,
+      OfflineAreaRepository offlineAreaRepository,
+      @ApplicationContext Context context) {
     this.navigator = navigator;
+    this.geocoder = new GeocodingManager(context);
```
shobhitagarwal1612, line 49: Instead of creating an instance here, how about converting `GeocodingManager` into an injectable dependency?

gino-m: +1. I believe all you need is to add `@Inject` to the constructor.

scolsen: 56b49e5

PR #565, opened 2020-08-14
[View/add/edit observations] Sort fields by `index` defined in form

### LINKED ISSUE #396: [View/edit observation] Order of fields not consistent
**To Reproduce**
1. Select "Plastic Frog Finder" project
2. Click on "+" to add observation

**Expected behavior**
Photo field should be at the last 

**Actual behavior**
Photo field is at position 3 

**Screenshots**
image

gino-m:
Fixes #396.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/model/form/Element.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/form/Element.java
@@ -17,6 +17,7 @@
 package com.google.android.gnd.model.form;
 
 import com.google.auto.value.AutoOneOf;
+import java8.util.Optional;
 import javax.annotation.Nullable;
 
 /**
@@ -28,6 +29,21 @@ public abstract class Element {
   public enum Type {
     UNKNOWN,
     FIELD
+  }
+
+  public int getIndex() {
+    switch (getType()) {
+      case FIELD:
+        // Fall back to hash code id on missing index; the order won't necessarily be correct
+        // but at least it will be stable.
+        return Optional.ofNullable(getField().getIndex())
+            .orElseGet(() -> getField().getId().hashCode());
+      case UNKNOWN:
+        // Intentional fall-through.
+      default:
+        // Fall back for unknown / bad index.
+        return hashCode();
+    }
   }
 
   public abstract Type getType();
--- a/gnd/src/main/java/com/google/android/gnd/model/form/Element.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/form/Element.java
@@ -17,7 +17,6 @@
 package com.google.android.gnd.model.form;
 
 import com.google.auto.value.AutoOneOf;
-import java8.util.Optional;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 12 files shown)


gnd/src/main/java/com/google/android/gnd/model/form/Form.java
```diff
@@ -35,6 +37,12 @@
 
   public abstract ImmutableList<Element> getElements();
 
+  public ImmutableList<Element> getElementsSorted() {
```
shobhitagarwal1612, line 40: How about `getElementsSorted()` --> `getSortedElements()`?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FieldEntity.java
```diff
@@ -52,6 +52,11 @@
   @ColumnInfo(name = "id")
   public abstract String getId();
 
+  @CopyAnnotations
+  @Nullable
+  @ColumnInfo(name = "index")
+  public abstract Integer getIndex();
```
shobhitagarwal1612, line 58: Can we please make this `@NonNull`? It would remove the extra complexity of having to deal with all fallbacks

gnd/src/main/java/com/google/android/gnd/model/form/Element.java
```diff
@@ -30,6 +31,21 @@
...
+      case FIELD:
+        // Fall back to hash code id on missing index; the order won't necessarily be correct
+        // but at least it will be stable.
+        return Optional.ofNullable(getField().getIndex())
+            .orElseGet(() -> getField().getId().hashCode());
+      case UNKNOWN:
+        // Intentional fall-through.
+      default:
+        // Fall back for unknown / bad index.
+        return hashCode();
```
shobhitagarwal1612, line 45: What if we make index a required field? If the online form generator always adds an index then there would be no need to take care of adding this attribute manually. Wdyt?

gnd/src/main/java/com/google/android/gnd/model/form/Form.java
gino-m: The naming was chosen to emphasize that this is a variant of the `getElements()` method -  it also returns elements, but sorted. I've seen this naming convention used elsewhere but can't find it now. I don't feel strongly about it though. Is there precedency for this pattern that you know of?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FieldEntity.java
gino-m: Done!

gnd/src/main/java/com/google/android/gnd/model/form/Form.java
shobhitagarwal1612: I don't know if there is a rule/pattern for that. It just seemed much more natural to add "sorted" first. I am fine with either as it is just a minor nit. Let's keep it the same way.

gnd/src/main/java/com/google/android/gnd/model/form/Element.java
gino-m: Good idea. Let's handle missing `index` in remote closer to the source: eff2520c74fd4db0ac66837d5479965f1b9c40ef.

gino-m:
Thanks @shobhitagarwal1612, RFAL!

shobhitagarwal1612 (approved):
Thanks for the fix!

gino-m:
Working now, thanks for the help! 🚀

shobhitagarwal1612 (approved):

PR #568, opened 2020-08-21
[Bug] Load all features after activating project

### LINKED ISSUE #470: Not all features are loaded on activating project
**To Reproduce**
1. Select project `Plastic Frog Finder`. (This one has a lot of dummy features)

**Expected behavior**
All features points (shown as map pins) should be displayed on the map

**Actual behavior**
Only a subset of points are shown. The count increases gradually whenever the app is restarted.

shobhitagarwal1612:
Fixes #470 

Added detailed explanation about the solution in the commit description.

Steps to verify bug:
1. Login into app with test user
1. Choose 'sample project' from the list of projects
1. All features should be loaded on first attempt

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/converter/ResponseJsonConverter.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/converter/ResponseJsonConverter.java
@@ -67,6 +67,6 @@ class ResponseJsonConverter {
         Timber.e("Error parsing JSONArray in db: %s", jsonArray);
       }
     }
-    return ImmutableList.copyOf(list);
+    return ImmutableList.<String>builder().addAll(list).build();
   }
 }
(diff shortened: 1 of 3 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/local/room/converter/ResponseJsonConverter.java
```diff
@@ -66,6 +67,6 @@ private static Object toJsonArray(MultipleChoiceResponse response) {
         Timber.e("Error parsing JSONArray in db: %s", jsonArray);
       }
     }
-    return list;
+    return ImmutableList.copyOf(list);
```
gino-m, line 70: Consider using ImmutableList.builder() instead of copying the list at the end.

shobhitagarwal1612: Done

gino-m (approved):
Sorry just getting back to this now; LGTM!

PR #569, opened 2020-08-25
Adding UI tests for adding a feature

### LINKED ISSUE #120: [Testing] Add instrumented test for adding a Feature
Espresso tests or similar that run the app and perform key flows. We could either mock out the Firebase API, or perform actual E2E tests against a live instance; for now would prefer whichever gets us some protection against regressions sooner.

Goal is to get one UI test working ("happy path") for adding a feature.

dturner:
Fixes #120 

Adds an instrumented test for adding a feature. Tests are constructed using the [GivenWhenThen style](https://martinfowler.com/bliki/GivenWhenThen.html). For example: 

- **Test name**: Add a feature
- **Given**: a logged in user - with an active project which doesn't direct the user to add an observation when adding a feature.
- **When**: they tap the "Add feature" FAB and choose a feature type.
- **Then**: the observation marker is displayed in the center of the map screen.

**Implementation details**
- [x] Refactor the authentication process so that a `FakeAuthenticationManager` can provide a fake user during tests
- [x] Create a `FakeRemoteDataStore` to provide fake `Project` and `Layer` objects to tests
- [x] Use trampoline as the `io` and `ui` Rx schedulers
- [x] Add [RxIdler](https://github.com/square/RxIdler)
- [x] Add databinding idling resource

**Tests:**
- [x] Verify that tapping on the map does nothing when initially loaded
- [x] Verify that adding a feature adds a map pin at the current crosshair position, and that tapping on that feature loads its details

**Verification steps**
- Run the tests on an emulator or real device

@gino-m @shobhitagarwal1612 @scolsen PTAL.


gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
```diff
@@ -0,0 +1,152 @@
...
+    new ActivityScenarioRule(MainActivity.class);
+
+  @Module
+  @InstallIn(ApplicationComponent.class)
+  abstract class TestDependenciesModule {
+
+    @Binds
+    @Singleton
+    abstract AuthenticationManager bindAuthenticationManager(
+      FakeAuthenticationManager authenticationManager);
```
JoseAlcerreca, line 84: Close parenthesis at the same level as the first character of the line that opens it. Not a hard rule but I think it's the trend now. 

To avoid bikeshedding like this I recommend you install spotless. It's 10 lines in your build.gradle file that will save a lot of time.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
```diff
@@ -0,0 +1,152 @@
...
+  public @Rule(order = 1)
+  SetPreferencesRule preferencesRule = new SetPreferencesRule();
+
+  public @Rule(order = 2)
+  ActivityScenarioRule scenarioRule =
+    new ActivityScenarioRule(MainActivity.class);
+
+  @Module
+  @InstallIn(ApplicationComponent.class)
+  abstract class TestDependenciesModule {
```
JoseAlcerreca, line 79: I don't think this is super idiomatic. I would move these modules to their own files or put them at the bottom of this file. I've opened this file to read the tests, not to read the setup. The only reason you would put a class inside another class is if you need to access the parent from the child.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
```diff
@@ -0,0 +1,152 @@
...
+    LocalDatabase localDatabase(@ApplicationContext Context context) {
+      return Room.inMemoryDatabaseBuilder(context, LocalDatabase.class).build();
+    }
+  }
+
+  // Given: a logged in user - with an active project with no map markers.
+  // When: they tap on the centre of the map.
+  // Then: nothing happens - the feature fragment is not displayed.
+  @Test
+  public void tappingCrosshairOnEmptyMapDoesNothing() throws InterruptedException {
```
JoseAlcerreca, line 115: Weird throws here, is it needed?

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
```diff
@@ -0,0 +1,152 @@
...
+    onData(allOf(is(instanceOf(String.class)), is(FakeData.LAYER_NO_FIELDS_NAME))).perform(click());
+
+    // Tap on the crosshair at the centre of the map.
+    onView(withId(R.id.map_crosshairs)).perform(click());
+
+    // TODO: figure out how to remove this
+    //  It's probably because the feature fragment animation does not block the test and the
+    //  feature title isn't updated until the animation has completed (and it isn't disabled by the
+    //  normal `transition_animation_scale 0` etc commands).
+    Thread.sleep(100);
```
JoseAlcerreca, line 146: Have you tried idling resources? Given you're dealing with maps and fragments it might be tricky

gnd/src/androidTest/java/com/google/android/gnd/CustomTestRunner.java
```diff
@@ -0,0 +1,36 @@
...
+import androidx.test.runner.AndroidJUnitRunner;
+import dagger.hilt.android.testing.HiltTestApplication;
+import timber.log.Timber;
+import timber.log.Timber.DebugTree;
+
+public class CustomTestRunner extends AndroidJUnitRunner {
+
+  @Override
+  public Application newApplication(ClassLoader cl, String className,
+    Context context)
```
JoseAlcerreca, line 30: Was this kind of indentation allowed in Java? D:

gnd/src/androidTest/java/com/google/android/gnd/persistence/remote/FakeRemoteDataStore.java
```diff
@@ -0,0 +1,87 @@
...
+    .putLayer(FakeData.LAYER_NO_FIELDS_ID, layerWithNoFields)
+    .build();
+
+  @Inject
+  FakeRemoteDataStore() {
+  }
+
+  @Override
+  public Single<List<Project>> loadProjectSummaries(
+    User user) {
```
JoseAlcerreca, line 60: same line

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
dturner: We're already using `checkstyle` with the google java style guide, however, there is no rule for this. I tried trawling through the docs to see if it's easy to add, but tbh after 15 mins of researching I didn't think it was worth it.

gnd/src/androidTest/java/com/google/android/gnd/CustomTestRunner.java
dturner: Nope :) All fixed now.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
dturner: I used to have a `Thread.sleep` in that method but it was removed so the `InterruptedException` can also be removed. Good spot, thanks.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
dturner: I think your idea for an `AnimationManager` which switches off animations for testing is probably the best one.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
dturner: @gino-m Any thoughts on this? If it's an easy add it's probably worth adding a rule for this, just for code consistency.

gnd/src/androidTest/java/com/google/android/gnd/AddFeatureTest.java
dturner: In iosched animation is disabled manually for the bottom sheet: https://github.com/google/iosched/blob/89df01ebc19d9a46495baac4690c2ebfa74946dc/mobile/src/main/java/com/google/samples/apps/iosched/widget/BottomSheetBehavior.kt#L209 (thanks @JoseAlcerreca)

dturner: I tried taking the current version of [BottomSheetBehavior](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java#L1312) and modifying it so that animations are disabled. This resulted in the test passing without `Thread.sleep` on my Pixel 4A but still failing on the emulator (tried API 23 and 29). I feel like there's a background thread being spun up somewhere after tapping on the map pin which delays the display of the bottom sheet just long enough for the test to fail.

gino-m:
Hi Don, what's the status on this? Did adding a longer wait in `Thread.sleep()` resolve the flakiness, at least temporarily?

gino-m:
Also, @JoseAlcerreca, is this the best we can do to work around async operations for now, or did we miss something?

dturner:
@gino-m Myself and @JoseAlcerreca did some extensive investigation into this ([PR here](https://github.com/dturner/ground-android/pull/1)). The current conclusion is that whilst we can minimise the flakiness of tests by using idling resources the `Thread.sleep` is unavoidable. 

I will selectively merge in the following code from the other PR:

- Idling resource for databindings
- Idling resource for Rx
- BottomSheetBehavior class which has the capability of animations being disabled for testing

I will also experiment with reducing the sleep time so that we're not waiting unnecessarily for tests to finish.

dturner:
I ended up removing the custom `BottomSheetBehavior` class because with the data binding idling resource the test will wait for the animation to finish. 

I also figured that it was better to rely on the material version of `BottomSheetBehavior` than our own implementation (especially given all the issues we've had with the bottom sheet). 

I'd like to merge this now as I think it provides a stable enough framework for UI testing.

gino-m (approved):
You've done some pretty serious research and a fair amount of refactoring and setup as well to get this to work, and wrote the code in way that even I could understand. Well done, @dturner! Looking forward to a future when we don't require as much boilerplate and explicit sleeps to get this to work; you've done a great job pushing us right up to the cutting edge. Thanks! Over to @shobhitagarwal1612 for one final look.

shobhitagarwal1612 (approved):

shobhitagarwal1612:
🎉 🎉 🎉

PR #574, opened 2020-09-01
Rename Tile -> TileSource

### LINKED ISSUE #400: [Code health] Rename "Tile" class to "TileSource"
Since each "Tile" instance actually represents a way of fetching multiple tiles for an extent.

scolsen:
TileSource better captures what this object represents--some URL and eventually file path that points to a file that contains tileset data that we can render on the map as an overlay. The old name, Tile, suggests that the object represents a single tile, at a single zoom level, while the tileset files we use actually are pyramidal and contain tiles at multiple zoom levels.

@gino-m @shobhitagarwal1612 

Fixes #400.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/sync/TileSourceDownloadWorker.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/sync/TileSourceDownloadWorker.java
@@ -23,8 +23,8 @@ import androidx.hilt.work.WorkerInject;
 import androidx.work.Data;
 import androidx.work.WorkerParameters;
 import com.google.android.gnd.R;
-import com.google.android.gnd.model.basemap.tile.Tile;
-import com.google.android.gnd.model.basemap.tile.Tile.State;
+import com.google.android.gnd.model.basemap.tile.TileSource;
+import com.google.android.gnd.model.basemap.tile.TileSource.State;
 import com.google.android.gnd.persistence.local.LocalDataStore;
 import com.google.android.gnd.persistence.remote.TransferProgress;
 import com.google.android.gnd.system.NotificationManager;
@@ -46,19 +46,19 @@ import timber.log.Timber;
  * provided in a {@link Data} object. This worker should only run when the device has a network
  * connection.
  */
-public class TileDownloadWorker extends BaseWorker {
+public class TileSourceDownloadWorker extends BaseWorker {
   private static final int BUFFER_SIZE = 4096;
 
   private final Context context;
   private final LocalDataStore localDataStore;
 
   @WorkerInject
-  public TileDownloadWorker(
+  public TileSourceDownloadWorker(
       @Assisted @NonNull Context context,
       @Assisted @NonNull WorkerParameters params,
       LocalDataStore localDataStore,
       NotificationManager notificationManager) {
-    super(context, params, notificationManager, TileDownloadWorker.class.hashCode());
+    super(context, params, notificationManager, TileSourceDownloadWorker.class.hashCode());
     this.context = context;
     this.localDataStore = localDataStore;
   }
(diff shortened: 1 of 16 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/sync/TileSourceDownloadWorker.java
```diff
@@ -170,7 +172,7 @@ private Completable processTiles(ImmutableList<Tile> pendingTiles) {
                 case IN_PROGRESS:
                 case FAILED:
                 default:
-                  return downloadTile(t);
+                  return downloadTileSource(t);
```
gino-m, line 175: Naming nit: You download tiles from a tile source, so maybe this would make more sense as "downloadTiles"?

gnd/src/main/java/com/google/android/gnd/persistence/sync/TileSourceDownloadWorker.java
```diff
@@ -209,8 +212,8 @@ public String getNotificationTitle() {
     return getApplicationContext().getString(R.string.downloading_tiles);
   }
 
-  static class TileDownloadException extends RuntimeException {
-    TileDownloadException(String msg, Throwable e) {
+  static class TileSourceDownloadException extends RuntimeException {
```
gino-m, line 215: Same comment here - the issue isn't that tiles couldn't be download _from_ the tile source, so I'd still expect this to be a `TileDownloadException`.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -47,7 +47,7 @@
 import timber.log.Timber;
 
 public class OfflineAreaRepository {
-  private final TileDownloadWorkManager tileDownloadWorkManager;
+  private final TileSourceDownloadWorkManager tileSourceDownloadWorkManager;
```
gino-m, line 50: Same here.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -94,36 +94,36 @@ private File downloadOfflineBaseMapSource(OfflineBaseMapSource offlineBaseMapSou
     return localFile;
   }
 
-  /** Enqueue a single area and its tiles for download. */
-  private Completable enqueueDownload(OfflineArea area, ImmutableList<Tile> tiles) {
+  /** Enqueue a single area and its tile sources for download. */
```
gino-m, line 97: Same - we download tiles, not tile sources.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -164,25 +164,26 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
-        // If no tiles are found, we report the area takes up 0.0mb on the device.
-        .doOnError(throwable -> Timber.d(throwable, "no tiles found for area %s", offlineArea))
+        // If no tile sources are found, we report the area takes up 0.0mb on the device.
+        .doOnError(
+            throwable -> Timber.d(throwable, "no tile sources found for area %s", offlineArea))
         .onErrorReturn(__ -> ImmutableSet.of());
   }
 
-  public Flowable<ImmutableSet<Tile>> getDownloadedTilesOnceAndStream() {
+  public Flowable<ImmutableSet<TileSource>> getDownloadedTileSourcesOnceAndStream() {
```
gino-m, line 180: This case is a little be trickier - it's not clear how to describe that we're "getting all tile sources for which all tiles have been downloaded", so I think it's ok as-is.

gnd/src/main/java/com/google/android/gnd/persistence/sync/TileSourceDownloadWorker.java
scolsen: Fair point. I guess the "Tile > TileSet" renaming issue isn't really a duplication then.

One could argue that we are in fact downloading the tile source--we're simply copying the file specified at a given URL (the tile source). We don't really know anything about individual tiles or care about them until we get down to the tile provider level, which looks at tiles specified in a tile source.

But I also don't want to argue semantics--if it's clearer to use "Tiles" or "TileSet" here I think that's fine--technically speaking though, if we are calling the thing specified in the geoJSON a tile source, that is what we are downloading.

gino-m: Ahhh... sorry, at first glance I may have though this was actually downloading the tiles themselves. Please disregard, this makes perfect sense. If we really wanted to future-proof our semantics, we could define "tile source" to be an arbitrary way of getting tiles, and this case as an "mbtiles tile source's geojson extents file" but we can cross that bridge if we ever get to it.

PR #575, opened 2020-09-03
Hide soft keyboard when saving form

### LINKED ISSUE #553: After adding an observation the map is fully zoomed in and it's not possible to zoom out
After adding an observation the map is fully zoomed in and it's not possible to zoom out. 

**Steps to reproduce**
- Load the app and tap on the "Add feature" FAB
- Choose the feature type
- Complete the fields for the observation and tap Save

**Expected behavior**
- The feature marker to appear on the map and for it to be possible to zoom in/out using pinch 

**Actual behavior**
- The feature marker doesn't appear and the map appears to be fully zoomed in. It's also not possible to zoom out. 

**Screenshots**
Here's the view I see after adding a feature and observation
image

Tested on Pixel 2 running Android 11 beta 2.5

### LINKED ISSUE #556: [Add/edit observation] Blue screen after returning to feature sheet from add/edit observation
1. Add/edit an observation
2. Click "Back", or save observation

Result:
Screenshot_1596492154

Expected result:
* Return to the feature sheet

shobhitagarwal1612:
Fixes #556, Fixes #553 

Steps to verify:
1. Select any project and add a new feature
2. On "edit observation" screen, tap on text field to show the soft keyboard (input)
3. Click "save"
4. Blue screen shouldn't appear

Trying the above steps on master branch results in a blue screen.

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
@@ -20,10 +20,12 @@ import static com.google.android.gnd.ui.editobservation.AddPhotoDialogAdapter.Ph
 import static com.google.android.gnd.ui.editobservation.AddPhotoDialogAdapter.PhotoStorageResource.PHOTO_SOURCE_STORAGE;
 
 import android.app.AlertDialog;
+import android.content.Context;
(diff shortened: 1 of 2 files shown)


gino-m (approved):
Thanks for the fix! 🥇

PR #580, opened 2020-09-22
[Offline base maps] Add support for deleting areas

### LINKED ISSUE #573: Downloaded Areas have a size of 0.0mb even when files are present on device
## Description
The offline area list should accurately report the amount of space consumed by offline area tiles. Currently, it reports 0.0mb for all tiles.

**To Reproduce**
1. Open Ground
2. Click on 'Offline Areas'
3. Click the plus icon to add a new area
4. Click download to download imagery for an area
5. After the download begins, select the area name from the areas list
6. Area size is constantly reported as 0.0mb.

## Expected Behavior
Area sizes should be equal to the amount of memory consumed by downloaded tile files.

## Actual Behavior
Area sizes are always reported as 0.0mb

scolsen:
**EDITED:** We use reference counting to remove tile sources from the device when offline areas are deleted.

It also updates the offline area viewer to trigger deletions upon clicking the remove button.

I verified this behavior using the device file explorer and stetho--everything seems to be removed correctly.

Also fixes #573


scolsen:
@gino-m, @shobhitagarwal1612, @dturner, this is ready for a review! If anyone has the bandwidth

gnd/src/main/java/com/google/android/gnd/persistence/local/room/dao/OfflineAreaTileSourceCrossRefDao.java
```diff
@@ -1,9 +1,25 @@
+/*
+ * Copyright 2019 Google LLC
```
gino-m, line 2: =>2020

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -187,7 +188,7 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
                     .collect(toImmutableSet()));
   }
 
-  //** Delete an offline area and the unique tile sources associated with it. */
+  // ** Delete an offline area and the unique tile sources associated with it. */
```
gino-m, line 191: Nit: Extra '/ ' here

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -583,4 +604,50 @@ public Completable insertOrUpdateOfflineArea(OfflineArea area) {
...
+                        tileSourceDao
+                            .delete(tile)
+                            .andThen(
+                                offlineAreaTileSourceCrossRefDao.delete(
+                                    OfflineAreaTileSourceCrossRef.builder()
+                                        .setOfflineAreaId(offlineAreaEntity.getId())
+                                        .setTileSourcePath(tile.getPath())
+                                        .build()))))
+        .doOnComplete(() -> Timber.d("Removed tiles"))
+        .doOnError(throwable -> Timber.e(throwable, "Failed to delete tiles"));
```
gino-m, line 638: Imo logging should be done in the subscriber (where possible) to avoid duplicate logging messages - lmk what you think!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -583,4 +604,50 @@ public Completable insertOrUpdateOfflineArea(OfflineArea area) {
         .toSingle()
         .subscribeOn(schedulers.io());
   }
+
+  private Completable deleteTiles(OfflineAreaEntity offlineAreaEntity) {
+    return tileSourceDao
```
gino-m, line 609: Is there any way to split this stream up to make it more readable and self documenting?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -583,4 +604,50 @@ public Completable insertOrUpdateOfflineArea(OfflineArea area) {
...
+                stream(sources)
+                    .filter(
+                        source ->
+                            // We only remove tiles when the given area is the only one that depends
+                            // on them.
+                            source.offlineAreaEntities.contains(offlineAreaEntity)
+                                && source.offlineAreaEntities.size() == 1)
+                    .map(source -> source.tileSourceEntity)
+                    .collect(toImmutableList()))
+        .flatMapObservable(Observable::fromIterable)
```
gino-m, line 623: Iirc there are some iterable Rx operations, like `flatMapIterable`. Do any of those help here?

gino-m (changes requested):
Added a few nits, but zooming out a sec:

Could finding non-overlapping areas be achieved by using either reference counting? i.e., increment a counter when a region is added, then decrement when the region is deleted, deleting tiles with count of zero? It should reduce the amount of code needed to maintain data consistency. Wdyt?

scolsen:
> Added a few nits, but zooming out a sec:
> 
> Could finding non-overlapping areas be achieved by using either reference counting? i.e., increment a counter when a region is added, then decrement when the region is deleted, deleting tiles with count of zero? It should reduce the amount of code needed to maintain data consistency. Wdyt?

That would be a LOT simpler and I think it would work just fine. Thanks, Gino! I can revert and make this change instead.

gino-m:
> > Added a few nits, but zooming out a sec:
> > Could finding non-overlapping areas be achieved by using either reference counting? i.e., increment a counter when a region is added, then decrement when the region is deleted, deleting tiles with count of zero? It should reduce the amount of code needed to maintain data consistency. Wdyt?
> 
> That would be a LOT simpler and I think it would work just fine. Thanks, Gino! I can revert and make this change instead.

@scolsen I had a thought while falling asleep last night (🤔 ) - alternatively, could we just load all the tiles and tilesets into memory and identify which ones need to be deleted? There would rarely be more than a few hundred tiles imo.. wdyt?

scolsen:
> > > Added a few nits, but zooming out a sec:
> > > Could finding non-overlapping areas be achieved by using either reference counting? i.e., increment a counter when a region is added, then decrement when the region is deleted, deleting tiles with count of zero? It should reduce the amount of code needed to maintain data consistency. Wdyt?
> > 
> > 
> > That would be a LOT simpler and I think it would work just fine. Thanks, Gino! I can revert and make this change instead.
> 
> @scolsen I had a thought while falling asleep last night (🤔 ) - alternatively, could we just load all the tiles and tilesets into memory and identify which ones need to be deleted? There would rarely be more than a few hundred tiles imo.. wdyt?

I think that's valid too, but I do think reference counting will be simpler. W/ reference counting, the approach is roughly:

```
delete(area) -> getIntersectingTiles(area) -> filter(refCount <= 1) -> deleteTiles(filteredTiles)
```

Whereas I think the other approach requires:

```
delete(area) -> getAllAreas() -> map(lambdaArea : getIntersectingTiles(area)) {we now have a list of lists of tiles} -> flatten() -> filter(singleInstance {filter out any tiles that appear n > 1}) -> filter(tile isMember of getIntersectingTiles(area))
```

which I think will be more complicated. Unless there's a really strong reason for us to avoid adding another field in the DB I'd prefer the reference counting approach for its simplicity. Prior to this PR, there is no relationship between tiles+areas encoded anywhere, it's completely determined by the `getIntersectingTiles() :: Area -> [Tile]` function. I think because of that the in-memory approach is trickier to get right and possibly more brittle if we wind up changing the way we handle certain things.

gino-m:
> which I think will be more complicated. Unless there's a really strong reason for us to avoid adding another field in the DB I'd prefer the reference counting approach for its simplicity. Prior to this PR, there is no relationship between tiles+areas encoded anywhere, it's completely determined by the `getIntersectingTiles() :: Area -> [Tile]` function. I think because of that the in-memory approach is trickier to get right and possibly more brittle if we wind up changing the way we handle certain things.

I agree. It also obviously scales better; this may be relevant if we end up with a dataset with many smaller titles (e.g., very high-res). Reference-counting SGTM!

scolsen:
@gino-m sorry for the delay--ready for another look! This also fixes #573, so two birds one stone on this one.

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
```diff
@@ -22,6 +22,11 @@
 @AutoValue
 public abstract class TileSource {
 
+  //** Increment the area reference count of a tile source by one. */
```
gino-m, line 25: Nit: Extra leading `/`

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
```diff
@@ -22,6 +22,11 @@
...
+  public static TileSource incAreaCount(TileSource tileSource) {
```
gino-m, line 26: Style nit: Please use full name "incrementAreaCount()"

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
```diff
@@ -41,6 +40,7 @@
 
   @CopyAnnotations
   @NonNull
+  @PrimaryKey
```
gino-m, line 43: It's generally safer to use artificial ids (`id`) as PK, and to declare other unique columns as unique if necessary. If there's no specific reason for making this PK, can we move it back to `id`?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
```diff
@@ -117,6 +131,8 @@ public static Builder builder() {
 
     public abstract Builder setState(TileEntityState newState);
 
+    public abstract Builder setAreaCount(Integer areaCount);
```
gino-m, line 134: If area count can never be null please use `int` instead.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -96,12 +97,21 @@ private File downloadOfflineBaseMapSource(OfflineBaseMapSource offlineBaseMapSou
...
   /** Enqueue a single area and its tile sources for download. */
   private Completable enqueueDownload(OfflineArea area, ImmutableList<TileSource> tileSources) {
+
     return localDataStore
         .insertOrUpdateOfflineArea(area.toBuilder().setState(State.IN_PROGRESS).build())
         .andThen(
             Completable.merge(
                 stream(tileSources.asList())
-                    .map(localDataStore::insertOrUpdateTileSource)
+                    .map(
```
gino-m, line 106: My intuition is that there's a simpler way of doing this using RxJava fromIterable and recombining the result (`combine()`?) but would take some tinkering to figure it out. If you don't want to explore that now please add a TODO to simplifying this block.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -170,13 +180,28 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
         .flatMapPublisher(
             tiles ->
                 getDownloadedTileSourcesOnceAndStream()
-                    .map(ts -> stream(ts).filter(tiles::contains).collect(toImmutableSet())))
+                    .map(
```
gino-m, line 183: Any way we can reduce nesting here? Maybe just refactor part of this into its own method?

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -186,4 +211,27 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
                     .filter(tileSource -> tileSource.getState() == TileSource.State.DOWNLOADED)
                     .collect(toImmutableSet()));
   }
+
+  /** Delete an offline area and the unique tile sources associated with it. */
```
gino-m, line 215: A better explanation might be "Delete an offline area, including any tile sources that don't overlap with other existing offline areas."

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -186,4 +211,27 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
+  public Completable deleteArea(String offlineAreaId) {
+    return localDataStore
+        .getOfflineAreaById(offlineAreaId)
+        .flatMapMaybe(this::getIntersectingDownloadedTileSourcesOnce)
+        .flatMapCompletable(
```
gino-m, line 220: Same here.. Any way to make this more readable?

gino-m (changes requested):
There are a few merge conflicts that now need to be resolved. Also added a few readability/style nits, overall looks great!

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
scolsen: e542aae

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
scolsen: e542aae

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
scolsen: Sorry, I was under the impression we'd need to use this as the PK if we want `insertOrUpdate` to work correctly and replace tiles with incremented area counts--will setting URL to Unique achieve the same?

scolsen: for reference 

> The implementation of the method will update its parameters in the database if they already exists (checked by primary keys). If they don't already exists, this option will not change the database.

https://developer.android.com/reference/androidx/room/Update

scolsen: @gino-m (just to ensure it winds up in your notifications! ...I can never tell w/ github)

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
scolsen: 048dd80

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
scolsen: b86a87a

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
scolsen: b86a87a

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
scolsen: b86a87a

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
scolsen: b86a87a

scolsen:
@gino-m ready for another looksee!

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
gino-m: I think what you want here is the following in the DAO:
@Query("UPDATE tile_source SET reference_count=:newCount WHERE id = :id")
void updateReferenceCount(String id, int newCount);

You probably could also do the increment directly in SQL, but that might be overkill here.
Note that we've been following the convention that the table or entity is named after what each row describes, in this case "tile_source" instead of "tile_sources'.

gino-m (changes requested):
Thanks for the edits. Reply inline.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
scolsen: We can do that, but we'd have to change our tile source retrieval logic. At the moment it roughly looks like this:

1. Fetch the basemap JSON from the URL specified in the project
2. Use the JSON parser to parse tile sources that intersect with the selected area
3. Construct a tile source for each source
4. Download tiles sources and add them to the DB.

Step 3. is the problematic one. We generate a new UID every time we determine intersecting tiles, so, if a user adds an area that overlaps with an existing area, we can't rely on UIDs to update the reference counts of existing tiles, since these are uniquely determined at JSON parsing time. Thus assuming tile source A exists in the db and has and equivalent URL and extent as tile B (newly parsed), the UIDs for each of them will still be unique thus any updates to the DB that rely on a UID primary key will fail and instead fallback to inserting a new entry (for the same. tile). 

Hopefully that explanation makes sense, we can still keep UIDs as the primary key if that's better, we'd just need to add an additional DB call at JSON parsing time to eliminate disregard parsed tiles whose URLs already collide w/ those in the db (since this indicates the reference count should be incremented). This also puts us into a a weird situation in which it more than likely makes the most sense to perform count incrementation in the JSON parser...which seems to be a bad crossing of boundaries imo.

Hopefully this explanation makes sense. Happy to walk through the code as well if its easier.

gino-m:
@scolsen Gentle ping! :)

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/TileSourceEntity.java
gino-m: Hi @scolsen sorry I didn't see this before - I think we resolved this in our VCs, please cmiiw!

scolsen:
@gino-m finally fixed, sorry for the delay!

gino-m (changes requested):
One tweak, otherwise LGTM!

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
```diff
@@ -22,6 +22,11 @@
 @AutoValue
 public abstract class TileSource {
 
+  /** Increment the area reference count of a tile source by one. */
+  public static TileSource incrementAreaCount(TileSource tileSource) {
```
gino-m, line 26: Since this operates on a tile source, can you make this non-static and modify `this` instead?

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
```diff
@@ -48,6 +48,32 @@
...
+      // TODO: Separate parsing and intersection checks, make asyc (single, completable).
+      JSONObject geoJson = new JSONObject(fileContents);
+      // TODO: Make features constant.
+      JSONArray features = geoJson.getJSONArray(FEATURES_KEY);
+
+      return stream(toArrayList(features))
+          .map(GeoJsonTile::new)
+          .filter(tile -> tile.boundsIntersect(bounds))
+          .map(this::jsonToTileSource)
+          .map(tileSource -> tileSource.toBuilder().setAreaCount(1).build())
```
gino-m, line 67: then here you could just call `TileSource::incrementAreaCount`

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
```diff
@@ -37,6 +42,8 @@
 
   public abstract State getState();
 
+  public abstract int getAreaCount();
```
gino-m, line 45: Nit: Calling it "reference count" might make the purpose of this field more obvious.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -164,6 +164,18 @@
   /** Returns all queued, failed, and completed offline areas from the local data store. */
   Flowable<ImmutableList<OfflineBaseMap>> getOfflineAreasOnceAndStream();
 
+  /** Delete an offline area and its *unique* tiles. */
+  Completable deleteOfflineArea(String offlineAreaId);
```
gino-m, line 168: " Delete an offline area and any tiles that are no longer needed." might be clearer here.

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -164,6 +164,18 @@
...
+
   /** Returns the offline area with the specified id. */
   Single<OfflineBaseMap> getOfflineAreaById(String id);
+
+  /**
+   * Update the area count of an existing tile source in the local data store with the area count of
+   * {@param tilesource}.
+   */
+  Single<TileSource> updateTileSourceAreaCountByUrl(TileSource tilesource);
```
gino-m, line 177: I'd expect this to accept the URL and new count; this makes what's happening more obvious and reduces the surface area of the internal API.

Also, returning a `Completable` here may be more correct.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -585,4 +586,32 @@ public Completable insertOrUpdateOfflineArea(OfflineBaseMap area) {
...
         .subscribeOn(schedulers.io());
   }
+
+  @Override
+  public Completable deleteOfflineArea(String id) {
+    return offlineBaseMapDao
+        .findById(id)
+        .toSingle()
+        .doOnSubscribe(__ -> Timber.d("Deleting offline area: %s", id))
+        .flatMapCompletable(offlineBaseMapDao::delete)
```
gino-m, line 596: Why do we need to find the area in order to delete it? We should be able to just deleted by id, no?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -585,4 +586,32 @@ public Completable insertOrUpdateOfflineArea(OfflineBaseMap area) {
...
+        .subscribeOn(schedulers.io());
+  }
+
+  @Override
+  public Single<TileSource> updateTileSourceAreaCountByUrl(TileSource tileSource) {
+    return tileSourceDao
+        .updateAreaCount(tileSource.getAreaCount(), tileSource.getUrl())
+        .map(__ -> tileSource);
```
gino-m, line 604: Can we return a `Completable` here instead?

gnd/src/main/java/com/google/android/gnd/ui/util/FileUtil.java
```diff
@@ -84,4 +84,14 @@ public File getFileFromRawResource(@RawRes int resourceId, String filename) thro
     }
     return file;
   }
+
+  /** Attempts to delete a file relative to the app's file directory when it exists. */
+  public void deleteFile(String filename) {
```
gino-m, line 89: Naming nit: ->`deleteFileIfExists`

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerFragment.java
```diff
@@ -109,6 +112,11 @@ public void onRemoveClick() {
       return;
     }
 
-    viewModel.onRemoveClick();
+    Timber.d("Removing offline area %s", viewModel.getOfflineArea());
+    viewModel
+        .onRemoveClick()
+        .doOnComplete(() -> navigator.navigateUp())
```
gino-m, line 118: Please move this logic into subscribe() instead of side effect.

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
```diff
@@ -186,4 +212,23 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
+
+  /**
+   * Delete an offline base map and any tile sources associated with it that do not overlap with
+   * other offline base maps .
+   */
+  public Completable deleteArea(String offlineAreaId) {
+    return localDataStore
+        .getOfflineAreaById(offlineAreaId)
+        .flatMapMaybe(this::getIntersectingDownloadedTileSourcesOnce)
+        .flatMapPublisher(Flowable::fromIterable)
```
gino-m, line 224: Can we avoid casting to Flowable here to simplify a bit?

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
```diff
@@ -186,4 +212,23 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
+        .map(
+            tileSource ->
+                tileSource.toBuilder().setAreaCount(tileSource.getAreaCount() - 1).build())
```
gino-m, line 227: It might be nice to have a `tileSource.decrementTileCount()` called here instead (better encapsulated and for consistency with increment..

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
```diff
@@ -186,4 +212,23 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
+        .flatMapCompletable(
```
gino-m, line 228: Also, if the previous stream is a `Completable`, here we could just use `andThen`.

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
```diff
@@ -186,4 +212,23 @@ public Completable addAreaAndEnqueue(LatLngBounds bounds) {
...
+            tile ->
+                Completable.fromSingle(localDataStore.updateTileSourceAreaCountByUrl(tile))
+                    .andThen(localDataStore.deleteTileByUrl(tile)))
```
gino-m, line 231: Same here.

gino-m (changes requested):
A few readability suggestions, otherwise looks great! Note that build was failing on last CI run, but could have been do to the previous bad merge.

scolsen:
> A few readability suggestions, otherwise looks great! Note that build was failing on last CI run, but could have been do to the previous bad merge.

Yep, defining a function twice (the bad merge) is a compile time failure.

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
scolsen: I actually find `referenceCount` more ambiguous--references to what, exactly?

Maybe `associatedBasemapCount` or `basemapReferenceCount`?

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
scolsen: Unless I'm missing something we'll either need to cast the initial iterable into a flowable or flatten a list of resulting  Completables using `merge`. I find converting an iterable to a stream of items and mapping over the stream to be more idiomatic and simpler to understand. For comparison, here's the current approach:

```java
.getOfflineAreaById(offlineAreaId)
        .flatMapMaybe(this::getIntersectingDownloadedTileSourcesOnce)
        .flatMapPublisher(Flowable::fromIterable) // convert List<TileSource> to a stream of <TileSource>
        .map(TileSource::decrementAreaCount) // from here on, just use straightforward RX stream operations to manipulate each item
        .flatMapCompletable(
            tile ->
                localDataStore
                    .updateTileSourceAreaCountByUrl(tile.getAreaCount(), tile.getUrl())
                    .andThen(localDataStore.deleteTileByUrl(tile)))
        .andThen(localDataStore.deleteOfflineArea(offlineAreaId));
```

versus the alternative explicit iteration over the iterable:

```java
Completable.merge(
  localdatastore.getOfflineAreaById(offlineAreaId)
        .flatMapMaybe(this::getIntersectingDownloadedTileSourcesOnce)
        .map(tileSources -> 
                    stream(tileSources).map( // we could have just converted this into an RX stream but instead we use a Java 8 stream...within a stream
                       tile ->
                          localDataStore
                          .updateTileSourceAreaCountByUrl(tile.getAreaCount(), tile.getUrl())
                          .andThen(localDataStore.deleteTileByUrl(tile))).collect(toImmutableList))) // list of Completable
        .andThen(localDataStore.deleteOfflineArea(offlineAreaId));
```

or something similar. I think the first, simply casting an iterable into terms of RX java streams has a few advantages:

- eliminate stream util code (since rx java operates on streams and provides conversions from iterables into streams, isn't somewhat silly for us to stream(listofThings).map...etc. within a stream, we're effectively using two different but equivalent stream languages (java 8 and RX) at different levels)?
- eliminate recollection and flattenting
- operate on a stream of values rather than a pseudo stream of a single value which is itself a streamable collection.

I think the interior stream approach makes sense when we want to continue talking about a collection of values in the RX stream (for instance, if we were collecting these after transforming them into ints) but when the final value is something like a Completable I think it makes more sense just to up-level the items to the RX stream level.

That said I'm not sure what the relative (if any) performance trade offs are between each approach, so this is mostly a stylistic preference. I just think thinking in terms of streams of tiles at a single level (RX) is simpler than having to think in terms of streams at two levels (the RX stream *and* the stream of tiles within that stream) and have to flatten an iterable of stream/observable types (which arguably I think you should only be doing when you have some more meaningful combination of multiple observables, such as observables that emit different items that need to be combined)

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
scolsen: done

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
scolsen: done

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
scolsen: Done, thanks!

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
scolsen: done, thanks!

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
scolsen: Done, nice edit, thanks!

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
scolsen: Changed to `basemapReferenceCount`

gnd/src/main/java/com/google/android/gnd/persistence/geojson/GeoJsonParser.java
scolsen: done, thanks!

gnd/src/main/java/com/google/android/gnd/model/basemap/tile/TileSource.java
scolsen: Done, thanks!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerFragment.java
scolsen: done!

gnd/src/main/java/com/google/android/gnd/repository/OfflineBaseMapRepository.java
scolsen: @gino-m

gino-m: Sorry, I should have been more clear - I think unpacking the collection makes sense, I was wondering if we can use `Observable::fromIterable` here instead of `Flowable::fromIterable`

scolsen: oh! Indeed we can! And it's better since we've no real need for backpressure here.

scolsen: 364195f

scolsen:
@gino-m one more check?!

gino-m (approved):
Super readable, thanks a lot for the careful edits. 🚢 !

PR #581, opened 2020-10-12
[Feature] Relocate point features

### LINKED ISSUE #109: [Edit feature] Allow users to reposition point features
Allow feature creators and project managers to move point features around on the map after they've been added.

The related sync and merge feature mutation functionality will also need to be implemented.

shobhitagarwal1612:
Fixes #109

Add UI flow for updating position of features

@gino-m  PTAL?


gnd/src/main/res/values/strings.xml
```diff
@@ -159,4 +159,6 @@
   <string name="offline_area_download_started">Download started.</string>
   <string name="offline_area_download_failed">Couldn\'t download area.</string>
   <string name="added_by">Added by %s</string>
+  <string name="reposition_location">Do you want to reposition to this location?</string>
```
gino-m, line 162: So far we've been using "point" elsewhere, and you technically can't reposition a location, so let's stick to "point".

Also, I think "Move point" might be more easily understood and less technical-sounding than "Reposition". Sorry I didn't mention this earlier.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -63,14 +65,18 @@
...
   private final MutableLiveData<Event<Nil>> showMapTypeSelectorRequests = new MutableLiveData<>();
   private final LiveData<ImmutableSet<String>> mbtilesFilePaths;
+
   // TODO: Create our own wrapper/interface for MbTiles providers
   // The impl we're using unfortunately requires calling a `close` method explicitly
   // to clean up provider resources; `close` however, is not defined by the `TileProvider`
   // interface, preventing us from treating providers generically.
   private final List<MapBoxOfflineTileProvider> tileProviders = new ArrayList<>();
 
+  @Nullable private Feature feature;
```
gino-m, line 78: Is this the currently selected feature? If so we should call it as such to be clear.

Also, since the selected feature is accessible via public methods, let's use `Optional<>` instead of `nullable` to expose that API.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -131,6 +135,32 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
     mapContainerViewModel
         .getShowMapTypeSelectorRequests()
         .observe(getViewLifecycleOwner(), __ -> showMapTypeSelectorDialog());
+    mapContainerViewModel.getViewMode().observe(getViewLifecycleOwner(), this::updateUI);
+  }
+
+  private void updateUI(Mode mode) {
+    switch (mode) {
+      case DEFAULT:
+        binding.mapTypeBtn.setVisibility(View.VISIBLE);
```
gino-m, line 144: Rather than setting the visibility or other dynamic states of view explicitly, we've started moving towards using data bindings on ViewModel. Would it be possible to expose the visibility of these elements in the corresponding ViewModel and binding to them in the layout XML?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -95,6 +101,27 @@
...
+    map.put(feature, point);
+    updateFeatureRequests.onNext(map);
+  }
+
+  public LiveData<Boolean> getUpdateFeature() {
+    return updateFeature;
+  }
+
+  private Completable updateFeaturePosition(Map<Feature, Point> map) {
+    Feature feature = map.keySet().iterator().next();
```
gino-m, line 122: Why do we only take the first feature in the map here?

gino-m (changes requested):
A few initial comments. Looks great so far!

shobhitagarwal1612:
@gino-m Thanks for the comments. Back to your queue.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -95,6 +102,27 @@
...
                 __ -> deleteActiveFeature().toSingleDefault(true).onErrorReturnItem(false)));
+
+    updateFeature =
+        LiveDataReactiveStreams.fromPublisher(
+            updateFeatureRequests.switchMapSingle(
+                map -> updateFeaturePosition(map).toSingleDefault(true).onErrorReturnItem(false)));
+  }
+
+  public void updateFeature(Feature feature, Point point) {
+    HashMap<Feature, Point> map = new HashMap<>();
```
gino-m, line 113: Please use `ImmutableMap` to ensure it can't be modified while in the stream.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -95,6 +102,27 @@
...
+    map.put(feature, point);
+    updateFeatureRequests.onNext(map);
+  }
+
+  public LiveData<Boolean> getUpdateFeature() {
+    return updateFeature;
+  }
+
+  private Completable updateFeaturePosition(Map<Feature, Point> map) {
```
gino-m, line 122: Why is this a Map and not just two args? If it's because you wanted to avoid a custom class to bind Feature and Point, you can just use the Feature with the updated position.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -147,12 +152,40 @@ private void onMapReady(MapAdapter map) {
...
+
+  private void moveToNewPosition(Point point) {
+    mapContainerViewModel
+        .getSelectedFeature()
+        .ifPresentOrElse(
+            feature -> {
+              homeScreenViewModel.updateFeature(feature, point);
+              setDefaultMode();
+            },
+            () -> Timber.e("Feature is null, can't move to new position"));
```
gino-m, line 182: A better error might be "No feature selected", since it described _why_ we might be in this state.

gino-m (changes requested):
Some more questions/suggestions.

+dturner could you also PTAL?

shobhitagarwal1612:
@gino-m PTAL

gino-m (approved):
LGTM! 🚀

PR #585, opened 2020-10-13
Rename OfflineArea -> OfflineBaseMap

### LINKED ISSUE #442: [Offline areas] Rename to Offline satellite imagery?
@scolsen, looking at the app I had some second thoughts about the term "Offline areas", so putting this out there for discussion.

I originally suggested this name because I thought we would also use the offline areas to selectively cache features and other data, but in hindsight, I probably really should have been more aggressively applying YAGNI. 

The main reason we may want to reconsider this is that it doesn't seem immediately obvious that "Offline areas" is where you go to download satellite imagery, but I could be wrong.

@coreyleamon, @dturner, @shobhitagarwal1612, any thoughts?

scolsen:
This commit renames all instances of "OfflineArea" to "OfflineBaseMap", in fragment, model, persistence classes etc. Additionally, all user facing strings and layouts have been updated to use the term "offline base map" instead of "offline area".

Fixes #442 

@gino-m @shobhitagarwal1612 @dturner


gino-m:
That's odd.. looking into it now.

gino-m (dismissed):
LGTM! Thanks for the change!

gino-m:
An arbitrary change also caused the same error. Either one of our CI deps changed, or the cache is corrupt. I'll try to clear the cache now.

gino-m:
Looks like `master` failed after version upgrades in #584. I wonder why it passed checks but then broke master..

gino-m:
Unfortunately rebuilding the image failed as well. Filed #588 to resolve this.

shobhitagarwal1612 (dismissed):

scolsen:
Ah, looks like the failure now is from some checkstyle errors, will fix!

gino-m (approved):
Thanks for your patience, and thanks @shobhitagarwal1612 for fixing the CI checks!

PR #589, opened 2020-10-15
Fix build

### LINKED ISSUE #588: [CI] Builds failing with 'repository 'tmp' does not exist'
On GCB #585 was failing with:
```
Already have image (with digest): gcr.io/cloud-builders/git
Cloning into 'tmp'...
mv: cannot move 'tmp/.git' to './.git': Directory not empty
```

I tried pushing a no-op change and received the same error in #587.

I deleted all caches in Cloud Storage, and now builds fail with:
```
Starting Step #0 - "Copying gradle build cache tar from GCS"
Starting Step #2 - "Copy .git directory for versioning"
Step #2 - "Copy .git directory for versioning": Already have image (with digest): gcr.io/cloud-builders/git
Step #0 - "Copying gradle build cache tar from GCS": Already have image (with digest): gcr.io/cloud-builders/gsutil
Step #2 - "Copy .git directory for versioning": fatal: repository 'tmp' does not exist
Finished Step #2 - "Copy .git directory for versioning"
ERROR
ERROR: build step 2 "gcr.io/cloud-builders/git" failed: step exited with non-zero status: 128
```

I tried rebuilding the image with `gcloud builds submit --config=cloudbuild.yaml --substitutions=_ANDROID_VERSION=28` and received the same error.

@shobhitagarwal1612 could you kindly take a look or point me in the right direction? Could this be related to #584?

shobhitagarwal1612:
Fixes #588 

.git directory already exists so no need to pull one manually. I also had to re-create the cloud buckets and republish the docker images. Maybe GCB changed its policies or there was an update we were unaware of.

@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -18,14 +18,13 @@
 #
 #  1. Copy Gradle cache
 #  2. Extract tarball
-#  3. Copy .git dir
-#  4. Build debug APK
-#  5. Run code quality checks (suppress failure if check fails)
-#  6. Run unit tests (suppress failure if tests fail)
-#  7. Save reports to GCS
-#  8. Update badge status and fail the build if any failures were suppressed in previous steps
-#  9. Compress Gradle files
-#  10. Save tarball
+#  3. Build debug APK
+#  4. Run code quality checks (suppress failure if check fails)
+#  5. Run unit tests (suppress failure if tests fail)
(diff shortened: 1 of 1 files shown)


scolsen (approved):
Looks like this did the trick, thank you, Shobhit!

PR #590, opened 2020-10-15
Revert "Fix build"

shobhitagarwal1612:
Reverts google/ground-android#589

For some reason, sometimes `.git` is initialized by GCB and sometimes not. 
So, to be on safer side. This PR removes the existing dir (if any) and then attempts to mv safely.


scolsen (approved):
lgtm, thanks for investigating, Shobhit!

PR #591, opened 2020-10-16
[Code cleanup] Refactor HomeScreenViewModel

shobhitagarwal1612:
* Replace `disposeOnClear` with `LiveData` for add feature requests
* Expose handler to fragment for UI interactions
* Add tests for `FeatureRepository`

@gino-m  PTAL


gino-m:
Please avoid force-pushing if possible, since it erases your edit history and it makes incremental reviews impossible ;)

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -121,6 +123,15 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
     projectSelectorViewModel = getViewModel(ProjectSelectorViewModel.class);
   }
 
+  private void onFeatureAdded(Optional<Feature> feature) {
+    feature.ifPresentOrElse(
+        viewModel::onAddFeature,
+        () -> {
+          // TODO: Show an error message to the user.
+          Timber.e("Couldn't add feature.");
```
gino-m, line 131: A more specific error message would be helpful here, but broader question - why would we ever call this with Optional.empty() in the first place?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -50,46 +47,41 @@
...
   private final MutableLiveData<Action> openDrawerRequests;
   private final MutableLiveData<BottomSheetState> bottomSheetState;
 
+  private final FlowableProcessor<Feature> addFeatureClicks = PublishProcessor.create();
+  private final FlowableProcessor<Feature> updateFeatureRequests = PublishProcessor.create();
   private final FlowableProcessor<Feature> deleteFeatureRequests = PublishProcessor.create();
-  private final LiveData<Boolean> deleteFeature;
 
-  private final FlowableProcessor<Feature> updateFeatureRequests = PublishProcessor.create();
+  private final LiveData<Optional<Feature>> addFeature;
```
gino-m, line 60: Same question here.

gino-m (changes requested):
Thanks for the cleanup! One comment/question, otherwise LG!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
shobhitagarwal1612: I agree that this is non-conventional. Ideally, we should return success state with result or error state with error message. I'll fix that. Thanks!

shobhitagarwal1612:
@gino-m Resolved the comments and added tests for the new behavior

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -127,21 +127,33 @@ private FeatureMutation fromFeature(Feature feature, Type type) {
   }
 
   // TODO(#80): Update UI to provide FeatureMutations instead of Features here.
-  public Completable createFeature(Feature feature) {
-    return applyAndEnqueue(fromFeature(feature, Type.CREATE));
+  public Flowable<Loadable<Feature>> createFeature(Feature feature) {
```
gino-m, line 130: I've experimented with pushing the Flowable<Loadable<>> down in the stack like this in the past, and while it simplifies usage, it also complicates the semantics of the internal API; these methods no longer simply do work and then either complete or return an error, but rather they stream status. This is slightly harder for clients of these methods to understand, and is really only introduced here because of how the UI streams emit state. It also is questionable whether this simplifies the code, since the transformation to a Flowable still has to happen somewhere.

Can we wrap the state higher up in the stack instead, as close to the subscription as possible?

Also note that `Loadable` is modeled around the action of something being loaded from the database, not creating, updating, and deleting.

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
```diff
@@ -58,17 +52,10 @@ private Loadable(LoadState state, @Nullable T data, Throwable error) {
   }
 
   public static <T> Loadable<T> error(Throwable t) {
+    Timber.e(t);
```
gino-m, line 55: This is a transformation method that will return a new instance. It's assumed that the error in the Loadable will be handled somewhere, so printing the error to the log here will result in duplicate outputs in the log.

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
```diff
@@ -98,6 +85,15 @@ public boolean isLoaded() {
         .startWith(Loadable.loading());
   }
 
+  public LoadState getState() {
+    return state;
+  }
+
+  public boolean isLoaded() {
+    return state == LoadState.LOADED;
+  }
```
gino-m, line 94: In the future, please reorder method in a separate PR if possible, since mixing actual changes with reorders makes it hard to tell what was changed vs moved.

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
```diff
@@ -106,4 +102,29 @@ public String toString() {
...
+      return false;
+    }
+    Loadable<?> loadable = (Loadable<?>) o;
+    return getState() == loadable.getState();
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(getState());
+  }
```
gino-m, line 121: What are equals() and hasCode() ever used? If not, please remove.

gnd/src/main/java/com/google/android/gnd/rx/ValueOrError.java
```diff
@@ -72,4 +62,38 @@ public String toString() {
...
+      return false;
+    }
+    ValueOrError<?> that = (ValueOrError<?>) o;
+    return Objects.equals(value, that.value) && Objects.equals(error, that.error);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(value, error);
+  }
```
gino-m, line 98: What are equals() and hasCode() ever used? If not, please remove.

gino-m (changes requested):
Thanks for the refactor, @shobhitagarwal1612! Are the new tests working now?

Please be sure to merge once comments are addressed.

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
shobhitagarwal1612: These are used in the unit tests

gnd/src/main/java/com/google/android/gnd/rx/ValueOrError.java
shobhitagarwal1612: Same as above, used in unit test

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
shobhitagarwal1612: Sure

gnd/src/main/java/com/google/android/gnd/rx/Loadable.java
shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m PTAL. I've updated the PR with the requested changes

gnd/src/main/java/com/google/android/gnd/rx/RxCompletable.java
```diff
@@ -45,4 +49,17 @@ public static Completable completeOrError(
           }
         });
   }
+
+  public static Single<Boolean> toSingle(
```
gino-m, line 53: This method is actually doing a lot more than just converting the Completable to a Single; it's returning whether the Single completed successfully, and is causing a side effect on the provided Consumer. I'd suggest a few improvements to make this clearer:

* Give the method a more explicit name, like `toBooleanSingle()` to indicate this is doing something special.
* Call the callback param for when it's actually invoked (`onError`).
* If this is every called with null `exceptionConsumer` (I didn't see it on a quick scan), create a separate override of this method without the Consumer (i.e. toBooleanSingle(Completable) and toBooleanSingle(Completable, onErro).
*  Add minimal Javadoc explaining these semantics (esp. since the method is public().

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -121,22 +124,37 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
   private void onFeatureUpdated(Boolean result) {
     if (result) {
-      Toast.makeText(getContext(), "Done", Toast.LENGTH_SHORT).show();
       mapContainerFragment.setDefaultMode();
-    } else {
-      Timber.e("Failed to update feature");
     }
   }
 
+  /** Hide the bottom sheet as deleting a feature also removes child observations. */
```
gino-m, line 145: This Javadoc isn't needed since the code is self-explanatory, and the method is internal/private.

gino-m (changes requested):
Thanks for the edits so far! Just a few more minor suggestions, thanks for your patience!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -53,74 +52,63 @@
...
   private final MutableLiveData<Action> openDrawerRequests;
   private final MutableLiveData<BottomSheetState> bottomSheetState;
 
+  private final FlowableProcessor<Feature> addFeatureClicks = PublishProcessor.create();
+  private final FlowableProcessor<Feature> updateFeatureRequests = PublishProcessor.create();
   private final FlowableProcessor<Feature> deleteFeatureRequests = PublishProcessor.create();
-  private final LiveData<Boolean> deleteFeature;
 
-  private final FlowableProcessor<Feature> updateFeatureRequests = PublishProcessor.create();
+  private final LiveData<Feature> addFeature;
```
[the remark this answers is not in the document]
shobhitagarwal1612: That makes it much more readable as well. Thanks!!!

shobhitagarwal1612:
@gino-m Ping.

gino-m:
> @gino-m Ping.

Sorry didn't know if it was ready! Note for next time you can also re-request review to get my attention:
image

gino-m (approved):
LGTM! 💯

PR #597, opened 2020-10-19
Add nullability annotations

### LINKED ISSUE #50: [Code health] Add explicit nullness checking
Set up lint or compile errors

dturner:
WIP. Fixes #50

TODO: Add wiki page with instructions. 

Discarding this in favour of: https://github.com/google/ground-android/pull/617


gnd/src/androidTest/java/com/google/android/gnd/persistence/remote/FakeRemoteDataStore.java
```diff
@@ -69,11 +73,13 @@
     return Flowable.empty();
   }
 
+  @Nullable
   @Override
   public Single<ImmutableList<ValueOrError<Observation>>> loadObservations(Feature feature) {
     return null;
   }
 
+  @Nullable
```
gino-m, line 82: If `@Nullable` is the default, can we remove this annotation throughout for brevity? Or is it recommended regardless?

gino-m (changes requested):
Wow, that's a big change! And I think it'll be a big improvement as well.

Is it possible to make the default `@NonNull` instead of `@Nullable`? Generally not allowing nulls is safer, so it might make more sense to make that the default imo - not sure if it's possible and how that will jibe with existing APIs though.

Also, once strict nullness checking is enabled, when should we use `@Nullable` vs `Optional`? My approach so far has been to use `Optional` in our model and internal APIs when "empty" is an expected outcome/state, and @Nullable for lower level impl when a variable may simply not have been initialized. Curious to hear your thoughts.

dturner:
@gino-m Thanks for your comments. The new PR using NullAway seems to fit our requirements much better.

PR #600, opened 2020-10-22
[Polygon] Add support for rendering polygon Features

### LINKED ISSUE #534: Show polyline and polygon features
Add GeoJson data to new `geoJson` field in Feature object (remotely and locally). For pilot they will be read-only.

Layer style should be used to style (color) the lines and polygons defined in the GeoJson.

shobhitagarwal1612:
Fixes #534 

- [x] Read geoJson from Feature object
- [x] Save geoJson to local db
- [x] Render all polygons from geoJson to map
- [x] Use layer color for styling polygons

Steps to verify:
Select "example" project > Drag to Australia

Screenshot_1603383329


shobhitagarwal1612:
Steps to verify:

Select "example" project > Drag to Australia

Screenshot_1603383329

shobhitagarwal1612:
@gino-m @scolsen It is ready for a review!

gnd/src/main/java/com/google/android/gnd/model/feature/Feature.java
```diff
@@ -39,8 +52,12 @@
   @Nullable
   public abstract String getCaption();
 
+  @Nullable
```
gino-m, line 55: This may be a good use of polymorphism, since PointFeature and GeoJsonFeature are mutually exclusive feature types. Wdyt?

Note that I don't know whether we will want to support drawing of multi-polygons (vs one polygon per feature) in the future; imported GeoJson might be treated differently from polygons that can be edited in the UI (at least for now), which is why I suggested calling the subclass specifically `GeoJsonFeature` instead of something like `MultiPolygonFeature`.

gnd/src/main/java/com/google/android/gnd/model/feature/Feature.java
```diff
@@ -39,8 +52,12 @@
...
   public abstract Point getPoint();
 
+  @Nullable
```
gino-m, line 58: If we define this in a subclass of `Feature`, this accessor can be `@NonNull.` Same for `getPoint`.

gino-m (changes requested):
Hi @shobhitagarwal1612, nice work so far!

I see you're unpacking the GeoJson yourself in this PR - any reason not to use https://developers.google.com/maps/documentation/android-sdk/utility/geojson to do this for us? 

On the one hand, your approach gives us more control over rendering of GeoJson, but on the other it comes with a higher impl & carrying cost. Can you share your rationale for going this route?

shobhitagarwal1612:
> I see you're unpacking the GeoJson yourself in this PR - any reason not to use https://developers.google.com/maps/documentation/android-sdk/utility/geojson to do this for us?
On the one hand, your approach gives us more control over rendering of GeoJson, but on the other it comes with a higher impl & carrying cost. Can you share your rationale for going this route?

I wasn't aware of this utility class. Thanks for pointing it out!
Trying it out now

gnd/src/main/java/com/google/android/gnd/model/feature/Feature.java
shobhitagarwal1612: This is what I went for in the first place. But then I realized that 'point' is a non-null property right now.
So this change cannot be made until point becomes optional wherever it is being used.

gnd/src/main/java/com/google/android/gnd/model/feature/Feature.java
shobhitagarwal1612: Same as above

gnd/src/main/java/com/google/android/gnd/model/feature/Feature.java
gino-m: That's why I was suggesting subclassing Feature; PointFeature could have `@NonNull getLocation()`, and GeoJsonFeature could have `@NonNull getPolygons()`, for example. Would that work? Might require some changes here and there though, but it would allow us to enforce the presence of a point or polygon at compile time instead of runtime.

gino-m:
> > I see you're unpacking the GeoJson yourself in this PR - any reason not to use https://developers.google.com/maps/documentation/android-sdk/utility/geojson to do this for us?
> > On the one hand, your approach gives us more control over rendering of GeoJson, but on the other it comes with a higher impl & carrying cost. Can you share your rationale for going this route?
> 
> I wasn't aware of this utility class. Thanks for pointing it out!
> Trying it out now

Gotcha! Please don't worry about rewriting it to use this lib if you've already unpacked the polygons in the GeoJson; we can always do that later if it turns out to be useful.

gino-m (changes requested):
Thanks for the edits!

shobhitagarwal1612:
@gino-m I already pushed the changes for rendering GeoJSON using SDK. While doing so, the part where we parsed geoJson and created custom models is cleaned up. But the part which is used for rendering is still there (which can be used in the future PRs while adding support for PolygonFeature/PolylineFeature)

I think this PR is ready for review now.

gino-m (approved):
LGTM! Thanks for the edits and the implementation!

PR #602, opened 2020-10-26
[Add feature] Hide "+" button if no layers are present in the active project

### LINKED ISSUE #33: [Add feature] Hide "Add" button when project has no user-editable layers

shobhitagarwal1612:
Fixes #33 

Steps to verify:

1. Select "example" project
2. "+" button appears as expected
3. Select "project with no layers" project
4. "+" button should no longer be visible

@gino-m   PTAL?


gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -110,6 +115,25 @@
...
+        Optional.of(project)
+            .filter(Loadable::isLoaded)
+            .map(p -> containsLayers(p.value()))
+            .orElse(false);
+    addFeatureBtnVisibility.postValue(addBtnVisibility ? View.VISIBLE : View.GONE);
+  }
+
+  /** Returns true if the given project contains layers, otherwise false. */
+  private boolean containsLayers(Optional<Project> project) {
+    return project.map(p -> !p.getLayers().isEmpty()).orElse(false);
```
scolsen, line 130: I think you can drop `orElse` here and call `flatMap` on line 123.

scolsen (approved):
Thanks, Shobhit!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
shobhitagarwal1612: Done

gino-m (changes requested):
#33 asks that the button is hidden when "no user editable layers" are present. This will only make sense once we actually implement per-layer permissions as implemented as described in https://github.com/google/ground-platform/issues/228. For now, a project with no layers should never happen, since it would be useless. Can we rename where appropriate and add a TODO to make it clear that we should finish implementing this feature by checking on the layer permissions to determine whether the project has user-editable layers? Thanks again for the PR!

gnd/src/androidTest/java/com/google/android/gnd/ui/home/AddFeatureButtonVisibilityTest.java
```diff
@@ -0,0 +1,124 @@
+/*
```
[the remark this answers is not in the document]
shobhitagarwal1612: Acknowledged

shobhitagarwal1612:
@gino-m Can I get a re-review?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -110,6 +115,28 @@
...
+    }
+
+    // TODO: Also check if the project has user-editable layers.
+    //  Pending feature, https://github.com/google/ground-platform/issues/228
+
+    // Project must contain at least 1 layer.
+    return project.value().map(p -> !p.getLayers().isEmpty()).orElse(false);
+  }
+
+  public LiveData<Integer> getAddFeatureBtnVisibility() {
```
gino-m, line 136: Nit: Btn -> Button throughout

gino-m (changes requested):
Thanks! One nit, otherwise LGTM!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
shobhitagarwal1612: Done

gino-m (approved):
LGTM.. Boom! 🚀

PR #614, opened 2020-10-31
[Points] Use marker manager for handling click events

### LINKED ISSUE #603: [Feature sheet] Bottom sheet does not appear when tapping pins
This is expected for features for layers without forms but this happens for all layers.

shobhitagarwal1612:
Fixes #603 

The reason it didn't get caught by the instrumentation tests is because this error only happens once a GeoJSON layer is added to the map. Our test data doesn't contain any GeoJSON in the feature

Verified by clicking on any pin in "example" project

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
@@ -45,6 +45,8 @@ import com.google.android.gnd.ui.map.MapGeoJson;
 import com.google.android.gnd.ui.map.MapPin;
 import com.google.android.gnd.ui.map.MapPolygon;
 import com.google.common.collect.ImmutableSet;
+import com.google.maps.android.collections.MarkerManager;
+import com.google.maps.android.collections.MarkerManager.Collection;
 import com.google.maps.android.data.Layer;
 import com.google.maps.android.data.geojson.GeoJsonLayer;
 import com.google.maps.android.data.geojson.GeoJsonLineStringStyle;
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -80,11 +82,20 @@
...
+   */
+  private final MarkerManager markerManager;
+  // TODO: Add managers for polyline and polygon layers
+
   /**
    * References to Google Maps SDK Markers present on the map. Used to sync and update markers with
    * current view and data state.
    */
-  private Set<Marker> markers = new HashSet<>();
+  private final Collection markers;
```
gino-m, line 98: Could we qualify Collection with its outer class name (`MapObjectManager.Collection`)? I automatically assumed this was a java.util.Collection at first :)

gino-m (changes requested):
Thanks for the fix! One minor edit request, otherwise LGTM.

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
shobhitagarwal1612: Done

gino-m (approved):
Ship it! 🚀

shobhitagarwal1612:
CLA is stuck. Maybe you can force merge this?

gino-m:
> CLA is stuck. Maybe you can force merge this?

Added it back manually. Merging now!

PR #615, opened 2020-11-03
[Bug] Fix regressions for adding/removing a MapPin

shobhitagarwal1612:
These bugs were introduced in the last PR.

Tested manually using examples project.

@gino-m  PTAL?


gino-m (approved):
Thanks for the fix!

PR #617, opened 2020-11-03
Adding NullAway plugin for checking null handling

dturner:
This change adds the [NullAway plugin](https://github.com/uber/NullAway). From their docs: 

> NullAway is a tool to help eliminate NullPointerExceptions (NPEs) in your Java code. To use NullAway, first add `@Nullable` annotations in your code wherever a field, method parameter, or return value may be null. Given these annotations, NullAway performs a series of type-based, local checks to ensure that any pointer that gets dereferenced in your code cannot be null. 

I have configured NullAway to run during the build process and to issue errors whenever it finds a possible null object dereference. Generated classes from Hilt, SafeArgs and DataBindings are excluded, as well as classes relating to Room and Firebase.  

I have fixed the remaining errors (around 100) by: 

1) Adding `@Nullable` and adding null checks: `if (thing==null) throw new NPE("error message");`  
2) Removing `@Nullable` on fields which should always be there, such as `Project.id`
3) Converting `@Nullable` to an `Optional<Thing>` and adding empty/present checks: `if (thing.isPresent()) doSomething(thing);`
4) Suppress the field initialization errors using `@SupressWarnings("NullAway.Init")` where there was an obvious initialization method which NullAway wasn't aware of. For example: 

```
public class AbstractFieldViewModel extends AbstractViewModel {
...
    @SuppressWarnings("NullAway.Init")
    private Field field;
    
    void init(Field field, Optional<Response> response) {
        this.field = field;
    }
```

`init` could be added to the list of methods which NullAway considers to be intialization methods if we want. 

5) There were a few points where I needed to suppress the NullAway errors entirely using `@SuppressWarnings("NullAway")`. Here's an example: 

```
private Completable insertOrUpdateField(String formId, Element.Type elementType, Field field) {
    return fieldDao
        .insertOrUpdate(FieldEntity.fromField(formId, elementType, field))
        .andThen(
            Observable.just(field)
                .filter(__ -> field.getMultipleChoice() != null)
                .flatMapCompletable(
                    __ -> insertOrUpdateMultipleChoice(field.getId(), field.getMultipleChoice())))
  }
```
This causes an error because `field.getMultipleChoice()` has a nullable return type, however, we're already filtering on `field.getMultipleChoice != null`, but NullAway isn't clever enough to figure that out. 

6) Where an overridden method has different annotations to the superclass method I've fixed that. A common example was `Fragment.onCreateView` which must be changed from: 
 
```
public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
            Bundle savedInstanceState)
```
to
```
public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
            @Nullable Bundle savedInstanceState)
```
   
Other changes: 

- I've removed many instances of `@NonNull` since this is no longer required (all objects are non-null by default)

Concerns/questions: 

- This is adding _another_ plugin dependency and extra time to the build process. In my experience it wasn't that much, maybe a few seconds. Is this something we're happy with? 

TODO: 

- Convert this into a wiki page when PR approved.


gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
```diff
@@ -211,7 +209,13 @@ private void observeMultipleChoiceClicks(MultipleChoiceFieldViewModel viewModel)
 
   private void onShowDialog(
       Field field, Optional<Response> currentResponse, Consumer<Optional<Response>> consumer) {
-    Cardinality cardinality = field.getMultipleChoice().getCardinality();
+
+    MultipleChoice multipleChoice = field.getMultipleChoice();
+    if (multipleChoice == null){
```
gino-m, line 214: Can we use Guava's [checkNotNull](https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkNotNull(T)) instead? It can be statically imported from `Preconditions`.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -149,7 +150,11 @@ void initialize(EditObservationFragmentArgs args) {
 
   Optional<Response> getSavedOrOriginalResponse(String fieldId) {
     if (responses.isEmpty()) {
-      return originalObservation.getResponses().getResponse(fieldId);
+      if (originalObservation == null){
```
gino-m, line 153: Alternatively:
`return Optional.fromNullable(originalObservation).map(o -> o.getResponses().getResponse(fieldId));`

gnd/src/main/java/com/google/android/gnd/ui/editobservation/MultipleChoiceFieldLayout.java
```diff
@@ -53,6 +58,9 @@ protected void onFinishInflate() {
         .setOnClickListener(
             v -> {
               if (editText.isFocused()) {
+                if (showDialogListener == null){
```
gino-m, line 61: Please run formatter to harmonize formatting.

gino-m (changes requested):
> I have configured NullAway to run during the build process and to issue errors whenever it finds a possible null object dereference. Generated classes from Hilt, SafeArgs and DataBindings are excluded, as well as classes relating to Room and Firebase.

Brilliant! Caveats noted, thanks for outlining them clearly here.

> 1. There were a few points where I needed to suppress the NullAway errors entirely using `@SuppressWarnings("NullAway")`. Here's an example:
> 
> ```
> private Completable insertOrUpdateField(String formId, Element.Type elementType, Field field) {
>     return fieldDao
>         .insertOrUpdate(FieldEntity.fromField(formId, elementType, field))
>         .andThen(
>             Observable.just(field)
>                 .filter(__ -> field.getMultipleChoice() != null)
>                 .flatMapCompletable(
>                     __ -> insertOrUpdateMultipleChoice(field.getId(), field.getMultipleChoice())))
>         .subscribeOn(schedulers.io());
>   }
> ```

In this case, can we wrap `field.getMultipleChoice()` with Guava's `checkNotNull` instead?
 
> * This is adding _another_ plugin dependency and extra time to the build process. In my experience it wasn't that much, maybe a few seconds. Is this something we're happy with?

If it doesn't add a noticeable change in build time I think it's ok.

Sending a few initial comments, will circle back for more detailed ones once resolved.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationFragment.java
dturner: Great suggestion. Replaced `if (x == null){ throw NPE(message); }` with `Preconditions.checkNotNull(x, message)` here and throughout.

gnd/build.gradle
```diff
@@ -103,6 +107,42 @@ android {
...
+                        "com.google.android.gnd.databinding," +
+                        "com.google.android.gnd.persistence.local.room," +
+                        "com.google.android.gnd.persistence.remote.firestore," +
+                        "com.google.android.gnd.persistence.remote.firestore.schema")
+
+                // TODO: Figure out a better way of excluding classes generated by the SafeArgs
+                //  plugin. Suggestions: patch NullAway to ignore methods annotated with
+                //  @SuppressWarnings("unchecked") or patch SafeArgs to annotate the generated
+                //  classes with @Generated
+                option("NullAway:ExcludedClasses",
```
shobhitagarwal1612, line 133: Is it possible to exclude patterns instead of full class/package name?

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
dturner: I don't think this works because when `originalObservation` is null an `Optional<Observation>` would be returned rather than the required `Optional<Response>`.

gnd/build.gradle
dturner: Annoyingly no. It is possible to use regex to exclude package names but not for class names.  

Config options are here: https://github.com/uber/NullAway/wiki/Configuration

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
gino-m: Out of curiosity, have you tried and it doesn't compile? When `originalObservation` is null, Optional.fromNullable(originalObservation) will be Optional.empty(). I think the generic closure is inferred from the type returned by the lambda in `map()`.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/MultipleChoiceFieldLayout.java
```diff
@@ -40,11 +41,15 @@ public void setOnShowDialogListener(Runnable showDialogListener) {
   protected void onFinishInflate() {
     super.onFinishInflate();
 
+
     TextInputEditText editText = findViewById(R.id.multiple_choice_input_edit_text);
 
     editText.setOnFocusChangeListener(
         (v, hasFocus) -> {
           if (hasFocus) {
+            if (showDialogListener == null){
```
gino-m: Here's another one for checkNotNull()

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -141,7 +143,8 @@ private Completable insertOrUpdateField(String formId, Element.Type elementType,
             Observable.just(field)
                 .filter(__ -> field.getMultipleChoice() != null)
                 .flatMapCompletable(
-                    __ -> insertOrUpdateMultipleChoice(field.getId(), field.getMultipleChoice())))
+                    __ -> insertOrUpdateMultipleChoice(field.getId(),
+                        Preconditions.checkNotNull(field.getMultipleChoice()))))
```
gino-m, line 147: Can we static import `Preconditions` throughout to make these more inline friendly and reduce line-wrapping?

gino-m (changes requested):
Looking good! One question and a comment.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
dturner: Ah right, sorry I get what you mean now. I'll update throughout.

cloudbuild.yaml
```diff
@@ -33,6 +33,13 @@
 
 steps:
 
+  - name: 'gcr.io/$PROJECT_ID/android:28'
+      id: 'Check javac version'
+      args:
+        - '-c'
+        - |
+          javac -version
```
shobhitagarwal1612, line 41: This should not be indented.

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
dturner: I tried it and it doesn't compile. 

image

gino-m: Ok, thanks for checking!

gino-m (dismissed):
LGTM! This will hopefully prevent a great many NPEs, and reduce uncertainty when writing new code!

dturner:
I've run into some issues with this change because `javac` throws a `NullPointerException` during the cloud build (ironic given the nature of this change). 

I filed an issue with Oracle who directed me to this bug: https://bugs.openjdk.java.net/browse/JDK-8235564 which was fixed in JDK15 (we are running 8). 

I'm pretty sure there's some code issue which is causing the compiler to hang but in order to find out what it is I need some way of using a different compiler. 

I tried switching to the latest official version from Docker: `openjdk:16-jdk-alpine3.12`, however, when I attempt to create the cloud image using: 

```
gcloud builds submit --config=cloudbuild.yaml --substitutions=_ANDROID_VERSION=28
```

I get the following error:

```
Step #0: Step 3/17 : RUN echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90builder   && echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90builder
Step #0:  ---> Running in 17ca573c491b
Step #0: /bin/sh: can't create /etc/apt/apt.conf.d/90builder: nonexistent directory
Step #0: The command '/bin/sh -c echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90builder   && echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90builder' returned a non-zero code: 1
Finished Step #0
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 1
Step #0: 
```

@shobhitagarwal1612 Any thoughts?

shobhitagarwal1612:
> Step #0: Step 3/17 : RUN echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90builder   && echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90builder
Step #0:  ---> Running in 17ca573c491b
Step #0: /bin/sh: can't create /etc/apt/apt.conf.d/90builder: nonexistent directory
Step #0: The command '/bin/sh -c echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90builder   && echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90builder' returned a non-zero code: 1
Finished Step #0
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 1
Step #0: 

`alpine` version of linux doesn't support `apt` commands. 

I was able to get around that problem by using this image `adoptopenjdk/openjdk13`
https://hub.docker.com/r/adoptopenjdk/openjdk13

But now am I getting another error which seems to related to > jdk8
```
Step #0: Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlSchema
Step #0:        at com.android.repository.api.SchemaModule$SchemaModuleVersion.<init>(SchemaModule.java:156)
Step #0:        at com.android.repository.api.SchemaModule.<init>(SchemaModule.java:75)
Step #0:        at com.android.sdklib.repository.AndroidSdkHandler.<clinit>(AndroidSdkHandler.java:81)
Step #0:        at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:73)
Step #0:        at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:48)
Step #0: Caused by: java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlSchema
Step #0:        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:602)
Step #0:        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
Step #0:        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
Step #0:        ... 5 more
Step #0: The command '/bin/sh -c yes | sdkmanager --licenses && sdkmanager --update' returned a non-zero code: 1
```

This [answer](https://stackoverflow.com/questions/46402772/failed-to-install-android-sdk-java-lang-noclassdeffounderror-javax-xml-bind-a) suggests downgrading to jdk8 (which tells us why the default docker image uses jdk8 and not the latest version).

shobhitagarwal1612:
I was able to solve the above error by updating the `sdkmanager`.

https://github.com/shobhitagarwal1612/ground-android/commit/a3581e16fa083d28bcbd623f9e56996eef0d0bc5

You can either create new images by patching the above yaml config or use the one I just uploaded `gcr.io/ground-android-gcb/android: jdk_13_28`

Please let me know if this works you!

dturner:
It worked - thanks so much @shobhitagarwal1612.

gino-m:
> Awesome! If @gino-m and @scolsen don't have any objections with moving to the newer jdk and sdkmanager then we can split this PR into 2 PRs:

With this affect the min compatible Android OS version on which the app can run? If not, no objections here!

shobhitagarwal1612:
> With this affect the min compatible Android OS version on which the app can run? If not, no objections here!

No. It shouldn't affect that. We are still using the same build tools version.

dturner:
image

dturner:
Note: I couldn't figure out a way of disabling the error prone warnings for unit tests. This doesn't affect whether the build passes but it does add a bit of extra log output during the unit test build phase of the cloud build.

cloud-builder/Dockerfile-base
```diff
@@ -69,6 +69,6 @@ RUN yes | sdkmanager \
     "extras;android;m2repository" \
     "extras;google;m2repository" \
     "extras;google;google_play_services" \
-    "build-tools;28.0.3"
+    "build-tools;30.0.2"
```
shobhitagarwal1612, line 72: Please revert

cloud-builder/Dockerfile-sdk
```diff
@@ -19,7 +19,7 @@
 # base Docker image.
 
 ARG project_id
-FROM gcr.io/$project_id/android:base
+FROM gcr.io/$project_id/android:newjavac_base
```
shobhitagarwal1612, line 22: Please revert

gnd/build.gradle
```diff
@@ -43,7 +47,7 @@ if (secretsFile.exists()) {
 def googleMapsApiKey = secrets.getProperty('GOOGLE_MAPS_API_KEY', '')
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 30
```
shobhitagarwal1612, line 50: Is this change required for the plugin? If not, lets revert this.

gnd/build.gradle
```diff
@@ -103,6 +107,49 @@ android {
     hilt {
         enableTransformForLocalTests = true
     }
+
+    // ErrorProne and NullAway configuration.
```
shobhitagarwal1612, line 111: Can you please add a link to the wiki/github repo of this plugin?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -337,11 +340,12 @@ public Completable updateMutations(ImmutableList<Mutation> mutations) {
   }
 
   @Override
-  public Completable finalizePendingMutations(ImmutableList<Mutation> mutations) {
+  public Completable finalizePendingMutations(@Nullable ImmutableList<Mutation> mutations) {
```
shobhitagarwal1612, line 343: We can always use an empty immutable list `ImmutableList.of()` instead of null. Why `@Nullable`?

shobhitagarwal1612 (changes requested):

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
dturner: Good idea. I've added a null check inside `LocalMutationSyncWorker::processMutations` and removed the `@Nullable` annotation.

gnd/build.gradle
dturner: Done

gnd/build.gradle
dturner: Reverted

cloud-builder/Dockerfile-sdk
dturner: Done

cloud-builder/Dockerfile-base
dturner: Done

shobhitagarwal1612 (approved):

gino-m (approved):
I've spot checked the changes and all looks good. Let's merge! 🚀 

Looking forward to trying this out when writing new code, thanks again for this huge improvement!

PR #622, opened 2020-11-10
[Map] Fix NPE when rotating screen after selecting map type

### LINKED ISSUE #550: App force quits after downloading offline areas
Repro TBD...

shobhitagarwal1612:
Fixes #550 

* Move subscription for clicks from onViewCreated() to onMapReady()
* Replace MutableLiveData with a PublishSubject to debounce clicks faster than 250ms

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -148,9 +148,10 @@ public class MapContainerFragment extends AbstractFragment {
     binding.moveFeature.cancelButton.setOnClickListener(__ -> setDefaultMode());
     enableLocationLockBtn();
     mapContainerViewModel.getMbtilesFilePaths().observe(this, map::addTileOverlays);
-    mapContainerViewModel.getSelectMapTypeClicks()
-        .as(autoDisposable(getViewLifecycleOwner()))
-        .subscribe(empty -> showMapTypeSelectorDialog());
+    mapContainerViewModel
+        .getSelectMapTypeClicks()
+        .observe(getViewLifecycleOwner(),
+            action -> action.ifUnhandled(this::showMapTypeSelectorDialog));
   }
 
   private void showMapTypeSelectorDialog() {
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
@@ -28,7 +28,6 @@ import android.widget.Toast;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
-import androidx.appcompat.app.AlertDialog;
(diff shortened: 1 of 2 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
```diff
@@ -163,6 +146,24 @@ private void onMapReady(MapAdapter map) {
     binding.moveFeature.cancelButton.setOnClickListener(__ -> setDefaultMode());
     enableLocationLockBtn();
     mapContainerViewModel.getMbtilesFilePaths().observe(this, map::addTileOverlays);
+    mapContainerViewModel.getSelectMapTypeClicks()
+        .as(autoDisposable(getViewLifecycleOwner()))
```
shobhitagarwal1612, line 150: I also considered using a flowable processor which converts events to LiveData. But the problem with that is, it always emits the last value on resubscription. So, the dialog would pop-up again and again after every rotation (or any config change).

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -300,8 +305,8 @@ public void setSelectedFeature(Optional<Feature> selectedFeature) {
 
   static class CameraUpdate {
 
-    private Point center;
-    private Optional<Float> minZoomLevel;
+    private final Point center;
+    private final Optional<Float> minZoomLevel;
```
shobhitagarwal1612, line 309: These changes were unintentional. IDE added these after running "reformat file".

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
gino-m: That shouldn't be the case; Flowables would only emit the last item if you use buffer strategy "latest". I think this is the most likely cause, since `selectMapTypeClicks` is a `PublishSubject`, which doesn't cache the last value.

If you're still having trouble with duplicate requests, consider using `com.google.android.gnd.rx.Event` to wrap the value, and only processing it using its `ifUnhandled()` method.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
gino-m: Ok

gino-m (changes requested):
Response to your comment, otherwise LGTM.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerFragment.java
shobhitagarwal1612: Done

gino-m (approved):
@shobhitagarwal1612 Force push makes incremental review impossible, and erases the thought process that went into getting where we are. Please avoid rewriting history and use normal push instead! 👍 

Thanks for the changes, looks good!

PR #623, opened 2020-11-10
[Map] Map type should persist after re-opening the app

shobhitagarwal1612:
The map type resets after app restart. So, this is to resolve that problem.

@gino-m  PTAL?


gino-m (approved):
Nice work, thanks for the quick fix!

Issue #451, opened 2020-04-21
Project list is not loaded when on a slow internet connection


shobhitagarwal1612:
An empty dialog is shown when using the app with slow internet connection.

**To Reproduce**
1. Set connection speed as GSM/GPRS in android emulator.
2. Open app
3. Project loading dialog appears
4. After 10 seconds a blank dialog appears

**Expected behavior**
Project should be loaded or correct message should be displayed while fetching the data (maybe Internet speed very low?)

Stacktrace:

```
2020-04-15 13:52:32.448 25062-25282/com.google.android.gnd D/ProjectRepository: Failed to load project list from remote
    java.util.concurrent.TimeoutException: The source did not signal an event for 10 seconds and has been terminated.
        at io.reactivex.internal.operators.single.SingleTimeout$TimeoutMainObserver.run(SingleTimeout.java:123)
        at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:38)
        at io.reactivex.internal.schedulers.ScheduledDirectTask.call(ScheduledDirectTask.java:26)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:919)
     Caused by: com.akaita.java.rxjava2debug.extensions.RxJavaAssemblyException
     Caused by: com.akaita.java.rxjava2debug.extensions.RxJavaAssemblyException
2020-04-15 13:52:32.453 25062-25189/com.google.android.gnd I/Firestore: (21.4.1) [OnlineStateTracker]: Could not reach Cloud Firestore backend. Backend didn't respond within 10 seconds
```

gino-m:
Also, we should only be loading project summaries rather than the entire project config for all projects. But that might not be necessary either depending on how we answer #270.

shobhitagarwal1612:
#270 is related to cloud build. Are you referring to some other issue?

gino-m:
Sorry, I meant #434! Updating comment in place.

gino-m:
@sergeydolgov1 @gino-m To discuss.

[...]

PR #624, opened 2020-11-10
Increase remote query timeout

shobhitagarwal1612:
Related #451 

Increase timeout when fetching project summaries from remote db.

@scolsen  PTAL?


scolsen (approved):
LGTM, thanks, Shobhit!

PR #625, opened 2020-11-10
Set GPS lock icon using data bound tint

### LINKED ISSUE #593: [Location lock] Location lock fab not updating when lock is enabled
To repro, tap the location lock button shown on the map above the "+" fab. The map follows the current location, but the fab does not turn blue as expected.
Screenshot_1602860690

gino-m:
Fixes #593.


gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -99,6 +100,11 @@
     this.locationLockState =
         LiveDataReactiveStreams.fromPublisher(
             locationLockStateFlowable.startWith(BooleanOrError.falseValue()));
+    this.iconTint =
+        LiveDataReactiveStreams.fromPublisher(
+            locationLockStateFlowable
+                .map(locked -> locked.isTrue() ? R.color.colorMapBlue : R.color.colorGrey800)
+                .startWith(R.color.colorGrey800));
```
shobhitagarwal1612, line 107: Can we move this to the `MapContainerFragment`? Only the true/false should be exposed and then the color can be set using data binding. We can also get rid of the BindingAdapter then. Wdyt?

gino-m: How would we use data binding without the BindingAdapter? What would that look like?

Also note that we aim to have as little logic as possible in the View components (MapContainerFragment) to improve testability, so moving this in this into the Fragment might be consider a move in the opposite direction. Open to ideas though!

shobhitagarwal1612: On second thought, let's stick with the current approach. Sorry for the confusion!

shobhitagarwal1612 (approved):
Thanks!

PR #630, opened 2020-11-18
[Docs] Initial Android architecture docs and diagram

mmisim:
Added the initial version of the Android architecture docs. I'll also add this page to the Android wiki once this PR is merged.


gino-m (changes requested):
Hi @mmisim thanks for kicking off our technical architecture docs!! Some comments/suggestions re the graph:

* It would be helpful to emphasize the unidirectionality of dependencies by adding arrows, i.e.: MainActivity->Fragments->ViewModels, etc. 
* The "model" is referenced everywhere from the UI layer to the data store, so it may make sense to omit it from this high-level graph.
* The graph looks almost identical to the arch graph in the Android docs; Is there anything more Ground-specific that we can add to this graph to make it more useful than what's already provided? I know we agreed we would just include a high level, so maybe just including a few indicative few examples somehow (e.g., HomeScreenFragment, HomeScreenViewModel, FeatureRepository)?
* LocalDataStore and RemoteDataStore exist as classes in the code base. Would it make sense to label them with their camel-case names so it's clearer that that's a specific class and not a group of classes (as we you did with MainActivity)?

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
+# Ground Android app architecture
+
+The Ground Android app is written in Java using the [Android Jetpack](https://developer.android.com/jetpack) and [RxJava](https://github.com/ReactiveX/RxJava) libraries. This page provides an overview of the high-level app architecture, including key components and libraries.
```
gino-m, line 3: It relies heavily on these two, but they don't form an exhaustive list of libraries used; can this be reworded somehow to make that more clear?

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+
+<!-- Editable image source: https://docs.google.com/drawings/d/1UjetRJsudLHg3YsWaf0sKqHfmrPNNBH4UKI7Ivajx-0/ -->
+![Ground Android app architecture diagram](android-architecture-diagram.png)
+
+## Components
+
+The app is comprised of the following components:
+
+- **MainActivity**: Ground is a single-activity app. The main [activity](https://developer.android.com/guide/components/activities/intro-activities) contains all other app components.
+- **Fragment**: Each part of the app interface is defined as a [fragment](https://developer.android.com/guide/components/fragments). Examples of fragments are the home screen and the map interface.
```
gino-m, line 13: s.->pl. "Fragments" so that it's clearer this is a class of thing rather than a specific instance.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **ViewModel**: Each fragment has a corresponding [ViewModel](https://developer.android.com/topic/libraries/architecture/viewmodel), which encapsulates its UI state, logic, and functionality, as well as the lifecycle callback methods.
+- **Managers**: The app managers abstract access to system services directly for the ViewModels. Examples of services handled by managers are location services and camera access.
+- **Models**: Entity representations used by ViewModels, such as the app's representations of map concepts, are defined as models. Models allow the app to remain platform agnostic because we can update model definitions without affecting ViewModels or other app components. Examples of entities defined models are the app's representation of offline imagery.
```
gino-m, line 16: "used by ViewModels" -> "used throughout the app"

gino-m, line 16: Models -> Model (singular, since there's only one "model")

> Models allow the app to remain platform agnostic because we can update model definitions without affecting ViewModels or other app components.

I'd replace this with something like "the .model package represents data across the various components of app",.

More here:
https://developer.android.com/jetpack/guide#drive-ui-from-model

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **Repository**: Each model has a corresponding repository that defines the APIs for interaction between ViewModels and data stores.
```
gino-m, line 17: There isn't necessarily a 1:1 representation between model classes and repositories; repositories are organized by thematic area.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **ViewModel**: Each fragment has a corresponding [ViewModel](https://developer.android.com/topic/libraries/architecture/viewmodel), which encapsulates its UI state, logic, and functionality, as well as the lifecycle callback methods.
+- **Managers**: The app managers abstract access to system services directly for the ViewModels. Examples of services handled by managers are location services and camera access.
+- **Models**: Entity representations used by ViewModels, such as the app's representations of map concepts, are defined as models. Models allow the app to remain platform agnostic because we can update model definitions without affecting ViewModels or other app components. Examples of entities defined models are the app's representation of offline imagery.
+- **Repository**: Each model has a corresponding repository that defines the APIs for interaction between ViewModels and data stores. 
+    - **Local data store**: Local data is stored in a SQLite database and the app uses the [Room persistence library](https://developer.android.com/topic/libraries/architecture/room) to manage interaction with the library.
```
gino-m, line 18: The data store encapsulates details about how data is stored locally. The default implementation uses Room and SQLite to do so, but the architecture attempts to remain clean by not exposing these implementation details to the rest of the app.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+    - **Remote data store**: Remote data is stored in [Firebase Cloud Firestore](https://firebase.google.com/docs/firestore). For more information on how remote data is structured, see [Cloud Firestore Representation](https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation)
```
gino-m, line 19: Same applies for the remote data store.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **Workers** (not pictured in the diagram): Jobs that run on a background thread are implemented via [workers](https://developer.android.com/reference/androidx/work/Worker). Examples of tasks handed by workers are syncing local and remote data when a network connection is available.
```
gino-m, line 20: background thread -> jobs that can run while the app is closed.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Data flows
+
+- Data flows between components is managed via [LiveData](https://developer.android.com/topic/libraries/architecture/livedata) and [RxJava](https://github.com/ReactiveX/RxJava). 
+- Dependencies between components are managed via [Hilt](https://developer.android.com/training/dependency-injection/hilt-android) (Dagger).
```
gino-m, line 25: I would reference Dependency Injection here specifically, since "dependencies" can have several meanings in this context.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **Repository**: Each model has a corresponding repository that defines the APIs for interaction between ViewModels and data stores. 
+    - **Local data store**: Local data is stored in a SQLite database and the app uses the [Room persistence library](https://developer.android.com/topic/libraries/architecture/room) to manage interaction with the library.
+    - **Remote data store**: Remote data is stored in [Firebase Cloud Firestore](https://firebase.google.com/docs/firestore). For more information on how remote data is structured, see [Cloud Firestore Representation](https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation)
+- **Workers** (not pictured in the diagram): Jobs that run on a background thread are implemented via [workers](https://developer.android.com/reference/androidx/work/Worker). Examples of tasks handed by workers are syncing local and remote data when a network connection is available.
+
+## Data flows
+
+- Data flows between components is managed via [LiveData](https://developer.android.com/topic/libraries/architecture/livedata) and [RxJava](https://github.com/ReactiveX/RxJava).
```
gino-m, line 24: "Data flows" - Perhaps a more precise description here might be "application components rely heavily on LiveData and RxJava to represent asynchronous flows of events and data through the app."

gino-m (changes requested):
A few more suggestions. 💯

docs/architecture-docs.md
mmisim: Done.

docs/architecture-docs.md
mmisim: Done.

docs/architecture-docs.md
mmisim: Done.

mmisim: Done

docs/architecture-docs.md
mmisim: Do you think the implementation details could still be useful, even if they're not exposed to the rest of the app? Or is there some other information about the LocalDataStore that we should provide here?

docs/architecture-docs.md
mmisim: Is there some other information that we should provide here instead?

docs/architecture-docs.md
mmisim: Done

docs/architecture-docs.md
mmisim: Done.

docs/architecture-docs.md
mmisim: Done.

docs/architecture-docs.md
mmisim: Reworded. Let me know if you think it makes sense.

mmisim:
@gino-m Thanks for the review and sorry for the delay. I've updated the text and diagram. PTAL.

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+
+<!-- Editable image source: https://docs.google.com/drawings/d/1UjetRJsudLHg3YsWaf0sKqHfmrPNNBH4UKI7Ivajx-0/ -->
+![Ground Android app architecture diagram](android-architecture-diagram.png)
+
+## Components
+
+The app is comprised of the following components:
+
+- **MainActivity**: Ground is a single-activity app. The main [activity](https://developer.android.com/guide/components/activities/intro-activities) contains all other app components.
+- **Fragments**: Each part of the app interface is defined as a [fragment](https://developer.android.com/guide/components/fragments). Examples of fragments are the home screen and the map interface.
```
scolsen, line 13: optional nit: `interface` -> `user interface` (since Java supports interfaces as a programming concept as well).

scolsen, line 13: optional nit: `interface` -> `user interface` (since Java supports interfaces as a programming concept as well).

docs/architecture-docs.md
```diff
@@ -0,0 +1,29 @@
...
+- **Repository**: The model classes are linked to the repositories, which are organized by thematic area, that define the APIs for interaction between ViewModels and data stores. 
+    - **Local data store**: Local data is stored in a SQLite database and the app uses the [Room persistence library](https://developer.android.com/topic/libraries/architecture/room) to manage interaction with the library.
+    - **Remote data store**: Remote data is stored in [Firebase Cloud Firestore](https://firebase.google.com/docs/firestore). For more information on how remote data is structured, see [Cloud Firestore Representation](https://github.com/google/ground-platform/wiki/Cloud-Firestore-Representation)
+- **Workers** (not pictured in the diagram): Jobs that that can run while the app is closed are implemented via [workers](https://developer.android.com/reference/androidx/work/Worker). Examples of tasks handed by workers are syncing local and remote data when a network connection is available.
```
scolsen, line 20: nit: Since there's only one example here, make the structure singular, rather than plural? `For example, syncing local and remote data when a network connection is available is a task that can run while the app is closed and is handled by a worker.`

scolsen (approved):

docs/architecture-docs.md
mmisim: Changed it to UI. Thanks

docs/architecture-docs.md
mmisim: Done. Nice catch. Thanks

docs/architecture-docs.md
gino-m: I think this should suffice.

gino-m:
@mmisim Sorry I really dropped the ball on this one. I'm going to merge, we can refine it iteratively going forward. Thanks again!

PR #639, opened 2020-11-25
[WIP] Cleanup optionals

gino-m:
Fixes several unchecked uses of `Optional.get()` that could result in runtime exceptions.


gino-m:
Splitting out instead a series of PRs. Closing.

PR #640, opened 2020-11-25
[Code health] Fix unchecked instances of Optional.get()

gino-m:
I split this out of #639 to make review easier. This should contain only instances where `get()` is called without `isEmpty()` or `isPresent()` being called in the same context. There are likely more - wondering if there's a PMD check for this?

@dturner care to have a look-see?

@shobhitagarwal1612 @scolsen FYI.


gino-m:
@shobhitagarwal1612 we're seeing new PMD errors on existing in the PR, but config hasn't changed. Any idea what could be going on?

shobhitagarwal1612:
> @shobhitagarwal1612 we're seeing new PMD errors on existing in the PR, but config hasn't changed. Any idea what could be going on?

Gradle version was recently updated. It could be because of that.

gino-m:
> > @shobhitagarwal1612 we're seeing new PMD errors on existing in the PR, but config hasn't changed. Any idea what could be going on?
> 
> Gradle version was recently updated. It could be because of that.

Sorry, these were real PMD errors; fixed!

shobhitagarwal1612 (dismissed):

gnd/src/main/java/com/google/android/gnd/util/StreamUtil.java
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Executes the specified Supplier, writing throw Error exceptions to debug logs. If an error
+   * occurs, an empty Stream is returned, otherwise a Stream with only the result of the Supplier is
+   * returned.
+   */
+  public static <R> Stream<R> logErrorsAndSkip(Supplier<R> supplier) {
+    try {
+      return stream(Collections.singleton(supplier.get()));
+    } catch (RuntimeException e) {
+      Timber.d(e);
```
shobhitagarwal1612, line 38: `Timber.e`

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -268,8 +266,8 @@ public Completable applyAndEnqueue(FeatureMutation mutation) {
     return featureDao
         .findById(featureId)
         .map(f -> FeatureEntity.toFeature(f, project))
-        .filter(Optional::isPresent)
-        .map(Optional::get)
+        .doOnError(e -> Timber.d(e))
```
shobhitagarwal1612, line 269: `Timber.e`

shobhitagarwal1612 (approved):
Some minor nits.

gnd/src/main/java/com/google/android/gnd/util/StreamUtil.java
gino-m: Not sure at what level these should actually be logged, but will change to e() to err [:))] on the side of caution.

shobhitagarwal1612: All error type logs are uploaded to crashlytics. So it would be a nice thing for us to keep track of.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
shobhitagarwal1612: This one is left.

shobhitagarwal1612 (approved):

PR #644, opened 2020-12-12
[Bug] Use "staging" buildType for CI

### LINKED ISSUE #643: [Build] Login broken at HEAD
This is due to the overriding of the debug keystore done in #616; the SHA-1 would need to be added any Firebase instances that the app might use. This is undesirable, not lead because we wouldn't want GCB to access the test builds we share with partners. Instead, local builds should use the default ~/.android/debug.keystore instead. 

@shobhitagarwal1612, is it possible to set up a different Gradle target (not `debug`, perhaps `test`) for use in GCB?

For now, we can remove the `signupConfigs` block in gnd/build.gradle locally as temporary workaround (without checking in the patch). @scolsen @dturner FYI.

shobhitagarwal1612:
Fixes #643 

@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -79,7 +79,7 @@ steps:
     args:
       - '-c'
       - |
-        ./gradlew -PdisablePreDex --no-daemon androidDependencies assembleDebug assembleDebugAndroidTest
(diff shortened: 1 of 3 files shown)


gino-m (approved):
LGTM. Thanks for fixing this!

PR #645, opened 2020-12-15
[Bug] Fix NPE when moving feature

shobhitagarwal1612:
Steps to reproduce the error in `master`:
1. Select any project
2. Click on any point feature
3. Rotate the screen twice
4. Open the overflow menu and click on "Move Feature"

@gino-m  PTAL?


gino-m (approved):
LGTM! In fact it was strange that we restore the child state but never save it. Thanks for the fix!

PR #647, opened 2020-12-15
Don't throw runtime exception if id not found

shobhitagarwal1612:
This doesn't fix any issues, just prevents the throwing of exception.

@dturner  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/common/Navigator.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/common/Navigator.java
@@ -30,6 +30,7 @@ import io.reactivex.Observable;
 import io.reactivex.subjects.PublishSubject;
 import io.reactivex.subjects.Subject;
 import javax.inject.Inject;
+import timber.log.Timber;
 
 /**
  * Responsible for abstracting navigation from fragment to fragment. Exposes various actions to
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/ui/common/Navigator.java
```diff
@@ -110,7 +110,9 @@ public void showHomeScreen(int currentNavDestinationId) {
         navigate(SignInFragmentDirections.proceedToHomeScreen());
         break;
       default:
-        throw new IllegalArgumentException(currentNavDestinationId + " id not found");
+        // Do nothing, probably a config change
```
gino-m, line 113: What happens if we add a one-sided direction to the home screen and use that here?

Also, could we log an error (Timber.e) here to make this easier to debug?

shobhitagarwal1612: > What happens if we add a one-sided direction to the home screen and use that here?

During config change, the navigation doesn't happen via Navigator. So the destination id is still unset.

> Also, could we log an error (Timber.e) here to make this easier to debug?

Done

gino-m (approved):
Thanks for the workaround!

PR #649, opened 2020-12-17
Restore map position and zoom level on configuration change.

### LINKED ISSUE #605: [Map] Viewport is reset after returning from other Fragments
For example after hitting "Save" on the add/edit observation fragment.

dturner:
Fixes #605. 

The map position and zoom level are now stored in the `MapContainerViewModel` and are used to restore the map state after configuration changes. 

@gino-m thanks very much for your help with this. LMK what you think. I'm not entirely convinced by the class name `GroundCameraPosition` but couldn't think of anything better.


gnd/src/main/java/com/google/android/gnd/ui/map/GroundCameraPosition.java
```diff
@@ -0,0 +1,27 @@
+package com.google.android.gnd.ui.map;
```
gino-m, line 1: Please add copyright header.

gnd/src/main/java/com/google/android/gnd/ui/map/GroundCameraPosition.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.android.gnd.model.feature.Point;
+
+public class GroundCameraPosition {
+
+  private final Point target;
+  private final Float zoomLevel;
+
+  public GroundCameraPosition(Point target, Float zoomLevel) {
```
gino-m, line 10: We haven't been prefixing classes with `Ground` since the package name sufficiently disambiguates. Could we use `CameraPosition` instead?

gnd/src/main/java/com/google/android/gnd/ui/map/MapAdapter.java
```diff
@@ -37,15 +37,21 @@
    */
   Observable<Point> getDragInteractions();
 
-  /** Returns a stream that emits the viewport center on each camera movement. */
-  Observable<Point> getCameraMoves();
+  /** Returns a stream that emits the viewport camera position on each camera movement. */
```
gino-m, line 40: Strike "viewport", since it implies the bounds are returned; better just "camera position".

gnd/src/main/java/com/google/android/gnd/ui/map/MapAdapter.java
```diff
@@ -37,15 +37,21 @@
...
+  Observable<GroundCameraPosition> getCameraMoves();
 
   /** Enables map gestures like pan and zoom. */
   void enable();
 
   /** Disables all map gestures like pan and zoom. */
   void disable();
 
+  /**
+   * Repositions the viewport camera position.
```
gino-m, line 50: Repositions the camera.

gino-m (changes requested):
LGTM, just a few nits. Thanks for fixing this!

gino-m:
Note that I filed #650 and #651 to fix the architectural issues which led to this issues.

gnd/src/main/java/com/google/android/gnd/ui/map/GroundCameraPosition.java
dturner: Sure. There's already a `CameraPosition` class in `com.google.android.gms.maps.model` so I added the `Ground` prefix to our class to avoid having to qualify usage of both classes in `GoogleMapsMapAdapter`, however, it's only one line so actually `CameraPosition` makes much more sense.

gnd/src/main/java/com/google/android/gnd/ui/map/MapAdapter.java
dturner: Thanks. Btw did you know that you can insert a suggestion like this using the +/- symbol above?

gino-m (approved):
LGTM! Thanks for the fix!

PR #657, opened 2020-12-20
[Code health] Move authentication handling into MainViewModel

gino-m:
Contributes to #9, @dturner as discussed. PTAL?


dturner (approved):
Nice change!

PR #666, opened 2020-12-23
[Feature] Fixes new point can't be added

### LINKED ISSUE #661: [Add feature] Can't add more than one point
The first point to be added works ok, but subsequent ones cause the edit observation screen to appear before the layer selection dialog. The layer selection popup is shown _after_ the observation is saved. A new feature is not created - instead the new observation is added to the last selected point.

@shobhitagarwal1612 could you PTAL? This may be related to #509.

shobhitagarwal1612:
Fixes #661 

Things done:
 - Replace Subject with interface
 - Remove extra Rx call to invoke dialog
 - Remove duplicate null check for project
 - Reuse parameter "point" instead of injecting
 - Limit AddFeatureDialogFragment to only return layer and remove uuid generator and auth manager
 - Create new feature object in HomeScreenViewModel

@gino-m  PTAL?


gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.java
```diff
@@ -21,107 +21,74 @@
...
-  }
+  public AddFeatureDialogFragment() {}
 
-  public Maybe<Feature> show(FragmentManager fragmentManager) {
+  public void show(
+      @NonNull Project project,
+      @NonNull FragmentManager fragmentManager,
+      @NonNull LayerSelectedListener listener) {
+    this.project = project;
+    this.listener = listener;
```
gino-m, line 54: The listener reference should be removed to prevent memory leaks when the dialog is closed to prevent memory leaks. @dturner Please cmiiw.

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.java
```diff
@@ -21,107 +21,74 @@
...
-    this.homeScreenViewModel = getViewModel(HomeScreenViewModel.class);
-    this.mapContainerViewModel = getViewModel(MapContainerViewModel.class);
-  }
+  public AddFeatureDialogFragment() {}
 
-  public Maybe<Feature> show(FragmentManager fragmentManager) {
+  public void show(
+      @NonNull Project project,
+      @NonNull FragmentManager fragmentManager,
+      @NonNull LayerSelectedListener listener) {
```
gino-m, line 52: Instead of defining a custom functional interface, you should be able to use `Consumer<Layer>`.

gino-m (changes requested):
This is a big improvement! A few comments/suggestions, one question for @dturner.

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.java
shobhitagarwal1612: Removed the reference on destroy.

gino-m (approved):
Great work, @shobhitagarwal1612, thanks for the fix and for the edits! 💯

PR #672, opened 2020-12-24
Re-enable build version label

shobhitagarwal1612:
@gino-m  PTAL?


gino-m (approved):
This will really come in handy, thanks Shobhit!

PR #673, opened 2020-12-24
Show progress dialog when signing in app

shobhitagarwal1612:
@gino-m  PTAL?


gino-m (approved):
Well done! Thanks for the fix!

PR #676, opened 2020-12-28
Update auto-value plugin

shobhitagarwal1612:
* Removes redundant `@Nonnull`
* Removes `@Nullable` from Builders that are being complained by the updated plugin

@scolsen  PTAL?


gino-m (approved):
Awesome, nice cleanup!

PR #681, opened 2020-12-30
[cleanup]

shobhitagarwal1612:
Fixes #

@... PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -35,7 +35,8 @@ steps:
     # we use rsync and not cp so that this step doesn't fail the first time it's run
     args: [ 'rsync', 'gs://${_CACHE_BUCKET}/', './' ]
 
-  - name: 'gcr.io/$PROJECT_ID/android:base'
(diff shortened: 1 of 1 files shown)


gino-m:
@shobhitagarwal1612 Is this ready for a look? If so, please add a more meaningful description, mark as "ready for review", and assign a reviewer. Thanks!

shobhitagarwal1612:
One of the issues is that GCB does a shallow git clone. So the version history is incomplete and leads to wrong version number in the app. So, I'm on the fence of whether to continue with this change or abort this.

The above is fine if we are just running static code analysis and unit test. But the apk won't show the correct app version if used in a production flow later. (e.g. deploy apk to playstore or copy to public bucket for publishing)

shobhitagarwal1612:
Let's close it for now. I don't think the reduction of one line justifies the change in app version.

gino-m:
Makes sense. Perhaps going forward we'd like to use git tags for versioning and something like [this](https://medium.com/@prxtl/using-git-to-generate-versionname-and-versioncode-for-android-apps-aaa9fc2c96af) to apply them to our releases.

PR #686, opened 2021-01-03
[lint] Remove unused drawables

shobhitagarwal1612:
Adds a lint rule to detect all unused non xml resources.

@gino-m  PTAL?


gino-m (approved):
Nice!

PR #671, opened 2020-12-23
[Code health] Add annotations for observables

gino-m:
@scolsen Would you be able to help with the wording and concept here? I was hoping this could help us clarify usage of streams throughout our code base and help us with code-as-documentation for our contract w/internal APIs. Any thoughts/edits greatly appreciated! 🙏

PR #687, opened 2021-01-04
[Code health] Cleanup Project observables and apply annotations

gino-m:
Depends on #671.


gnd/src/main/java/com/google/android/gnd/rx/annotations/Events.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.Target;
+
+/**
+ * Denotes an observable that emits events.
+ *
+ * <p>Recommended base types:
+ *
+ * <ul>
+ *   <li>{@link io.reactivex.Observable}
```
scolsen, line 30: Since `Observables` are cold by default, should we also mention `Subject` and/or `ConnectableObservable`, or another class that's "hot" by default?

gnd/src/main/java/com/google/android/gnd/rx/annotations/Hot.java
```diff
@@ -0,0 +1,54 @@
...
+ * <p>Note: Observables that replay values to late subscribers are considered hot because
+ * subscribing to a these observables does not create a new producer. For example, subscribing to
+ * {@link io.reactivex.subjects.ReplaySubject} will not cause upstream producers to be recreated;
+ * upstream operations will not be reexecuted. Similarly, {@link
+ * io.reactivex.subjects.BehaviorSubject} are not derived from other observables, and their
+ * producers are independent of their existence.
+ *
+ * <p>Some operations on hot observables will result in cold ones. <code>switchMap</code> or <code>
+ * flatMap</code> with a cold observable, for example, will result in a new cold observable, since
+ * the act of subscribing causes to source stream to start emitting values to an observer which
```
scolsen, line 46: s/to source stream/the source stream

gnd/src/main/java/com/google/android/gnd/rx/annotations/States.java
```diff
@@ -0,0 +1,43 @@
...
+ * state on subscribe, then continues emitting the latest state each time it changes.
+ *
+ * <p>Operations on items emitted by this observable should be idempotent; the same item being
+ * emitted multiple times should have the same result. Observers shouldn't depend on values being
+ * emitted in any particular order.
+ *
+ * <p>Recommended base types:
+ *
+ * <ul>
+ *   <li>{@link io.reactivex.Flowable} with {@link io.reactivex.BackpressureStrategy#LATEST}
```
scolsen, line 35: Since this is `Hot`, should we also include classes that are "hot" by default, `Subjects`, `Connectables`, `Processor`? I realize these aren't base types, but it feels worthwhile mentioning them here in some capacity since it's the only way to actually instantiate a "hot" source (esp. since the rxJava docs don't even mention their hot/cold property! (though it's captured in marble diagrams))

gnd/src/main/java/com/google/android/gnd/rx/annotations/Events.java
gino-m: Moved this into our doc so we can iterate there instead.

gnd/src/main/java/com/google/android/gnd/rx/annotations/Hot.java
gino-m: Removed this class.

gnd/src/main/java/com/google/android/gnd/rx/annotations/States.java
gino-m: Removed this class as well.

scolsen (approved):
Thanks! I think having `terminates = false` be the default for `@Hot` makes a lot more sense.

gino-m:
Thanks for the quick review, @scolsen ! Moving forward!

PR #689, opened 2021-01-11
[Scrap] [Code health] Experimenting with alternate observable annotations

gino-m:
@shobhitagarwal1612 @scolsen I was playing with using an array of enum values to indicate our observable properties, but defaults aren't always explicit in this case. Sharing purely as FYI.


gino-m:
Abandoned in favor of #671.

PR #698, opened 2021-01-17
[Code health] Clean up and document observables in MainViewModel

gino-m:
- [x] Make `onSignInStateChange()` fully lazy and `@Cold`
- [x] Add JavaDoc and missing Rx annotations
- [x] Make schedulers specific
- [x] Simplify observers slightly
- [x] Small rename refactor
- [x] Introduce custom operator `switchMapIf()`.

Work towards #699.


gino-m:
@shobhitagarwal1612 PTAL?

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -42,7 +44,12 @@
 @SharedViewModel
 public class MainViewModel extends AbstractViewModel {
 
+  /** The window insets determined by the activity. */
+  @Hot(replays = true)
   private final MutableLiveData<WindowInsetsCompat> windowInsets = new MutableLiveData<>();
+
+  /** The state of sign in progress dialog visibility. */
+  @Hot(replays = true)
```
shobhitagarwal1612, line 54: Same as above.
Annotation already added to the public accessible method. Let's keep it at only one place.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -42,7 +44,12 @@
 @SharedViewModel
 public class MainViewModel extends AbstractViewModel {
 
+  /** The window insets determined by the activity. */
+  @Hot(replays = true)
```
shobhitagarwal1612, line 50: Annotation already added to the public accessible method. Let's keep it at only one place.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -67,11 +75,16 @@ public MainViewModel(
 
     // TODO: Move to background service.
     disposeOnClear(
-        projectRepository.getActiveProject().switchMapCompletable(this::syncFeatures).subscribe());
+        projectRepository
+            .getActiveProject()
+            .observeOn(schedulers.io())
```
shobhitagarwal1612, line 82: Do we need to pass the `subscribeOn` scheduler as well?

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -67,11 +75,16 @@ public MainViewModel(
...
+            .switchMapCompletable(this::syncFeatures)
+            .subscribe());
 
     disposeOnClear(
         authenticationManager
             .getSignInState()
+            .observeOn(schedulers.ui())
```
shobhitagarwal1612, line 90: `.observeOn(schedulers.ui())` 
Why is this change needed?

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -95,8 +109,19 @@ void onApplyWindowInsets(WindowInsetsCompat insets) {
     windowInsets.setValue(insets);
   }
 
+  @Cold
   private Completable onSignInStateChange(SignInState signInState) {
     Timber.d("Auth status change: %s", signInState.state());
+    Optional<User> user = signInState.getUser();
+
+    if (user.isPresent()) {
```
shobhitagarwal1612: We should also check for state == `SIGNED_IN`

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -138,9 +159,11 @@ void onSignedOut() {
     navigator.navigate(SignInFragmentDirections.showSignInScreen());
   }
 
-  void onSignedIn() {
-    hideProgressDialog();
-    navigator.navigate(HomeScreenFragmentDirections.showHomeScreen());
+  private Completable onSignedIn(User user) {
+    return userRepository
+        .saveUser(user)
```
shobhitagarwal1612: Can we move the user creation under if block `onSignInStateChange` and call `onSignedIn` from there?

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -111,13 +112,8 @@ void onApplyWindowInsets(WindowInsetsCompat insets) {
...
   @Cold
-  private Observable<SignInState> saveUserOnSignIn(SignInState signInState) {
-    Optional<User> user = signInState.getUser();
-    if (user.isPresent()) {
-      return userRepository.saveUser(user.get()).andThen(Observable.just(signInState));
-    } else {
-      return Observable.just(signInState);
-    }
+  private Completable saveUser(SignInState signInState) {
+    return userRepository.saveUser(signInState.getUser().get());
```
shobhitagarwal1612, line 116: `.get()` -> `.orElseThrow()` ?

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -133,7 +133,7 @@ public ProjectRepository(
     return projectLoadingState.map(Loadable::value);
   }
 
-  public void onSelectProject(String projectId) {
+  public void activateProject(String projectId) {
```
shobhitagarwal1612, line 136: How about adding a `@NonNull` to the parameter and removing `Optional` below? 

I think all the usages of this method are already checking for non-null project ids. So we can simplify it here.

shobhitagarwal1612 (changes requested):
Sorry for the delay. Some comments for your consideration.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
gino-m: I think we should annotation both fields and accessors. In many case the class' impl will interact with the field directly (not via the public accessor), in which case it will be important to understand the observable's semantics.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
gino-m: See previous comment.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
gino-m: The WIP best practices doc prescribes that we specify `subscribeOn()` in the method that defines the background operation (in this case `getActiveProject()`).

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
gino-m: Our best practices guidelines recommend specifying `observeOn()`  downstream close to `subscribe()`. But I think in this case it's irrelevant to place it here since we only do work inside the `switchMap` operation, which has its own subscription and threads. I'll remove it to avoid confusion.

gino-m: It's needed to ensure we're actually using the UI thread to process values handled by `subscribe()`.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
gino-m: `@NonNull` is the default in our current setup, so we don't need to specify it.

The `Optional` isn't being used to check for nulls here.. not sure I follow?

gino-m (commented):
Thanks @shobhitagarwal1612 for reviewing. Sent replies and follow up questions. PTAL.

gnd/src/main/java/com/google/android/gnd/MainViewModel.java
gino-m: Good point.. we shouldn't use `get()` unless we explicitly check whether the Optional is empty. The problem here is that the check would need to be done in the caller, not here together with the call to `get()`.

What if we consolidate this logic into our custom operators? Then the check and accessor are called in the same place, and the code is a bit more readable. Added a commit to demonstrate. Wdyt?

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
shobhitagarwal1612: `selectProjectEvent` wraps the `projectId` as an Optional. As you already mentioned, the string can't be null. So I was wondering if we can remove the optional now. Line 138

shobhitagarwal1612 (approved):

Issue #699, opened 2021-01-17
[Code health] Ensure correct and readable usage of LiveData and RxJava2 


gino-m:
Status | com.google.android.gnd | Assignee | PR(s)
-- | -- | -- | --
 🔀 Merged | [package root] | @gino-m | #698
  | persistence.local.* | @shobhitagarwal1612 | #731 
 🔀  Merged   | persistence.remote.* | @gino-m | #704
  | persistence.sync | @shobhitagarwal1612
  | repository | @scolsen | #731
 🚧  In progress  | system | @gino-m | #724, #725, #727, #729
  | system.auth | @shobhitagarwal1612
  | system.rx | @scolsen
  | ui.common | @gino-m
  | ui.editobservation | @shobhitagarwal1612
  | ui.home | @gino-m
  | ui.home.featuredetails | @gino-m
  | ui.home.mapcontainer | @gino-m
  | ui.map.* | @scolsen
  | ui.observationdetails | @shobhitagarwal1612
  | ui.offlinebasemap.* | @scolsen | #731
  | ui.projectselector | @scolsen

Checklist template (work in progress):
```
ViewModels

- [ ] Make exposed Rx streams @Hot(replays=false) wherever possible
- [ ] Subscribe in constructor, destroy onClear()
- [ ] Use Observable for Flowable for events (not LiveData)
- [ ] Handle errors and backpressure before fromPublisher()

Activities and Fragments

- [ ] Subscribe in onCreate() only
- [ ] Use autoDispose(this) in subscribe
- [ ] Observe in onCreate() only
- [ ] Use “this” as LiveData lifecycle owner

Usage
- [ ] Call subscribeOn() upstream in layer encapsulating async operation
- [ ] Call observeOn() downstream before subscribe()
- [ ] Define Flowable backpressure before subscribe()

Naming

- [ ] States: fooVisibility, loadingState, etc.
- [ ] Events: fooClicks, barInteractions, bazRequests, etc.

General readability

- [ ] Split complex streams into methods or variables
- [ ] Replace complex operations with side effects
- [ ] Instantiate Subjects in field initializers instead of constructors
- [ ] Expose observables via accessors, field never public

Documentation

- [ ] Add JavaDoc to RxJava2 and LiveData fields and methods
- [ ] Annotate RxJava2 and LiveData fields and methods
```

gino-m:
@shobhitagarwal1612 @scolsen Gentle ping! :)

gino-m:
[...]

PR #725, opened 2021-02-01
[Code health] Document PermissionsManager

gino-m:
Work towards #699.

@scolsen PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/system/PermissionsManager.java
+++ b/gnd/src/main/java/com/google/android/gnd/system/PermissionsManager.java
@@ -22,11 +22,13 @@ import android.content.pm.PackageManager;
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import com.google.android.gnd.rx.RxCompletable;
+import com.google.android.gnd.rx.annotations.Cold;
 import io.reactivex.Completable;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


gnd/src/main/java/com/google/android/gnd/system/PermissionsManager.java
```diff
@@ -36,15 +38,25 @@
 
   @Inject
   public PermissionsManager(Application app, ActivityStreams activityStreams) {
-    context = app.getApplicationContext();
+    this.context = app.getApplicationContext();
     this.activityStreams = activityStreams;
   }
 
+  /**
+   * Asynchronously requests the app be granted the specified permission. If the permission has
```
scolsen, line 46: "the specified permission" -> "the specified {@param permission}"?

scolsen (approved):

gnd/src/main/java/com/google/android/gnd/system/PermissionsManager.java
gino-m: I meant "the permission" in general, otherwise we might say "synchronously requests the app be granted {@param permission}".

gino-m: If you don't mind I'll merge as-is to avoid needing reapproval. Happy to reword if you feel it should be changed!

PR #728, opened 2021-02-01
[Code health] Remove obsolete DeviceCapabilities class

gino-m:
One method was unused, and one was always true since we only support Jellybean and above.

@scolsen PTAL?


scolsen (approved):
Great!

PR #730, opened 2021-02-02
[Logging] Enable logging earlier in app lifecycle

gino-m:
Timber currently doesn't work before the end of `onCreate` where it's initialized. This PR initializes it in the application constructor so it can be used throughout the whole app lifecycle.


shobhitagarwal1612 (approved):
Thanks!

PR #731, opened 2021-02-03
Rx refactors + annotations

scolsen:
Refactors and annotates a few RX constructs in Offline Areas and repositories.


gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -101,9 +101,20 @@ private double tileSourceStorageSize(TileSource tileSource) {
   /**
    * Removes the offline area associated with this viewmodel from the device by removing all tile
    * sources that are not included in other areas and removing the area from the db.
+   *
+   * <p>Terminates the upstream click processor so that the resulting stream completes.
    */
-  public Completable onRemoveClick() {
-    return offlineBaseMapRepository.deleteArea(this.offlineArea.getValue().getId());
+  @Hot(terminates = true)
```
scolsen, line 107: @gino-m This is the `Hot` terminating case I mentioned in our last standup--I was slightly inaccurate, as the reason the click stream *must* complete is because we `flatMap` into a Completable which will *only* complete granted both the returned completable *and* its upstream complete.

I think it makes sense to model this as a completable, since we're ultimately calling a completable DB operation, but doing so forces us to call `complete` on the click processor to in turn propagate the `complete` to the completable.

The old way this was written (more imperative style) also works; but anecdotally I've found that this approach actually makes the UI run faster (I guess because the call no longer blocks on the UI thread VS the some amount of blocking imposed by the original imperative solution?)

scolsen:
work toward #699

scolsen:
@gino-m ready for a review whenever

scolsen:
@gino-m fixed some PMD issues, should be good now!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -101,9 +101,20 @@ private double tileSourceStorageSize(TileSource tileSource) {
...
+  public Completable onRemoveArea() {
+    return this.removeClickProcessor.flatMapCompletable(
```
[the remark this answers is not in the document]
gino-m: There should only be one "remove result" stream, not one for every click to `onRemoveArea`. In that case, we'd want to define this stream in the constructor as something like:

```java
    disposeOnClear(
        this.removeClickProcessor
            .switchMapCompletable(
                __ ->
                    offlineBaseMapRepository.deleteArea(
                        Objects.requireNonNull(this.offlineArea.getValue(), "Null area").getId()))
            .subscribe(this::navigateUp));
```

Does that make sense?

gino-m (changes requested):
Initial thoughts / suggestions. Please take a look and we can do a deep dive over VC if it's helpful.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +45,43 @@
...
 
   @Inject
   public OfflineBaseMapViewerViewModel(
       OfflineBaseMapRepository offlineBaseMapRepository, @ApplicationContext Context context) {
+    this.argsProcessor = BehaviorProcessor.create();
+    this.removeClickProcessor = PublishProcessor.create();
     this.offlineBaseMapRepository = offlineBaseMapRepository;
     this.context = new WeakReference<>(context);
+    @Hot
+    Flowable<OfflineBaseMap> offlineAreaInternal =
```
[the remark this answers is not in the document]
scolsen: Gotcha, it's is a `Single` under the hood, and in fact even the args processor only needs to be a `SingleSubject`, we only need a `Flowable` to satisfy `LiveDataReactiveStreams.fromPublisher()`-- I've changed that to try to signal what's going on better and  renamed this `offlineAreaItemAsFlowable` (to signify it's not actually a sequence of multiple items).

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +45,43 @@
 public class OfflineBaseMapViewerViewModel extends AbstractViewModel {
 
   @Hot(replays = true)
-  private final FlowableProcessor<OfflineBaseMapViewerFragmentArgs> argsProcessor =
-      BehaviorProcessor.create();
+  private final BehaviorProcessor<OfflineBaseMapViewerFragmentArgs> argsProcessor;
+
+  @Hot private final PublishProcessor<Nil> removeClickProcessor;
```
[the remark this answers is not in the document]
scolsen: good call, thanks!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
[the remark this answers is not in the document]
scolsen: sgtm, thanks!

scolsen:

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google LLC
+ * Copyright 2021 Google LLC
```
gino-m, line 2: Please keep original copyright date on files. Optionally, you could write `Copyright 2019-2012...`

Same throughout.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
+import timber.log.Timber;
 
 /**
  * View model for the offline area manager fragment. Handles the current list of downloaded areas.
  */
 public class OfflineBaseMapsViewModel extends AbstractViewModel {
 
-  private LiveData<ImmutableList<OfflineBaseMap>> offlineAreas;
-  private LiveData<Integer> noAreasMessageVisibility;
+  @Cold(terminates = false, errors = false)
```
gino-m, line 37: Thinking out loud, I wonder if we shouldn't omit these qualifiers for `LiveData`, since they don't actually carry error or termination states. If anything they would apply to the upstream Flowable. Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
+
   private final Navigator navigator;
 
   @Inject
   OfflineBaseMapsViewModel(Navigator navigator, OfflineBaseMapRepository offlineBaseMapRepository) {
     this.navigator = navigator;
-    this.offlineAreas =
-        LiveDataReactiveStreams.fromPublisher(
-            offlineBaseMapRepository.getOfflineAreasOnceAndStream());
+    Flowable<ImmutableList<OfflineBaseMap>> offlineAreas =
```
gino-m, line 48: Throwing this out there for future work: What if we roll the `doOnError`, `onErrorReturnItem`, and `fromPublisher` into a `Converter` returned by a function, something like `handleErrors`. Then we could do use it to log and should a message in the UI in a consistent way, e.g.
```java
       offlineBaseMapRepository
            offlineBaseMapRepository.getOfflineAreasOnceAndStream())	            
              .getOfflineAreasOnceAndStream()
              .compose(handleErrors("Error loading base map", ImmutableList.of()));
```

`handleErrors` could both log the exception to`Timber.e` with the provided message, and show a toast. If that's helpful I can create a bug and do this in a followup.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
+                        throwable,
+                        "Unexpected error accessing offline basemaps in the local store."))
+            .onErrorReturnItem(ImmutableList.of());
+    this.offlineAreas = LiveDataReactiveStreams.fromPublisher(offlineAreas);
     this.noAreasMessageVisibility =
         LiveDataReactiveStreams.fromPublisher(
-            offlineBaseMapRepository
-                .getOfflineAreasOnceAndStream()
-                .map(baseMaps -> baseMaps.isEmpty() ? View.VISIBLE : View.GONE));
+            offlineAreas.map(baseMaps -> baseMaps.isEmpty() ? View.VISIBLE : View.GONE));
```
gino-m, line 60: A new view helpers could also simplify this:
            offlineAreas.compose(visibleIf(baseMaps -> baseMaps.isEmpty()));

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
   }
 
+  /** Navigate to the offline area selector UI from the Offline basemaps UI. */
```
gino-m, line 63: Nit: Offline -> offline

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
   public void showOfflineAreaSelector() {
     navigator.navigate(OfflineBaseMapsFragmentDirections.showOfflineAreaSelector());
   }
 
+  /**
+   * Returns an infinite stream of downloaded offline basemaps. If an unexpected error accessing the
+   * local store is encountered, emits an empty list downstream, circumventing the error.
+   */
```
gino-m, line 71: This sounds a bit abstract, and not sure "emits downstream( it's 100% accurate. How about something like this?

> Emits the current list of areas available for offline viewing, or an empty list if an error occurs.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
```diff
@@ -21,41 +21,64 @@
...
+  @Cold(terminates = false)
   LiveData<ImmutableList<OfflineBaseMap>> getOfflineAreas() {
     return offlineAreas;
   }
 
+  /**
+   * Returns the visibility of a no area message based on the current number of available offline
+   * basemaps.
+   */
```
gino-m, line 80: Nit: "no area" in quotes

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorFragment.java
```diff
@@ -56,24 +57,15 @@ public static OfflineBaseMapSelectorFragment newInstance() {
...
-        navigator.navigateUp();
-        break;
-    }
+    viewModel
+        .getDownloadMessages()
+        // Since we pop a toast, we need to observe on the main (UI) thread.
+        // Otherwise this subscription handler will trigger before Looper.prepare() has been called.
+        .observeOn(AndroidSchedulers.mainThread())
+        .as(autoDisposable(this))
+        .subscribe(this::onDownloadMessage);
```
gino-m, line 68: `LiveData` by default replays the last value on configuration changes. Since this isn't what we want in this case, we should use an `Observable` here instead.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorFragment.java
```diff
@@ -99,18 +91,32 @@ public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
...
     if (map == null) {
       return;
     }
 
-    viewModel.onDownloadClick(
-        map.getViewport(),
-        map.getCurrentZoomLevel(),
-        getContext().getString(R.string.unnamed_area));
+    viewModel.downloadBaseMap(
+        map.getViewport(), getContext().getString(R.string.unnamed_area));
```
gino-m, line 106: Can the VM inject `Resources` and get the string resource itself instead of passing here?

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorViewModel.java
```diff
@@ -38,39 +36,39 @@
...
-        LiveDataReactiveStreams.fromPublisher(
-            downloadClicks.switchMapSingle(
-                baseMap ->
-                    offlineBaseMapRepository
-                        .addAreaAndEnqueue(baseMap)
-                        .toSingleDefault(DownloadMessage.STARTED)
-                        .onErrorReturn(this::onEnqueueError)
-                        .map(Event::create)));
+    this.downloadMessage =
+        baseMapDownloads
```
gino-m, line 50: I find the interaction between values and completion in `baseMapDownloads`, the `Completable` returned by `addAreaAndEnqueue`, and the default value set by `toSingleDefault` hard to grok. Would things be simpler if we didn't rely on termination at all, and just used an `Observable` for `downloadMessage`? We can and should just debounce (or rather, only take the first) download button click.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +48,64 @@
...
     this.context = new WeakReference<>(context);
-    this.areaName =
-        LiveDataReactiveStreams.fromPublisher(
-            this.argsProcessor.switchMap(
+    this.navigator = navigator;
+    @Hot
+    // We only need to convert this single to a flowable in order to use it with LiveData.
+    // It still only contains a single offline area returned by getOfflineArea.
+    Flowable<OfflineBaseMap> offlineAreaItemAsFlowable =
+        this.fragmentArgs
```
gino-m, line 75: Can we refactor to avoid this deep nesting?

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +48,64 @@
...
+                .flatMap(
+                    offlineBaseMapRepository::getIntersectingDownloadedTileSourcesOnceAndStream)
+                .map(this::tileSourcesToTotalStorageSize));
+    this.offlineArea = LiveDataReactiveStreams.fromPublisher(offlineAreaItemAsFlowable);
+    disposeOnClear(
+        removeAreaClicks
+            .first(Nil.NIL)
+            .flatMapCompletable(
+                __ -> {
+                  if (this.offlineArea.getValue() == null) {
```
gino-m, line 102: I believe using `area = Objects.requireNonNull(this.offlineArea.getValue(), "Could not remove.....");` would have the same effect here.

gino-m (changes requested):
Hi @scolsen, so sorry this has taken me so long to review. The past few weeks have been challenging on a number of fronts. Most of my suggestions are more considerations for future work that I'm mentioning while the context is fresh. PTAL and let me know what you think?

gnd/src/main/java/com/google/android/gnd/persistence/local/LocalDataStore.java
scolsen: Whoops, for some reason I thought we needed to update theses. Fixed! 5ed2f340

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
scolsen: Sounds useful to me! I created a bug: https://github.com/google/ground-android/issues/743

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorFragment.java
scolsen: I'm missing something here, I think. The stream we're looking at `getDownloadMessages` is a `Single`, not LiveData.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
scolsen: sgtm!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/OfflineBaseMapsViewModel.java
scolsen: Indeed, esp. since we're not dealing w/ streams at this point. Fixed, thanks!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorViewModel.java
scolsen: I do think it would make it simpler in a certain sense, but I also feel using an `Observable` for this purpose and only taking the first emission is sort of like a square peg round hole type of situation.

From an outside reader's perspective w/o context, I actually feel using an `Observable` that we only pull a single emission from would give me more pause, I'd likely think "why not just use a single?" and then we'd have special explanation around why we're using less specific types, etc.

What I do think makes sense here is using a SingleSubject for the `baseMapDownloads` instead of a Publisher, as then we wouldn't need to call `complete` explicitly. Maybe we can see if that makes it clearer? We'd effectively get the debouncing behavior for free since late observers just receive the initial terminal event. (the behavioral effect of this iiuc, would be that if a user clicks download, has insane finger speed and beats the UI and quickly goes to another area and clicks it again, only the first area will be downloaded).

Lemme take another stab at it using these types and see if I can simplify, if not we can swap em out for an observer.

scolsen: Ok, I tried this out, but admittedly it's not that different, we only remove the explicit call to `complete` and the SingleSubject better communicates that we only care about the first basemap selected, so lmk if you feel strongly about using an `Observable` instead here and we can make the switch.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
scolsen: done, thanks!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
scolsen: Thanks! Unfortunately our min API is too low to use the form of that function that takes the string argument, but letting it throw the default null pointer exception and catching the error works just as well.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorViewModel.java
gino-m: I think in this case the question of whether to use `Single` or `Observable` comes down to levels of abstraction, or more specifically, whether the VM is aware of / models the fact that it's a single-use object. Also, experience with Rx in Android so far has shown that at the level of lifecycle-aware objects, it's almost always better to make no or few assumptions about the order in which things happen. It's also slightly easier to reason about "the VM will emit download status messages for the View to display" than "the VM will emit a single message when the download is complete". How/when the messages are generated could be considered a detail of the current flow; adding the ability a message when the download starts wouldn't break the former contract, but it would break the latter.

tl;dr: I'm wary of storing references to streams that terminate inside life-cycle aware objects due additional complexity due to lifecycle x termination state. I can't think of a good example now, so if you feel strongly please feel free to keep it for now till I can think of one!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/selector/OfflineBaseMapSelectorFragment.java
scolsen: Done, thanks!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +48,64 @@
 public class OfflineBaseMapViewerViewModel extends AbstractViewModel {
 
   @Hot(replays = true)
-  private final FlowableProcessor<OfflineBaseMapViewerFragmentArgs> argsProcessor =
-      BehaviorProcessor.create();
+  private final SingleSubject<OfflineBaseMapViewerFragmentArgs> fragmentArgs =
```
gino-m, line 49: Are we sure the view model won't be used more than once w/different args? If not it might be better to model this as an `Observable` (see previous comment).

scolsen:
@gino-m I think it's ready for another look!

gino-m (changes requested):
Looks great so far, thanks for all the edits/comments. Mostly a few suggestions, please take a look and I'll approve ASAP!

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
scolsen: Switch to `PublishSubject`.

gnd/src/main/java/com/google/android/gnd/ui/offlinebasemap/viewer/OfflineBaseMapViewerViewModel.java
```diff
@@ -42,46 +46,50 @@
...
+      SingleSubject.create();
+
+  @Hot private final SingleSubject<Nil> removeAreaClicks = SingleSubject.create();
```
scolsen, line 52: `PublishSubject`

gino-m (approved):

gino-m:
🥳 🎉 !

PR #732, opened 2021-02-04
[Photo] Add label to photo type fields

### LINKED ISSUE #684: [Add/edit observation] Field labels not showing for photo fields
@shobhitagarwal1612 Could you confirm that the labels show about photo fields? I inferred they might not from screenshots in your recent PR. Thanks!

shobhitagarwal1612:
Fixes #684

Adds label and background to photo fields to make it consistent with other input types.

@gino-m  PTAL?


gino-m:
Thanks @shobhitagarwal1612 for adding the field labels! What does it look like after a photo is added?

@coreyleamon Could you PTAL and advise on UX?

shobhitagarwal1612:
Also removed the material card.

Screenshot:

coreyleamon:
I would actually recommend a different treatment for images, rather than loading an image inside a text input. We'll also need to consider if an image has representation in the observation summary screen. I can mock something up in the next couple days.

gino-m:
> I would actually recommend a different treatment for images, rather than loading an image inside a text input. We'll also need to consider if an image has representation in the observation summary screen. I can mock something up in the next couple days.

@coreyleamon, that would be super helpful! I've create #738 to track and to unblock the incremental improvements made here. Thanks for taking a look!

gino-m:
@coreyleamon Gentle ping in case you're able to comment on this. If you can't no worries at all, just let me know and I'll look for precedent in similar Android Material apps!

gino-m:
Shobhit isn't currently available, so merging this PR for now. We can refine the UI details going forward. 

@coreyleamon Could you please create a feature request with mocks for the updated photo field treatment?

gino-m (approved):

PR #733, opened 2021-02-04
[Rx Cleanup] Terminate photo picker/capture observable

### LINKED ISSUE #726: [Photo capture] Report errors when capturing a photo
The `Observable` returned by `CameraManager.onCapturePhotoResult()` currently doesn't terminate when the result of the request was not OK or when no data is returned.  Instead, an error should be thrown, which is caught by the UI and reported in a toast.

@shobhitagarwal1612 Could you PTAL?

shobhitagarwal1612:
Fixes #726 

 * Adds tests to verify that streams are terminated

@gino-m  PTAL?


gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -73,30 +75,26 @@ private Completable sendCapturePhotoIntent() {
   public Observable<Bitmap> capturePhotoResult() {
     return activityStreams
         .getNextActivityResult(CAPTURE_PHOTO_REQUEST_CODE)
-        .flatMap(this::onCapturePhotoResult);
+        .flatMapMaybe(this::onCapturePhotoResult);
```
gino-m, line 78: Should this method maybe return a `Maybe` instead?

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -88,23 +90,26 @@ private Completable sendPhotoPickerIntent() {
   public Observable<Bitmap> photoPickerResult() {
     return activityStreams
         .getNextActivityResult(PICK_PHOTO_REQUEST_CODE)
-        .flatMap(this::onPickPhotoResult)
-        .map(uri -> Media.getBitmap(context.getContentResolver(), uri));
+        .flatMapMaybe(this::onPickPhotoResult)
```
gino-m, line 93: Please annotation this method with `@Hot`.

gino-m (changes requested):
Great job on the refactors and especially on the unit tests. A few nits, otherwise LGTM.

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
shobhitagarwal1612: Done

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
```diff
@@ -71,11 +70,12 @@ private Completable sendCapturePhotoIntent() {
   }
 
   /** Emits the result of the photo capture request. */
-  @Hot
-  public Observable<Bitmap> capturePhotoResult() {
+  @Hot(terminates = true)
+  public Maybe<Bitmap> capturePhotoResult() {
```
gino-m, line 74: Sorry if I created some confusion with my last comment. I think the API of this class could be simplified. Rather than sending the photo capture request and capturing it via separate methods, would it be possible to expose a single `@Cold Maybe<Bitmap> capturePhoto()` that launches the intent and waits for the result?

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
```diff
@@ -66,6 +67,7 @@ public StorageManager(
    * Requests for selecting a photo from the storage, if necessary permissions are granted.
    * Otherwise, requests for the permissions and then sends out the request.
    */
+  @Cold
```
gino-m, line 70: I may have missed this in previous iterations, but it appears this class isn't really necessary; that the photo picker logic should live in `CameraManager`, and the remote storage logic should live in `RemoteStorageManager`. Please correct me if I'm wrong!

gino-m (changes requested):
Two more general suggestions/comments. Thanks!

gnd/src/main/java/com/google/android/gnd/system/StorageManager.java
shobhitagarwal1612: I don't agree that this code should be moved to `CameraManager`. For photo picker, we don't need camera permissions or launch camera app. Only READ_EXTERNAL_STORAGE is needed and an intent is started to open any supported gallery/file manager app.

Refactored the changes related to `RemoteStorageManager` as suggested.

gnd/src/main/java/com/google/android/gnd/system/CameraManager.java
shobhitagarwal1612: Thanks for the suggestion. Changes applied.

gino-m (approved):
@shobhitagarwal1612 Sorry it took me so long to get back to this, great work. LGTM!

PR #740, opened 2021-02-16
[Add observation] Display bottom sheet instead on adding a new feature

### LINKED ISSUE #694: [Add observation] User is redirected to map after adding feature + observation
**To Reproduce**

* Add a feature and observation, click save
* User is returned to map instead of feature panel. User should instead see feature panel with new observation.

@shobhitagarwal1612 PTAL?

### LINKED ISSUE #695: [Add observation] Dialog reappears unexpectedly on config change
**To Reproduce**
1. Add a feature and observation, click save
2. User is returned to map instead of feature panel (see #694).
3. Rotate the device

Form reappears. Closing the form causes the app to force quit.

@shobhitagarwal1612 PTAL?

shobhitagarwal1612:
Fixes #694, Fixes #695

Instead of redirecting to a new observation, open the bottom sheet fragment.

Verified by following the steps mentioned in the bug.

@gino-m PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
@@ -48,7 +48,6 @@ import com.google.android.gnd.databinding.HomeScreenFragBinding;
 import com.google.android.gnd.model.Project;
 import com.google.android.gnd.model.feature.Feature;
 import com.google.android.gnd.model.feature.Point;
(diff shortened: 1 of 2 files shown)


scolsen (approved):
Looks great!

gino-m (changes requested):
Hi @shobhitagarwal1612! This is how the behavior was previously, before we updated the flow to transition directly to "Add observation" after a feature was added. The problem in #694 is a different one; the user should see the bottom sheet after _adding the observation_, not right after adding the feature. Could you PTAL?

PR #741, opened 2021-02-16
[Bug] Fix crashes on rotating ObservationDetails screen

shobhitagarwal1612:
Fixes app crashing on rotating ObservationDetails screen.

Steps to verify:
 * Add a new feature and observation.
 * Click on the added observation in the bottom sheet to open ObservationDetails screen
 * Rotate the device two times

@scolsen PTAL?


scolsen:
The rotation works fine, but when I click the back arrow (in the observation sheet, not android's native one) the app occasionally crashes, and if it doesn't, the map fragment switches to a different map mode (normal from hybrid) and doesn't render any observations-- @shobhitagarwal1612  any ideas why this might be happening/can you see it on your end?

gino-m:
> The rotation works fine, but when I click the back arrow (in the observation sheet, not android's native one) the app occasionally crashes, and if it doesn't, the map fragment switches to a different map mode (normal from hybrid) and doesn't render any observations-- @shobhitagarwal1612 any ideas why this might be happening/can you see it on your end?

Hi @scolsen ! Just circling back now. If you can repro, could you share the stack trace of the crash when you get a sec?

gino-m:
@scolsen Were you ever able to repro the bug you mentioned? I'm seeing a similar bug even without this fix, so I'll go ahead and merge this for now.

gino-m (approved):

gino-m:
@shobhitagarwal1612 Seems `testStagingUnitTest` is failing. Any chance you could take a look if/when you have a spare moment?

gino-m:
Please disregard, looks like that was a stale run!

PR #742, opened 2021-02-16
[UX] Open Github on clicking "Send feedback" under settings

shobhitagarwal1612:
* Uses custom tabs for opening urls
 * Opens the page to create new bugs / feature request on clicking the "send feedback" button

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/res/values/strings-untranslated.xml
+++ b/gnd/src/main/res/values/strings-untranslated.xml
@@ -21,5 +21,6 @@
   <string name="app_name" translatable="false">Ground</string>
   <string name="text_separator" translatable="false">·</string>
   <string name="ground_website" translatable="false">https://groundplatform.org/</string>
+  <string name="ground_android_github" translatable="false">https://github.com/google/ground-android/issues/new/choose</string>
 
 </resources>
--- a/gnd/build.gradle
+++ b/gnd/build.gradle
@@ -181,6 +181,9 @@ dependencies {
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'androidx.preference:preference:1.1.1'
 
+    // Custom tabs
+    implementation "androidx.browser:browser:1.3.0"
+
     // Stetho - A debug bridge for Android applications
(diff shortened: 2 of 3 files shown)


gnd/src/main/res/values/strings-untranslated.xml
```diff
@@ -21,5 +21,6 @@
   <string name="app_name" translatable="false">Ground</string>
   <string name="text_separator" translatable="false">·</string>
   <string name="ground_website" translatable="false">https://groundplatform.org/</string>
+  <string name="ground_android_github" translatable="false">https://github.com/google/ground-android/issues/new/choose</string>
```
scolsen, line 24: Maybe we should default to **bug report**? I suppose users could also want to request a feature, but they definitely won't report code health issues. I think reporting bugs is the most likely use case.

scolsen (approved):

scolsen:
I re-triggered the continuous build since it seemed to potentially be a transient error (gradle daemon crashed).

gnd/src/main/res/values/strings-untranslated.xml
shobhitagarwal1612: @gino-m Please share your thoughts too. My rationale behind this was to promote users to suggest features in addition to reporting bugs.

gino-m (changes requested):
Not so sure how useful this feature will be when working offline. Also, users need to have a GitHub account + permission to create issues in our GitHub repo. Is there another mechanism we can use to take feedback? In particular, we'd like to make it easy for users to report bugs, providing minimal friction.

shobhitagarwal1612:
That's a great point! We could instead take the feedback directly from the app and send it to a common email whenever internet gets available.

shobhitagarwal1612:
@gino-m @scolsen How about saving the feedback to a new firebase table similar to how responses are synced?

shobhitagarwal1612:
Closing this for now. Please reopen if this is still relevant

Issue #754, opened 2021-04-21
[Code health] Pass form metadata into deserializers to disambiguate data types


gino-m:
Currently, both local and remote deserialization infer Ground field types from the data types used in the persistent representation. This is fragile, limits backward compatibility, and is blocking new field types from being created without temporary workarounds (eg #748). Passing `Form`, `Element`, or `Field` into deserializers will unblock new field types using the same persistent representation (e.g., `string`, `long`, etc.), and will allow for better error checking (checking form definition against what's in the dbs). In the future, this will also allow backward compatible changes to persistent representations (e.g., replacing `string` with `long` for dates in the dbs if desired). @HemanParbhakar @atalekar  Fyi.

PR #757, opened 2021-04-21
[Code health] Use form definition when deserializing observations from remote db

gino-m:
Contributes to #754. This will be used to allow date, time, and other Ground form field types to use the same remote and local representation as other field types.

@atalekar @HemanParbhakar PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ObservationConverter.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ObservationConverter.java
@@ -42,10 +42,10 @@ class ObservationConverter {
       throws DataStoreException {
     ObservationDocument doc = snapshot.toObject(ObservationDocument.class);
     String featureId = checkNotNull(doc.getFeatureId(), "featureId");
-    String formId = checkNotNull(doc.getFormId(), "formId");
     if (!feature.getId().equals(featureId)) {
       throw new DataStoreException("Observation doc featureId doesn't match specified feature id");
     }
+    String formId = checkNotNull(doc.getFormId(), "formId");
     Form form = checkNotEmpty(feature.getLayer().getForm(formId), "form " + formId);
     // Degrade gracefully when audit info missing in remote db.
     AuditInfoNestedObject created =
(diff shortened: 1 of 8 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ObservationConverter.java
```diff
@@ -54,32 +56,61 @@ static Observation toObservation(Feature feature, DocumentSnapshot snapshot)
...
+      try {
+        Object obj = docResponses.get(fieldId);
+        putResponse(fieldId, form, obj, responses);
+      } catch (DataStoreException e) {
+        Timber.d("Field " + fieldId + "in remote db in observation " + observationId + ": " + e);
       }
     }
     return responses.build();
   }
+
```
atalekar, line 81: many of these classes are too low level to not have 100% unit test coverage. LGTM in general but strongly recommend having good unit test coverage here. In future across release if there are changes in this section, one will need to be quite a bit careful around backward compatibility etc.

atalekar (dismissed):

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/ObservationConverter.java
gino-m: @atalekar Very good points. We started out working under the assumption that everything we're building is a throw-away prototype/demo, so we started without test coverage. The initial assumption probably no longer holds, but our 20% team is so small that adding tests retroactively wasn't feasible. That said, I agree we probably shouldn't kick the can further down the road. I'll try to add some tests for this class!

gino-m: Tests still a WIP, but added basic skeleton to get started. I'll flesh them out tomorrow, but please feel free to take a look at the last few commits now if you have a cycle or two. Thanks!

atalekar: lgtm - thanks a lot!

gino-m (commented):
100% coverage of `ObservationConverter`. RFAL!

atalekar (approved):

gino-m:
@atalekar Sorry I forgot to push the PMD error fixes, so your LGTM got invalidated. Could you kindly restamp?

atalekar (approved):

PR #760, opened 2021-04-26
Expect "text_field" for text fields instead of "text" in remote db

### LINKED ISSUE #759: [Add/edit Observation] Form appears blank
Click "Add observation" on any feature:
Screenshot_1619471203

Likely caused by recent merge #757.

gino-m:
Fixes #759.

@scolsen Could you PTAL?


scolsen:
I'm not sure that I fully understand the root cause of the bug based on reading the issue (am I right in interpreting the problem stemmed from some call to `text_field.toUppercase()` resulting in the wrong type assignment? (Also wasn't aware of our Enums class and not entirely sure what that does either)), but otherwise the changes are sound on their own and lgtm.

scolsen (approved):

PR #762, opened 2021-04-26
[Local persistence] Call observation response converters directly rather than via Room type converters

gino-m:
Contributes to  #754. . This change is required to allow us to pass in form metadata in a subsequent PR.

Depends on #760. Please review that PR first.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/LocalDatabase.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/LocalDatabase.java
@@ -20,7 +20,6 @@ import androidx.room.Database;
 import androidx.room.RoomDatabase;
 import androidx.room.TypeConverters;
 import com.google.android.gnd.Config;
-import com.google.android.gnd.persistence.local.room.converter.ResponseDeltasTypeConverter;
 import com.google.android.gnd.persistence.local.room.converter.StyleTypeConverter;
 import com.google.android.gnd.persistence.local.room.dao.FeatureDao;
 import com.google.android.gnd.persistence.local.room.dao.FeatureMutationDao;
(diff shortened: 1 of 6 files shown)


gino-m:
> Changes look good. To summarize the end goal: we're effectively trying to decouple representations across different layers right? Am I correct in thinking the idea here is that our DB representation, remote store representation, and in-memory (view model) representations will all be independent and managed using custom conversions?
> 
> Was the issue here that `ResponseMap` proved too inflexible and we'd rather transform it into a literal JSON string, after which we can parse out the primitive types we need?

Hi @scolsen! We already do a fairly good job of separating the different representations. This PR is actually working towards #754. The idea is that when deserializing form responses in local and remote db, we shouldn't guess the form field type from the data type used in the db, but rather we should look at the form definition in `Project`. The current approach of mapping data store types to field types is fragile, and it breaks once field types<->data types are no longer 1:1, as is the case with "time" fields, which will also be strings in the form "HH:MM".

scolsen:
> > Changes look good. To summarize the end goal: we're effectively trying to decouple representations across different layers right? Am I correct in thinking the idea here is that our DB representation, remote store representation, and in-memory (view model) representations will all be independent and managed using custom conversions?
> > Was the issue here that `ResponseMap` proved too inflexible and we'd rather transform it into a literal JSON string, after which we can parse out the primitive types we need?
> 
> Hi @scolsen! We already do a fairly good job of separating the different representations. This PR is actually working towards #754. The idea is that when deserializing form responses in local and remote db, we shouldn't guess the form field type from the data type used in the db, but rather we should look at the form definition in `Project`. The current approach of mapping data store types to field types is fragile, and it breaks once field types<->data types are no longer 1:1, as is the case with "time" fields, which will also be strings in the form "HH:MM".

Ah, I see, thanks for the explanation!

PR #765, opened 2021-04-28
[Code health] Use form definition when deserializing observation mutations in local db

### LINKED ISSUE #754: [Code health] Pass form metadata into deserializers to disambiguate data types

gino-m:
Closes #754. Depends on #764.


gino-m:
@HemanParbhakar Could you PTAL?

HemanParbhakar (approved):

HemanParbhakar (approved):

Issue #774, opened 2021-05-09
[Data collection] Support "capture location" data collection task


gino-m:
Capture "latitude, longitude, altitude in meters, and accuracy radius"

Show values in form. In the future we can visualize the location on the map and show calculated distance from feature location.

Steps:

- [ ] Add scaffolding and basic UI - #1861 
- [ ] Parse the task from Firebase
- [ ] Saving the data back to local and remote storage

scolsen:
@JSunde to update the status on this -- please reassign if needed!

JSunde:
How does this differ from the DropAPinTask? @gino-m

JSunde:
+@shobhitagarwal1612 for context

shobhitagarwal1612:
We haven't created this task type yet. The idea is it will open the map with gestures disabled. The pin will get automatically point to the user's current location and show the accuracy number somewhere it the UI.

JSunde:
@shobhitagarwal1612 do you mind taking this one since you worked on the DropAPinTask?

shobhitagarwal1612:
Sure

shobhitagarwal1612:
@gino-m Do we need this for the Oct code freeze?

gino-m:
Hi @shobhitagarwal1612! Are you still working on this one? I'm creating several demos and realized this wasn't wired up yet. Lmk when it's ready and I'll add it to the demo survey!

PR #778, opened 2021-05-12
[DRAFT WIP] Add support for location fields

gino-m:
Contributes to #774.


gino-m:
Abandoning this change for now since it's no longer needed by pilots and other potentially better alternatives exist.

PR #783, opened 2021-05-14
[Map] Allow GeoJSON to be tappable

gino-m:
Builds on #782. 

Next steps in fast-follow PR(s):

* Parse GeoJSON at load time rather than rendering time.
* Use on shared GeoJSONLayer rather than one for each feature. Set parsed geometries on that layer.
* Let view model provide GeoJSON styles rather than setting them in MapAdapter.
* View  model updates GeoJSON styles to reflect whether feature is selected.

@scolsen This PR includes your commits as well. PTAL?


scolsen:
@googlebot I consent.

scolsen (approved):

PR #787, opened 2021-05-17
fix: stop unintended cursor movement on number fields

scolsen:
This commit fixes an issue whereby number fields would sporadically move
the cursor while adjusting values.

We move numeric conversions further down the stack and treat field
values as strings, additionally specifying the correct field type for
numeric field (numberDecimal)

@gino-m 

fixes #787


gino-m (approved):
At some point might be good to figure out why out domain class has to be a string. Overall LGTM, thanks for the quick fix!

PR #805, opened 2021-05-19
[Map] Prompt user to disambiguate on click of overlapping polygon features

scolsen:
This PR uses the S2 Geometry Library to add initial support for disambiguating feature taps, here's the behavior in action:

https://user-images.githubusercontent.com/11237600/118893724-e931de80-b8d0-11eb-8dd6-f5ec0efe4014.mov

## Limitations

Unfortunately, the maps API does not propagate clicks downwards through handlers. As a result, taps at higher zoom levels, even those that appear ambiguous will have a bias toward markers. We could force disambiguation on all marker clicks, but this would be a trade off at lower zoom levels, where the user would have to make a selection even for very precise marker clicks within a polygon bound.

Due to the way we currently represent geojson features, polygon bounds tend to be larger than one might expect. For example, in the video in this PR, clicking on another letter polygon would still trigger disambiguation between the polygon and the point, even though the loop representing this portion of the polygon is quite far away from the contained point—this is because we presently represent the entirety of the *json source* on the map as a single feature, not as individual features per feature entity in the geo json.


scolsen:
@gino-m

gino-m:
Hi @scolsen, thanks for the PR and clear explanation. My understanding was that the bounds would only be used to identify which polygons to check against `PolyUtil.containsLocation()`. In the video about, the behavior appears broken in that, as you mention, the user is asked whether they meant to tap a marker which isn't overlapping the tap target. My understanding was that we'd do the following:

1. When a polygon is added to the map, add it to a hash map keyed by the S2 cell number at some reasonably high zoom level. The map would be of type `Map<S2Cell, Set<GeoJsonPolygon>>` or similar. The polygon would be added to all cells that overlap with the bounding box of the polygon (include cells inside the polygon, not just those intersecting w/vertices).
2. When a user taps the map (anywhere but a marker), calculate the S2 cell of the tapped latlng and get polygons that intersect that cell.
3. Use `PolyUtil.containsLocation()` to determine which polygons intersect the tapped latlng.
4. Show these polygons in the disambiguation dialog.

In the case a marker is tapped, we'd do the same as above, but instead using the marker's latlng location. We would show the list of intersecting polygons plus the tapped polygon.

Does this make sense? Please lmkwyt!

scolsen:
> When a user taps the map (anywhere but a marker), calculate the S2 cell of the tapped latlng and get polygons that intersect that cell.

It's this part that proves problematic when using multiple click handlers--but this sounds like the right approach and I think we can do it this way if we drop the GMS json click handler! Just for posterity, here's why it's a problem when using the json click handler:

If the user clicks an *arbitrary* point on the map, this works, since we'll get a latlng. 

However, if the user clicks an actual GeoJson feature, we don't get a single latlng, we get an entire *Feature* (GMS.data, not ground), which contains *all the points* making up the corresponding polygon, not just the actual latlng that was clicked (unfortunately). This is why it made more sense to me initially to hash GeoJsonFeatures and their corresponding s2loops (the outer bounds of a polygon, ignoring holes). This means at really low zoom levels (e.g. like zoom 1, when the user is zoomed really far in), we're getting *the full extent* of the tapped feature, most of which might be off screen, and we have no way of knowing (afaict) where precisely the user tapped.

It really is a huge limitation when it comes to precision. It makes it tempting to just set a single map-wide click handler and maintain lists/queues of features and dispatch on our own, but then we'd have to handle zoom-level  adjustment/correction ourselves...which is also a mess.

Another middle ground, which might be our best option is to handle *geojson* clicks specifically ourselves, that is:

1. We parse GeoJson into either a mapping from GeoJsonFeature -> s2cells or s2loops.
2. We set an *onMapClick* handler instead of an onGeoJsonClick handler. The *onMapClick* handler gets the precise latlng the user tapped/clicked at: https://developers.google.com/maps/documentation/android-sdk/events#map_click_long_click_events (vs. the geojson handler which gets us an entire feature extent)
3. On map clicks we perform lookup ourselves and handling ourselves

This would give us the desired precision at lower zoom levels, and the behavior at higher zooms (zoomed out far) would remain the same, since maps seems to bias toward marker targets at high zoom anyway. 

So it sounds like we should drop the geojson click handler in favor of a map click handler and roll our own dispatch. I can try this out and see if it does the trick.

gino-m:
Hi @scolsen, that makes sense. From our last conversation I understood we were dropping the GeoJsonClickHandler for the reason you mention here. In that case, the only click handlers would be on map pins and map clicks, right?

scolsen:
> Hi @scolsen, that makes sense. From our last conversation I understood we were dropping the GeoJsonClickHandler for the reason you mention here. In that case, the only click handlers would be on map pins and map clicks, right?

Yep, that's right. I somehow missed the dropping the GeoJson handler part before pushing this PR, but the change shouldn't take too much effort.

gnd/src/main/java/com/google/android/gnd/ui/home/FeatureSelectorState.java
```diff
@@ -0,0 +1,45 @@
+package com.google.android.gnd.ui.home;
```
gino-m, line 1: Please add copyright header.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -108,18 +114,43 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
     viewModel.getUpdateFeatureResults().observe(this, this::onFeatureUpdated);
     viewModel.getDeleteFeatureResults().observe(this, this::onFeatureDeleted);
     viewModel.getErrors().observe(this, this::onError);
+    featureSelectorViewModel
+        .getFeatureSelections()
+        .as(autoDisposable(this))
+        .subscribe(this::onFeatureSelection);
+  }
+
+  private void onCandidateFeatures(ImmutableList<Feature> features) {
```
gino-m, line 139: Can we call this more closely related to the trigger of the event? e.g. `onTapOverlappingFeatures()`?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -261,4 +284,31 @@ public void showSettings() {
   public void onGeoJsonClick(MapGeoJson mapGeoJson) {
     showBottomSheet(mapGeoJson.getFeature());
   }
+
+  public Observable<ImmutableList<Feature>> getCandidateFeatures() {
```
gino-m, line 288: Same here re naming.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -261,4 +284,31 @@ public void showSettings() {
...
+  }
+
+  private void showFeatureSelector(ImmutableList<MapFeature> mapFeatures) {
+    ImmutableList<Feature> features =
+        stream(mapFeatures)
+            .map(
+                feature -> {
+                  if (feature instanceof MapPin) {
+                    Feature feat = ((MapPin) feature).getFeature();
+                    return feat;
```
gino-m, line 303: We don't need the intermediate variable `feat` here and below, you can just `return ((cast)) feature)...`

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,97 @@
+package com.google.android.gnd.ui.home.featureselector;
```
gino-m, line 1: Copyright header here

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Nullable private ArrayAdapter listAdapter;
+
+  public FeatureSelectorFragment(FeatureSelectorViewModel featureSelectorViewModel) {
+    this.viewModel = featureSelectorViewModel;
+  }
+
+  // public FeatureSelectorFragment(FeatureSelectorViewModel vm) {
+  //  this.viewModel = vm;
+  // }
```
gino-m, line 42: Please remove commented out code.

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorViewModel.java
```diff
@@ -0,0 +1,41 @@
+package com.google.android.gnd.ui.home.featureselector;
```
gino-m, line 1: Same here

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -327,6 +328,10 @@ public void onGeoJsonClick(MapGeoJson mapGeoJson) {
     // TODO: Move the camera and adjust styling appropriately.
   }
 
+  public void onFeatureClick(ImmutableList<MapFeature> mapFeatures) {
+    Timber.i("Got ambiguous feature click: %s", mapFeatures.size());
+  }
```
gino-m, line 333: Did you mean to remove this callback or is it still WIP?

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -167,6 +186,40 @@ private static LatLng toLatLng(Point point) {
...
+    if (loops == null) {
+      featureClicks.onNext(stream(candidates).collect(toImmutableList()));
+      return;
+    }
+
+    for (Marker marker : markers.getMarkers()) {
+      if (stream(loops).anyMatch(loop -> loop.contains(markerToS2Point(marker)))) {
+        candidates.add((MapFeature) marker.getTag());
+      }
+    }
```
gino-m, line 207: I think if the click was received by the map and not the marker, then we can assume the user wasn't trying to tap the marker. Is that right?

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -167,6 +186,40 @@ private static LatLng toLatLng(Point point) {
     return new LatLng(point.getLatitude(), point.getLongitude());
   }
 
+  // Handle taps on ambiguous features.
+  private void handleAmbiguity(MapGeoJson feature) {
+    List<S2Loop> loops = geoJsonPolygonLoops.get(feature);
+    Collection<MapFeature> candidates = new ArrayList<>();
```
gino-m, line 192: Instead of streaming to an immutable you can use `ImmutableList.builder()` and call `build()` at then end.

gino-m (changes requested):
Thanks, this looks like it was really tricky to get right! Two high-level comments:

* Can we import the S2 library in a separate PR?
* The algorithm you're using here seems different from what we discussed the other day; it appears to be built around S2Loop, and a click handler on the MapGeoJson rather than on an arbitrary point. Perhaps it'll be clearer when we discuss tomorrow!

scolsen:
@gino-m the map click handler approach isn't going to work, unfortunately. When I remove the geojson click handler the maps API still seems to bypass the map click handler entirely if one clicks in the bounds of a geojson polygon. 

This could be stemming from the fact that it seems to treat these polygons as the same things as the other maps polygons under the hood (one needs to add a polygon manager to handle clicks in the geojson layer, even). I don't know, but at any rate I can't get it to work. Maybe I'm doing something incorrectly, but the right approach is non-obvious at least.

https://user-images.githubusercontent.com/11237600/119441801-bb93cd80-bcf4-11eb-8349-b8b0f8bba864.mov

So, it seems like the map click handler approach is really all-or-nothing in terms of the way maps propagates (or rather doesn't) clicks across objects. 

I think at this point our best bet might be the following until a better solution crops up:

1. Keep the current approach using a geo json click handler.
2. Only disambiguate between overlapping polygons, on polygon taps.

scolsen:
Basically, it seems like the maps API will divert clicks to the right layer whether or not we have a handler for that layer, and it will *never* propagate clicks to another layer. So in this instance, it doesn't matter if we have alternative handlers. If the user taps an object, marker, polygon, json, w/e, Maps will *only* send the tap to that layers/object's management bits, ignoring all others.

In other words, we need to roll our own tap position getter if we want that amount of precision. Or I'm just doing something completely wrong. 

Is it somehow possible to "arrange" the layers so that the "map" layer is at the top? that would potentially solve our problem too, but then we'd need to handle markers as well.

scolsen:
https://user-images.githubusercontent.com/11237600/119577670-14ad4100-bd89-11eb-8299-0828030a6f60.mov

@gino-m ! All that remains is to remove the s2 import.

gino-m:
> @gino-m ! All that remains is to remove the s2 import.

Fantastic!!! 🥇 🗺️  👈  ❗ 

Please let me know when it's ready for another look! If you've address previous comments or they're not relevant please resolve or let me know and I'll do it. Great work!

scolsen:

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -95,8 +98,10 @@
   private BottomSheetBehavior<View> bottomSheetBehavior;
   private ProjectSelectorDialogFragment projectSelectorDialogFragment;
   private ProjectSelectorViewModel projectSelectorViewModel;
+  private FeatureSelectorViewModel featureSelectorViewModel;
```
gino-m, line 101: To make the purpose of this dialog clearer, would it be better to call it something like `FeatureDisambiguationDialog[ViewModel|Fragment` etc? It's wordy, but might be more self-documenting. Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -108,18 +113,45 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+    featureSelectorDialogFragment = new FeatureSelectorFragment(featureSelectorViewModel);
 
     viewModel = getViewModel(HomeScreenViewModel.class);
     viewModel.getProjectLoadingState().observe(this, this::onActiveProjectChange);
     viewModel
         .getShowAddFeatureDialogRequests()
         .observe(this, e -> e.ifUnhandled(this::onShowAddFeatureDialogRequest));
     viewModel.getBottomSheetState().observe(this, this::onBottomSheetStateChange);
+    viewModel
+        .getOverlappingFeatures()
```
gino-m, line 127: Naming nit: From the name it's not clear when this observable emits items; interpreted naively it would imply it always returns all overlapping features. Is there a more self-describing name we could use here?

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -108,18 +113,45 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
     viewModel.getUpdateFeatureResults().observe(this, this::onFeatureUpdated);
     viewModel.getDeleteFeatureResults().observe(this, this::onFeatureDeleted);
     viewModel.getErrors().observe(this, this::onError);
+    featureSelectorViewModel
+        .getFeatureSelections()
+        .as(autoDisposable(this))
+        .subscribe(this::onFeatureSelection);
+  }
+
+  private void onTapOverlappingFeatures(ImmutableList<Feature> features) {
```
gino-m, line 141: Modern mobile UX uses "tap", but Android uses "click". To be internally consistent I'd suggest "click".

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -266,8 +278,29 @@ public void showSettings() {
...
+                  if (feature instanceof MapPin) {
+                    return ((MapPin) feature).getFeature();
+                  } else if (feature instanceof MapGeoJson) {
+                    return ((MapGeoJson) feature).getFeature();
+                  }
+                  return null;
+                })
+            .filter(f -> f != null)
+            .collect(toImmutableList());
+    candidateFeatures.onNext(features);
```
gino-m, line 303: Since `candidateFeatures` is subscribed to by the same class, can we just get rid of it and call `onTapOverlappingFeatures` directly?

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,111 @@
...
+  private FeatureSelectorDialogBinding binding;
+
+  @Nullable private ArrayAdapter listAdapter;
+
+  public FeatureSelectorFragment(FeatureSelectorViewModel featureSelectorViewModel) {
+    this.viewModel = featureSelectorViewModel;
+  }
+
+  @Override
+  public @NotNull Dialog onCreateDialog(Bundle savedInstanceState) {
```
gino-m, line 57: `@NotNull` is the default in our codebase so not needed here.

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,111 @@
...
+    super.onCreateDialog(savedInstanceState);
+    AlertDialog.Builder dialog = new Builder(getContext());
+    dialog.setTitle("Select Feature");
```
gino-m, line 60: Please move into `strings.xml` and capitalize as "Select feature"

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  private void showFeatureList(ImmutableList<Feature> features) {
+    binding.listLoadingProgressBar.setVisibility(View.GONE);
+    checkNotNull(listAdapter, "listAdapter was null when attempting to show project list");
+
+    listAdapter.clear();
+    stream(features).map(this::setFeatureText).forEach(listAdapter::add);
+
+    if (features.size() == 1) {
```
gino-m, line 84: I assume the dialog doesn't flash open and close in this case? Since it's not entirely clear what would happen, it feels like this logic belongs in the caller rather than here.

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,111 @@
...
+      dismiss();
+      return;
+    }
+
+    binding.featureSelectorListView.setVisibility(View.VISIBLE);
+  }
+
+  private String setFeatureText(Feature feature) {
+    String text = "";
+    if (feature.isGeoJson()) {
```
gino-m, line 98: Please add a TODO to disambiguate different GeoJSON geometry types.

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
```diff
@@ -0,0 +1,111 @@
...
+      text = "Area\n";
```
gino-m, line 99: I'm leaning more towards being more explicit and consistent w/GIS terminology rather than trying to come up with more user friendly (but possibly incorrect) names, e.g.:

> Feature type: Polygon
> Layer: ____

Wdyt?

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -164,6 +173,37 @@ private static LatLng toLatLng(Point point) {
     return new LatLng(point.getLatitude(), point.getLongitude());
   }
 
+  // Handle taps on ambiguous features.
+  private void handleAmbiguity(LatLng latLng) {
+    Builder<MapFeature> candidates = ImmutableList.builder();
+    ArrayList<String> processed = new ArrayList<>();
```
gino-m, line 179: Declare vars using more generate interface until impl matters (ie `List<String> processed = new ArrayList<>();`

gino-m (changes requested):
Fantastic work, @scolsen! A few style/naming comments, but overall looking tip-top!! Congrats on getting this working!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
scolsen: I see your point, but it also feels unnecessarily narrow to me. I think it'd be ok to keep it more general since I think `FeatureSelector` still communicates what it is and it could also fit future scenarios (e.g. maybe we want to trigger it for some other future flow, like viewing a list of all features).

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
scolsen: The `Fragment` subscribes to it, not the view model. We could still remove it, but this would entail moving the locus of control over UI from the fragment into the VM, which I think we should avoid. This approach is also consistent with the way we pass other information across `map container -> home VM -> home fragement`

gnd/src/main/java/com/google/android/gnd/ui/home/featureselector/FeatureSelectorFragment.java
scolsen: Thanks!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -266,8 +278,29 @@ public void showSettings() {
...
   }
 
-  public void onGeoJsonClick(MapGeoJson mapGeoJson) {
-    showBottomSheet(mapGeoJson.getFeature());
+  public Observable<ImmutableList<Feature>> getOverlappingFeatures() {
+    return candidateFeatures;
+  }
+
+  public void onFeatureClick(ImmutableList<MapFeature> mapFeatures) {
+    showFeatureSelector(mapFeatures);
```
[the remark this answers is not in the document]
scolsen: Thanks!

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapAdapter.java
```diff
@@ -297,13 +334,29 @@ private void addMapGeoJson(MapGeoJson mapFeature) {
 
     layer.addLayerToMap();
 
-    layer.setOnFeatureClickListener(__ -> onGeoJsonClick(mapFeature));
+    for (GeoJsonFeature geoJsonFeature : layer.getFeatures()) {
```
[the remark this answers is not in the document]
scolsen: Done, thanks!

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
scolsen: However, I will use `LiveData`

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenViewModel.java
```diff
@@ -86,6 +91,9 @@
   @Hot(replays = true)
   private final MutableLiveData<Integer> addFeatureButtonVisibility = new MutableLiveData<>(GONE);
 
+  private final BehaviorSubject<ImmutableList<Feature>> candidateFeatures =
+      BehaviorSubject.create();
```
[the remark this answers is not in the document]
scolsen: Changed to `PublishSubject`

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
scolsen: I switched this to `LiveData` and retained the name, since it mirrors our other naming conventions.

gnd/src/main/res/layout/feature_selector_list_item.xml
```diff
@@ -0,0 +1,31 @@
...
+  android:layout_width="match_parent"
+  android:layout_height="match_parent">
+
+  <TextView
+    android:id="@+id/feature_name"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:padding="26dp"
+    android:textColor="@color/colorForeground"
+    android:textSize="16sp"/>
```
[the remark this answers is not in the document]
scolsen: Done!

scolsen:
@gino-m 

one-more-time

gino-m (approved):
LGTM. Ship it!🚀

PR #809, opened 2021-05-22
[Code health] Allow unhandled Rx errors to crash app

### LINKED ISSUE #808: [Code health] Crashlytics not reporting crashes in TerraBio project

gino-m:
Eating errors can cause the app to be an unstable state and masks potential problems. Allowing the app to crash will also allow us to pick capture errors in Crashlytics. Fixes #808.

@shobhitagarwal1612  Do you think this is a good idea?


shobhitagarwal1612 (approved):
I agree that this would definitely prevent unknown app states which may cause new problems. LGTM 👍

PR #810, opened 2021-05-22
Add script to swap out local credentials and keys

gino-m:
@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
How about creating a new directory next to the `/debug` which contains the actual credentials?
That way gradle will automatically choose the correct config based on current variant.

shobhitagarwal1612 (approved):
Let's merge it as is for the upcoming delivery. We can automate it later.

gino-m:
> Let's merge it as is for the upcoming delivery. We can automate it later.

Thanks @shobhitagarwal1612, this isn't required for the pilot, it was more for my own convenience. Imo this probably shouldn't be in the codebase anyway. Note that creating a separate build variant might not make sense either iiuc, since the only different would be to use different keys stored locally. The alternate keys are config specific to this use case, not a feature of Ground.

PR #825, opened 2021-05-25
[Map] GeoJSON feature support

### LINKED ISSUE #790: [Map] Increase border width of polygons when selected

gino-m:
Fixes #790. @kdyson FYI.

Normal:
Screenshot_1621912543

Selected:
Screenshot_1621912549

@scolsen Could you PTAL? 🤠


gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -131,17 +136,43 @@
...
 
+  private ImmutableSet<MapFeature> updateSelectedFeature(
+      ImmutableSet<MapFeature> features, Optional<Feature> selectedFeature) {
+    Timber.v("Updating selected feature style");
+    if (selectedFeature.isEmpty()) {
+      return features;
+    }
+    ImmutableSet.Builder updatedFeatures = ImmutableSet.builder();
+    String selectedFeatureId = selectedFeature.get().getId();
+    for (MapFeature feature : features) {
```
scolsen, line 161: It's not clear to me why we need to iterate here. Why wouldn't we just pass the feature directly to the VM via the click handler using an RX stream? e.g., if we pass via `selectedFeature`, wouldn't we just handle such a side-effect in a subscription?

```
selectedFeature.onNext(feat -> if isinstanceOf JSON setStroke.......etc.)
```

The only reason we maybe wouldn't want to do that is to avoid using old values (updated happened on the server somewhere so the repo contains new data post mutation), but this doesn't solve that case either, as to get around that we'd want to request a fresh copy from the DB, right?

In other words, wouldn't a simpler model leveraging RX be:

MapAdapter[onClick { featureProcessor.OnNext(someFeature)}] -> MapFragment[subscribe(Adapter.featureProcessor).onNext(change styles).emitToVMconnectingStream] -> VM.vmConnectingStream.onNext(do something w/ feature data)

or something like that

gino-m: Each emission in the stream of features coming from the data store contains the full set of features to show on the map. We're effectively trying to modify one item in that stream if it's selected; the `updateSelectedFeature` function does this for us. There isn't a stream of individual features to modify, so emitting a single value wouldn't help us here (unless I'm misunderstanding something, which is entirely possible!)

gino-m: @scolsen Since we spoke earlier, is it ok to proceed to merge? If so can I get a restamp?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -151,6 +159,30 @@ private void onProjectChange(Optional<Project> project) {
...
+      if (feature instanceof MapGeoJson) {
+        MapGeoJson geoJsonFeature = (MapGeoJson) feature;
+        String geoJsonFeatureId = geoJsonFeature.getFeature().getId();
+        if (geoJsonFeatureId.equals(selectedFeatureId)) {
+          Timber.v("Restyling selected GeoJSON feature " + selectedFeatureId);
+          // TODO: Make width configurable.
+          updatedFeatures.add(geoJsonFeature.toBuilder().setStrokeWidth(16).build());
```
scolsen, line 177: Add a resource `default_geo_json_selected_stroke_width`? or something similar

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
```diff
@@ -193,6 +225,7 @@ private static MapGeoJson toMapGeoJson(GeoJsonFeature feature) {
         .setId(feature.getId())
         .setGeoJson(jsonObject)
         .setStyle(feature.getLayer().getDefaultStyle())
+        .setStrokeWidth(4) // TODO: Make configurable.
```
scolsen, line 228: Same here, add a resource for default width?

scolsen (approved):

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
gino-m: Done!

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/MapContainerViewModel.java
gino-m: Done.

Issue #828, opened 2021-05-25
[Add/Edit Observation] App crash when clicking "X"


gino-m:
```
    Process: com.google.android.gnd, PID: 25267
    java.lang.IllegalArgumentException: offset must be greater than or equal to 0
        at com.google.android.material.bottomsheet.BottomSheetBehavior.setExpandedOffset(BottomSheetBehavior.java:864)
        at com.google.android.gnd.ui.home.HomeScreenFragment.onGlobalLayout(HomeScreenFragment.java:240)
        at android.view.ViewTreeObserver.dispatchOnGlobalLayout(ViewTreeObserver.java:945)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2250)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1392)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6752)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:911)
        at android.view.Choreographer.doCallbacks(Choreographer.java:723)
        at android.view.Choreographer.doFrame(Choreographer.java:658)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:897)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

gino-m:
Same when returning back from View Observation. For some reason in `BottomSheetMetrics`:
```
  /**
   * Calculates the expected height of the bottom sheet when fully expanded, assuming the sheet will
   * stop expanding just below the top toolbar.
   */
  public int getExpandedOffset() {
    return toolbarWrapper.getHeight() - marginTop;
  }
```

`marginTop > toolbarWrapper.getHeight()`

gino-m:
This was fixed by #832.

PR #832, opened 2021-05-26
Work around negative expanded offset bug

gino-m:
Workaround for #828. Logs message when the crash would occur and falls back to default value. @scolsen @HemanParbhakar Could you please keep an eye out for the error message and see if you can understand when it's happening? It's seems to be hard to repro.

@HemanParbhakar Could you PTAL?


HemanParbhakar (approved):

gino-m:
Thanks @HemanParbhakar !

PR #866, opened 2021-06-06
Improve feature labels in feature sheet and selector

### LINKED ISSUE #793: [Feature details] Show user-defined ID and/or caption in header when feature selected

gino-m:
Uses GeoJSON properties when available to guess label or id.

Closes #793.

Screenshot_1623013557
Screenshot_1623013494
Screenshot_1623013550

@scolsen PTAL? 🙏


gnd/src/main/java/com/google/android/gnd/ui/common/FeatureHelper.java
```diff
@@ -16,46 +16,73 @@
...
+    }
+    return feature.getCaption();
   }
 
-  private Optional<String> getLayerName(@NonNull Optional<Feature> feature) {
-    return feature.map(Feature::getLayer).map(Layer::getName);
+  private String getGeoJsonCaption(GeoJsonFeature feature) {
+    String caption = feature.getCaptionFromProperties();
+    return caption.isEmpty()
+        ? getFeatureType(feature) + " " + feature.getIdFromProperties()
```
scolsen, line 85: optional: Feels like we could declare some of these methods in the interface/abstract class for generic features and let the specific types (e.g. GeoJsonFeature) provide concrete implementations. This would reduce some of the duplication that occurs re: checking the feature type here and above.

scolsen (approved):

gnd/src/main/java/com/google/android/gnd/ui/common/FeatureHelper.java
gino-m: The trouble is that Feature and related subclasses don't have references to the application context required for i18n. Eventually the model classes will be dumb data model classes (either as Kotlin data classes, or protobufs), in which case the presenters for those classes will need to be separate anyway.

scolsen: Ah, ok, sounds good!

PR #867, opened 2021-06-06
[Feature details] Update add observation chip label

gino-m:
This makes the CTA label consistent with the web app UI.

Screenshot_1623014011

@coreyleamon PTAL?


coreyleamon:
Not sure if relevant anymore but I agree with this adjustment to the CTA. Consistency with web is great, and this is also better for accessibility. :)

PR #868, opened 2021-06-06
Map tweaks / improvements

### LINKED ISSUE #445: [Feature sheet] Basemap sometimes blank
To repo, add a feature in the Brazilian Amazon or any place where high res imagery is not available.

The feature sheet zooms the map to a level where no imagery is shown by Maps SDK:
Screenshot_1586835037

### LINKED ISSUE #836: Option to have no base map ("None") should be removed
**Describe the bug**
On Android version 8.0.0.
Signed in as non-owner

Under base map layer button, user has the ability to choose "None", which gives them a blank screen. The option to choose "None" should be removed.

**To Reproduce**
1. Click base map layer icon
2. Select "None"
3. See blank screen

**Expected behavior**
User should only be able to choose valid base maps

**Screenshots**
Screenshot_20210601-144510_Ground

### LINKED ISSUE #865: High battery usage
@kdyson reported that app consumed 1% of battery per 3-5 minutes of run time on Pixel 3. Profiling revealed the largest energy consumption in the app is from panning and zooming hybrid satellite maps. One option could be to use "Normal" mode as default basemap.

gino-m:
- [x] Closes #865: Reduce battery usage due to Maps SDK satellite imagery rendering by using "normal" map as default. Hybrid or satellite imagery can be enabled via the layers button in the top-right corner.
- [x] Closes #836: Remove "None" basemap type.
- [x] Closes #445: Reducing default zoom when features are tapped reduces (but does not eliminate) the risk of zooming into an area where tiles are no available (e.g., the ocean). More complex fixes are not feasible atm, but we can return to this in the future if necessary.

@shobhitagarwal1612 PTAL?


shobhitagarwal1612 (approved):
💯

PR #884, opened 2021-06-13
[Photos] Load image from local cache if available

### LINKED ISSUE #881: [Add/Edit Observation] Photo thumbnail doesn't appear when offline
1. Enable airplane mode
2. Add/edit an observation with photo
3. Take a photo

Screenshot_1623284962

@shobhitagarwal1612 Might you have a moment to take a look at this one?

shobhitagarwal1612:
Fixes #881

Default timeout for the remote fetch was very large. Which is why the local image wasn't being loaded instantly. Also, we were removing the local copy once the photo was uploaded successfully.

To solve that:
 * Stop removing the local copy
 * Attempt to load from local cache first for better UX

### Steps to verify bug

1. Enable airplane mode
2. Open app and create a new observation
3. Click a new photo and save it.
4. The image should be visible under view observation
5. Then disable the airplane mode and let the photo upload.
6. Kill the app and enable airplane mode again.
7. Open the app and load the same observation.
8. Image should still be visible

@gino-m  PTAL?


gino-m (approved):
LGTM. Great work! 🚀

PR #885, opened 2021-06-13
[Code health] Fix instrumentation tests not building locally

shobhitagarwal1612:
Make `testBuildType` configurable. 

Default value is "debug" for local testing/development. For GCB, using flag to set it to "staging"

@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -79,7 +79,7 @@ steps:
     args:
       - '-c'
       - |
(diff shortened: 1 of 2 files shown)


gino-m:
@dturner FYI

gino-m (approved):
Awesome thanks so much for fixing this. @dturner @JoseAlcerreca FYI.

PR #886, opened 2021-06-13
[Cleanup] Create MapsRepository

### LINKED ISSUE #872: [Code health] Move reference to LocalDataStore out of GoogleMapsMapAdapter
`MapAdapter` is a system-level abstraction with no access to application state. All access to the application state should be through the interface rather than injected directly into implementation.

@shobhitagarwal1612 Is this something you might be able to fix?

shobhitagarwal1612:
Fixes #872

@gino-m  PTAL?


gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapProvider.java
```diff
@@ -16,33 +16,41 @@
...
 import java.util.Map;
+import javax.inject.Inject;
 
 /** Ground map adapter implementation for Google Maps API. */
 public class GoogleMapsMapProvider implements MapProvider {
 
+  private final Context context;
   private final MarkerIconFactory markerIconFactory;
-  private final LocalValueStore localValueStore;
+  private final MapsRepository mapsRepository;
```
gino-m, line 40: The `MapProvider` is intended as a wrapper around the map API and the application, to keep the implementation clean it shouldn't depend on any other application layers or state. Instead of injecting and passing `MapsRepository`, can `MapContainerFragment` set it instead in `onMapReady`?

gino-m (changes requested):
Thanks for cleaning this up. One design suggestion, otherwise LGTM.

gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsMapProvider.java
shobhitagarwal1612: Done

gino-m (approved):
Nice work! Thanks for the fix 💯

PR #897, opened 2021-06-16
[Feature] Hide "move feature" option for non-point features

### LINKED ISSUE #861: [Feature details] "Move" option should only be shown for point features
It also appears for GeoJSON features.

shobhitagarwal1612:
Fixes #861

@gino-m  PTAL?


gino-m (approved):
LGTM. Thanks for the fix!

PR #912, opened 2021-06-22
[Stability] Fixes several exceptions reported by Crashlytics and in logs

gino-m:
Fixes #910.
Fixes #911.
Fixes #913.

@shobhitagarwal1612 Could you PTAL?

PR #917, opened 2021-06-25
[Code health] Stop errors when no viewport saved in local value store

gino-m:
When there was no viewport saved in SharedPreferences errors were being caught and logged to error log. Since this is a normal state, we should handle it.

@HemanParbhakar PTAL?


HemanParbhakar (approved):

shobhitagarwal1612:
Changes already merged in #912

PR #921, opened 2021-06-25
[Map] Only register map click listener once

gino-m:
The listener was getting reset on each change to GeoJSON on the map.

@scolsen PTAL?


scolsen (dismissed):
Doh! Nice fix.

Issue #922, opened 2021-06-27
GCB still failing


gino-m:
Potentially relevant log lines:
```
Already have image (with digest): gcr.io/ground-android-gcb/android:28
ERROR: (gcloud.beta.firebase.test.android.run) argument --results-dir: expected one argument
Usage: gcloud beta firebase test android run [ARGSPEC] [optional flags]
  optional flags may be  --additional-apks | --app | --app-package | --async |
                         --auto-google-login | --client-details | --device |
                         --device-ids | --directories-to-pull |
                         --environment-variables | --filter |
                         --grant-permissions | --help | --limit | --locales |
                         --network-profile | --num-flaky-test-attempts |
                         --num-uniform-shards | --obb-files | --orientations |
                         --os-version-ids | --other-files | --page-size |
                         --performance-metrics | --record-video |
                         --results-bucket | --results-dir |
                         --results-history-name | --robo-directives |
                         --robo-script | --scenario-labels |
                         --scenario-numbers | --sort-by | --test |
                         --test-package | --test-runner-class | --test-targets |
                         --test-targets-for-shard | --timeout | --type |
                         --use-orchestrator

For detailed information on this command and its flags, run:
  gcloud beta firebase test android run --help
```

@shobhitagarwal1612 PTAL? 🙏

shobhitagarwal1612:
I'm not sure why this is failing. Didn't see this happen while working on #904.
Can you please point me to the PR/build where it failed recently?

PR #925, opened 2021-06-28
[GCB] Revert cache save step

shobhitagarwal1612:
Prevents failing of remote build because of concurrent modification error.

`tar: .gradle/caches/build-cache-1: file changed as we read it`

@gino-m  PTAL?

Contributes to #922.


gino-m (approved):
Thanks for the fix!

PR #930, opened 2021-06-29
[Bug] Fix crash when creating a new feature

shobhitagarwal1612:
* Fixes a bug when adding a new feature
 * Menu icon tint
 * Hide missing feature from settings

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
@@ -132,7 +132,12 @@ public class HomeScreenFragment extends AbstractFragment
     viewModel.getBottomSheetState().observe(this, this::onBottomSheetStateChange);
     viewModel.getOverlappingFeatures().observe(this, this::showFeatureSelector);
     viewModel.getOpenDrawerRequests().as(autoDisposable(this)).subscribe(__ -> openDrawer());
-    viewModel.getAddFeatureResults().as(autoDisposable(this)).subscribe(this::onFeatureAdded);
+    viewModel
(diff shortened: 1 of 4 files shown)


scolsen:
@shobhitagarwal1612 do you have steps to reproduce? I can't seem to trigger the crash.

gnd/src/main/java/com/google/android/gnd/ui/home/HomeScreenFragment.java
```diff
@@ -132,7 +132,12 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
     viewModel.getBottomSheetState().observe(this, this::onBottomSheetStateChange);
     viewModel.getOverlappingFeatures().observe(this, this::showFeatureSelector);
     viewModel.getOpenDrawerRequests().as(autoDisposable(this)).subscribe(__ -> openDrawer());
-    viewModel.getAddFeatureResults().as(autoDisposable(this)).subscribe(this::onFeatureAdded);
+    viewModel
+        .getAddFeatureResults()
+        .subscribeOn(schedulers.io())
```
gino-m, line 137: As a general practice, we've been adding `subscribeOn` as far upstream as possible, ie as close to the IO or other async operation as possible to make it clearer why we're subscribing on IO thread. Would you be able to move the `subscribeOn` into the stream's definition in  `HomeScreenViewModel`?

shobhitagarwal1612: Done

gino-m (approved):
LGTM. Thanks again for the quick fix 🩹 !

PR #940, opened 2021-07-04
[Layers] Only show modifiable layers with "points"

shobhitagarwal1612:
Things done: 
 * Add enums for Role and Feature type
 * Match for "points" in `contributorsCanAdd` list
 * Stop accessing `AuthenticationManager` directly in fragment. Use user repository instead.

Scenarios tested:
 * Click add feature in project with owner ACL
 * Click add feature in project with contributor ACL with `contributorsCanAdd` set as `["points"]`
 * Click add feature in project with contributor ACL with `contributorsCanAdd` empty
 

@gino-m  PTAL?


gnd/src/main/java/com/google/android/gnd/model/feature/Type.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.model.feature;
+
+public enum Type {
+  UNKNOWN,
+  POINTS
```
gino-m, line 21: I think feature type should be singular, since the type is attached to each individual feature, i.e. `POINT`.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -174,20 +176,21 @@ public void clearActiveProject() {
...
   }
 
-  public ImmutableList<Layer> getModifiableLayers(Project project, User user) {
-    String role = project.getAcl().get(user.getEmail());
-    if (role == null) {
-      return ImmutableList.of();
+  public ImmutableList<Layer> getModifiableLayers(Project project, Type featureType) {
+    // TODO: Use enums instead of string values
+    String featureTypeValue = featureType.name().toLowerCase(Locale.getDefault());
+    switch (userRepository.getUserRole(project)) {
```
gino-m, line 182: This is clearer than `canAddFeatures`, thanks.

gnd/src/main/java/com/google/android/gnd/repository/ProjectRepository.java
```diff
@@ -174,20 +176,21 @@ public void clearActiveProject() {
     selectProjectEvent.onNext(Optional.empty());
   }
 
-  public ImmutableList<Layer> getModifiableLayers(Project project, User user) {
-    String role = project.getAcl().get(user.getEmail());
-    if (role == null) {
-      return ImmutableList.of();
+  public ImmutableList<Layer> getModifiableLayers(Project project, Type featureType) {
+    // TODO: Use enums instead of string values
+    String featureTypeValue = featureType.name().toLowerCase(Locale.getDefault());
```
gino-m, line 181: `featureTypeValue` is only used in the `CONTRIBUTOR` switch case. Move inside?

gnd/src/main/java/com/google/android/gnd/model/feature/Type.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.gnd.model.feature;
+
+public enum Type {
```
gino-m, line 19: Since this is now a top-level class, and `Type` is a very common class name, it may be worth renaming to `FeatureType` both for readability and to avoid clashes with similarly named classes in other packages.

gino-m (changes requested):
Excellent clean up! A few nits/suggestions.

gino-m (approved):
LGTM! 🚢  🚢  !

PR #945, opened 2021-07-08
[Docs] Add instructions for fixing Gradle failure on clone

gino-m:
Also includes minor formatting tweaks, including moving keytool command on one line so it works in Linux terminal.

@mmisim PTAL?


mmisim (approved):
LGTM

PR #954, opened 2021-07-10
[Refactor] Move options menu logic to FeatureDetailsFragment

### LINKED ISSUE #948: [Code health] "BottomState is null" error is logged when opening app
```
2021-07-08 16:23:54.209 14466-14466/com.google.android.gnd E/HomeScreenFragment: BottomSheetState is null
```

I assume this is an expected state while the app is initializing; if so, we shouldn't log an error.

@shobhitagarwal1612 Can you take a peek when you have a cycle?

shobhitagarwal1612:
Fixes #948 

* Options menu represents actions for the current selected feature. So, it is better to keep the related code closer to it.
* Removes hardcoded item index with resource id.
* Ideally, we should also move the options selected method. But it is currently tangled with HomeScreenFragment. Needs a bigger cleanup.

@gino-m  PTAL?


gnd/src/main/java/com/google/android/gnd/ui/home/featuredetails/FeatureDetailsViewModel.java
```diff
@@ -37,39 +37,41 @@
 public class FeatureDetailsViewModel extends ViewModel {
 
   @Hot
-  private final FlowableProcessor<Optional<Feature>> selectedFeature =
+  private final FlowableProcessor<Optional<Feature>> featureFlowable =
```
gino-m, line 40: I think a more consistent way to handle the sync vs async selectedFeature states might be to define a `LiveData<Boolean> moveMenuOptionVisible` and expose that to the view via a public access. If/when data binding on menu options is supported it will then be easy to switch. We could then get rid of the new `selectedFeature` member and restore this stream's original name. Wdyt?

gino-m (changes requested):

gnd/src/main/java/com/google/android/gnd/ui/home/featuredetails/FeatureDetailsViewModel.java
shobhitagarwal1612: Yes, data binding isn't supported for options menu. So should this suggestion be added as a TODO?

gnd/src/main/java/com/google/android/gnd/ui/home/featuredetails/FeatureDetailsFragment.java
```diff
@@ -88,7 +88,7 @@ public void onPrepareOptionsMenu(@NonNull Menu menu) {
         .forEach(
             menuItem -> {
               if (menuItem.getItemId() == R.id.move_feature_menu_item) {
-                menuItem.setVisible(viewModel.isSelectedFeatureOfTypePoint());
+                viewModel.getMoveMenuOptionVisible().observe(this, menuItem::setVisible);
```
gino-m, line 91: Cool, didn't this of this.

gino-m: [i.e. updating the visibility reactively rather than synchronously.]

gino-m (approved):
🎸  Rockin'! Thanks!

PR #957, opened 2021-07-12
Delete broken test

gino-m:
I can't get this test to work because it needs to wait for the UI to be fully initialized after remote and local I/O completes. The tests aren't failing because the button isn't [in]visible, but rather because the UI element is not yet created. This test is probably anyway overkill, and we should prefer unit tests wherever possible.

Unblocks #934.

@shobhitagarwal1612 WDYT?


gino-m:
@shobhitagarwal1612 Thanks for taking a look. I'm off for tonight but please feel free to merge if checks pass and you approve. Thanks!

shobhitagarwal1612:
Agreed. Since the visibility is controlled via ViewModel, it should be possible to convert this into a unit test.

shobhitagarwal1612 (approved):

PR #961, opened 2021-07-15
[GCB] Run task "dependencyUpdates" only when running on master

shobhitagarwal1612:
The generated reports are only used when the build is running against `master`.
So, run this task only when necessary.

@gino-m  PTAL?


gino-m (approved):
LGTM. Thanks!

PR #963, opened 2021-07-17
[Feature Repository] Add tests and cleanup ToDos

### LINKED ISSUE #80: [Code health] [Add feature] Use FeatureMutation instead of full Feature object on save
DataRepository currently receives full Feature objects on create and converts them to FeatureMutations itself. Since update feature will create FeatureMutations, API would be simpler to always use FeatureMutations directly.

shobhitagarwal1612:
Fixes #80 

The intention behind this PR is to cleanup pending TODOs and improve test coverage.

Followed TDD approach for this PR.

 * Cleaned up unit tests + new test library (power-mockito)
 * Increased test coverage of FeatureRepository to 100%
 * Added helper method : Create a new mutation from the feature via it's internal method
 * Provide "mutations" instead of "feature" when performing create/update/delete operations.
 * Simplify code by replacing `Maybe` with `Single` when attempting to get a feature from repo
 * Remove dependency : project id should be the active project when fetching a feature 

@gino-m PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
+++ b/gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
@@ -23,7 +23,6 @@ import com.google.android.gnd.model.feature.Feature;
 import com.google.android.gnd.model.feature.FeatureMutation;
 import com.google.android.gnd.model.feature.Point;
 import com.google.android.gnd.persistence.local.LocalDataStore;
-import com.google.android.gnd.persistence.remote.NotFoundException;
 import com.google.android.gnd.persistence.remote.RemoteDataEvent;
 import com.google.android.gnd.persistence.remote.RemoteDataStore;
 import com.google.android.gnd.persistence.sync.DataSyncWorkManager;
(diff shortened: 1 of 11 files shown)


gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
```diff
@@ -114,15 +113,10 @@ private Completable updateLocalFeature(RemoteDataEvent<Feature> event) {
...
-  // TODO: Don't require projectId to be the active project.
   @Cold
   public Single<Feature> getFeature(String projectId, String featureId) {
     return projectRepository
-        .getProjectLoadingState()
-        .compose(Loadable::values)
-        .firstElement()
-        .filter(project -> project.getId().equals(projectId))
-        .switchIfEmpty(Single.error(() -> new NotFoundException("Project not found: " + projectId)))
+        .getProject(projectId)
```
gino-m, line 119: What if the local db was deleted while the app was paused? In that case the project will not be in the local db and this will crash.

gino-m (changes requested):
Great work, improving test coverage and code health is critical to our success, thanks for the initiative. One comment/question.

gnd/src/main/java/com/google/android/gnd/repository/FeatureRepository.java
shobhitagarwal1612: It shouldn't crash if the downstream handles the exception and displays/logs it. But I'm not sure if that is currently being handled or not.

gino-m: Imo this shouldn't depend on the order in which the project and feature are loaded from remote, but even if we always try to sync from remote when this is called we'd need to handle the failure case when there's not network connection and the project isn't in the local db.

Instead, perhaps we can just document the assumption that this only works if the project and feature are already cached, and return a more informative error if either of them are not?

gino-m (changes requested):

gino-m (approved):
LGTM. Thanks for the fixes!

shobhitagarwal1612:
@gino-m Can I please get a re-stamp?

gino-m (approved):

PR #871, opened 2021-06-07
[Feature disambiguation dialog] Fix crash on device rotate

gino-m:
Also includes a few small refactors.

Fixes #858

@scolsen PTAL?

PR #966, opened 2021-07-20
[Map] Fix broken add feature button

### LINKED ISSUE #965: [Add point] Add feature fab not shown even though project contains editable layer
Internal test project ID `OPcfIlYMDy8MmV7jDaJX`, layer `vxzSK7sgoRqKBvRexMv7`

gino-m:
Fixes #965:

- [ ] Uses correct string values for `contributorsCanAdd` permissions.
- [ ] Restores add feature button callback accidentally removed while merging with upstream in #871.

Also does some minor cleanup.

@scolsen or @shobhitagarwal1612 PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/LayerEntity.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/LayerEntity.java
@@ -23,6 +23,7 @@ import androidx.room.Entity;
 import androidx.room.ForeignKey;
 import androidx.room.Index;
 import androidx.room.PrimaryKey;
+import com.google.android.gnd.model.feature.FeatureType;
 import com.google.android.gnd.model.layer.Layer;
 import com.google.android.gnd.model.layer.Style;
 import com.google.android.gnd.persistence.local.room.relations.FormEntityAndRelations;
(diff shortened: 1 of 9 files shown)


gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/LayerEntity.java
```diff
@@ -92,22 +93,31 @@ static Layer toLayer(LayerEntityAndRelations layerEntityAndRelations) {
...
+        builder.add(toFeatureType(value));
       }
     }
     return builder.build();
   }
 
+  private static FeatureType toFeatureType(String stringValue) {
+    switch (stringValue) {
+      case "points":
+        return FeatureType.POINT;
```
shobhitagarwal1612, line 115: It would be simpler to convert string -> enum using `toEnum()` if both stringValue and Enum were either "point" or "points".

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/LayerConverter.java
```diff
@@ -40,12 +43,25 @@ static Layer toLayer(String id, LayerNestedObject obj) {
...
     return layer.build();
   }
 
+  private static FeatureType toFeatureType(String stringValue) {
+    switch (stringValue) {
+      case "points":
+        return FeatureType.POINT;
+      default:
+        return FeatureType.UNKNOWN;
+    }
```
shobhitagarwal1612, line 62: Same as above

gnd/src/main/res/layout/map_home_layout.xml
```diff
@@ -61,45 +61,50 @@
...
       android:layout_height="wrap_content"
       android:onClick="@{() -> homeScreenViewModel.openNavDrawer()}"
       android:src="@drawable/ic_menu_black_24dp"
-      app:tint="@{viewModel.getIconTint()}"
       app:backgroundTint="@color/colorBackground"
-      app:layout_constraintStart_toStartOf="parent"
       app:fabSize="mini"
+      app:layout_constraintStart_toStartOf="parent"
       app:layout_constraintTop_toTopOf="parent"
+      app:tint="@{viewModel.getIconTint()}"
```
shobhitagarwal1612, line 107: We should use color value rather than `getIconTint`. This gets changed with location lock state

shobhitagarwal1612 (approved):

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/LayerEntity.java
gino-m: It would be less code, but without making the mapping explicit, a rename of the enum would break the code. Ideally the enum values should be named for what's more readable in code, and the values in the db for whatever makes most sense there. Coupling them could force a compromise on either end.

gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema/LayerConverter.java
gino-m: See previous comment. If you feel strongly about it please let's discuss!

gnd/src/main/res/layout/map_home_layout.xml
shobhitagarwal1612: `getIconTint` supplies the color based on location lock state (i.e. blue if active, else gray)
Hamburger icon shouldn't change its tintColor.

So, I was referring to replacing it with `app:tint="@color/some_color"`

PR #970, opened 2021-07-26
[Refactor] Create custom view and view model for feature reposition

shobhitagarwal1612:
Things done:
* Creates a new view for `FeatureReposition`
* Moves all related logic to `FeatureRepositionViewModel`
* Make MapProvider a `singleton`

Changes verified manually for regression:
 * Move feature

@gino-m  PTAL?


shobhitagarwal1612:
@gino-m Please let me know if this draft looks fine. I'll try to finalise it then. Otherwise, I'll remove it.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionView.java
```diff
@@ -0,0 +1,33 @@
...
+  public FeatureRepositionView(Context context, @Nullable AttributeSet attrs) {
+    super(context, attrs);
+    init();
+  }
+
+  private LayoutInflater getLayoutInflater() {
+    return (LayoutInflater) getContext().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+  }
+
+  private void init() {
```
gino-m, line 25: If this is only used in the constructor can we move it inline there?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionView.java
```diff
@@ -0,0 +1,33 @@
...
+    LayoutInflater inflater = getLayoutInflater();
+    binding = DataBindingUtil.inflate(inflater, R.layout.map_move_feature_layout, this, true);
+  }
+
+  public void setViewModel(FeatureRepositionViewModel viewModel) {
```
gino-m, line 30: Usually views create their own VMs. Can we do that here as well? Perhaps by defining a method that gets called when the right lifecycle hook in the parent fragment is called?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionViewModel.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public void onConfirmButtonClick() {
+    confirmButtonClicks.onNext(mapProvider.getCameraTarget());
+  }
+
+  public void onCancelButtonClick() {
+    cancelButtonClicks.onNext(Nil.NIL);
+  }
+
+  public Observable<Point> getConfirmButtonClicks() {
```
gino-m, line 33: Please add `@Hot` annotations here and below as well.

gino-m (changes requested):
I think this is a lot cleaner, please do proceed!

shobhitagarwal1612:
@gino-m Ready for your review!

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionViewModel.java
```diff
@@ -0,0 +1,60 @@
...
+@SharedViewModel
+public class FeatureRepositionViewModel extends AbstractViewModel {
+
+  private final MapProvider mapProvider;
+
+  @Hot private final Subject<Point> confirmButtonClicks = PublishSubject.create();
+  @Hot private final Subject<Nil> cancelButtonClicks = PublishSubject.create();
+
+  @Inject
+  FeatureRepositionViewModel(MapProvider mapProvider) {
```
shobhitagarwal1612, line 39: @gino-m Should we directly use `MapProvider` or use `MapRepository`?

shobhitagarwal1612: Feels like there should be a separate abstraction which should provide location updates rather than exposing `getCameraTarget` indirectly all the way to the top.
`GoogleMapsAdapter` -> `GoogleMapsProvider` -> `VM`

Wdyt?

gino-m: This is definitely wrong.. ViewModels should never reference Views since the lifecycle of VMs is longer. Even though MapProvider isn't a view class, it returns a the map Fragment which is. I think the right thing to do here would be for `FeatureRepositionView` to subscribe to camera movements, calling a setter on `FeatureRepositionViewModel` on each change to update the current VM state. Wdyt?

gino-m (changes requested):
See inline comment.

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionViewModel.java
shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m This is ready for another pass!

gnd/src/main/java/com/google/android/gnd/ui/common/AbstractView.java
```diff
@@ -39,20 +38,19 @@ public AbstractView(@NonNull Context context, @Nullable AttributeSet attrs) {
...
+  protected FragmentActivity getActivity() {
     Context context = getContext();
     while (context instanceof ContextWrapper) {
       if (context instanceof FragmentActivity) {
         return (FragmentActivity) context;
       }
       context = ((ContextWrapper) context).getBaseContext();
     }
-    return null;
+    throw new NullPointerException("Activity is null");
```
gino-m, line 49: Not sure this is the right exception/error to report here. `IllegalStateException` is probably more appropriate, with a message like "View is not contained in FragmentActivity".

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionViewModel.java
```diff
@@ -21,27 +21,26 @@
...
   @Inject
-  FeatureRepositionViewModel(MapProvider mapProvider) {
-    this.mapProvider = mapProvider;
-  }
+  FeatureRepositionViewModel() {}
 
+  // TODO: Disable the confirm button until the map has not been moved
   public void onConfirmButtonClick() {
-    confirmButtonClicks.onNext(mapProvider.getCameraTarget());
+    confirmButtonClicks.onNext(Objects.requireNonNull(lastKnownCameraTarget));
```
gino-m, line 43: In addition to disabling the button as per your TODO, we can be more defensive here by ignoring the click if there's not target rather than throwing a runtime errors (i.e. if (lastKnownCameraTarget != null).

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/FeatureRepositionViewModel.java
```diff
@@ -21,27 +21,26 @@
...
 import javax.inject.Inject;
 
 @SharedViewModel
 public class FeatureRepositionViewModel extends AbstractViewModel {
 
-  private final MapProvider mapProvider;
-
   @Hot private final Subject<Point> confirmButtonClicks = PublishSubject.create();
   @Hot private final Subject<Nil> cancelButtonClicks = PublishSubject.create();
+  @Nullable private Point lastKnownCameraTarget;
```
gino-m, line 36: You can probably drop "lastKnown", since all states carry their last known values.

gino-m (changes requested):
Minor suggestions, otherwise LGTM!

gino-m (approved):
LGTM. Thanks for the refactor!

PR #981, opened 2021-08-04
[Code health] Add unit tests for FeatureHelper

shobhitagarwal1612:
I've added all possible cases by looking at the code. PLMK if i've missed any.

@scolsen  PTAL?


gino-m (approved):
Great effort! I think this covers all paths. Are you able to run code coverage check in AS to verify?

shobhitagarwal1612:
I couldn't run code coverage for this particular test file. Works for the rest of the tests though.

PR #982, opened 2021-08-07
[Code health] Lint checks & CI fix

shobhitagarwal1612:
@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -26,6 +26,10 @@
 #                                /                   \                                                                     /
 #  Fetch google-services.json -->                      --> Authorize Gcloud --> Build test apks --> Instrumented tests -->
 #
+# Notes:
+# ------
+# ${_PUSH_TO_MASTER} is true when the commit is pushed to master branch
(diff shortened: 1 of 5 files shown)


cloudbuild.yaml
```diff
@@ -74,6 +74,10 @@ steps:
         ./gradlew -PdisablePreDex checkCode 2> check-logs.txt || echo "fail" > build-status.txt
         cat check-logs.txt
 
+        if [[ "${_PUSH_TO_MASTER}" ]]; then
```
gino-m, line 77: Could you please add a comment explaining when _PUSH_TO_MASTER is true? Is it run before GitHub merges an approved PR with master? Will updates block merge?

gino-m (changes requested):

cloudbuild.yaml
shobhitagarwal1612: Added notes. No, it won't block merges.

shobhitagarwal1612:
Needs a re-approval

gino-m (approved):
Restamp!

PR #984, opened 2021-08-10
Use stable version of hilt/work

shobhitagarwal1612:
Things done:
 * Replaces `ApplicationComponent` with `SingletonComponent`
 * Use stable versions of hilt, hilt-worker libraries
 * Introduce a new module for `WorkManager`
 * Use new suggested methods for injecting dependencies in `*Worker` and `*Manager` classes

@gino-m PTAL?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@
 
 buildscript {
     ext {
-        hiltVersion = "2.29.1-alpha"
+        hiltVersion = "2.37"
     }
     repositories {
         google()
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@
 
 buildscript {
     ext {
-        hiltVersion = "2.37"
+        hiltVersion = "2.38"
     }
     repositories {
         google()
--- a/config/spotbugs/spotbugs-filter.xml
+++ b/config/spotbugs/spotbugs-filter.xml
@@ -35,6 +35,9 @@
       <!-- Auto-generated NavDirections classes -->
       <Class name="~.*Args.*" />
       <Class name="~.*Directions\$.*" />
+
+      <!-- Auto-generated Hilt classes -->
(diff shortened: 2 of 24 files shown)


gino-m (dismissed):
Thanks for the upgrade and cleanup!

gnd/src/androidTest/java/com/google/android/gnd/persistence/sync/FakeWorkManager.java
```diff
@@ -0,0 +1,174 @@
...
+import androidx.work.WorkContinuation;
+import androidx.work.WorkInfo;
+import androidx.work.WorkManager;
+import androidx.work.WorkQuery;
+import androidx.work.WorkRequest;
+import com.google.common.util.concurrent.ListenableFuture;
+import java.util.List;
+import java.util.UUID;
+
+public class FakeWorkManager extends WorkManager {
```
gino-m, line 36: Could you kindly remind me why Mockito mocks don't work here? It seems this fake has the same behavior as the Mockito's mock. Any reason we can't just use that?

gino-m (changes requested):

gnd/src/androidTest/java/com/google/android/gnd/persistence/sync/FakeWorkManager.java
shobhitagarwal1612: I tried using mockito first. But the mocks couldn't be generated. It expects a proper test runner.

gino-m: Sgtm. I don't really understand the relationship between test runner and providing a worker - let's discuss offline?

gino-m (approved):
LGTM!

PR #985, opened 2021-08-11
Append commit sha1 to version name

shobhitagarwal1612:
It is difficult to find the commit using the current version name.
So, append the current commit's SHA1 for better debugging.

e.g. `7539-SNAPSHOT(2 +5 -1) a345bc3f96dc369c32eb1b540890ac3fb4eba7b2`

@gino-m PTAL?

### DIFF
--- a/gnd/build.gradle
+++ b/gnd/build.gradle
@@ -46,6 +46,15 @@ if (secretsFile.exists()) {
 }
 def googleMapsApiKey = secrets.getProperty('GOOGLE_MAPS_API_KEY', '')
 
+def getCommitSha1 = { ->
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        commandLine 'git', 'show', '-s', '--format=%h'
+        standardOutput = stdout
+    }
+    return stdout.toString().trim()
+}
+
 android {
     compileSdkVersion 28
     defaultConfig {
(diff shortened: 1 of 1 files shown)


gnd/build.gradle
```diff
@@ -56,7 +56,7 @@ android {
         // fix before increasing this value.
         targetSdkVersion 26
         versionCode gitVersioner.versionCode
-        versionName gitVersioner.versionName
+        versionName gitVersioner.versionName + " " + gitVersioner.currentSha1
```
gino-m, line 59: Is there a way to get the shortened commit ID instead? From the command line the latest commit can be shown with `git show -s --format=%h`.

gino-m (changes requested):

gnd/build.gradle
shobhitagarwal1612: The git versioner plugin doesn't provide that out of the box. I'll try doing it directly like you suggested. Maybe we can remove this dependency too.

gino-m: The short version is just the first 7 characters of the commit hash. Is it possible to truncate to only include those?

gino-m (dismissed):

gnd/build.gradle
shobhitagarwal1612: Done

shobhitagarwal1612: sample output : `7571-version+3 6714acd75`

gino-m (approved):
Cool trick, learning a lot from these PRs, thanks again!

PR #986, opened 2021-08-13
Update Maps SDK version to 17.0.1

### LINKED ISSUE #946: [Map] Map freezes when upgrading to targetSdk 27 or higher
We haven't been able to upgrade past 26 because of issues with Maps SDK when updating to 27 or higher: panning and zooming the map gets progressively slower until the app completely freezes.

Repro:

* Update `targetSdkVersion` to `27` in `gnd/build.gradle`
* Sync, build, and run app on Nexus 5X API 29 in emulator.
* Sign in and join a project.
* Try to pan and zoom the map, app is either frozen or slow to the point of being unusable.

nmhung00:
Fixes #946

Update Maps SDK version to 17.0.1 to avoid having the map frozen after updating targetSdkVersion  to 27 or higher.

Maps SDK version 17.0.0 doesn't seem to work well with StrictMode at Android SDK level 27+.

Verified that the map works fine by updating the targetSdkVersion to 27 locally with an emulator.


gino-m (approved):
Great, thanks! Care to up the targetSdkVersion in a follow-up PR?

nmhung00:
@gino-m : sure, will create a separate one to update the targetSdkVersion.

PR #987, opened 2021-08-13
[Regression] Make MapProvider singleton

shobhitagarwal1612:
MapProvider is injected into both `FeatureRepositionView` and `MapContainerFragment`. Without a singleton, each would have a different copy of MapProvider.

@gino-m PTAL?


gino-m (commented):
The provider actually contains a reference to a map inside `MapAdapter`. Do the offline areas area selector or thumbnail view need a new instance, or is ok for them to reuse the same one?

shobhitagarwal1612:
I tested locally and the offline areas selected is working fine with the above changes.

gino-m (approved):
Thanks for the fix!

PR #1001, opened 2021-09-01
Downgrade compile and target SDK version to avoid ANR problem in Dev build

nmhung00:
The strict mode doesn't seem to work well when there are many GeoJsonLayer on the map that results in a lot of setting listener calls to Maps SDK.


gino-m (approved):
Thanks for the quick fix!

PR #1007, opened 2021-09-05
[Testing] Simplify tests configuration

shobhitagarwal1612:
Use [@TestInstallIn](https://dagger.dev/hilt/testing.html) in module instead of uninstalling in each test.

@gino-m  PTAL?


gino-m (approved):
Thanks for the fix, LGTM.

PR #1027, opened 2021-10-23
[Testing] Simplify testing by pulling boilerplate to base class

shobhitagarwal1612:
This PR aims at reducing the boiler-plate code needed to write tests and provide a structure for writing new ones.

Currently, we are overusing mocks. Some of that could be replaced by using actual instances or fake implementations. Mocks should only be used if doing it via fakes is too complex.

Things done:
 * Moves all fake implementations to the shared test dir. Previously, only integration tests used them.
 * Configure Hilt to apply fakes by default for all unit/integration tests. This removes the need to manually install/uninstall modules per test.
 * Create base class with all necessary configs to prevent boilerplate code in final test classes.
 * Apply the below guidelines to tests.

General guidelines:
-------------------

 * Prefer fake implementations over mocks. Write one if not already present.
 * Mocks should only be used where we want to verify that a particular method was called with certain params.
 * Use DI for android specific tests to benefit from pre-configured dependencies.

Writing unit tests:
------------------

 * If the class that needs to be tested does not use any android specific code, then simply use Junit runner or MockitoJunit runner.
 * Otherwise use newly created `HiltTestWithRobolectricRunner` or `HiltTestWithParameterizedRobolectricRunner` as the base class which is preconfigured to support mocking, DI and synchronous background threads.
 * If a particular test demands a mock instead of pre-configured fake implementation, then use `@BindValue @Mock` to replace that dependency with mock only for that class. If that dependency is originally configured via modules, then explicit `@UninstallModule` is also neeeded. (see examples in PR)

Writing Integration tests:
-------------------------

 * Use newly created `BaseHiltTest` as the base class. It is preconfigured for run tests that require main activity and DI.

@gino-m @scolsen @nmhung00  PTAL?


gnd/src/androidTest/java/com/google/android/gnd/BaseMainActivityTest.java
```diff
@@ -0,0 +1,79 @@
...
+import dagger.hilt.android.testing.HiltAndroidRule;
+import javax.annotation.OverridingMethodsMustInvokeSuper;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+
+/**
+ * Configures all necessary rules and {@link MainActivity} needed for running instrumentation tests.
+ */
+public abstract class BaseMainActivityTest {
```
nmhung00, line 31: Would a new test rule that includes all of these logic be better to favor composition over inheritance?

gnd/src/test/java/com/google/android/gnd/HiltTestWithParameterizedRobolectricRunner.java
```diff
@@ -0,0 +1,28 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.ParameterizedRobolectricTestRunner;
+import org.robolectric.RobolectricTestRunner;
+
+/**
+ * Injects Hilt dependencies during setUp and uses {@link RobolectricTestRunner} for
+ * instrumentation.
+ */
+@RunWith(ParameterizedRobolectricTestRunner.class)
+public abstract class HiltTestWithParameterizedRobolectricRunner extends BaseHiltTest {}
```
nmhung00, line 28: I'm not sure if it's worth creating a new class to reuse one annotation. It's probably easier to read if we just apply the annotation at the concrete class level instead.

gnd/src/test/java/com/google/android/gnd/persistence/local/DateTimeResponseTest.java
```diff
@@ -16,39 +16,36 @@
...
   @Test
-  public void testTimeFormat() {
-    Date originalDate = new Date();
-    String timeString = timeFormat.format(originalDate);
-    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("HH:mm", Locale.getDefault());
-    LocalTime lt = LocalTime.parse(timeString, formatter);
-    String formattedTime = lt.format(formatter);
-    Truth.assertThat(timeString).isEqualTo(formattedTime);
+  public void testTimeResponse_textDetails() {
+    Truth.assertThat(new TimeResponse(getExpectedDate()).getDetailsText()).isEqualTo("07:30");
```
nmhung00, line 44: For self-describing methods like "assertThat", it's better for readability to just static import the method instead.

nmhung00: It's probably not important since this class is small, but I would prefer to have the time showing up somewhere in this test method to show the readers on why we are expecting "07:30" text.

nmhung00 (changes requested):

gnd/src/test/java/com/google/android/gnd/HiltTestWithParameterizedRobolectricRunner.java
shobhitagarwal1612: Done

gnd/src/test/java/com/google/android/gnd/persistence/local/DateTimeResponseTest.java
shobhitagarwal1612: Done

gnd/src/androidTest/java/com/google/android/gnd/BaseMainActivityTest.java
nmhung00: What do you think about this suggestion?

Just want to make sure you considered this option. The code base seems to already have a bunch of abstract classes for this kind of code re-use so I guess that could be fine to have it here, too. I personally feel too much usage of inheritance could make the code a bit difficult to read because of the hidden inherited code. Using composition looks more straightforward to me from the point of a reader, so I often use composition whenever I can.

shobhitagarwal1612: I haven't written a lot of integration tests. So I'm not sure which is preferred. We can revert the changes for integration tests and handle them separately if you prefer. Wdyt?

nmhung00: That's probably fine. Just want to check if we can agree on some best practices moving forward.

gino-m: +1 to preferring [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) except where inheritance provides concrete value. I think in this case since the class is inheriting annotations it might make sense, but I don't know enough about how Hilt rules work. In either case, this lgtm for now, we can circle back later if necessary.

nmhung00 (approved):

gnd/src/androidTest/java/com/google/android/gnd/BaseMainActivityTest.java
shobhitagarwal1612: I tried converting it into a custom test rule. It had 2 issues:
 1. Hilt injection requires `@HiltAndroidTest` annotation. So that couldn't be put inside the test rule. [Minor issue, could be ignored by keeping this in the main test class]
 2. Tests run without waiting for the binding to complete. Needs more investigation.

Created https://github.com/google/ground-android/issues/1029 to track this separately.

PR #1031, opened 2021-11-04
[Unit tests] Improve test coverage

shobhitagarwal1612:
Refactor remaining unit tests based on https://github.com/google/ground-android/pull/1027#issue-1034130097. This improves test coverage as we reduce dependency on mocks and use real instances.

Things done:
 * Replaces mocks with actual implementation / fakes. Only `BitmapUtil` is mocked right now as it couldn't be converted.
 * Adds a util class to mock system permissions instead of mocking `PermissionsManager`
 * Uses static imports for consistency

@nmhung00 PTAL?


gnd/src/test/java/com/google/android/gnd/MainViewModelTest.java
```diff
@@ -16,134 +16,136 @@
...
+
+  private void setupUserPreferences() {
+    sharedPreferences.edit().putString("foo", "bar").apply();
+  }
+
+  private void assertUserPreferencesCleared() {
+    assertThat(sharedPreferences.getAll()).isEmpty();
+  }
+
+  private void assertUserSaved() {
```
gino-m, line 73: Nit: Methods that execute the unit under test are not only asserting, they're actually tests themselves. The easiest way to make this clear would be to rename them with something like "verify*"

gino-m (changes requested):
Very nice! One nit, otherwise over to @nmhung00

gnd/src/test/java/com/google/android/gnd/MainViewModelTest.java
shobhitagarwal1612: Done

shobhitagarwal1612:
ping!

gino-m (approved):
Great example of how we can add more test coverage. I wasn't aware of the ShadowApplication util, good to know. Thanks for these!

PR #1063, opened 2021-11-18
Add tests for PolygonDrawing flow

### LINKED ISSUE #1052: Refactor and add unit tests for PolygonDrawingViewModel
This class has a lot of internal states. It would be nice to add some unit tests covering all possible cases. 
Also refactor the class as needed to simplify the code.

shobhitagarwal1612:
Fixes #1052

Adds full coverage for polygon drawing flow. The remaining code coverage is related to location lock. IMO, we should separate it out into an UI component and use it everywhere.

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
@@ -30,7 +30,6 @@ import com.google.android.gnd.model.feature.Point;
 import com.google.android.gnd.model.feature.PolygonFeature;
 import com.google.android.gnd.model.layer.Layer;
 import com.google.android.gnd.persistence.uuid.OfflineUuidGenerator;
-import com.google.android.gnd.repository.FeatureRepository;
 import com.google.android.gnd.rx.BooleanOrError;
 import com.google.android.gnd.rx.Nil;
 import com.google.android.gnd.rx.annotations.Hot;
@@ -82,7 +81,6 @@ public class PolygonDrawingViewModel extends AbstractViewModel {
 
   private final OfflineUuidGenerator uuidGenerator;
   private final AuthenticationManager authManager;
-  private final FeatureRepository featureRepository;
   @Nullable private Point cameraTarget;
 
   /**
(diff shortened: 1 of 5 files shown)


gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
```diff
@@ -188,7 +186,17 @@ private void addVertex(Point vertex, boolean isNotSelectedByUser) {
     updateDrawnPolygonFeature(ImmutableList.copyOf(vertices));
   }
 
+  @VisibleForTesting
```
gino-m, line 189: Can we test using the existing public methods rather than exposing internal state?

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
```diff
@@ -203,16 +211,25 @@ private void updateDrawnPolygonFeature(ImmutableList<Point> vertices) {
   }
 
   public void onCompletePolygonButtonClick() {
+    if (vertices.size() < 3 || !getFirstVertex().equals(getLastVertex())) {
```
gino-m, line 214: Does the last vertex snap to the exact value as the first index? I assume it does, just wanted to check.

gino-m (changes requested):
Some questions. Thanks for the tests!

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
shobhitagarwal1612: Fixed

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
shobhitagarwal1612: Yes, the last vertex and first vertex are exactly same.

gino-m (approved):

PR #1064, opened 2021-11-18
fix: show deletion mutations in sync status UI

### LINKED ISSUE #1059: [Sync status] DELETE mutations are not shown in sync status UI
Found by @rachaprince

scolsen:
Before, we cascaded deletions on the foreign keys for
feature/observation mutations, meaning all mutations for a
feature/observation were wiped from the DB, eradicating the user's
edit/sync history.

This change alters that behavior so that we keep mutations around even
after a feature is deleted. Additionally, it removes a redundant call
to deleteFeature when handling mutations (these are handled server side
and should not be finalized until deleted on the server--the local
project receives this update via syncFeatures).

fixes #1059
Screen Shot 2021-11-18 at 2 27 23 PM

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
@@ -412,9 +412,11 @@ public class RoomLocalDataStore implements LocalDataStore {
         .flatMapCompletable(
             mutation -> {
               if (mutation instanceof ObservationMutation) {
-                return deleteObservation(((ObservationMutation) mutation).getObservationId());
+                // Deletions are finalized once the remote receives the delete and propagated locally
+                // via Project.syncFeatures
+                return Completable.complete();
               } else if (mutation instanceof FeatureMutation) {
-                return deleteFeature(mutation.getFeatureId());
+                return Completable.complete();
               } else {
                 return Completable.error(new RuntimeException("Unknown type : " + mutation));
               }
--- a/gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
@@ -17,6 +17,7 @@
 package com.google.android.gnd.persistence.local.room.entity;
 
 import static androidx.room.ForeignKey.CASCADE;
+import static androidx.room.ForeignKey.SET_DEFAULT;
 import static com.google.android.gnd.persistence.local.room.entity.FeatureEntity.formatVertices;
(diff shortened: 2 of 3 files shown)


scolsen:
@gino-m ptal! this will fix #1059 but gives us a new problem to solve: It persists mutations longer than the features/observations they belong to. We'll have to determine a good strategy for deleting these over time.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
```diff
@@ -49,7 +50,7 @@
             entity = FeatureEntity.class,
             parentColumns = "id",
             childColumns = "feature_id",
-            onDelete = CASCADE),
+            onDelete = SET_DEFAULT),
```
gino-m, line 53: Removing `CASCADE` would lead to mutations in the db with no corresponding entities. Are we sure this is what we want? Two possible alternatives:

1. Mark features and observations as "DELETED" rather than actually removing them locally
2. Delete feature and observations, and write history of changes to a separate MUTATION_LOG table. 

IMO option 2 is preferable since it doesn't involve keeping entities in the db after they're deleted in order show the mutation log, and it doesn't rely on other components filtering out "deleted" items correctly. Wdyt?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -412,9 +412,11 @@ private Completable finalizeDeletions(ImmutableList<Mutation> mutations) {
         .flatMapCompletable(
             mutation -> {
               if (mutation instanceof ObservationMutation) {
-                return deleteObservation(((ObservationMutation) mutation).getObservationId());
+                // Deletions are finalized once the remote receives the delete and propagated locally
+                // via Project.syncFeatures
+                return Completable.complete();
               } else if (mutation instanceof FeatureMutation) {
-                return deleteFeature(mutation.getFeatureId());
+                return Completable.complete();
```
gino-m, line 419: If we go this route, `finalizeDeletions` would then be a no-op and could be removed. But before that, please see my question below.

gino-m (changes requested):
Thanks for the fix! A few questions.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
scolsen: Agreed--I think option 2 is better. Let's do that!

gino-m:
Hey @scolsen do you think it's possible to merge with HEAD? If they've diverged too much should we close this one? :cry:

gino-m:
@scolsen Is this PR still active? If not can we close it?

gino-m:
Closing for now due to inactivity. Please reopen if necessary!

PR #1065, opened 2021-11-18
refactor: clean up mbtiles parser; use RX

scolsen:
This commit cleans up the mbtiles footprint parser by:

- deleting unused methods.
- wrapping return values in RX constructs
- making use of collection helpers to reduce stream boilerplate.

It also adds two helper functions to ImmutableList collector to simplify
working with collections in the context of RX streams.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/mbtiles/MbtilesFootprintParser.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/mbtiles/MbtilesFootprintParser.java
@@ -16,6 +16,8 @@
 
 package com.google.android.gnd.persistence.mbtiles;
 
+import static com.google.android.gnd.util.ImmutableListCollector.filterAndRecollect;
+import static com.google.android.gnd.util.ImmutableListCollector.mapAndRecollect;
 import static com.google.android.gnd.util.ImmutableListCollector.toImmutableList;
 import static java8.util.stream.StreamSupport.stream;
 
@@ -24,8 +26,8 @@ import com.google.android.gnd.model.basemap.tile.TileSet;
 import com.google.android.gnd.model.basemap.tile.TileSet.State;
 import com.google.android.gnd.persistence.uuid.OfflineUuidGenerator;
 import com.google.common.collect.ImmutableList;
+import io.reactivex.Single;
 import java.io.File;
-import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
```diff
@@ -271,35 +270,38 @@ public Completable deleteOfflineArea(String offlineAreaId) {
         .map(ImmutableList::asList)
         .flatMap(Flowable::fromIterable)
         .firstOrError()
-        .map(this::identifyAndHandleSource)
+        .flatMap(this::identifyAndHandleSource)
         .doOnError(t -> Timber.e(t, "Couldn't retrieve basemap sources for the active project"));
   }
 
-  private ImmutableList<TileSet> identifyAndHandleSource(BaseMap baseMap) throws IOException {
+  private Single<ImmutableList<TileSet>> identifyAndHandleSource(BaseMap baseMap)
```
gino-m, line 277: It's not immediately clear what this method does from its name. My understanding is that it gets the TileSets comprising the base map (and, that identifying the type of BaseMap is an implementation detail included in that process). Perhaps something like `getTilesSets(BaseMap)` would make more sense? JavaDoc should mention that this may make a remote call.

gnd/src/main/java/com/google/android/gnd/util/ImmutableListCollector.java
```diff
@@ -37,11 +38,37 @@ private ImmutableListCollector() {}
...
   /**
    * Returns a {@link Collector} that accumulates the input elements into a new ImmutableList, in
    * encounter order.
    */
   public static <E> Collector<E, ?, ImmutableList<E>> toImmutableList() {
     return (Collector) TO_IMMUTABLE_LIST;
   }
+
+  /**
+   * Returns a function for use in rx streams that maps a function over the contents of an immutable
```
gino-m, line 53: rx streams - > RxJava observables here and below for consistency with our docs.

gnd/src/main/java/com/google/android/gnd/persistence/mbtiles/MbtilesFootprintParser.java
```diff
@@ -110,20 +68,35 @@
...
+    return getJsonTileSets(file).map(mapAndRecollect(this::jsonToTileSet)).doOnError(Timber::e);
+  }
+
+  /**
+   * Returns the immutable list of tiles specified in {@param geojson} that intersect {@param
+   * bounds}.
+   */
+  public Single<ImmutableList<TileSet>> intersectingTiles(LatLngBounds bounds, File file) {
+    return getJsonTileSets(file)
+        .map(filterAndRecollect(tile -> tile.boundsIntersect(bounds)))
```
gino-m, line 97: Perhaps `compose()`  would be better suited for such higher-order functions?

gino-m (changes requested):
Very cool! Some initial thoughts.

gnd/src/main/java/com/google/android/gnd/persistence/mbtiles/MbtilesFootprintParser.java
scolsen: @gino-m can you clarify a bit here? Do you mean `rx.compose` or `Function.compose`?

RX compose would be nice....if it were generic on on observables, but it's not afaict. One needs to pass an explicit `XTransformer` type to `compose` in which case the signature for these helpers would need to change to `SingleTransformer....` which would reduce the utility of these helpers a lot imo. I may be missing some way to make it generic using java's generics system though.

gnd/src/main/java/com/google/android/gnd/repository/OfflineAreaRepository.java
scolsen: Good call, thanks! Done!

gnd/src/main/java/com/google/android/gnd/util/ImmutableListCollector.java
scolsen: Done!

gino-m (dismissed):
LGTM! Things are getting easier to understand, awesome!

gino-m:
@scolsen Looks like checkStyle is failing

scolsen:
> @scolsen Looks like checkStyle is failing

Thanks! (missing newline) Should be fixed after this run!

gino-m (approved):

PR #1069, opened 2021-11-23
[Add/edit observation] Minor cleanup

gino-m:
@shobhitagarwal1612 I did some renaming and minor refactoring to make things easier to reason about. PTAL?

Contributes to #1067, #1068.

PR #1070, opened 2021-11-23
[Add/edit observation] Save/restore form states to bundle when activity is paused

### LINKED ISSUE #1067: [Add/edit observation] After launching camera underlying form losing state
On launching camera, taking a photo, and coming back sometimes the form looses state (all previously entered data is lost, and a blank form is presented to the user with only one photo that was just taken). In case there are multiple photo entries required, the form can't be filled ever if this keeps happening.

Most probably the underlying fragment / activity is not handling recreation properly.

After trying a few things out - it seems that if you rotate screen when the camera is on, take a photo and come back, the activity is destroyed and recreated and state is lost.
https://stackoverflow.com/questions/16014930/android-activity-getting-destroyed-after-calling-camera-intent

### LINKED ISSUE #1068: Cleared photo reappears on rotating device
**Describe the bug**
The photo response reappears even after removing it manually. This only happens if the screen is rotated after removing the photo.

**To Reproduce**
1. Select any project which contains a photo field
2. Create a new observation and click on add photo
3. Select a new photo 
4. The selected/captured photo should be visible in the edit fragment.
5. Now, remove the photo and rotate the screen.

**Expected behavior**
Removed photo should not reappear

**Actual behavior**
We can see that the photo is visible.

gino-m:
Depends on #1069. [Relevant diffs](https://github.com/google/ground-android/pull/1070/files/a782a64bd1e41cf3b355033cad5419a8c4782a74..980e7ee26889a967e4f6c721cfd9241b3087b249).
Fixes #1067.
Contributes to #1068.

@shobhitagarwal1612 PTAL?

@atalekar FYI.

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/model/form/MultipleChoice.java
+++ b/gnd/src/main/java/com/google/android/gnd/model/form/MultipleChoice.java
@@ -20,11 +20,14 @@ import static java8.util.stream.StreamSupport.stream;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
+import java.io.Serializable;
 import java8.util.Optional;
 
 /** A {@link Field} with pre-defined options for the user to choose from. */
 @AutoValue
-public abstract class MultipleChoice {
+public abstract class MultipleChoice implements Serializable {
+  private static final long serialVersionUID = 1L;
+
   public enum Cardinality {
     SELECT_ONE,
     SELECT_MULTIPLE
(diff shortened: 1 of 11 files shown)


gino-m:
Transient failure.

/gcbrun

gino-m:
@shobhitagarwal1612 RFAL!

shobhitagarwal1612 (approved):
Tested locally. Works as expected. 👍

PR #1076, opened 2021-11-23
[Add/edit observation] Rewrite capture and select photo flows

### LINKED ISSUE #1068: Cleared photo reappears on rotating device

gino-m:
Fixes #1068. Merges commits from #1073. Also

* Minor cleanup and refactors
* Delegate storing photo to `PhotoFieldViewModel`.
* Store field id across activity re-creation.


rachaprince (approved):

rachaprince (approved):

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
```diff
@@ -336,9 +315,53 @@ private boolean hasValidationErrors() {
...
+  @Nullable
+  public String getFieldWaitingForPhoto() {
+    return fieldWaitingForPhoto;
+  }
+
+  public void setFieldWaitingForPhoto(@Nullable String fieldWaitingForPhoto) {
+    this.fieldWaitingForPhoto = fieldWaitingForPhoto;
+  }
+
+  public Observable<Pair<String, Optional<Bitmap>>> getLastPhotoResult() {
```
JSunde, line 326: Would it be worth creating a new class to encapsulate this result so that the fields could be accessed by name in OnPhotoResult() instead of .first and .second?

JSunde (changes requested):

gnd/src/main/java/com/google/android/gnd/ui/editobservation/EditObservationViewModel.java
gino-m: Sure. Done!

JSunde (approved):

Issue #1056, opened 2021-11-17
[Add polygon] Nothing changes when first vertex is added


gino-m:
There is no indication that a vertex was added when Add Point is first clicked. 

@shobhitagarwal1612 Could you PTAL when you have a sec?

shobhitagarwal1612:
A couple of ideas:

1. Setting the default text of button as "Start" which changes to "Add point" after clicking it?

2. Showing selected vertex counter at the top

3. Adding a marker on the start position which gets removed after completing the polygon drawing. 

@gino-m Thoughts?

gino-m:
Hi @shobhitagarwal1612 I'd be in favor of option #3, since it's consistent with the rest of the flow.

We will probably need to add ephemeral round markers (markers that are shown only in add/edit polygon) to the map anyway for two reasons: 1) Maps SDK only shows the round markers as end caps on the poly line while drawing, not on all vertices as expected and 2) the edit flow will need round markers on all vertices as affordances for selecting a vertex to edit.

Wdyt?

shobhitagarwal1612:
That sounds good to me. Since we need handle rendering of transient polygons separately, i feel this should be the order of changes:

1. Separate the rendering of saved polygons and temporary polygons.
2. Create feature from the temporary polygon at the end rather than from the beginning to prevent recreation of feature on every vertex update. Also decouple selected project and layer from PolygomDrawingView, if possible.
3. Add logic to add a marker for each temporary vertex (except the last one).

Wdyt?

gino-m:
Sounds reasonable. To achieve (1), would we decouple `Feature` from the maps interface, and pass points and polygons instead?

PR #1088, opened 2021-11-25
Draw marker for each drawn polygon vertex

shobhitagarwal1612:
Towards #1056

Adds a temporary marker for each selected point while drawing polygon.

@atalekar  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
+++ b/gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
@@ -89,6 +89,9 @@ public class PolygonDrawingViewModel extends AbstractViewModel {
    */
   boolean isLastVertexNotSelectedByUser;
 
+  /** Avoid creating a new uuid to prevent re-drawing everything on render. Reset to */
+  @Nullable private String uuid;
+
   @Inject
   PolygonDrawingViewModel(
       LocationManager locationManager,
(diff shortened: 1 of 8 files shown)


gnd/src/main/java/com/google/android/gnd/ui/map/gms/GoogleMapsFragment.java
```diff
@@ -421,9 +421,14 @@ public void enableCurrentLocationIndicator() {
 
   @Override
   public void setMapFeatures(ImmutableSet<MapFeature> features) {
-    Timber.d("Set map features called : %s", features.size());
+    if (features.isEmpty()) {
```
shobhitagarwal1612, line 424: Is this correct? I feel this might be wrong because it would prevent cleaning up the map if the new activated project has 0 features.

shobhitagarwal1612: Fixed.

atalekar (commented):
quick pass

gnd/src/main/java/com/google/android/gnd/ui/home/mapcontainer/PolygonDrawingViewModel.java
```diff
@@ -185,6 +188,13 @@ private void addVertex(Point vertex, boolean isNotSelectedByUser) {
     updateDrawnPolygonFeature(ImmutableList.copyOf(vertices));
   }
 
+  private String getId() {
+    if (uuid == null) {
+      uuid = uuidGenerator.generateUuid();
```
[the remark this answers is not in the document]
shobhitagarwal1612: I'm not 100% sure since the code gets invoked using RxJava and has multiple operators in between. But, it should be mostly thread safe since the events downstream are serialized and sent sequentially. Also, there is only 1 subject that writes to that stream. Also, this code would be further refactored to move UUID generation at the beginning of polygon drawing in later PRs. Thanks!

atalekar (approved):

PR #1093, opened 2021-12-05
Improve CI build time

### LINKED ISSUE #270: [Cloud Build] Save gradle cache only if dependencies have been updated
Currently, we zip and copy gradle cache to GCS with every successful trigger. This takes ~ 40 seconds and doesn't add any value if the dependencies haven't been updated in the current pull request.

Ideas:
 * Generate hash (md5) for `build.gradle` files (both project and app level)
 * Run `Compress gradle build cache` and `Save gradle cache to GCS` steps only if the combined hash has been modified.
 * Add another step to save the new hashes to GCS

shobhitagarwal1612:
Improve overall run time:
 * Pull & extract cache only if dependencies have not changed
 * Build & save cache only when dependencies have changed

Fixes #270

Current run time: ~13.5 mins
New run time:
Run # | Cache Present in bucket | Cache pulled & extracted | Cache built & saved | Time taken
------|-------------------------|---------------------------|----------------------|------------
 1 a1c24b7 | No / Obsolete | No | Yes | 00:09:37
 2 4b98ce9  | Yes | Yes | No | 00:06:52

Notes:
 * Since the md5 depends on file size, it might result in a collision if only a number is changed in the gradle file. But the improvements are still huge and we can tackle that minor problem later if needed.
 * We should also consider adding `retention policy` to the cache bucket to avoid storing obsolete caches.

@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -44,7 +44,7 @@ steps:
         # 1. List all files except hidden dirs.
         # 2. Filter by filename ending with .gradle
         # 3. Convert continuous spaces into a single space
-        # 4. Pick 5th column containing file size. This causes a collision if only a number is flipped (eg 2.4 -> 2.5). Look for a better alternative.
+        # 4. Pick 5th column containing file size. This causes a collision if only a number is flipped (eg "v2.4" -> "v2.5"). Look for a better alternative.
         # 5. Calculate md5sum (for mac, use md5)
         # 6. Strip ending " -" if present
         ls -lR | grep -e "\.gradle$" | tr -s " " | cut -d " " -f5 | md5sum | sed 's/  -//' > hashsum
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
@gino-m PTAL

cloudbuild.yaml
```diff
@@ -200,13 +212,26 @@ steps:
...
     waitFor:
       - *compress_cache
-    args: [ 'cp', 'cache.tgz', 'gs://${_CACHE_BUCKET}/cache.tgz' ]
+    args:
+      - '-c'
+      - |
+        if [ -f $(< hashsum)-cache.tgz ]; then
+          echo "Cache already exists, skipping save cache"
+        else
+          # TODO: Do not persist indefinitely. Add a retention policy to GCS bucket.
```
gino-m, line 232: Could you create a bug for this and reference as TODO(#..): ?

gino-m (changes requested):
Nice work, @shobhitagarwal1612 . One nit, otherwise LGTM.

cloudbuild.yaml
shobhitagarwal1612: Done

shobhitagarwal1612:
I've cleaned git history now. It got accidently closed due to pushing empty branch.

JSunde (approved):

PR #1100, opened 2022-04-11
Add kotlin support

shobhitagarwal1612:
Add gradle dependency for Kotlin

This is a no-op change.


gino-m (changes requested):
Thanks, Shobhit! One nit, otherwise LGTM

gino-m (approved):
👏

Issue #532, opened 2020-07-07
Migrate to Kotlin


gino-m:

shobhitagarwal1612:
Is there a recommendation for how to approach this? Should we start migrating classes which already have unit tests or should it be used for new classes only?

gino-m:
> Is there a recommendation for how to approach this? Should we start migrating classes which already have unit tests or should it be used for new classes only?

Hi Shobhit, we discussed starting with new classes, tests, and necessary dependencies, and once we’re comfortable, migrating existing classes and tests as we go. Wdyt?

shobhitagarwal1612:
Since I don't have any active PRs, I thought of migrating one or two of the existing classes to kotlin. But adhering to using Kotlin for new/modified code also sounds reasonable. Can't wait to get my hands dirty.

scolsen:
just an FYI, I'm already altering all of the mutation objects, so just a heads up to avoid selecting those -- I'll have a PR for that out today

scolsen:
Think it's safe to consider this one done?

shobhitagarwal1612:
:+1:

scolsen:
Closing, while we're not 100% Kotlin just yet, much of the codebase has been migrated. We can continue to migrate classes over time if it makes sense.

PR #1101, opened 2022-04-11
Convert TermsOfServiceRepository to Kotlin

shobhitagarwal1612:
Towards #532 

Tested for regressions by running the app locally and verified that terms of service is displayed normally.

@JSunde PTAL?


gnd/src/main/java/com/google/android/gnd/repository/TermsOfServiceRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+import io.reactivex.Maybe
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+private const val LOAD_REMOTE_PROJECT_TERMS_OF_SERVICE_TIMEOUT_SECS: Long = 30
+
+class TermsOfServiceRepository @Inject constructor(
+    private val remoteDataStore: RemoteDataStore, private val localValueStore: LocalValueStore) {
+
+    val termsOfService: @Cold Maybe<TermsOfService>
```
scolsen, line 32: you can hide the setter for this if we don't call it elsewhere:

`private set`

scolsen: oops my bad, this is a val so there is no setter

scolsen:
lgtm

shobhitagarwal1612:
Thanks Scott!

PR #1102, opened 2022-04-11
convert mutation models to kotlin

scolsen:
Converts the mutation model classes to kotlin classes. The mutation's have an interesting inheritance relationship that I think is modeled well using Kotlin's sealed classes.

This PR also includes the requisite call site changes. The current interface for these classes is a bit awkward as we need to define builders and methods for using the builders to prevent breakage at existing call sites. Once we're fully a kotlin codebase we won't need builders anymore for the vast majority of cases (if not all of them).


gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
```diff
@@ -0,0 +1,88 @@
...
+    override val projectId: String = "",
+    override val featureId: String = "",
+    override val layerId: String = "",
+    override val userId: String = "",
+    override val clientTimestamp: Date = Date(),
+    override val retryCount: Long = 0,
+    override val lastError: String = "",
+    var observationId: String = "",
+    var form: Form? = null,
+    var responseDeltas: ImmutableList<ResponseDelta> = ImmutableList.of()
```
shobhitagarwal1612, line 22: Is it allowed to create a new `ObservationMutation` with these defaults?

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
```diff
@@ -53,36 +53,14 @@
...
-  @CopyAnnotations
-  @Nullable
-  @Embedded
-  public abstract Coordinates getNewLocation();
-
-  /** Non-empty if a polygon's vertices were updated, null if unchanged. */
-  @CopyAnnotations
-  @Nullable
-  @ColumnInfo(name = "polygon_vertices")
-  public abstract String getNewPolygonVertices();
```
shobhitagarwal1612, line 76: Could you please revert these unrelated code refactors to reduce the noise? Also applicable everywhere else. Thanks!

scolsen: good catch, not even sure why this stuff was moved

gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
scolsen: Technically yes. It's what we want in Kotlin, since that allows us to bypass the builder in Kotlin callers. However, for Java callers I should probably mark it as `JvmSynthetic` so they can only use the builder. Right now Java callers could technically use the constructor too, but they'd need to provide all arguments. It's also possible to use `JvmOverloads` to provide constructors on the Java side where only N arguments use the default

gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
```diff
@@ -0,0 +1,88 @@
...
+            it.responseDeltas = this.responseDeltas
+        }.fromMutation(this) as Builder
+    }
+
+    override fun toString(): String =
+        super.toString() + "deltas= $responseDeltas"
+
+    inner class Builder : Mutation.Builder<ObservationMutation>() {
+        var observationId: String = ""
+            @JvmSynthetic set
```
gino-m, line 38: Is there a plan to get rid of `@JvmSynthetic` once we're 100% Kotlin? If so, can we track the post-migrations TODOs in an issue?

gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
scolsen: In kotlin, we need to initialize all properties, either in the field declaration or in a constructor for the class. Using defaults in a primary constructor like this is basically just shorthand for doing the same thing w/ field declarations. Kotlin doesn't have default initial values like Java does and requires you to initialize things explicitly. (in java uninitialized props are just set to 0 values behind the scenes (or null for reference types)).

So, callers either have to provide every single argument to a constructor, or constructors need to provide sane defaults in the case of long argument lists, in other words, providing default values is the idoimatic way to do what Java builders do in Kotlin, namely making it easier to work w/ long argument lists, e.g.:

```kotlin
ObservationMutation(type = Type.Create, projectId = someProjectId) // all others will be set to constructor defaults
```

here we provide both this approach for Kotlin callers and the explicit Builder for Java callers, mostly so I don't have to change every current call site that used to use the AutoValue builder to a non-builder style.

see: https://stackoverflow.com/questions/36140791/how-to-implement-builder-pattern-in-kotlin

that said, the initial values should really be declared in `Mutation` and not in the inheritors imo. The problem is, you cannot construct sealed classes directly in Kotlin, which bars us from defining these on the parent Mutation class. That said, using a `sealed` class here doesn't benefit us to much (even in the cases where it should (when clauses) the current usage doesn't seem to benefit us) so perhaps the right approach here is to drop `sealed` and declare the defaults on `Mutation` so that they are shared across Feature and Observation mutations. 

But maybe it doesn't even make sense to do this. Like you mention, since default values are an error for this class we should just accept the fact that the class will have a really big constructor that needs a lot of arguments? This would be better in terms of correctness, though less ergonomic.

basically I'm just trying to provide idiomatic kotlin for kotlin callers w/o disrupting the existing Java callers. if you finda a better way to go about it lmk!

gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
scolsen: yeah we should be able to drop them post-migration

gnd/src/main/java/com/google/android/gnd/model/mutation/ObservationMutation.kt
scolsen: though we should provide defaults for the args where it makes sense, e.g. `retryCount`

shobhitagarwal1612: Thanks for the detailed explanation! The current approach lgtm.

gnd/src/main/java/com/google/android/gnd/persistence/local/room/entity/FeatureMutationEntity.java
scolsen: done, for some reason when I run the formatter now it wants to move all this stuff -- could it have something to do with kotlin settings?

shobhitagarwal1612 (approved):

PR #1105, opened 2022-04-14
Converts model classes to Kotlin

shobhitagarwal1612:
Converts the non-complex model classes `ActivityResult`, `Action`, `ValueOrError` and their subclasses to Kotlin.

The classes were first converted to Kotlin using the Android Studio's action "Convert to Kotlin" and then simplified manually following common practices. Some unused code has also been cleaned up as a part of the refactor.


gnd/src/main/java/com/google/android/gnd/MainViewModel.java
```diff
@@ -92,7 +92,7 @@ public MainViewModel(
     disposeOnClear(
         authenticationManager
             .getSignInState()
-            .compose(switchMapIfPresent(SignInState::getUser, userRepository::saveUser))
+            .compose(switchMapIfPresent(SignInState::value, userRepository::saveUser))
```
scolsen, line 95: From a reader's perspective, I feel that there's some information loss here.

Before, I could read the call on SignInState and immediately know it contains a `User` -- now this information is lost since we're just calling the generic method from the superclass.

Maybe we can override the `getter` for `value` in `SignInState` to keep the name `getUser` or at least provide a wrapper with this name to help readability?

gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataEvent.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.android.gnd.rx.ValueOrError
+
+/**
+ * An event returned by the remote data store indicating either an entity was successfully loaded,
+ * or that it was modified or removed in the remote data store.
+ *
+ * @param <T> the type of entity being loaded, modified, or removed.
+</T> */
+class RemoteDataEvent<T> private constructor(
```
scolsen, line 26: Once we fully convert to Kotlin, we should remove this in favor of [kotlin.Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) and just keep the ENUM. Just an FYI. I don't think we can do this until we don't have any Java callers.

scolsen (approved):

gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataEvent.kt
shobhitagarwal1612: Added a TODO

shobhitagarwal1612:
@scolsen Can I get a restamp?

scolsen (approved):

scolsen:
done, thanks, @shobhitagarwal1612!

PR #1106, opened 2022-04-15
Fix license header check for kotlin files

shobhitagarwal1612:
Tested locally by verifying that checkstyle fails if header is missing for kotling files

Fixes the detected errors

@gino-m  PTAL?


gino-m (approved):

gino-m:
Thanks for fixing this!

PR #1108, opened 2022-04-25
Convert AddFeatureDialogFragment to Kotlin

JSunde:
- [x] Convert AddFeatureDialogFragment to Kotlin
- [x] Add AddFeatureDialogFragmentTest
- [x] Set up kapt to get the hilt codegen working properly for @AndroidEntryPoint Kotlin files

@shobhitagarwal1612 @gino-m PTAL


gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.core.util.Consumer
+import androidx.fragment.app.FragmentManager
+import com.google.android.gnd.model.layer.Layer
+import java.lang.NullPointerException
+
+@AndroidEntryPoint
+class AddFeatureDialogFragment @Inject constructor() : AbstractDialogFragment() {
+
+    private var layerConsumer: Consumer<Layer>? = null
+    private var layers: List<Layer>? = null
```
shobhitagarwal1612, line 35: Should we mark these as "lazy init"? That would get rid of NPE and throw a more meaningful exception if anything is accessed before the vars are assigned. We can also cleanup the code by removing the try-catch block for NPE and !! operators.

Thoughts?

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    override fun onDestroyView() {
+        super.onDestroyView()
+        layers = null
+        layerConsumer = null
+    }
+
+    companion object {
+        private val TAG = AddFeatureDialogFragment::class.java.simpleName
```
shobhitagarwal1612, line 81: `companion object` creates an extra instance object. Since this is only used within the same class, should we move this to top-level as `private const`?

gnd/src/test/java/com/google/android/gnd/ui/home/AddFeatureDialogFragmentTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.truth.Truth.assertThat
+import com.google.android.gnd.R
+import com.google.android.gnd.model.feature.FeatureType
+import com.google.common.collect.ImmutableList
+import org.junit.Test
+
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+class AddFeatureDialogFragmentTest : BaseHiltTest() {
+    var addFeatureDialogFragment: AddFeatureDialogFragment? = null
```
shobhitagarwal1612, line 38: `lazy init`?

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.kt
JSunde: Removed companion object, but I was unable to make it const since the initializer is not a constant value

gnd/src/test/java/com/google/android/gnd/ui/home/AddFeatureDialogFragmentTest.kt
JSunde: Done!

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.kt
shobhitagarwal1612: Sorry, I meant `lateinit`. 

I think we can drop the check as well. Right now, we are failing in the else-block anyways. So, `lateinit` would take care of throwing the exception if the var is not initialized. 


```
UninitializedPropertyAccessException: lateinit property foo has not been initialized
```

gnd/src/main/java/com/google/android/gnd/ui/home/AddFeatureDialogFragment.kt
shobhitagarwal1612: Since this is a private var, we could also move it out of the class but in the same file. Then we can define it as `const`.

shobhitagarwal1612 (approved):

PR #1113, opened 2022-05-15
Refactor location lock

### LINKED ISSUE #383: [Main screen] Try to enable location lock when the app is opened
Currently, there viewport is not explicitly defined when the app is opened. We assume users will be collecting data in situ, so it generally makes sense to pan/zoom the map to their current location.

Note: the project selector will be moved to its own fragment in which the map will not be shown, so for now it's ok to show attempt location lock as soon as the HomeScreenFragment is shown.

shobhitagarwal1612:
Closes #383

Things done:

- [x] Create a shared layout for location lock 
- [x] Create a shared view model for location lock
- [x] Cleanup existing usages. This also ensures that the state is maintained across screens
- [x] Trigger location lock toggle when app starts.
- [x] Create a module for LocaionManager. Then uninstall the module for unit test and replace with a mock


gnd/src/main/res/layout/location_lock_btn.xml
```diff
@@ -0,0 +1,40 @@
...
+    <variable
+      name="viewModel"
+      type="com.google.android.gnd.ui.home.LocationLockViewModel" />
+  </data>
+
+  <com.google.android.material.floatingactionbutton.FloatingActionButton
+    android:id="@+id/location_lock_btn"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:enabled="@{viewModel.locationLockEnabled}"
```
JSunde, line 32: Huh, I don't think I've seen this syntax in XML before, is that Kotlin?

gnd/src/main/java/com/google/android/gnd/ui/home/LocationLockViewModel.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.google.android.gnd.ui.common.SharedViewModel
+import com.google.android.gnd.util.toLiveData
+import io.reactivex.BackpressureStrategy
+import io.reactivex.Flowable
+import io.reactivex.subjects.PublishSubject
+import io.reactivex.subjects.Subject
+import javax.inject.Inject
+
+@SharedViewModel
+class LocationLockViewModel @Inject constructor(private val locationManager: LocationManager) :
```
JSunde, line 36: Could you add tests for this file?

JSunde (changes requested):

gino-m:
> @... PTAL?

Nit: Please remember to update/remove PR description template strings before sending :)

gnd/src/main/res/layout/location_lock_btn.xml
shobhitagarwal1612: This is data binding. (not related to Kotlin)

https://developer.android.com/topic/libraries/data-binding

gnd/src/main/java/com/google/android/gnd/ui/home/LocationLockViewModel.kt
shobhitagarwal1612: Done

gnd/src/main/res/layout/location_lock_btn.xml
JSunde: Gotcha, thanks for the info!

JSunde (approved):

shobhitagarwal1612:
Adding coarse permissions also didn't help.

https://alexzh.com/ui-testing-of-android-runtime-permissions/

shobhitagarwal1612:
Closing this PR for now

Issue #1114, opened 2022-05-16
[Migration] Rename classes and members in code


JSunde:
This change should be restricted to references at the code level and should not be applied to the collection and document names in Firestore:

- [x] Project -> Survey [@shobhitagarwal1612]
- [x] Layers -> Job [@shobhitagarwal1612]
- [x] Forms -> Task [@jsunde]
- [x] Feature -> LocationOfInterest [@scolsen]
- [x] Element -> Step [@JSunde]
- [x] Observation -> Submission [@JSunde]
-  User roles:
  - [x] manager -> survey-organizer  [@JSunde]
  - [x] contributor -> data-collector [@JSunde]

JSunde:
I'll start working on renaming Observation -> Submission

JSunde:
I don't really see any results when searching for "Question" in the codebase. @gino-m Was there something else we should be renaming to Step or is it a newer concept that doesn't already exist in the codebase?

gino-m:
> I don't really see any results when searching for "Question" in the codebase. @gino-m Was there something else we should be renaming to Step or is it a newer concept that doesn't already exist in the codebase?

I think `Question` was never explicitly modeled because in the prototype it was 1:1 with form `Element`s. Updating the GitHub issue accordingly.

JSunde:
What are some examples of `manager` that need to be renamed? I'm having a hard time finding it with all of the other usages like FragmentManager, LayoutManager, AuthenticationManager...

gino-m:
> What are some examples of `manager` that need to be renamed? I'm having a hard time finding it with all of the other usages like FragmentManager, LayoutManager, AuthenticationManager...

Ah, that was referring to renaming of User roles, not the keyword "manager" anywhere it occurs. Sorry for the confusion.

PR #1131, opened 2022-05-19
Rename Observation -> Submission

JSunde:
Addresses one of the planned renames from #1114


shobhitagarwal1612:
We should also update the db schema version in Config.

JSunde:
> We should also update the db schema version in Config.

Is that as simple as bumping the DB_VERSION?

shobhitagarwal1612:
> Is that as simple as bumping the DB_VERSION?

Yes lets increment that by +1. This would ensure that the existing apps auto-clear the db. Otherwise the app would fail to load.

shobhitagarwal1612 (approved):
LGTM

PR #1141, opened 2022-06-04
Remove "+" fab

### LINKED ISSUE #1123: [Map] Remove "+" fab
This is no longer useful, since LOIs can only be added to the map by survey organizers; data collectors cannot draw LOIs on the map. Once “Capture GPS”, “Drop a pin”, etc. data collection steps are implemented, they’ll be able to submit geospatial data as part of the task. Note that those points/polygons will not show on the default map; they’ll only be shown when reviewing the data submitted for a particular task.

shobhitagarwal1612:
Fixes #1123 

1. Removes the "+" button in home screen, used for adding new features. This won't be allowed in the new flow.
2. Removes the related instrumentation test.
3. Doesn't remove the unused dialog fragment classes yet. Hoping some of these could be repurposed in the new UX

@JSunde  PTAL?


shobhitagarwal1612:
> Do we still need the AddFeatureDialogFragment? Or can that be removed as well?

Cleaned up.

JSunde (approved):

PR #1144, opened 2022-06-15
Update User Role names and convert enum to Kotlin

JSunde:
Towards #1114, rename User Roles

@gino-m PTAL?


JSunde:
I'm guessing we don't want to update the roles in ProjectsCollectionReference just yet since that is interacting with Firestore and we haven't updated them in Firestore yet. Is that right??

gino-m:
> It looks like UserRepositoryTest.testGetUserRole() is failing because we map from the Firestore roles to the Roles enum, so this whole change might need to happen in tandem with the web client update.

If it's not too much code, feel free to explicitly convert between role strings and enum values. We probably shouldn't be relying on the enum values anyway!

JSunde:
> > It looks like UserRepositoryTest.testGetUserRole() is failing because we map from the Firestore roles to the Roles enum, so this whole change might need to happen in tandem with the web client update.
> 
> If it's not too much code, feel free to explicitly convert between role strings and enum values. We probably shouldn't be relying on the enum values anyway!

Actually I think I was wrong, we simply had the role strings hardcoded in the tests, so updating those got the tests to start passing again.

JSunde:
@gino-m PTAL

gnd/src/sharedTest/java/com/google/android/gnd/FakeData.java
```diff
@@ -60,7 +60,7 @@ public static Survey.Builder newSurvey() {
         .setId("SURVEY")
         .setTitle("Survey title")
         .setDescription("Test survey description")
-        .setAcl(ImmutableMap.of(FakeData.USER.getEmail(), "contributor"));
+        .setAcl(ImmutableMap.of(FakeData.USER.getEmail(), "data_collector"));
```
gino-m, line 63: Perhaps roles in the db it should in slug case (eg `data-collector`, dash instead of underscore) for consistency with examples in Firestore docs? This seems to be more idiomatic for the Firestore's JSON-like data structure.

If this is because you plan to add explicit mappings between enum names and db strings in a followup PR please disregard.

gnd/src/main/java/com/google/android/gnd/model/Role.kt
```diff
@@ -13,12 +13,8 @@
...
-
-public enum Role {
-  UNKNOWN,
-  OWNER,
-  MANAGER,
-  CONTRIBUTOR
-}
+enum class Role {
+    UNKNOWN, OWNER, SURVEY_ORGANIZER, DATA_COLLECTOR
+}
```
gino-m, line 20: Are you using `google-java-format`? It seems odd that LF is messing at EOF.

gino-m (changes requested):

gnd/src/sharedTest/java/com/google/android/gnd/FakeData.java
JSunde: I think we need the underscore for now for the existing enum name mapping to work.

Do we need to update the mapping to map from "data-collector" -> Role.DATA_COLLECTOR instead of "data_collector" -> Role.DATA_COLLECTOR?

gino-m: > I think we need the underscore for now for the existing enum name mapping to work.

Ok, SGTM.

> Do we need to update the mapping to map from "data-collector" -> Role.DATA_COLLECTOR instead of "data_collector" -> Role.DATA_COLLECTOR?

Yep, once we have the explicit mapping in place.

gino-m (approved):
LGTM!

PR #1145, opened 2022-06-16
fix: execute navigations on the main thread

scolsen:
This fixes an issue on startup whereby navigations were executed on the
IO thread (which android doesn't allow). We compose the auth manager
stream with the userRepository. The userRepository returns an observable
that observes on the IO thread; post-composition, the resulting stream
takes the observation thread of the compose argument; so the final
stream was being observed on the IO thread.

The fix is straightforward: set the resulting stream to observe on the
UI thread after the composition is performed.

@gino-m @shobhitagarwal1612

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/MainViewModel.kt
+++ b/gnd/src/main/java/com/google/android/gnd/MainViewModel.kt
@@ -73,8 +73,8 @@ class MainViewModel @Inject constructor(
         disposeOnClear(
             authenticationManager
                 .signInState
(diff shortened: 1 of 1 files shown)


scolsen:
This seems...easy to get wrong. Apparently order matters when it comes to calling `observerOn` and `compose`. This makes sense when we think of the meaning of `compose` but it still seems easy to get wrong (imo). Maybe this won't matter if we can eventually drop RX in favor of coroutines.

shobhitagarwal1612 (approved):
Thanks!

Issue #1122, opened 2022-05-18
[Migration] Rename references to data in Firestore


gino-m:
Most code that defines our implicit "schema" lives in: 
https://github.com/google/ground-android/tree/master/gnd/src/main/java/com/google/android/gnd/persistence/remote/firestore/schema

To be applied to Android only, not to actual data itself:

- [x] projects -> surveys [@shobhitagarwal1612]
- [x] layers -> jobs [@shobhitagarwal1612] 
- [x] forms -> tasks  [@shobhitagarwal1612] 
- [x] features -> lois [@shobhitagarwal1612] 
- [x] question -> step [@JSunde]
- [x] observation -> submission [@JSunde]
- [x] manager -> survey-organizer [@JSunde]
- [x] contributor -> data-collector [@JSunde]

This should be coordinated carefully with https://github.com/google/ground-platform/issues/867.

PR #1155, opened 2022-06-24
Rename references of "feature" to "loi" in firebase schema

shobhitagarwal1612:
Towards #1122 

@gino-m  PTAL?

### DIFF
--- a/gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataStore.java
+++ b/gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataStore.java
@@ -57,16 +57,15 @@ public interface RemoteDataStore {
   Maybe<TermsOfService> loadTermsOfService();
 
   /**
-   * Returns all locationOfInterests in the specified survey, then continues to emit any remote
-   * updates to the set of locationOfInterests in the survey until all subscribers have been
-   * disposed.
+   * Returns all lois in the specified survey, then continues to emit any remote updates to the set
+   * of lois in the survey until all subscribers have been disposed.
    */
   @Cold(stateful = true, terminates = false)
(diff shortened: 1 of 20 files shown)


gino-m (changes requested):
I would consider calling these "lois" in Firestore instead of "locationOfInterest". Naming conventions I've seen in docs seem to prefer shorter, one word names. It should also be clear from code that these are associated with the LocationOfInterest class, hopefully making this name self-documenting.

@scolsen Wdyt? Does that sound reasonable?

scolsen:
Sounds good to me!

I really slacked on my in-progress renaming and have some merging to sort out now 😅

shobhitagarwal1612:
@gino-m PTAL! I've updated the references

gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataStore.java
```diff
@@ -57,15 +57,15 @@ public interface RemoteDataStore {
   Maybe<TermsOfService> loadTermsOfService();
 
   /**
-   * Returns all features in the specified survey, then continues to emit any remote updates to the
-   * set of features in the survey until all subscribers have been disposed.
+   * Returns all lois in the specified survey, then continues to emit any remote updates to the set
```
gino-m, line 60: This is an acronym, so in comments this should be capitalized as "LOI" or "LOIs".

gino-m (approved):
LGTM, just one nit.

gnd/src/main/java/com/google/android/gnd/persistence/remote/RemoteDataStore.java
shobhitagarwal1612: Done

Issue #1150, opened 2022-06-23
[Code health] Rename gnd to ground


gino-m:
- [x] Rename package
- [x] Rename root dir

gino-m:
Sorry, we missed one two classes:

- [ ] `GndApplication`
- [ ] `GndApplicationModule`

@shobhitagarwal1612 Could you PTAL?

PR #1157, opened 2022-06-24
Rename package to com.google.android.ground

shobhitagarwal1612:
Towards #1150

@gino-m  PTAL?

PR #1158, opened 2022-06-24
Rename root dir to groumd

shobhitagarwal1612:
Towards #1150 

@gino-m PTAL?


gino-m (approved):
I assume a restamp and big merge will be needed one #1157 is merged? As with that PR, please coordinate with @JSunde and @scolsen.

PR #1162, opened 2022-06-28
Rename Feature -> LocationOfInterest

scolsen:
Renames all instances of Feature to LocationOfInterest *except* for classes in `persistence/remote/` (model classes in these packages are renamed, however, to ensure the code builds).

Also renames `PointFeature` to `PointOfInterest` and `PolygonFeature` to `PolgyonOfInterest`.

Contributes to #1114


gino-m:
Cool, thanks! How does spelling out LOI feel so far? Acceptable or burdensome?

> Also renames PointFeature to PointOfInterest and PolygonFeature to PolgyonOfInterest.

Going forward I don't think we'll need separate classes for different typing of LOIs; they'll just have Geometry which will contain that complexity. We'll still have to model MultiPolygon, Point, etc. but I believe we already have those modeled elsewhere. No changed needed in this PR, just wanted it to be on everyone's radar.

scolsen:
oops, not sure why there's a bunch of errorprone errors for generated code

shobhitagarwal1612:
gentle ping!

shobhitagarwal1612 (approved):
Thanks!

PR #1163, opened 2022-06-29
 Fix locate me on polygon drawing view

rachaprince:
Fixes an issue where locate me does not work when drawing a polygon. 

To verify bug fix:
1. Enter polygon drawing mode
2. Tap the location lock button

screen recording of the fix:
https://user-images.githubusercontent.com/7172296/177991023-0217dc4a-59b9-44ea-a378-6dc02f3b31e2.mov


gino-m:
@shobhitagarwal1612 PTAL? Also, is this a fix we might be able to merge into HEAD as well?

shobhitagarwal1612 (approved):

shobhitagarwal1612:
> Also, is this a fix we might be able to merge into HEAD as well?

I haven't tested with the latest version. Will file a bug to track this.

Issue #1146, opened 2022-06-21
[Data collection] New data collection flow


gino-m:
@ngosue Please to mocks or a public link here when ready!

@jsunde This issue is intended to only include the basic "answer a question" (text) and "take a picture" data collection tasks. Other tasks can be split among the team.

JSunde:
I believe the data collection flow for all of the following P0 task types has been completed.

P0 Task Types:
- [x] Take Photo
- [x] Answer a Question
    -  [x] Text
    -  [x] Select one (radio button)
    -  [x] Select multiple (checkbox)
    
I think the only thing I can think of that is remaining before we can close this issue is the summary screen at the end of the data collection flow. Currently the job is completed when the user hits "Continue" on the final task in the job, but I believe we want to show a summary screen with a list of the tasks that the user responded to before the job is completed.

@gino-m can you remind me if that screen is in scope for the current prototype? And if so do you think it should be included in this issue or a follow up one?

gino-m:
> I believe the data collection flow for all of the following P0 task types has been completed.
> 
> P0 Task Types:
> 
> * [x]  Take Photo
> * [x]  Answer a Question
>   
>   * [x]  Text
>   * [x]  Select one (radio button)
>   * [x]  Select multiple (checkbox)
> 
> I think the only thing I can think of that is remaining before we can close this issue is the summary screen at the end of the data collection flow. Currently the job is completed when the user hits "Continue" on the final task in the job, but I believe we want to show a summary screen with a list of the tasks that the user responded to before the job is completed.
> 
> @gino-m can you remind me if that screen is in scope for the current prototype? And if so do you think it should be included in this issue or a follow up one?

I can't find the mocks for that screen, and it's not essential for the e2e flow to work. I think we can close the data collection flow and file that as a separate P2 FR.

PR #1165, opened 2022-07-07
Get submission in DataCollectionViewModel and set title based on the Job's name

JSunde:
Towards #1146


gino-m:
Note that you'll have an existing submissionId only in the case of editing an existing submission, which is a P2 edge case. The critical path is for submitting new data, in which case the outcome of this flow will be a new submission.

gino-m:
> Gotcha, yea that makes sense. In that case I think I can revert my changes to Task and pull the name from `submission.locationOfInterest.job.name`

The submission should be associated with both a job and a location, so it would be: `submission.job.name` and `submission.locationOfInterest.name`.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.java
```diff
@@ -409,8 +409,15 @@ private void showSurveySelector() {
   }
 
   private void showDataCollection() {
+    String dummySurveyId = "123";
```
gino-m, line 412: Maybe add a TODO referencing the GitHub issue mentioning these are temporary.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -15,10 +15,57 @@
...
  * View model for the Data Collection fragment.
  */
-class DataCollectionViewModel @Inject internal constructor() : AbstractViewModel()
\ No newline at end of file
+class DataCollectionViewModel @Inject internal constructor(
+    private val submissionRepository: SubmissionRepository,
+    private val locationOfInterestHelper: LocationOfInterestHelper
+) : AbstractViewModel() {
+    val submission: @Hot(replays = true) LiveData<Loadable<Submission>>
+    val title: @Hot(replays = true) LiveData<String>
```
gino-m, line 39: I think we can model these more specifically as `jobName` and `loiName` or `locationOfInterestName`. [I think abbreviations are fine as variable names]

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -15,10 +15,57 @@
...
+import io.reactivex.processors.BehaviorProcessor
+import io.reactivex.processors.FlowableProcessor
 import javax.inject.Inject
 
 /**
  * View model for the Data Collection fragment.
  */
-class DataCollectionViewModel @Inject internal constructor() : AbstractViewModel()
\ No newline at end of file
+class DataCollectionViewModel @Inject internal constructor(
```
gino-m, line 34: Please add unit tests for this class

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -49,11 +46,19 @@ class DataCollectionFragment : AbstractFragment() {
         super.onCreateView(inflater, container, savedInstanceState)
         val binding = DataCollectionFragBinding.inflate(inflater, container, false)
 
+        val args = getDataCollectionFragmentArgs()
```
gino-m, line 49: Would it be possible to add a test for this behavior? (load submission on fragment creation)

gino-m (changes requested):
Lg so far, a few suggestions.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.java
JSunde: Good idea. Done!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionViewModelTest.kt
```diff
@@ -0,0 +1,123 @@
...
+                any(),
+                any(),
+                any()
+            )
+        ).doReturn(Single.create {
+            submission
+        })
+    }
+
+    @Test
```
JSunde, line 106: For some reason both of these tests are always failing because the actual jobName and loiName are both null.

I wonder if I might be missing something in terms of waiting for the result of the load to propagate? Let me know if you have any ideas!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionViewModelTest.kt
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright 2021 Google LLC
```
gino-m, line 2: 2022

gino-m (changes requested):

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionViewModelTest.kt
gino-m: Actually, that's probably expected, since `observe()` always replays the latest value. If there's a race condition with `loadSubmissionDetails()` then the second call will get the initial value as well. Since this is async logic, we probably need something that captures value up to some timeout.

gino-m: Looks like you need to add and remove and `Observer`, and to use `InstantTaskExecutorRule` to ensure all background tasks are run on the same blocking thread:
https://developer.android.com/codelabs/advanced-android-kotlin-training-testing-basics#8

JSunde: So I tried following that codelab, but the test fails with the message "LiveData value was never set." after I use the LiveDataTestUtil class they provide and the `InstantTaskExecutorRule`.

I also tried the step before the LiveDataTestUtil which was basically the same as my previous test, but with the `InstantTaskExecutorRule` and the values were still null.

gino-m: @dturner Any idea what we might be missing here?

dturner: > So I tried following that codelab, but the test fails...

Is it the codelab which is failing, or just when you copy/paste code from the codelab into the ground codebase?

JSunde: The tests in DataCollectionViewModelTest are failing when I am trying to use the copy pasted LiveDataTestUtil code from the codelab.

The jobName and loiName LiveData are apparently never getting set (at least according to the LiveDataTestUtil).

dturner: Few thoughts off the top of my head: 

1) Is 

```
@get:Rule
var instantTaskExecutorRule = InstantTaskExecutorRule()
```

the first line in `DataCollectionViewModelTest`?

2) From the codelab: 
 
> This rule runs all Architecture Components-related background jobs in the same thread 

On what thread are you writing the initial livedata values? It could be that `InstantTaskExecutorRule` is not forcing that thread to run synchronously. 

3) Are you using the androidx version of [`InstantTaskExecutorRule`](https://developer.android.com/reference/androidx/arch/core/executor/testing/InstantTaskExecutorRule) - the one in the codelab is deprecated?

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun setUp() {
+        super.setUp()
+
+        val activityController = Robolectric.buildActivity(
+            MainActivity::class.java
+        ).setup()
+
+        val activity = activityController.get()
+
+        // TODO(jsunde): Figure out the best way to instantiate a DataCollectionFragment here with
```
JSunde, line 61: @gino-m Do you have any suggestions for launching the DataCollectionFragment directly from the test here? I tried using the Navigator, but that doesn't seem to work because it is on the startup page by default and can only navigate to the DataCollectionFragment from the HomeScreen.

I also tried navigating to the HomeScreen first, but I got an error saying "Main looper has queued unexecuted runnables...".

JSunde: I found out you can call `.toBundle()` on the FragmentArgs so I was able to launch a DataCollectionFragment with the args that way.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
JSunde: Done!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -15,10 +15,57 @@
...
+
+    private val argsProcessor: @Hot(replays = true) FlowableProcessor<DataCollectionFragmentArgs> =
+        BehaviorProcessor.create()
+
+    init {
+        val submissionStream: Flowable<Loadable<Submission>> =
+            argsProcessor.switchMapSingle { args ->
+                submissionRepository.createSubmission(
+                    args.surveyId, args.locationOfInterestId, args.submissionId
+                ).map { Loadable.loaded(it) }.onErrorReturn { Loadable.error(it) }
```
scolsen, line 50: Instead of using `Loadable` here, can we just make the return value nullable and leverage nullability checks?

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Great idea! Done!

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -35,32 +35,32 @@ class DataCollectionViewModel @Inject internal constructor(
...
         BehaviorProcessor.create()
 
     init {
-        val submissionStream: Flowable<Loadable<Submission>> =
+        val submissionStream: Flowable<Submission?> =
             argsProcessor.switchMapSingle { args ->
                 submissionRepository.createSubmission(
                     args.surveyId, args.locationOfInterestId, args.submissionId
-                ).map { Loadable.loaded(it) }.onErrorReturn { Loadable.error(it) }
+                ).map { it }.onErrorReturn { null }
```
shobhitagarwal1612, line 50: Is `map()` operator required if we are not transforming the object? Same question for `onErrorReturn()`

What would happen if we remove both?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -35,32 +35,32 @@ class DataCollectionViewModel @Inject internal constructor(
...
             }
 
         submission = LiveDataReactiveStreams.fromPublisher(submissionStream)
 
-        jobName = LiveDataReactiveStreams.fromPublisher(submissionStream.map { submission ->
-            submission.value().map { it.locationOfInterest.job.name }.orElse("")
+        jobName = LiveDataReactiveStreams.fromPublisher(submissionStream.map {
+            it.locationOfInterest.job.name ?: ""
```
shobhitagarwal1612, line 56: Out of curiosity, this LiveData is used only by the text field in UI, right? Isn't that already null safe?

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Looks like we can remove them both. Thanks for the suggestion!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: I think for loiName it's safe because it's used in the TwoLineToolbar which accepts a `String?` in `setTitle()`, however the jobName is used in a `TextView` which doesn't accept null values, so I think I'd rather just keep the strings non-null here.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Actually it looks like I get an error after removing the onErrorReturn so I added that back in.

PR #1169, opened 2022-07-20
Refactor: Migrate LOI models to Kotlin

scolsen:
Migrates the LOI classes to Kotlin. We define temporary builders to
prevent breaking existing Java callers.

Furthermore, this commit renames PolygonOfInterest->AreaOfInterest and
updates names for fake test data.


scolsen:
lol, I have yet to open a PR that didn't fail on the first run thanks to import order problems...should be fixed now

ground/src/main/java/com/google/android/ground/model/TestModelBuilders.java
```diff
@@ -17,9 +17,9 @@
 package com.google.android.ground.model;
 
 import com.google.android.ground.model.job.Job;
+import com.google.android.ground.model.locationofinterest.AreaOfInterest;
```
gino-m, line 20: Unrelated: I'd be in favor of using `loi` for package and variable names.

gino-m (approved):

ground/src/main/java/com/google/android/ground/model/TestModelBuilders.java
scolsen: sgtm

gino-m: Also, let's discuss how we can consolidate our representation of point, polygon, and geojson in code and in the db.

PR #1170, opened 2022-07-20
[Code Health] Replace string maps with strings

### LINKED ISSUE #1132: [Code health] Replace string maps with simple strings
Previously, `Map<String, String>` was used for forward compatibility with multilingual surveys. This leads to a lot of boilerplate and additional complexity throughout the codebase. Instead, we can simply use `String`.

If and when we introduce multilingual surveys, we can allow the user to select a single language when opening the survey. We can store alternative translations separately in the db (at top level in the survey, for example), making additional of multilingual surveys an additive, non-breaking change.

gino-m:
Fixes #1132.

@scolsen PTAL?

@shobhitagarwal1612 @JSunde FYI.


scolsen (approved):
~So we're dropping localization?~ EDIT: read the issue :) sg!

PR #1173, opened 2022-07-21
Add GeoTools Dependency

scolsen:
[GeoTools](https://www.geotools.org/) is an open source library for
processing and working with geospatial data. We'll use it to handle
geometries and geojson going forward. This is a first step toward
simplifying and improving our model of map features and locations of
interest.


gino-m (approved):
🗺️ ! Thanks!

Issue #929, opened 2021-06-28
[Code health] Harmonize geometry representation


gino-m:
Shapes are represented in remote db in features as `point` (GeoPoint), `geoJson` (string) and `geometry` (map). There's also some confusion in the implementation about geometries vs LOIs. This design wasn't intentional and we can improve it.

Ideas:
- [x] Define generic geometry types independent of Ground model (perhaps import ~[GeoTools](https://docs.geotools.org/latest/userguide/library/jts/geometry.html)?~ [JTS](https://github.com/locationtech/jts))
- [x] Instead of extending `LocationOfInterest`, have it contain an arbitrary `Geometry`
- [x] Replace separate point, geoJson db format mirror GeoJSON where appropriate, but use Firebase `GeoPoint` for coordinates. Avoid nested arrays since they're unsupported by Firebase.
- [x] Load/save in local db (@scolsen)
- [x] Update map to render new Geometry types (@jsunde)
- [x] Delete old subclasses of `LocationOfInterest` (@scolsen)

scolsen:
> Instead of extending LocationOfInterest, have it contain an arbitrary Geometry

This is a definite yes.

> Define generic geometry types independent of Ground model (perhaps import [GeoTools](https://docs.geotools.org/latest/userguide/library/jts/geometry.html)?)

I think we should probably do this too, but I'm not sure if we should use GeoTools or not—I'd like to refactor LOIs first sticking with our own representation then figure it out from there.

gino-m:
> > Define generic geometry types independent of Ground model (perhaps import [GeoTools](https://docs.geotools.org/latest/userguide/library/jts/geometry.html)?)
> 
> I think we should probably do this too, but I'm not sure if we should use GeoTools or not—I'd like to refactor LOIs first sticking with our own representation then figure it out from there.

Good idea. Not sure either, but I see GeoTools has a permission license (LGPL), and includes [de]serialization of their model to/from GeoJSON.

scolsen:
Yeah, I looked at it a bit more and it seems like its very robust. At the moment, I'm just trying to get gradle set up to try it out.

gino-m:
We spoke. I'll summarize here:

- [ ] Define model classes and GeoJSON (de)serialization (import GeoTools?) @scolsen 
- [ ] (De)serialize geometries as GeoJSON in local db
[...]

PR #1174, opened 2022-07-21
Convert Geometries to Firestore compatible format

gino-m:
Work towards #929.


gino-m:
@scolsen  Should be ready for review! PTAL?

ground/build.gradle
```diff
@@ -206,7 +206,9 @@ dependencies {
     implementation "com.google.maps.android:android-maps-utils:2.2.0"
 
     // GeoTools
-    implementation "org.geotools:gt-main:$project.geotoolsVersion"
+    implementation "org.locationtech.jts:jts-core:1.19.0"
+    implementation "org.locationtech.jts.io:jts-io-common:1.19.0"
+    //implementation "org.geotools:gt-geojson-core:$project.geotoolsVersion"
```
scolsen, line 211: Delete? Also, I think we can remove the repo URL in the other gradle file if we remove this.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
```diff
@@ -0,0 +1,83 @@
...
+    fun toFirestoreMap(geometry: Geometry): Map<*, *> {
+        val writer = GeoJsonWriter()
+        writer.setEncodeCRS(false)
+        val jsonString = writer.write(geometry)
+        return toFirestoreValue(Gson().fromJson<MutableMap<String, *>>(jsonString))
+    }
+
+    private fun toFirestoreValue(value: Map<*, *>): Map<*, *> {
+        return value.mapValues {
+            toFirestoreValue(it.value!!)
```
scolsen, line 57: We should avoid `!!` Instead:

```
private fun toFirestoreValue(value: Map<*,*>): Map<*,*> = 
    value.filterValues { it != null }.mapValues { toFirestoreValue(it.value as Any) }
```

Unfortunately the kotlin stdlib doesn't define `mapValuesNotNull` which is what we want here.

scolsen: You might be able to enforce the same thing by changing the type of Map s.t. the value place is `Any` (v.s. `*` which includes `Any?` in this case). Note that this might require changing the bound on `T` to `Any` (exclude nullable types).

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    private fun toFirestoreValue(value: Any): Any {
+        return when (value) {
+            is ArrayList<*> -> {
+                if (value.size == 2 && value.all { it is Double }) {
+                    GeoPoint(value[0] as Double, value[1] as Double)
```
scolsen, line 65: Instead of casting here, why not match against a more specific type?

```
is ArrayList<Double> ->
```

Or does the compiler dislike that for some reason?

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
```diff
@@ -0,0 +1,83 @@
...
+                toFirestoreValue(value)
+            }
+            else -> {
+                value
+            }
+        }
+    }
+
+    private fun listToMap(list: List<*>): Map<Int, Any> {
+        return list.mapIndexed { index: Int, value -> index to toFirestoreValue(value!!) }
```
scolsen, line 80: Avoid

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
scolsen: Then you'd remove the `if` and just have a separate `when` branch:

```
is ArrayList<Double> -> ...
is ArrayList<*> -> listToMap(value) 
```

Also, should we be using an interface like `Collection` here instead of `ArrayList`?

scolsen (approved):

ground/build.gradle
gino-m: Ah yes, I missed this one!

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
gino-m: The latter worked, thanks for the suggestion!

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
gino-m: How about this?

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
gino-m: Similar to `instanceof` in Java, you can't check the type of generic with `is`:
```
Cannot check for instance of erased type: kotlin.collections.ArrayList<Double> /* = java.util.ArrayList<Double> */
```

gino-m (commented):
RFAL!

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
scolsen: oh, yeah, erasure. sigh

scolsen (approved):

PR #1175, opened 2022-07-22
Refactor Geometry and LOI models.

scolsen:
We now decouple the notion of a location of interest's geometry from our
representation of geometrical objects. We have a single LOI type which
contains various geometrical types (point, polygon, etc.).

Note that the new Polygon representation is stricter than our previous
representation. The outershell of a Polygon now *must* form a linear
ring. I've had to update a few tests to reflect this.

This PR also removes the GeoJSON object, we'll only use GeoJSON as a transport representation.


build.gradle
```diff
@@ -68,7 +68,7 @@ allprojects {
         mavenLocal()
         maven { url "https://maven.google.com" }
         maven { url "https://jitpack.io" }
-        maven { url "https://repo.osgeo.org/repository/release/" } // GeoTools
+        // maven { url "https://repo.osgeo.org/repository/release/" } // GeoTools
```
gino-m, line 71: Remove?

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.google.android.ground.model.geometry
+
+import org.locationtech.jts.geom.Coordinate
+import org.locationtech.jts.geom.GeometryFactory
+import org.locationtech.jts.geom.Point
+
+/** A single point given by x and y coordinates in a Cartesian plane. */
+data class Point(val x: Double, val y: Double) : Geometry {
```
gino-m, line 24: Why do we need these adapter classes? Can we just reference JTS Geometry directly in LOI?

gino-m (changes requested):
Initial comments (high-level).

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
scolsen: We could, but there are some benefits to wrapping:

1. For comparisons, we want `data class` (a.k.a AutoValue) behavior. e.g. two points are equal by value, not by reference. (at least this was our previous behavior and it seems correct).
2. If we decide JTS doesn't work down the line, we can swap it out for a different underlying library (though this is unlikely, from an OSS perspective it also gives forks the opportunity to do this).
3. The wrapper classes give us a convenient place to provide utilities JTS doesn't provide as built-ins (e.g. if a PolyLine is a Ring, convert it to a Polygon).
4. When it comes to converting coordinates to map impl representations (e.g. LatLng) these classes give us a convenient boundary for middle-man conversion to a general base-type representation (Coordinate -> Double), v.s. needing to implement this logic for every Map, we only implement the second half (Java type to Map types) (Double -> LatLng).

That said, we could just use them directly. Based on working with the code so far, this feels a bit cleaner. LOI and the rest of the application doesn't need to know anything about JTS.

gino-m: That all makes sense, though then it begs to question - if we have our own representations, why use JTS at all? GeoJSON (de)serialization is easy enough to implement if we're only defining a small subset of geometries. My understanding was that we'd use JTS as if it were a standard library like Java/Kotlin Collections to simplify the model.

One alternative could be to use Kotlin extension functions to simplify JTS usage. Let's discuss offline.

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    override val geometry: Polygon = factory.createPolygon(coordinates.toTypedArray())
+    override val type: GeometryType = GeometryType.POLYGON
+    override val isClosed: Boolean = true
+
+    val vertices: ImmutableList<Point>
+        get() = geometry.coordinates.map { Point(it.x, it.y) }.toImmutableList()
+
+    /** Returns the first point in this polygon's coordinates, or null if the polygon has no coordinates.*/
+    val firstVertex: Point?
```
gino-m, line 42: Can the caller work with `Coordinate` directly? Then they can just call `get()` directly.

Also, is it possible for a `Polygon` to have no coordinates?

ground/src/main/java/com/google/android/ground/model/geometry/MultiPolygon.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.locationtech.jts.geom.MultiPolygon
+
+/** A grouping of one or more [Polygon]s. */
+data class MultiPolygon(val polygons: Collection<Polygon>) : Geometry {
+    private val factory: GeometryFactory = GeometryFactory()
+    private val geometries = polygons.map { it.geometry }
+
+    override val geometry: MultiPolygon = factory.createMultiPolygon(geometries.toTypedArray())
+    override val type: GeometryType = GeometryType.MULTIPOLYGON
+    override val isClosed: Boolean = true
```
gino-m, line 29: What is this used for? Did it leak from the polygon drawing implementation? If so, whether the shape needs to be a LinearRing/Polygon or LineString is a detail that should be handle in the ViewModel, not in the actual model, IMO.

scolsen: Ok, stgm -- I was approaching it from the opposite angle (since closure is a property of the geometry itself not a property of the viewmodel, I felt it belonged in the geometry, but it works either way, esp. if we just use "ephemeral" incomplete objects in the VM)

gino-m: I see.. I think the key is that closed rings are modeled as different classes from open rings altogether, since they're defined as different geometry types by OGC. Contradicting the standard definition of a polygon by allowing the to be LineStrings as well may (will?) lead to confusion, so it's probably better to be consistent. Thanks for all the updates!

gino-m:
Iiuc you're breaking this PR into separate smaller steps? If so can we close this one?

gino-m:
y=

PR #1176, opened 2022-07-22
[Model] Implement Firestore map ->`Geometry` conversion

gino-m:
Towards #929. Cleans up existing tests in the process.

@scolsen PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
@@ -111,11 +111,7 @@ class GeometryConverter {
         return if (map.entries.all { it.key is Int && it.value != null }) {
             indexedMapToList(map as Map<Int, Any>).map(::fromFirestoreValue)
         } else {
-            map
-                .filter { it.value != null }
-                .mapValues {
-                    fromFirestoreValue(it.value as Any)
-                }
+            map.mapValues { it.value?.let(::fromFirestoreValue) }
         }
     }
 
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
@@ -49,7 +49,8 @@ class GeometryConverter {
         val writer = GeoJsonWriter()
         writer.setEncodeCRS(false)
         val jsonString = writer.write(geometry)
-        return toFirestoreValue(Gson().fromJson<MutableMap<String, Any>>(jsonString))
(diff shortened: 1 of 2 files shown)


ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
```diff
@@ -76,8 +78,52 @@ class GeometryConverter {
...
+            else -> {
+                value
+            }
+        }
+    }
+
+    private fun fromFirestoreValue(map: Map<*, *>): Any {
+        // If all keys are non-null Ints, assume it refers to an indexed map.
+        // If heuristic breaks, we may also want to check keys are in order starting at 0.
+        return if (map.entries.all { it.key is Int && it.value != null }) {
```
scolsen, line 111: NIT: this implementation will need to (presumably) iterate over the complete map multiple times (the call to `all` plus the call to `filter`, plus the call to `map`.

Instead, you could write this as:

```kotlin
return if (map.entries.any { it.key !is Int || it.value == null } ) {
  map.mapNotNull { it.value?.let { fromFirestoreValue(it as Any)} }
} else {
  indexedMapToList(...)...
}
```

Since `map g . filter f === map (g . f)` but it only traverses the structure once, instead of twice. `.any()` is also better than `.all()` since it can return immediately on the first instance of the predicate, rather than traversing the entire structure.

That said, I do think the original impl is more readable, and if the compiler is smart enough it can perform the `filter+map` optimization itself.

scolsen: Feel free to take just a piece of this suggestion too. And, actually the `value?.let` should allow you to drop the cast to any (though it may be dependent on compiler state) Perhaps:

```kotlin
private fun fromFirestoreValue(map: Map<*, *>): Any {
        // If all keys are non-null Ints, assume it refers to an indexed map.
        // If heuristic breaks, we may also want to check keys are in order starting at 0.
        return if (map.entries.all { it.key is Int && it.value != null }) {
          indexedMapToList(map as Map<Int, Any>).map(::fromFirestoreValue)
        } else {
            map.mapNotNull { it.value?.let(::fromFirestoreValue) }
        }
    }
```

would be best.

scolsen (approved):

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
gino-m: Good catch btw!

scolsen: oh, then this is great! I should've suspected as much.

PR #1183, opened 2022-07-30
[Map] Load geometries from db using new GeoJSON-like format

gino-m:
In this PR we "cheat" and map our quasi-GeoJSON Firestore format to either a `PointOfInterest` for points, or a `GeoJsonOfInterest` for Polygon and MultiPolygon.

Towards #929.

Note that this removes a test in `LoiConverterTest`, which will be rewritten in a follow-up PR after we migrate to data classes.

@scolsen PTAL?


ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
```diff
@@ -84,17 +86,29 @@ class GeometryConverter {
...
+        try {
+            val jsonString = Gson().toJson(jsonMap)
+            val reader = GeoJsonReader()
+            val geometry = reader.read(jsonString)
+            if (geometry.coordinates.isEmpty()) {
+                throw DataStoreException("Empty coordinates in $geometry")
+            }
+            return geometry
+        } catch (e: Throwable) {
+            throw DataStoreException("Invalid geometry")
```
scolsen, line 101: Why not log this error and just return `null` in this case?

scolsen: A better option: If we think it's important to distinguish `null` values from "something was malformed" values, we should use [`Result`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/). Result will also leave the decision about whether or not a failure in this code is a truly exceptional condition up to callers (see `exceptionOrNull`)

Exceptions are not properly in the class of things that are values and correspond to non-terminating conditions (even though they are really unexpected terminations--in kotlin they are typed as `Nothing`, the same type as an infinite loop, the type of all "unreachable" code: https://kotlinlang.org/docs/exceptions.html#the-nothing-type).

imo, in Kotlin code, we should be a bit more conservative about throwing and only do so when we really want to reflect that some program state effectively results in improper operation of the whole system. In this case I'd expect the app to continue to function just fine if one ore two geometries are malformed.

scolsen (approved):

scolsen:
Looks like the build failure was due to PMD violations.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/GeometryConverter.kt
gino-m: Applying suggestion to use `Result` instead

gino-m:
I applied your suggestion to use `Result` instead of thrown exceptions. A few things to note:

1. As discussed on chat, the `GeometryConverter` still throws exceptions internally to avoid needing to unpack and handle lists of `Result`.
2. I kicked the can down the road by throwing an exception on failure in `LoiConverter`, but added a TODO to return `Result` instead in the future.
3. Added a new file `ResultAssertions.kt` to sharedTest with custom asserts for `Result`, since I couldn't find any in the public docs.

scolsen (approved):

gino-m:
Looking into PMD failures now

PR #1186, opened 2022-07-31
Update compileSdkVersion and gradle plugin

shobhitagarwal1612:
This change was made using AS's tool. Also updates compile sdk version to 31

Updates gradle (and other dependencies) using Android Studio's AGP tool.

This also requires updating compile sdk version to 31.

Verified the changes by running the app locally.

@gino-m  PTAL?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,9 +18,9 @@
 
 buildscript {
     ext {
-        hiltVersion = '2.40.1'
+        hiltVersion = "2.40.1"
         kotlinVersion = "1.6.20"
-        navigationVersion = '2.4.1'
+        navigationVersion = "2.4.1"
     }
     repositories {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -18,9 +18,9 @@
 
 buildscript {
     ext {
-        hiltVersion = "2.38.1"
+        hiltVersion = '2.40.1'
```
gino-m, line 21: Please change to double quotes for consistency.

gino-m (changes requested):
Lgtm, just one nit.

build.gradle
shobhitagarwal1612: Done

gino-m (approved):
🆙

PR #1187, opened 2022-08-02
Move shared test classes to a new module

### LINKED ISSUE #1185: SharedTest reports warning "Duplicate content roots detected"
Instrumentation tests aren't able to find shared test classes due to the following error.

```
Duplicate content roots detected: Path [/Users/ashobhit/Desktop/Code/ground-android/ground/src/sharedTest/java] of module [ground-android.ground.unitTest] was removed from modules [ground-android.ground.androidTest]
```

Related article

shobhitagarwal1612:
Fixes #1185 

@... PTAL?


shobhitagarwal1612:
For some reason, the instrumentation tests are unable to use IDs. Looking into it.

shobhitagarwal1612:
Changing the namespace (`com.google.android.ground`) of the sharedTestCode to something else (`com.google.android.ground.sharedTestCode`) resolved the issue with resource ID. 

But now, another error is coming `CustomerTestRunner cannot be cast to org.robolectric.android.fakes.RoboMonitoringInstrumentation`

shobhitagarwal1612:
The solution is to remove robolectric dependency from the shared test module

shobhitagarwal1612:
Verified that both unit tests and instrumentation tests are working

shobhitagarwal1612:
@gino-m PTAL

ground/src/androidTest/java/com/google/android/ground/TestAuthenticationModule.java
```diff
@@ -18,7 +18,7 @@
 
 import com.google.android.ground.system.auth.AuthenticationManager;
 import com.google.android.ground.system.auth.AuthenticationModule;
-import com.google.android.ground.system.auth.FakeAuthenticationManager;
+import com.google.android.ground.test.system.auth.FakeAuthenticationManager;
```
gino-m, line 21: Do we need to use a different package name? One advantage to having test classes be in the same package as the unit under test is that they can see package-private variables and methods, which in some cases makes things easier to mock (e.g., overriding methods in test fakes).

sharedTest/build.gradle
```diff
@@ -0,0 +1,43 @@
+apply plugin: 'com.android.library'
```
gino-m, line 1: Please add copyright header.

gino-m (changes requested):
Some questions, overall LGTM!

ground/src/androidTest/java/com/google/android/ground/TestAuthenticationModule.java
shobhitagarwal1612: See https://github.com/google/ground-android/pull/1187#issuecomment-1207335080

With same package name, all instrumentation tests were failing due to unresolved resource ID.

sharedTest/build.gradle
shobhitagarwal1612: Done

gino-m (approved):
!!!

PR #1188, opened 2022-08-02
model: delete GeoJsonLocationOfInterest

scolsen:
Going forward, we will only model locations of interest with geometries,
and use geojson as a serialization format only. We'll no longer need the
GeoJsonLocationOfInterest class once this work is complete, so I've
removed it now to avoid requiring unnecessary changes to it as a
side-effect of base LOI changes.

@gino-m


ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
```diff
@@ -60,8 +60,10 @@ object LoiConverter {
                 return builder.build()
             }
             "Polygon", "MultiPolygon" -> {
-                val builder = GeoJsonLocationOfInterest.newBuilder()
-                builder.setGeoJsonString(GeoJsonWriter().write(geometry))
+                val builder = AreaOfInterest.newBuilder()
+                builder.setVertices(geometry.coordinates.map {
```
gino-m, line 64: I think this is incorrect for MultiPolygon; AreaOfInterest doesn't support multiple polygons. Instead, we'll need to set a `MultiPolygon` to `geometry`, creating it by passing Polygons to the MultiPolygon constructor.

gino-m (changes requested):
Lg, one comment

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
scolsen: I'm not fully grokking this yet—mind clarifying a bit here? What LOI type should we be returning or using for multipolygons in the interim before we move to the singular class representation?

gino-m: The only type that supports MultiPolygons at the moment (albeit indirectly) is GeoJsonLOI; AOI is only able to describe a single Polygon with no holes, since it only consists of a single set of vertices. For now I think we can just ignore the MultiPolygon case since it will be already covered by our effort to switch to  Geometry and subclasses, just wanted to raise the point. A TODO may be warranted here.

scolsen: Gotcha, will do!

scolsen: @gino-m, thanks! I added a call to `TODO()` for now (note that this throws).

gino-m (approved):
LGTM! Thanks for kicking this off!

PR #1196, opened 2022-08-08
Refactor: Single LocationOfInterestModel

scolsen:
This commit simplifies and updates our model for locations of interest:

- We retain only a single LocationOfInterest data class that accepts a
  geometry in its constructor.
- The Builder for this class has been removed.
- We now return the type of the LocationOfInterest based on its
  geometry.
- The AreaOfInterest and PointOfInterest classes have been removed.

In addition, I've added some extensions for working with the Point model
while it still exists, and I've converted FakeData to kotlin.


ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
```diff
@@ -22,37 +22,39 @@ import com.google.android.ground.model.mutation.LocationOfInterestMutation
...
-
+    val created: AuditInfo,
     /** User and time audit info pertaining to the last modification of this LOI. */
-    abstract val lastModified: AuditInfo
+    val lastModified: AuditInfo,
+    /** Geometry associated with this LOI. */
+    val geometry: Geometry,
+) {
+    /** Returns the type of this LOI based on its Geometry. */
+    val type: LocationOfInterestType =
```
gino-m, line 52: Do we need to have a separate LOI type? Can we not just differentiate using `geometry.geometryType`?

ground/src/main/java/com/google/android/ground/model/locationofinterest/Point.java
```diff
@@ -17,10 +17,14 @@
 package com.google.android.ground.model.locationofinterest;
 
 import com.google.auto.value.AutoValue;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.GeometryFactory;
 
 /** The location of a single point on the map. */
 @AutoValue
 public abstract class Point {
+  private final GeometryFactory geometryFactory = new GeometryFactory();
```
gino-m, line 26: Should we instead be using JTS `Point` instead? Or should we just create our own topology classes in the model as I'm doing in https://github.com/google/ground-platform/pull/928? Sorry to flip-flop, would like to think about it well now since we won't have time to revisit for awhile.

Note that going forward we may want to consider keeping conversion logic separate from data classes.

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
```diff
@@ -22,37 +22,39 @@ import com.google.android.ground.model.mutation.LocationOfInterestMutation
...
+    val caption: String? = null,
     /** User and time audit info pertaining to the creation of this LOI. */
-    abstract val created: AuditInfo
-
+    val created: AuditInfo,
     /** User and time audit info pertaining to the last modification of this LOI. */
-    abstract val lastModified: AuditInfo
+    val lastModified: AuditInfo,
+    /** Geometry associated with this LOI. */
+    val geometry: Geometry,
```
gino-m, line 49: Looking at this now, wondering if it would be more type-safe to keep the AOI and POI classes, having AOI contain `Polygon` and POI contain `Point`?

gino-m (changes requested):
A few questions/discussions.

ground/src/main/java/com/google/android/ground/model/locationofinterest/Point.java
scolsen: > Should we instead be using JTS Point instead? 

Yes. I decided to omit that from this PR, but the next one will delete this pointless `Point` class and use JTS directly.

But, if the broader plan is to use our own model, I think we should just do that. There's some benefit to having the representation on both sides be equivalent.

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
scolsen: It is more type safe, but we'd also be unable to construct LOIs, which is fine if we're ok with keeping that class abstract.

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
scolsen: Yes we could!

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
```diff
@@ -70,50 +72,22 @@ sealed class LocationOfInterest {
...
-        created = this@LocationOfInterest.created
-        lastModified = this@LocationOfInterest.lastModified
-    }
-
-    companion object {
-        @JvmStatic
-        fun newBuilder(): Builder = Builder()
-    }
+    // "Lens" style setter for geometries.
+    fun <T : Geometry> updateGeometry(geometry: T): LocationOfInterest = LocationOfInterest(
```
shobhitagarwal1612, line 83: Would this create a new instance of `LocationOfInterest` every time the `geometry` is updated? Is it not possible to update the property from the caller place?

ground/src/main/java/com/google/android/ground/persistence/remote/DataStoreException.kt
```diff
@@ -0,0 +1,48 @@
...
+        /**
+         * Checks if the provided object is of the same type as (or a subtype of) the specified type. If
+         * not, a `DataStoreException` is thrown with relevant details.
+         */
+        @JvmStatic
+        @Throws(DataStoreException::class)
+        fun <T : Any> checkType(expectedType: Class<*>, obj: T): T {
+            if (!expectedType.isAssignableFrom(obj.javaClass)) {
+                throw DataStoreException(
+                    "Expected " + expectedType.name + ", got " + obj.javaClass.name
```
shobhitagarwal1612, line 42: Can we use string template here as well?

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
```diff
@@ -37,52 +34,48 @@ object LoiConverter {
...
-                }.toImmutableList())
-                fillLocationOfInterest(builder, survey, loiId, loiDoc)
-                return builder.build()
-            }
+            "Polygon" -> return fillLocationOfInterest(
+                survey,
+                loiId,
+                loiDoc,
+                geometryFactory.createPolygon(geometry.coordinates)
+            )
```
shobhitagarwal1612, line 64: Instead of repeating the constructor code, can we use `.map()` on line 47 to convert to the appropriate geometry?

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/SubmissionCollectionReference.kt
```diff
@@ -36,7 +36,7 @@ class SubmissionCollectionReference internal constructor(ref: CollectionReferenc
     fun submissionsByLocationOfInterestId(
         locationOfInterest: LocationOfInterest
     ): @Cold Single<ImmutableList<ValueOrError<Submission?>>> {
-        return RxFirestore.getCollection(byLoiId(locationOfInterest.id))
+        return RxFirestore.getCollection(byLoiId(locationOfInterest.id ?: ""))
```
shobhitagarwal1612, line 39: Isn't this incorrect? Loi ID should never be empty.

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
```diff
@@ -143,37 +144,49 @@ class SubmissionRepository @Inject constructor(
...
+            applyAndEnqueue(
+                builder()
+                    .setJob(submission.job)
+                    .setSubmissionId(submission.id)
+                    .setResponseDeltas(ImmutableList.of())
+                    .setType(Mutation.Type.DELETE)
+                    .setSyncStatus(SyncStatus.PENDING)
+                    .setSurveyId(submission.survey.id)
+                    .setLocationOfInterestId(
+                        submission.locationOfInterest.id!!
```
shobhitagarwal1612, line 160: Should loi.id be non-nullable?

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
```diff
@@ -143,37 +144,49 @@ class SubmissionRepository @Inject constructor(
...
+        } else {
+            applyAndEnqueue(
+                builder()
+                    .setJob(submission.job)
+                    .setSubmissionId(submission.id)
+                    .setResponseDeltas(responseDeltas)
+                    .setType(if (isNew) Mutation.Type.CREATE else Mutation.Type.UPDATE)
+                    .setSyncStatus(SyncStatus.PENDING)
+                    .setSurveyId(submission.survey.id)
+                    .setLocationOfInterestId(submission.locationOfInterest.id!!)
```
shobhitagarwal1612, line 183: Same as above.

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
scolsen: Yes that's right, it always returns a new instance. This is generally how you update immutable structures in functional programming. The big difference here is that functional languages usually use sharing under the hood to mitigate cost (so even though it looks like an immutable copy, it's implemented as mutation under the hood).

I'm not sure whether or not Kotlin has similar optimizations, we'd have to look at the bytecode. Instead, I can change this to a `var` and make the setter private to give us immutability without the additional allocation costs.

gino-m (approved):
LGTM, please resolve Shobhit's comments where relevant before merging.

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
gino-m: We spoke. A single concrete LOI class with `Geometry` will work; we can specialize into subclasses iff we need to down the road.

ground/src/main/java/com/google/android/ground/model/locationofinterest/Point.java
gino-m: We spoke. Let's go with our own geometry impl, but please feel free to merge this PR first and retrofit.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
```diff
@@ -37,52 +34,48 @@ object LoiConverter {
...
 
         // As an interim solution, we map geometries to existing LOI types.
         // TODO: Get rid of LOI subclasses and just use Geometry on LOI class.
         when (geometry.geometryType) {
-            "Point" -> {
-                val builder = PointOfInterest.newBuilder()
-                builder.setPoint(
-                    Point.newBuilder().setLatitude(geometry.coordinate.x)
-                        .setLongitude(geometry.coordinate.y).build()
+            "Point" ->
```
gino-m, line 52: One more reason to create our own model classes (runtime String bindings are more error prone as opposed to static enums)

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
```diff
@@ -37,52 +34,48 @@ object LoiConverter {
...
-            }
-            "Polygon" -> {
-                val builder = AreaOfInterest.newBuilder()
-                builder.setVertices(geometry.coordinates.map {
-                    Point.newBuilder().setLatitude(it.x).setLongitude(it.y).build()
-                }.toImmutableList())
-                fillLocationOfInterest(builder, survey, loiId, loiDoc)
-                return builder.build()
-            }
+            "Polygon" -> return fillLocationOfInterest(
```
gino-m, line 59: Please rename `fillLocationOfInterest` to `createLocationOfInterest` or similar to make its new scope clear.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/SubmissionCollectionReference.kt
gino-m: +1. Though there are many cases where we could benefit from making model attributes non-null; we have a TODO in our planning doc to fix so feel free to address in a separate PR.

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
```diff
@@ -143,37 +144,49 @@ class SubmissionRepository @Inject constructor(
...
-                .setType(Mutation.Type.DELETE)
-                .setSyncStatus(SyncStatus.PENDING)
-                .setSurveyId(submission.survey.id)
-                .setLocationOfInterestId(submission.locationOfInterest.id)
-                .setClientTimestamp(Date())
-                .setUserId(authManager.currentUser.id)
-                .build()
-        )
+    fun deleteSubmission(submission: Submission): @Cold Completable {
+        return if (submission.locationOfInterest.id == null) {
```
gino-m, line 148: Nit: Imo short-circuiting is still more readable here (eg. `if (==null) return error; return applyAndEnqueue...`), mostly since the `applyAndEnqueue` expression is non-trivial. Is there any official style guidance on this?

gino-m (changes requested):
Undo, LGTM. Added a few more minor comments.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
scolsen: great idea!

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
scolsen: done!

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
scolsen: I'll hold off on this for now since we're going to roll our own Geometry model

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/SubmissionCollectionReference.kt
scolsen: sg, let's tackle this separately.

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
scolsen: Thanks! Will address in a follow up.

scolsen: oh- scratch that, this field is already non-null in this PR, these checks are redundant. Removed.

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
scolsen: the check was actually just redundant, so it's gone.

scolsen:

gino-m (approved):
LGTM, great work! Looking forward to our own model reducing some of the complexity further!

shobhitagarwal1612 (approved):

PR #1201, opened 2022-08-11
Convert Option to Kotlin

JSunde:


JSunde:
@gino-m Any ideas why the @Serializable annotation doesn't seem to be working properly?

JSunde:
> @gino-m Any ideas why the `@Serializable` annotation doesn't seem to be working properly?

Actually I think I was able to resolve this after following the setup instructions here: https://github.com/Kotlin/kotlinx.serialization#setup

gino-m:
> @gino-m Any ideas why the `@Serializable` annotation doesn't seem to be working properly?

I was just looking into this, I see you found a fix!

gino-m (approved):
LGTM!

PR #1202, opened 2022-08-12
Convert MultipleChoice to Kotlin

JSunde:


ground/src/main/java/com/google/android/ground/persistence/local/room/RoomLocalDataStore.java
```diff
@@ -128,7 +128,7 @@ private Completable insertOrUpdateOption(String taskId, Option option) {
         .subscribeOn(schedulers.io());
   }
 
-  private Completable insertOrUpdateOptions(String taskId, ImmutableList<Option> options) {
+  private Completable insertOrUpdateOptions(String taskId, List<Option> options) {
```
gino-m, line 131: This and other collections should be immutable.

Should we start using kotlinx immutable collections instead of Guava.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/local/room/RoomLocalDataStore.java
JSunde: Hmm, I tried switching to the kotlinx.collections.immutable.ImmutableList, but I'm having a hard time figuring out how to create a Kotlin ImmutableList from a java class, and right now we're constructing a MultipleChoice from the MultipleChoiceEntity.java.

The Kotlin ImmutableList doesn't seem to have a builder and I don't think we can access the toImmutableList extension function from a Java class.

Let me know if you have any ideas to workaround this, or if we should stick to the Guava Immutable Collections for now.

JSunde: Synced offline and decided to stick with Guava Immutable Collections for now.

ground/src/main/java/com/google/android/ground/model/submission/MultipleChoiceResponse.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.google.android.ground.model.task.Option
+import java8.util.Optional
+import kotlinx.serialization.Serializable
+
+/**
+ * User responses to a select-one (radio) or select-multiple (checkbox) field.
+ */
+@Serializable
+class MultipleChoiceResponse(
+    private val multipleChoice: MultipleChoice?, val selectedOptionIds: List<String>
```
shobhitagarwal1612, line 29: Why is `multipleChoice` nullable now? It wasn't in the original java class.

ground/src/main/java/com/google/android/ground/model/submission/MultipleChoiceResponse.kt
```diff
@@ -0,0 +1,68 @@
...
+        return false
+    }
+
+    override fun hashCode(): Int = selectedOptionIds.hashCode()
+
+    override fun toString(): String = selectedOptionIds.sorted().joinToString()
+
+    companion object {
+        @JvmStatic
+        fun fromList(multipleChoice: MultipleChoice?, codes: List<String>): Optional<Response> {
```
shobhitagarwal1612, line 60: Same as above

ground/src/main/java/com/google/android/ground/model/task/MultipleChoice.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    fun getOptionById(id: String): Option? = options.firstOrNull { it.id == id }
+
+    fun getIndex(id: String): Optional<Int> {
+        for (i in options.indices) {
+            if (options[i].id == id) {
+                return Optional.of(i)
+            }
+        }
+        return Optional.empty()
```
shobhitagarwal1612, line 41: nit: We generally use -1 as default value when index is not found. Any specific reason for using optional here?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/model/submission/MultipleChoiceResponse.kt
JSunde: I think I saw that there were callers that were passing null so I updated the field to be null to be more correct, but after looking again I realized that it was just in tests that null was passed. I've now updated the callers and made the field non nullable again.

ground/src/main/java/com/google/android/ground/model/submission/MultipleChoiceResponse.kt
JSunde: Resolved

ground/src/main/java/com/google/android/ground/model/task/MultipleChoice.kt
JSunde: I'm not sure of the original reasoning, but I was just converting the class as is so this was the behavior of the java class.

gino-m (dismissed):
LGTM! Please just fix build errors.

```
e: /workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ResponseJsonConverter.kt: (82, 53): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
e: /workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ResponseJsonConverter.kt: (85, 53): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
e: /workspace/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/SubmissionConverter.kt: (125, 41): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
```

JSunde:
> LGTM! Please just fix build errors.
> 
> ```
> e: /workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ResponseJsonConverter.kt: (82, 53): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
> e: /workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ResponseJsonConverter.kt: (85, 53): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
> e: /workspace/ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/SubmissionConverter.kt: (125, 41): Type mismatch: inferred type is MultipleChoice? but MultipleChoice was expected
> ```

Oh actually I'm realizing there were legitimate non-test use cases where the MultipleChoice field of MultipleChoiceResponse was null. I'm guessing we want to mark it as nullable then?

WDYT?

gino-m:
> 

It probably shouldn't be the case, but if that's the current state let's do that and fix the root cause in a follow-up.

shobhitagarwal1612 (approved):

PR #1206, opened 2022-08-16
Convert User to Kotlin

JSunde:


ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
```diff
@@ -126,11 +120,5 @@ class GoogleAuthenticationManager @Inject constructor(
...
         GoogleAuthProvider.getCredential(googleAccount.idToken, null)
 
-    private fun FirebaseUser.toUser(): User =
-        User.builder()
-            .setId(uid)
-            .setEmail(email)
-            .setDisplayName(displayName)
-            .setPhotoUrl(photoUrl.toString())
-            .build()
+    private fun FirebaseUser.toUser(): User = User(uid, email!!, displayName!!, photoUrl.toString())
```
gino-m, line 123: Are `email` and `displayName` guaranteed to be set? If so please add a comment, if not we should handle that case.

gino-m (changes requested):
LGTM! One question/nit.

ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
JSunde: I believe they are both nullable, previously they were passed to the User AutoValue.Builder even though those fields weren't nullable.

What do you think we should do to handle the case where they're null? Should we just replace null with an empty string for those params?

gino-m: I assume they're non-nullable in the builder because they need to be atm. If that's the case for now default to `""` is ok.

JSunde: SG, passing `""` in the case that `email` or `displayName` are null

gino-m (approved):
LGTM!

PR #1216, opened 2022-08-18
Convert DateResponse to Kotlin

JSunde:


JSunde:
PTAL @gino-m @scolsen?

gino-m (approved):

ground/src/main/java/com/google/android/ground/model/submission/DateResponse.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.serialization.Serializable
+import java.text.DateFormat
+import java.text.SimpleDateFormat
+import java.util.*
+
+/**
+ * A user-provided response to a date question task.
+ */
+@Serializable
+class DateResponse(val date: @Contextual Date) : Response {
```
scolsen, line 29: Should we make this a data class?

scolsen (approved):

ground/src/main/java/com/google/android/ground/model/submission/DateResponse.kt
JSunde: Good idea, I'm not sure I realized that data classes handled the equals() toString() and hashCode() methods so this ended up much cleaner after that!

scolsen: awesome!

scolsen (approved):

PR #1217, opened 2022-08-18
Implement our own geometry model

scolsen:
Defines new interfaces and data classes for representing geometrical
forms in ground. These classes should allow us to drop dependencies on
other geometry libraries and implement the minimum functionality we
need.

Note that this commit only adds the new model classes, they are not yet
used in other parts of the codebase. Adoption of these classes in favor
of the external ones will be handled in a separate commit.

@gino-m


ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+class PolygonExteriorHoleException(val shellMaximum: Coordinate, val holeMaximum: Coordinate) :
+    GeometryException("Invalid polygon. Holes must be contained within a polygon's shell, but the maximum coordinates of the shell $shellMaximum are lesser than the maximum coordinates of a hole $holeMaximum")
```
JSunde, line 21: Can this be split to multiple lines and then use trimIndent() to remove the extra whitespace from the final string. It ends up being much more readable.

```
"""
    Example of a long string
    that is written over multiple lines
""".trimIndent()

**Review by JSunde — APPROVED** [review]

File: ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt (hunk 1 above)
Comment (scolsen) [reply ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt:21]: sgtm, thanks for the suggestion!

File: ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+/** The class of geometry errors.
```
gino-m, line 20: Please run `google-java-format` on all files, I believe it will place the first phrase in multiline Javadoc on the following line.

Also, if you mean 'class' as in 'a group, set, or kind sharing common attributes' it may be better to rephrase to avoid collision with the OOP term of the same name

ground/src/main/java/com/google/android/ground/model/geometry/LinearRing.kt
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+class OpenLinearRingException(first: Coordinate, last: Coordinate) :
+    GeometryException(
+        """Invalid linear ring. Linear rings must be closed, but the first coordinate 
+        |$first and last coordinate $last of the ring are no equal""".trimMargin()
```
gino-m, line 23: Typo: no equal -> not equal

ground/src/main/java/com/google/android/ground/model/geometry/LinearRing.kt
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+class LinearRingLengthException(numVertices: Int) :
+    GeometryException("Invalid linear ring. Expected 3 or more vertices but got: $numVertices")
```
gino-m, line 27: Nit: Remove ":"

ground/src/main/java/com/google/android/ground/model/geometry/MultiPolygon.kt
```diff
@@ -0,0 +1,29 @@
...
+
+/** A collection of [Polygon]s.*/
+data class MultiPolygon(val polygons: List<Polygon>) : Geometry {
+    // TODO: Enforce definitional constraints on initialization: polygons may not overlap and can only touch at single points.
+
+    /** Returns the first coordinate of the first polygon in this multipolygon. */
+    override val coordinate: Coordinate = polygons.first().coordinate
+
+    /** Returns the sequence of coordinates associaited with the first polygon in this multipolygon. */
+    override val coordinates: ImmutableList<Coordinate> = polygons.first().coordinates
```
gino-m, line 28: Same here.

ground/src/main/java/com/google/android/ground/model/geometry/MultiPolygon.kt
```diff
@@ -0,0 +1,29 @@
...
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+/** A collection of [Polygon]s.*/
+data class MultiPolygon(val polygons: List<Polygon>) : Geometry {
+    // TODO: Enforce definitional constraints on initialization: polygons may not overlap and can only touch at single points.
+
+    /** Returns the first coordinate of the first polygon in this multipolygon. */
+    override val coordinate: Coordinate = polygons.first().coordinate
```
gino-m, line 25: What is this used for? Is it needed?

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+/** An individual point. */
+data class Point(override val coordinate: Coordinate) : Geometry {
+    override val coordinates: ImmutableList<Coordinate> = ImmutableList.of(coordinate)
```
gino-m, line 22: Consider not using `coordinates` in all Geometry implementations, instead modeling the semantics of geometry try, as we do in https://github.com/google/ground-platform/tree/master/web/src/app/shared/models/geometry

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+import com.google.common.collect.ImmutableList
+
+class PolygonExteriorHoleException(val shellMaximum: Coordinate, val holeMaximum: Coordinate) :
```
gino-m, line 20: Will callers ever need to access the shell min and hole max programmatically? If not, can instead maintain a single GeometryException whose message provides the necessary details for a human to debug the source of the problem?

gino-m (changes requested):
A few initial thoughts. Excited to see our model coming together!

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
scolsen: Hm, when I run the formatter on kotlin files it doesn't seem to automatically handle doc comments for some reason. Good suggestion, changed the terminology.

ground/src/main/java/com/google/android/ground/model/geometry/LinearRing.kt
scolsen: good catch, thanks!

ground/src/main/java/com/google/android/ground/model/geometry/LinearRing.kt
scolsen: done.

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
scolsen: I was trying to model the different particular errors, per Geometry type, as separate exception types all subclassing the same parent--do you think we don't need that? It seems like it'd be helpful to encode this when dealing with different types of Geometries at the same time.

ground/src/main/java/com/google/android/ground/model/geometry/MultiPolygon.kt
scolsen: Yes, we'll need coordinates implementations for our Geometry types since our Firestore conversion routines and other parts of the codebase rely on them. This is modeled directly after JTS. e.g. see `GeometryConverter.kt`

scolsen:
@gino-m, could use another review -- I'll move validation out of the data class constructors in a separate PR.

ground/src/main/java/com/google/android/ground/model/geometry/MultiPolygon.kt
gino-m: What would `coordinate` `MultiPolygon` and `Polygon` be used for? If we don't need it we should leave it out. Afaik the conversion routines work differently for different geometry types (e.g., for `Point` we just look for a single set of coordinates, for `Polygon` we look at the outer shell and inner shells, etc.)

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
gino-m: I think we might be trying to force polymorphism by adding a common attribute to all geometries. I know the JTS libraries do this, but a) I haven't seen a need for these in our code yet, and b) it's unclear how those would be useful at all.

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
gino-m: If there's some difference in error state that we need/want compile-time checks for then yes, but if it's only to pass additional information about the type of error to the logs, then the additional types are redundant wrt the exception messages.

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
scolsen: Ah yeah, you're right, I was looking at the converter, which I believe we're keeping as JTS for JSON reads, right? ok sg, I can drop these.

gino-m (changes requested):
Replies incoming!

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
scolsen: Gotcha, I'll remove these for now then, maybe we're fine with a single exception check.

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
gino-m: The converters read/write to GeoJSON, which is what JTS does as well. In that case, `coordinates` has a different meaning and structure for each geometry type. Ideally, we wouldn't leak the serialized representation into our domain model; those details are currently encapsulated in `GeometryConverter`, so no need to add them here as well.

scolsen:
@gino-m ok! let's see if this looks any better

ground/src/main/java/com/google/android/ground/model/geometry/Coordinate.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+/** A coordinate on a two-dimensional Cartesian plane. */
+data class Coordinate(val x: Double, val y: Double) : Comparable<Coordinate> {
+    companion object {
+        /** Returns the zero coordinate: (0,0) */
+        fun zero() = Coordinate(0.0, 0.0)
```
gino-m, line 22: What is this used for?

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+/**
+ * Represents types of geometry errors.
+ *
+ * Typically thrown when a construction does not satisfy definitional constraints for a given geometry.
+ */
+sealed class GeometryException(override val message: String) : Throwable(message)
```
gino-m, line 23: Nit: Maybe "InvalidGeometryException" might be more accurate here?

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+/** An individual point. */
```
gino-m, line 18: It may be helpful to clarify how this is different from `Coordinate`. Feel free to copy my wording from our TS impl if it helps.

gino-m (changes requested):
One more round of comments, we're close!

ground/src/main/java/com/google/android/ground/model/geometry/Coordinate.kt
scolsen: it's not needed anymore, I was going to use it to return default values in some cases instead of throwing exceptions. removed it.

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
scolsen: sgtm, done

ground/src/main/java/com/google/android/ground/model/geometry/Point.kt
scolsen: Copied the TS doc

ground/src/main/java/com/google/android/ground/model/geometry/Polygon.kt
gino-m: Thanks!

gino-m (approved):
🚀 🥇!

PR #1218, opened 2022-08-18
Convert Submission classes to Kotlin

JSunde:


JSunde:
PTAL @scolsen @gino-m

JSunde:
It looks like there's a test that's failing because the toString of the TextResponse changed from "updated response" to "TextResponse(text=updated response)"

Is it important that we preserve the Response toString methods? Or is the new Data Class default implementation sufficient? (In which case I would just update the test)

gino-m (approved):
LGTM! 🐰

gino-m:
> It looks like there's a test that's failing because the toString of the TextResponse changed from "updated response" to "TextResponse(text=updated response)"
> 
> Is it important that we preserve the Response toString methods? Or is the new Data Class default implementation sufficient? (In which case I would just update the test)

We chatted, ok to get rid of the custom `toString`!

PR #1229, opened 2022-08-23
Remove instance of Survey from Submission model

shobhitagarwal1612:
Towards #1228

Replace with "id" of the Survey

@gino-m PTAL?


gino-m (approved):
That was fast..thanks!

PR #1230, opened 2022-08-23
Replace Survey with surveyId in Loi model

### LINKED ISSUE #1228: [Code health] Stop referencing Survey and contained classes from model objects to avoid stale references; instead get from repo as needed

shobhitagarwal1612:
Fixes #1228

Things done:
1. Add a new field "surveyId" to Loi model
2. Directly replace "getSurvey().getId()" with "surveyId"
3. Remove remaining dependencies of by directly fetching the instance using survey id

@gino-m  PTAL?


gino-m:
@shobhitagarwal1612 Does nothing actually use the `Survey` reference, ie `id` is sufficient? I remember we needed them in several places, but perhaps this has changed?

shobhitagarwal1612:
It was only being used in SubmissionRepository and UserRepository. Rest were just using it for the ID

gino-m (approved):
Great stuff, LGTM!

PR #1237, opened 2022-08-24
Convert Worker classes to Kotlin

JSunde:


ground/src/main/java/com/google/android/ground/persistence/sync/BaseWorkManager.kt
```diff
@@ -0,0 +1,76 @@
...
+ *
+ *
+ * By default, the only constraint is availability of any type of internet connection, as it is
+ * assumed that all background tasks need at least some sort of connectivity.
+ *
+ *
+ * In case the required criteria are not met, the next attempt uses LINEAR backoff policy with a
+ * backoff delay of 10 seconds.
+ */
+abstract class BaseWorkManager {
```
shobhitagarwal1612, line 34: nit: Should this be marked as internal?

shobhitagarwal1612 (approved):
One small nit, rest looks good!

ground/src/main/java/com/google/android/ground/persistence/sync/BaseWorkManager.kt
JSunde: Good idea, done!

PR #1238, opened 2022-08-25
Convert Response interface to Kotlin

JSunde:


ground/src/main/java/com/google/android/ground/model/submission/DateResponse.kt
```diff
@@ -31,8 +31,6 @@ data class DateResponse(val date: @Contextual Date) : Response {
     private val dateFormat: @Contextual DateFormat =
         SimpleDateFormat("yyyy-MM-dd", Locale.getDefault())
 
-    override fun getSummaryText(): String = detailsText
```
gino-m, line 34: Could you kindly file a cleanup bug to get rid of `getDetailsText()` as well? That should be handle in the view layer, where we're able to correctly localize the response depending on its type and the user's locale.

gino-m (approved):

ground/src/main/java/com/google/android/ground/model/submission/DateResponse.kt
JSunde: Done, filed #1240

gino-m (approved):

PR #1239, opened 2022-08-25
No More JTS

scolsen:
Updates the codebase to use our custom geometry model throughout. 

In addition, I converted a few files to kotlin and performed minor refactors where necessary.

This PR also removes our JTS dependencies.


JSunde (approved):

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/LocationOfInterestEntity.java
```diff
@@ -130,9 +132,10 @@ public static LocationOfInterestEntity fromLocationOfInterest(
             .setCreated(AuditInfoEntity.fromObject(locationOfInterest.getCreated()))
             .setLastModified(AuditInfoEntity.fromObject(locationOfInterest.getLastModified()));
     if (locationOfInterest.getType() == LocationOfInterestType.POINT) {
-      entity.setLocation(Coordinates.fromPoint(locationOfInterest.getCoordinatesAsPoint()));
+      entity.setLocation(
```
gino-m, line 135: Do we still need `LocationOfInterestType`? Perhaps we can remove it in favor of `getGeometry()` and instanceof in the future?

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/LocationOfInterestEntity.java
```diff
@@ -130,9 +132,10 @@ public static LocationOfInterestEntity fromLocationOfInterest(
...
+          Coordinates.fromPoint(locationOfInterest.getGeometry().getVertices().get(0)));
     } else {
-      entity.setPolygonVertices(formatVertices(locationOfInterest.getCoordinatesAsPoints()));
+      entity.setPolygonVertices(formatVertices(locationOfInterest.getGeometry().getVertices()));
```
gino-m, line 138: This will only work for polygons with no holes. Could you find a TODO + bug to load/save the new model in the local db?

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
```diff
@@ -15,12 +15,9 @@
...
- * Typically thrown when a construction does not satisfy definitional constraints for a given geometry.
- */
-sealed class InvalidGeometryException(override val message: String) : Throwable(message)
+import com.google.common.collect.ImmutableList
 
 /** A common ancestor for all geometry types. */
-sealed interface Geometry
\ No newline at end of file
+sealed interface Geometry {
+    abstract val vertices: ImmutableList<Point>
```
gino-m, line 22: Do we ever need vertices independent of what type of geometry and structure we're dealing with? I don't think the list of vertices is sufficient for our purposes, except to act as a workaround for legacy code that supported polygon.

Instead of breaking the semantic in callers, I think it would be preferable to have clients use the specific polygon and point coordinates where needed, and in cases where the geometry is a multi-polygon or a polygon with hole, throw a `UnsupportedOperationException` with a TODO(#Issue) to add support. Wdyt?

gino-m (changes requested):
This looks like a big improvement over using JTS directly in our model. A few comments/suggestion.

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/LocationOfInterestEntity.java
scolsen: Once we have only kotlin callers, we can remove it and exclusively use `when` statements. I kept it around for the moment for java callers since it's a bit shorter than the long if/else instanceof statements.

gino-m: That makes sense! Would you mind filing a cleanup bug for that? 🙏

scolsen: will do!

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
scolsen: Once we can match on the `geometry` type statically (once all callers are kotlin) we don't, but at the moment, we need to rely on either instanceof checks and casts or go through a unified interface in Geometry. The latter seemed cleaner to me as a temporary state. I'll add a TODO to move toward using specific classes only.

scolsen: Another option would be to move back to subclassing LocationOfInterest, where subclasses have specific geometries.

gino-m: Sgtm, thanks!

gino-m: > Another option would be to move back to subclassing LocationOfInterest, where subclasses have specific geometries.

Unless LOIs need to have different structure depending on the type of geometry they can't contain I can't think of a reason to have the extra complexity. Also, if there are places where views can't use `instanceof` or need a value object (persistence?) we can always add an enum to Geometry to represent the type.

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/LocationOfInterestEntity.java
scolsen: #1247

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/LocationOfInterestEntity.java
scolsen: #1246

scolsen:
@gino-m added some TODOs and issues for further improvements!

PR #1245, opened 2022-08-26
Convert TransferProgress to data class

shobhitagarwal1612:
@JSunde  PTAL?


ground/src/main/java/com/google/android/ground/persistence/remote/TransferProgress.kt
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.persistence.remote
+
+data class TransferProgress constructor(
+    val state: UploadState,
+    val byteCount: Int = 0,
+    val bytesTransferred: Int = 0
+) {
```
JSunde, line 22: I think you can omit "constructor"

So this would just be
```
data class TransferProgress(
    val state: UploadState,
    val byteCount: Int = 0,
    val bytesTransferred: Int = 0
) {
```

I've only needed to include the constructor keyword for data classes when I needed to provide overloads for java callers

JSunde (approved):

ground/src/main/java/com/google/android/ground/persistence/remote/TransferProgress.kt
shobhitagarwal1612: Thanks! Done

Issue #1224, opened 2022-08-20
[Code health] Set up Kotlin linter 


shobhitagarwal1612:
https://github.com/pinterest/ktlint
https://github.com/jlleitschuh/ktlint-gradle

gino-m:
Let's also be sure to make CheckReturnValue the default, it could save us many headaches.

PR #1253, opened 2022-08-30
Add ktfmt plugin and format all classes

shobhitagarwal1612:
Towards #1224

One of the key differences between ktfmt and intellij formatter is that the line indentation is 2

Things done:
1. Make the plugin required and also enforce the plugin configuration (requires a restart for changes to take effect)
2. Run the code formatter on all java and kotlin files. 

@gino-m  PTAL?


gino-m:
I see some .java files were reformatted as well. Should we not be using `google-java-format` anymore?

shobhitagarwal1612:
This is causing failure with integration tests. Investigating into it.

shobhitagarwal1612:
Unable to repro the failures after invalidating caches

PR #1254, opened 2022-08-31
Fix IllegalStateException thrown at app startup

JSunde:
On app startup, Navigator.navigate() was causing an IllegalStateException because it was not run on the main thread.


gino-m (approved):
Thanks for the fix and the migration!

PR #1258, opened 2022-08-31
Fix conversion of geometry db map keys

### LINKED ISSUE #929: [Code health] Harmonize geometry representation

JSunde:
Fixes #929 as well as fixing a ClassCastException in GeometryConverter and returns Result from LoiConverter.


ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiConverter.kt
```diff
@@ -34,18 +35,15 @@ object LoiConverter {
   const val GEOMETRY_COORDINATES = "coordinates"
   const val GEOMETRY = "geometry"
 
-  fun toLoi(survey: Survey, doc: DocumentSnapshot): LocationOfInterest {
+  fun toLoi(survey: Survey, doc: DocumentSnapshot): Result<LocationOfInterest> {
```
gino-m, line 38: You'll need to update the two callers to check the result and on failure log a debug warning that the LOI was skipped (id + message).

JSunde: I think that actually already happens. In the QuerySnapshotConverter I'm returning a Result.error() when the Result<LocationOfInterest> is a failure, and then in LocationOfInterestRepository#updateLocalLocationOfInterest() it skips and logs those invalid LOIs

JSunde: The Result<LocationOfInterest> ends up getting wrapped in a RemoteDataEvent<LocationOfInterest> in the QuerySnapshotConverter.

gino-m (approved):
:1st_place_medal: LGTM!

PR #1261, opened 2022-09-03
Add ktfmt gradle plugin

shobhitagarwal1612:
Towards #1224

Adds ktlint gradle plugin for code analysis.
https://github.com/cortinico/ktfmt-gradle

This plugin adds 2 new tasks: `ktfmtCheck` and `ktfmtFormat`

@gino-m @JSunde  PTAL?


shobhitagarwal1612:
It seems that ktlint doesn't agree with "Google Internal" formatting. Making it work with the current configuration is turning out to be lot harder than expected.

shobhitagarwal1612:
Ktlint isn't as configurable as I imagined it to be. Either we need to switch to some other configuration that ktlint also understands or find another linter which is more configurable.

shobhitagarwal1612:
Found a new plugin (https://github.com/cortinico/ktfmt-gradle) which is based on ktfmt. Force pushing to remove old commits.

PR #1263, opened 2022-09-06
Convert DAO interfaces to Kotlin

shobhitagarwal1612:
Towards #532 

Converted `insertOrUpdate()` method to an extension function as all inner methods of DAO must be annotated with `Insert`, `Update`, `Delete` or `Query`

@JSunde  PTAL?


shobhitagarwal1612:
@JSunde Gentle ping!

PR #1268, opened 2022-09-12
Remove MapPin and replace with LocationOfInterest containing a single point

JSunde:
Added a SimpleMapLocationOfInterest to allow the wrapping of a LocationOfInterest and implement MapLocationOfInterest. Once we finish removing the MapPin, MapPolygon, and MapGeoJson it should be easier to remove the MapLocationOfInterest wrapper altogether and then we can remove the SimpleMapLocationOfInterest as well.


JSunde:
Towards #929

shobhitagarwal1612 (approved):

PR #1273, opened 2022-09-14
Converts instrumentation tests to kotlin

shobhitagarwal1612:
Towards #532 

Verified by running the instrumentation tests locally.

@JSunde  PTAL?


ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
```diff
@@ -0,0 +1,117 @@
...
+ * Used to make Espresso work with DataBinding. Without it the tests will be flaky because
+ * DataBinding uses Choreographer class to synchronize its view updates hence using this to monitor
+ * a launched fragment in fragment scenario will make Espresso wait before doing additional checks
+ */
+class DataBindingIdlingResource : IdlingResource {
+  // Holds whether isIdle was called and the result was false. We track this to avoid calling
+  // onTransitionToIdle callbacks if Espresso never thought we were idle in the first place.
+  private var wasNotIdle = false
+  private var activity: FragmentActivity? = null
+  override fun getName(): String {
```
JSunde, line 37: Could be simplified as:
```
override fun getName(): String = "DataBinding $ID"
```

ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  private fun getBinding(view: View?): ViewDataBinding? {
+    return DataBindingUtil.getBinding(view!!)
+  }
+
+  private val bindings: List<ViewDataBinding?>
+    get() {
+      val fragments =
+        if (activity == null) emptyList() else activity!!.supportFragmentManager.fragments
```
JSunde, line 86: I think this could be simplified to:
```
activity?.supportFragmentManager.fragments ?: emptyList()
```

ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
```diff
@@ -0,0 +1,117 @@
...
+      val bindings: MutableList<ViewDataBinding?> = ArrayList()
+      for (f in fragments) {
+        if (f.view == null) {
+          continue
+        }
```
JSunde, line 91: I think this might be easier to follow if we stored these views in local variables
```
val view = f.view ?: continue
```

Then the whole if statement can be collapsed and we can reference the new non null view variable below

JSunde (approved):

ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
shobhitagarwal1612: Done

ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
shobhitagarwal1612: Done

ground/src/androidTest/java/com/google/android/ground/DataBindingIdlingResource.kt
shobhitagarwal1612: Done

PR #1277, opened 2022-09-15
Refactor: store polygon holes in db; add converters

scolsen:
We now decouple model->DB entity conversions from the models and entities themselves and handle them in dedicated Converter classes.

Additionally, we now write polygon holes to the database as a list of lists of coordinates. This also required updating the handling of geometries in mutations.

### DIFF
--- a/ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
+++ b/ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.converter
+
+import com.google.android.ground.model.geometry.*
+import com.google.android.ground.persistence.local.room.entity.GeometryEntity
+import com.google.android.ground.persistence.local.room.models.Coordinates
+import com.google.android.ground.persistence.local.room.models.GeometryType
+import com.google.android.ground.util.toImmutableList
+import com.google.common.collect.ImmutableList
+import com.google.common.reflect.TypeToken
+import com.google.gson.Gson
+
+/** Converts [Geometry] model objects to/from a persisted database representation. */
+object GeometryModelToLocalDbConverter : Converter<Geometry, GeometryEntity> {
+  override fun convertTo(model: Geometry): GeometryEntity? =
+    when (model) {
+      is Point -> fromPointModel(model)
+      is Polygon -> fromPolygonModel(model)
+      else -> null
+    }
+
+  override fun convertFrom(entity: GeometryEntity): Geometry? =
+    when (entity.geometryType) {
+      GeometryType.POINT.name -> toPointModel(entity)
+      GeometryType.POLYGON.name -> toPolygonModel(entity)
+      else -> null
+    }
+
+  private fun toPointModel(record: GeometryEntity): Geometry? = record.location?.toPoint()
+
+  private fun toPolygonModel(record: GeometryEntity): Geometry {
+    val shell = LinearRing(parseVertices(record.vertices).map { it.coordinate })
+    val holes = parseHoles(record.holes).map { LinearRing(it.map(Point::coordinate)) }
+
+    return Polygon(shell, holes)
+  }
+
+  private fun fromPointModel(point: Point): GeometryEntity =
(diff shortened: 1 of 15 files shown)


ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
```diff
@@ -0,0 +1,115 @@
...
+      return null
+    }
+    val gson = Gson()
+    val verticesArray =
+      vertices.map { (coordinate): Point -> ImmutableList.of(coordinate.x, coordinate.y) }.toList()
+    return gson.toJson(verticesArray)
+  }
+
+  private fun parseHoles(holes: String?): List<ImmutableList<Point>> {
+    if (holes == null || holes.isEmpty()) {
```
shobhitagarwal1612, line 85: nit: Can be replaced with `.isNullOrEmpty()`

ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
```diff
@@ -0,0 +1,115 @@
...
+        object : TypeToken<List<List<List<Double?>?>?>?>() {}.type
+      )
+
+    return holesArray.map {
+      it.map { vertex -> Point(Coordinate(vertex[0], vertex[1])) }.toImmutableList()
+    }
+  }
+
+  fun parseVertices(vertices: String?): ImmutableList<Point> {
+    if (vertices == null || vertices.isEmpty()) {
```
shobhitagarwal1612, line 102: nit: Can be replaced with `.isNullOrEmpty()`

ground/src/main/java/com/google/android/ground/converter/LocationOfInterestModelToLocalDbConverter.kt
```diff
@@ -0,0 +1,90 @@
...
+        lastModified = AuditInfoEntity.toObject(entity.lastModified),
+        geometry = geometry,
+        job =
+          this.survey.getJob(jobId = entity.jobId).orElseThrow {
+            LocalDataConsistencyException(
+              "Unknown jobId $entity.jobId in location of interest $entity.id"
+            )
+          }
+      )
+    }
```
shobhitagarwal1612, line 88: Is there any specific reason for not moving this to companion object?

ground/src/main/java/com/google/android/ground/converter/LocationOfInterestModelToLocalDbConverter.kt
```diff
@@ -0,0 +1,90 @@
...
+        id = model.id,
+        surveyId = model.surveyId,
+        jobId = model.job.id,
+        state = EntityState.DEFAULT,
+        created = AuditInfoEntity.fromObject(model.created),
+        lastModified = AuditInfoEntity.fromObject(model.lastModified),
+        geometry = GeometryModelToLocalDbConverter.convertTo(model.geometry)
+      )
+
+    override fun convertFrom(entity: LocationOfInterestEntity): LocationOfInterest? = null
```
shobhitagarwal1612, line 63: Why is there an inconsistency b/w convertTo and convertFrom? Should this be no-op?

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiMutationConverter.kt
```diff
@@ -40,14 +41,17 @@ internal object LoiMutationConverter {
...
+    when (mutation.geometry) {
+      is Point -> map.put(LoiConverter.LOCATION, toGeoPoint(mutation.geometry))
+      is Polygon -> {
+        val geometryMap: MutableMap<String, Any> = HashMap()
+        geometryMap[LoiConverter.GEOMETRY_COORDINATES] =
+          toGeoPointList(mutation.geometry.shell.vertices)
+        geometryMap[LoiConverter.GEOMETRY_TYPE] = LoiConverter.POLYGON_TYPE
+        map.put(LoiConverter.GEOMETRY, geometryMap)
+      }
+      else -> {}
```
shobhitagarwal1612, line 53: Can we add a TODO here?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
scolsen: done, thanks!

ground/src/main/java/com/google/android/ground/converter/GeometryModelToLocalDbConverter.kt
scolsen: done, thanks!

ground/src/main/java/com/google/android/ground/converter/LocationOfInterestModelToLocalDbConverter.kt
scolsen: This one only exists to satisfy the interface...which is bad. I need to mark this private and find a better way.

ground/src/main/java/com/google/android/ground/persistence/remote/firestore/schema/LoiMutationConverter.kt
scolsen: For the `else` branch? In this case it's working as intended since this `when` only performs side-effects--the default side-effect is to do nothing.

shobhitagarwal1612 (approved):

PR #1279, opened 2022-09-16
Remove MapPolygon

JSunde:
Replace MapPolygon with a SimpleMapLocationOfInterest wrapping a Polygon Geometry


JSunde:
Do our Polygon geometries support incomplete Polygons? I'm realizing we might have a disconnect here between the MapPolygons which supported incomplete polygons and the new Polygon Geometry which may not.

scolsen:
> Do our Polygon geometries support incomplete Polygons? I'm realizing we might have a disconnect here between the MapPolygons which supported incomplete polygons and the new Polygon Geometry which may not.

good question, I think technically they did but it's something we can retrofit if needed. iirc we didn't let users draw incomplete polygons, so the only case incomplete ones would have existed is if the project owner had imported some from elsewhere, which I don't think ever really happened -- they're pretty much always complete.

plus we have a more accurate model now--our polygons are necessarily complete by definition, we'll have to use linestrings instead for incomplete polys

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/PolygonDrawingViewModel.kt
```diff
@@ -158,20 +159,26 @@ internal constructor(
...
   private fun updateVertices(newVertices: ImmutableList<Point>) {
-    mapPolygon =
-      mapPolygon.map { polygon: MapPolygon -> polygon.toBuilder().setVertices(newVertices).build() }
-    mapPolygonFlowable.onNext(mapPolygon)
+    polygonLocationOfInterest =
+      polygonLocationOfInterest.map {
+        it.copy(
+          locationOfInterest =
+            it.locationOfInterest.copy(
+              geometry = Polygon(LinearRing(newVertices.map { point -> point.coordinate }))
```
scolsen, line 167: I should probably just update the `LinearRing` constructor to take points. we keep having to write this everywhere.

scolsen (approved):

JSunde:
> > Do our Polygon geometries support incomplete Polygons? I'm realizing we might have a disconnect here between the MapPolygons which supported incomplete polygons and the new Polygon Geometry which may not.
> 
> good question, I think technically they did but it's something we can retrofit if needed. iirc we didn't let users draw incomplete polygons, so the only case incomplete ones would have existed is if the project owner had imported some from elsewhere, which I don't think ever really happened -- they're pretty much always complete.
> 
> plus we have a more accurate model now--our polygons are necessarily complete by definition, we'll have to use linestrings instead for incomplete polys

Gotcha, so do I need to update the PolygonDrawingViewModel to use LineStrings to represent the incomplete polygons as the user is drawing them?

scolsen:
> > > Do our Polygon geometries support incomplete Polygons? I'm realizing we might have a disconnect here between the MapPolygons which supported incomplete polygons and the new Polygon Geometry which may not.
> > 
> > 
> > good question, I think technically they did but it's something we can retrofit if needed. iirc we didn't let users draw incomplete polygons, so the only case incomplete ones would have existed is if the project owner had imported some from elsewhere, which I don't think ever really happened -- they're pretty much always complete.
> > plus we have a more accurate model now--our polygons are necessarily complete by definition, we'll have to use linestrings instead for incomplete polys
> 
> Gotcha, so do I need to update the PolygonDrawingViewModel to use LineStrings to represent the incomplete polygons as the user is drawing them?

In theory we'd have to; in practice our `LinearRing` don't throw any exceptions if you try to construct an incomplete/invalid ring—they probably should if we want things to be correct, but this means that we could just take the shortcut of using `LinearRing` directly while drawing and checking for their completeness.

I'd say it might be worth chatting a bit about it before we land on something.

JSunde:
> > > > Do our Polygon geometries support incomplete Polygons? I'm realizing we might have a disconnect here between the MapPolygons which supported incomplete polygons and the new Polygon Geometry which may not.
> > > 
> > > 
> > > good question, I think technically they did but it's something we can retrofit if needed. iirc we didn't let users draw incomplete polygons, so the only case incomplete ones would have existed is if the project owner had imported some from elsewhere, which I don't think ever really happened -- they're pretty much always complete.
> > > plus we have a more accurate model now--our polygons are necessarily complete by definition, we'll have to use linestrings instead for incomplete polys
> > 
> > 
> > Gotcha, so do I need to update the PolygonDrawingViewModel to use LineStrings to represent the incomplete polygons as the user is drawing them?
> 
> In theory we'd have to; in practice our `LinearRing` don't throw any exceptions if you try to construct an incomplete/invalid ring—they probably should if we want things to be correct, but this means that we could just take the shortcut of using `LinearRing` directly while drawing and checking for their completeness.
> 
> I'd say it might be worth chatting a bit about it before we land on something.

We decided offline to update the PolygonDrawingViewModel to use a LineString to represent the incomplete polygon, so I've made that change and then convert that LineString to a LinearRing in PolygonDrawingViewModel#onCompletePolygonButtonClick.

PTAL, Thanks!

scolsen (approved):

Issue #1288, opened 2022-09-20
[Code health] Remove "builder" methods from mutation models


shobhitagarwal1612:
With most classes migrated to Kotlin, we can start simplifying the mutation model classes by removing the legacy builder pattern

PR #1289, opened 2022-09-20
Remove builder() from LocationOfInterestMutation

shobhitagarwal1612:
Towards #1288 

Replaces references of `builder()` with kotlin object instantiation and `copy()`

@scolsen  PTAL?


scolsen (approved):
Thanks, Shobhit!

PR #1290, opened 2022-09-20
Remove builder() from SubmissionMutationEntity

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/1288

Replaces references of `builder()` with kotlin object instantiation and `copy()`

@scolsen PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/mutation/SubmissionMutation.kt
+++ b/ground/src/main/java/com/google/android/ground/model/mutation/SubmissionMutation.kt
@@ -56,14 +56,6 @@ data class SubmissionMutation(
     var responseDeltas: ImmutableList<ResponseDelta> = ImmutableList.of()
       @JvmSynthetic set
 
-    fun setJob(job: Job): Builder = apply { this.job = job }
-
-    fun setSubmissionId(id: String): Builder = apply { this.submissionId = id }
-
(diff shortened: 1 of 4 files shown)


ground/src/main/java/com/google/android/ground/model/mutation/SubmissionMutation.kt
```diff
@@ -56,14 +56,6 @@ data class SubmissionMutation(
...
 
-    fun setJob(job: Job): Builder = apply { this.job = job }
-
-    fun setSubmissionId(id: String): Builder = apply { this.submissionId = id }
-
-    fun setResponseDeltas(deltas: ImmutableList<ResponseDelta>): Builder = apply {
-      this.responseDeltas = deltas
-    }
-
     override fun build(): SubmissionMutation =
```
scolsen, line 59: Can we remove the `Builder` class entirely at this point?

scolsen (approved):

ground/src/main/java/com/google/android/ground/model/mutation/SubmissionMutation.kt
shobhitagarwal1612: There is one usage left which is in Java. So, we need to convert it to Kotlin and then perform the same steps to get rid of Builder

PR #1291, opened 2022-09-21
Converts LocalMutationSyncWorker to Kotlin and removes Builder from Mutation

### LINKED ISSUE #1288: [Code health] Remove "builder" methods from mutation models

shobhitagarwal1612:
Fixes #1288

* Converts `LocalMutationSyncWorker` to Kotlin which contains the last usage of `builder()`
* Remove builder() and related methods from `Mutation`

@scolsen  PTAL?


ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -0,0 +1,161 @@
...
+  ): Map<String, List<Mutation>> = pendingMutations.groupBy { it.userId }
+
+  private fun incrementRetryCounts(
+    mutations: ImmutableList<Mutation>,
+    error: Throwable
+  ): ImmutableList<Mutation> =
+    mutations.map { m: Mutation -> incrementRetryCount(m, error) }.toImmutableList()
+
+  private fun incrementRetryCount(mutation: Mutation, error: Throwable): Mutation =
+    when (mutation) {
```
scolsen, line 146: Since the branches are the same, can we drop the `when`?

scolsen (approved):

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
shobhitagarwal1612: `copy()` function isn't available with `Mutation` as it is not a data class. This is the only solution I could come up with. Any suggestions?

scolsen: ah, that makes sense. I think this is fine for now then. I have a few ideas but they're probably more complicated than this approach so we can maybe investigate them later.

shobhitagarwal1612: 👍

PR #1296, opened 2022-09-24
Add controllers for map and location

shobhitagarwal1612:
Fixes #

@... PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/map/CameraUpdate.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/map/CameraUpdate.kt
@@ -17,28 +17,23 @@ package com.google.android.ground.ui.map
 
 import com.google.android.ground.model.geometry.Point
 import com.google.android.ground.ui.home.mapcontainer.MapContainerViewModel
-import java8.util.Optional
 
-data class CameraUpdate(
-  val center: Point,
-  val zoomLevel: Optional<Float>,
-  val isAllowZoomOut: Boolean
-) {
+data class CameraUpdate(val center: Point, val zoomLevel: Float?, val isAllowZoomOut: Boolean) {
 
   override fun toString(): String =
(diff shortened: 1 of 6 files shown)


shobhitagarwal1612:
@scolsen Could you also take a pass at it?

scolsen:
I'll take a look now!

ground/src/main/java/com/google/android/ground/ui/map/CameraUpdate.kt
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.map
+
+import com.google.android.ground.model.geometry.Point
+import com.google.android.ground.ui.home.mapcontainer.MapContainerViewModel
+
+data class CameraUpdate(
```
scolsen, line 21: Maybe we merge this with `CameraPosition`?

I only suggest this since we're not really making observable updates here. We construct some data we need to perform an observable update, but the actual updating part (applying the new camera location) is handled elsewhere.

We already have `CameraPositon` and some of the data are shared. I find it slightly difficult to understand the difference between these two classes at a glance, and it would simplify the number of classes we have.

wdyt?

scolsen: either way, feel free to do in another PR

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/map/CameraUpdate.kt
shobhitagarwal1612: Thanks for the suggestion! I'll do it in a separate PR.

PR #1301, opened 2022-09-28
Fix survey not visible even when ACLs are present

shobhitagarwal1612:
Firebase uses "survey_organizer" whereas android app expects "survey-organizer". 

Verified by manually setting the ACL to "survey_organizer" in firebase and running the app locally.

@JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/Constants.kt
+++ b/ground/src/main/java/com/google/android/ground/model/Constants.kt
@@ -17,7 +17,7 @@ package com.google.android.ground.model
 
 object Constants {
   const val OWNER = "owner"
-  const val SURVEY_ORGANIZER = "survey-organizer"
-  const val DATA_COLLECTOR = "data-collector"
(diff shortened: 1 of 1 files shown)


JSunde:
Hmm, I seem to remember Gino telling me to change them from "survey_organizer" and "data_collector" to "survey-organizer" and "data-collector". Let me see if I can find the context on that.

JSunde:
Hmm, well unfortunately I didn't provide more context in the PR where I made that change... Let me see if I can find anything in chat.

JSunde:
Here is the context from our chat:

```
it seems survey_organizer isn't recognized as a role
...
Are you ok with "survey-organizer" as role key?
If so I'll manually edit it in the db, but you'll need to update it in the Android app
same for "data-collector" (instead of data_collector)
...
Firebase's schema is JSON like so seems tk make more sense to use slug case
```

So I'm wondering if we need to update the roles in Firebase?

shobhitagarwal1612:
Thanks for the context! Will update the db manually

PR #1306, opened 2022-10-03
Handle continue click in QuestionDataCollectionFragment

JSunde:
- Handle continue click in QuestionDataCollectionFragment
- Store user's response in memory

A future PR will persist the List<Response> when the user clicks "Continue" on the final task.


JSunde:
@scolsen PTAL

PR #1313, opened 2022-10-07
refactor: convert MapFragment to Kotlin; rename methods

scolsen:
Converts the MapFragment interface to Kotlin and renames several methods in efforts to clarify what they mean/do and decouple them from the particular map SDK we use.

For example, we don't assume *what* kind of interactions a map supports on LOIs, instead naming the stream that captures them generically: locationOfInterestInteractions.


shobhitagarwal1612 (approved):
Thanks!

PR #1320, opened 2022-10-12
Map Model

scolsen:
Moves a bunch of model objects in the Map UI package into the model package. Also defines a few abstractions over map types.


JSunde (approved):

ground/src/main/java/com/google/android/ground/model/map/Map.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.map
+
+import com.google.android.ground.model.geometry.Coordinate
+import com.google.common.collect.ImmutableList
+
+interface Map {
```
gino-m, line 21: These classes seem to be abstractions of view behaviors rather than the domain model, and as such I'd think they'd live in the view layer. Please cmiiw!

ground/src/main/java/com/google/android/ground/ui/map/gms/ConverterExt.kt
```diff
@@ -0,0 +1,31 @@
...
+
+fun LatLng.toModelObject(): Coordinate = Coordinate(this.latitude, this.longitude)
+
+fun Coordinate.toGoogleMapsObject(): LatLng = LatLng(this.x, this.y)
+
+fun LatLngBounds.toModelObject(): Bounds =
+  Bounds(this.southwest.toModelObject(), this.northeast.toModelObject())
+
+fun Bounds.toGoogleMapsObject(): LatLngBounds =
+  LatLngBounds(this.southwest.toGoogleMapsObject(), this.northeast.toGoogleMapsObject())
```
shobhitagarwal1612, line 31: We have another converter file `java/com/google/android/ground/ui/map/ConverterExt.kt`. Can we merge these two?

ground/src/main/java/com/google/android/ground/model/map/Map.kt
scolsen: Yes, that makes sense -- I was trying to clean up the organization in our ui.map section a bit -- would it make sense to have:

`ui.map.model` -- or do we feel that's overkill?

ground/src/main/java/com/google/android/ground/ui/map/gms/ConverterExt.kt
scolsen: Yes!

ground/src/main/java/com/google/android/ground/model/map/Map.kt
scolsen: I'll just move these back into `ui.map` for now. We can revisit this later.

shobhitagarwal1612 (approved):

PR #1326, opened 2022-10-13
Handle Continue clicks in Data Collection tasks, move more of state handling to DataCollectionViewModel

JSunde:
- Store List<AbstractTaskViewModel> in DataCollectionViewModel to track the view models of the Tasks the user has filled out so far
- Track the currentPosition in the DataCollectionViewModel as well and subscribe to this position from the DataCollectionFragment, allowing the ViewModel to drive the active tasks and update the position when the user hits continue on a completed task
- Show an error popup if the user's input is invalid when Continue is clicked

Towards #1146


shobhitagarwal1612:
@JSunde The unit tests are failing. Is it ready to be reviewed?

shobhitagarwal1612:
I think the reason they are failing is because we are using a `@Mock DataCollectionViewModel` instead of an actual instance. This causes an NPE while instantiating the fragment as mock isn't properly set. That being said, I don't think we should mock the `DataCollectionViewModel` as that is the class being tested here. Instead, we should test the behavior using exposed live data and repositories

JSunde:
Okay, I think the tests should be working now 👍

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -66,7 +89,34 @@ internal constructor(
...
+      //  include persisting the list of responses to the database
+      currentPosition.postValue(currentPosition.value!! + 1)
+
+      responses[currentTask.task.id] = currentTask.response.value?.orElse(null)
+      return Single.never()
+    } else {
+      popups.get().showError(validationError)
+    }
+
+    return Single.just(validationError)
```
shobhitagarwal1612, line 120: It seems like we are displaying the error twice:

1. First time when we directly do, `popups.get().showError(validationError)`
2. Second time in `handleContinueClickError()`

Please fix.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewPagerAdapterFactory.kt
```diff
@@ -0,0 +1,15 @@
+package com.google.android.ground.ui.datacollection
```
shobhitagarwal1612, line 1: Please add license header

ground/src/main/java/com/google/android/ground/ui/datacollection/QuestionDataCollectionFragment.kt
```diff
@@ -45,11 +46,9 @@ class QuestionDataCollectionFragment constructor(private val task: Task) : Abstr
     super.onCreateView(inflater, container, savedInstanceState)
     val binding = QuestionDataCollectionFragBinding.inflate(inflater, container, false)
 
-    binding.task = task
     binding.lifecycleOwner = this
+    binding.setVariable(BR.viewModel, viewModel)
```
shobhitagarwal1612, line 50: We can directly set the variable 

`binding.viewModel = viewModel`

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -1,3 +1,5 @@
+package com.google.android.ground.ui.datacollection
+
```
shobhitagarwal1612, line 2: Please move this after the license header or else checkstyle will fail

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -13,108 +15,120 @@
...
+    assertThat(fragment.onBack()).isFalse()
+  }
+
+  private fun setupFragment() =
+    DataCollectionTestData.args.toBundle().let {
+      launchFragmentInHiltContainer<DataCollectionFragment>(it) {
+        fragment = this as DataCollectionFragment
+      }
+    }
```
shobhitagarwal1612, line 133: For simplicity, can we move this to the `setup()`?

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -13,108 +15,120 @@
...
+  @BindValue @Mock lateinit var surveyRepository: SubmissionRepository
+  lateinit var fragment: DataCollectionFragment
 
   @Before
   override fun setUp() {
     super.setUp()
 
-    val activityController = Robolectric.buildActivity(MainActivity::class.java).setup()
-    activity = activityController.get()
+    whenever(
```
shobhitagarwal1612, line 56: As discussed, please add a TODO to add more error cases.

ground/src/main/res/layout/question_data_collection_frag.xml
```diff
@@ -21,8 +21,8 @@
 
   <data>
     <variable
-      name="task"
-      type="com.google.android.ground.model.task.Task" />
+      name="viewModel"
+      type="com.google.android.ground.ui.editsubmission.TextTaskViewModel" />
```
shobhitagarwal1612, line 25: Would this type eventually be dynamic?

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -13,108 +15,120 @@
...
+          DataCollectionTestData.surveyId,
+          DataCollectionTestData.loiId,
+          DataCollectionTestData.submissionId
+        )
+      )
+      .thenReturn(Single.just(DataCollectionTestData.submission))
   }
 
   @Test
   fun created_submissionIsLoaded() {
```
shobhitagarwal1612, line 67: Please update the test name accordingly

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -13,108 +15,120 @@
...
-    assertThat(viewPager.currentItem).isEqualTo(0)
-    dataCollectionFragment.view!!
-      .findViewById<Button>(R.id.data_collection_continue_button)
-      .performClick()
-    assertThat(viewPager.currentItem).isEqualTo(1)
-    assertThat(dataCollectionFragment.onBack()).isTrue()
-    assertThat(viewPager.currentItem).isEqualTo(0)
+    onView(withId(R.id.data_collection_continue_button)).perform(click())
+
+    assertThat(ShadowToast.getTextOfLatestToast()).isEqualTo("This field is required")
```
shobhitagarwal1612, line 93: Can we also assert that the task hasn't changed?

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -13,108 +15,120 @@
...
-  fun onBack_firstViewPagerItem_returnsFalse() {
+  fun onContinueClicked_newTaskIsShown() {
     setupFragment()
+    onView(allOf(withId(R.id.user_response_text), isDisplayed())).perform(typeText("user input"))
+
+    onView(withId(R.id.data_collection_continue_button)).perform(click())
 
-    assertThat(dataCollectionFragment.onBack()).isFalse()
+    onView(withText(DataCollectionTestData.task1Name)).check(matches(not(isDisplayed())))
+    onView(withText(DataCollectionTestData.task2Name)).check(matches(isDisplayed()))
```
shobhitagarwal1612, line 104: Similar to the above case, can we assert that no toast was shown in this case?

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionTestData.kt
```diff
@@ -28,11 +28,13 @@ import java.util.*
 import java8.util.Optional
 
 object DataCollectionTestData {
```
shobhitagarwal1612, line 30: Is this class shared between tests? 

We already have a FakeData class that contains generic objects. Tests use that as baseline and modify then according to their needs. Would it possible to remove this in favor of `FakeData`?

Everything which is specific to the test should live inside that. Only common test data should be moved to FakeData.

shobhitagarwal1612 (changes requested):

shobhitagarwal1612:
If `DataCollectionViewModelTest` is already covered by the added test, then we can get rid of that class as well.

ground/src/main/res/layout/question_data_collection_frag.xml
JSunde: I don't think so, since we will have different layouts for the different types of Tasks

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
JSunde: Done!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
JSunde: Done!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
JSunde: Done!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionTestData.kt
JSunde: This class is now only used by this test.

I think we could remove this and use FakeData, but FakeData doesn't have a Submission, so I think we would either have to add that to the DataCollectionFragmentTest or to FakeData.

Let me know which option you prefer!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Done!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -17,31 +17,54 @@ package com.google.android.ground.ui.datacollection
...
-  private val locationOfInterestHelper: LocationOfInterestHelper
+  private val locationOfInterestHelper: LocationOfInterestHelper,
+  private val popups: Provider<EphemeralPopups>,
 ) : AbstractViewModel() {
   val submission: @Hot(replays = true) LiveData<Loadable<Submission>>
   val jobName: @Hot(replays = true) LiveData<String>
   val loiName: @Hot(replays = true) LiveData<String>
+  /** "Continue" button clicks. */
+  private val continueClicks: @Hot PublishProcessor<Nil> = PublishProcessor.create()
+  /** Outcome of user clicking "Save". */
```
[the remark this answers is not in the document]
JSunde: Oops, this is supposed to be "continue", updated!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -17,31 +17,54 @@ package com.google.android.ground.ui.datacollection
...
+  val continueResults: Observable<String?>
```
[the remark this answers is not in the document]
JSunde: I'm not actually sure, I was following the same pattern as the saveResults observable in EditSubmissionViewModel.

Would LiveData be better here?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -60,35 +68,37 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
...
-    return if (viewPager.currentItem == 0) {
+  override fun onBack(): Boolean =
+    if (viewPager.currentItem == 0) {
       // If the user is currently looking at the first step, allow the system to handle the
       // Back button. This calls finish() on this activity and pops the back stack.
       false
     } else {
       // Otherwise, select the previous step.
-      viewPager.currentItem = viewPager.currentItem - 1
+      viewModel.currentPosition.value = viewModel.currentPosition.value!! - 1
```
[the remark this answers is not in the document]
JSunde: The ViewModel needs to keep track of the current position so that we can keep track of which task is active, currently this is used in onContinueClicked to get validate the results of the current task. Then the ViewPager item is updated by the DataCollectionFragments currentPosition observer.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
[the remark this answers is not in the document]
shobhitagarwal1612: LiveData ensures that the value gets rendered after config changes or fragment pause/resume. We would have used observables where we didn't want to handle the event after config changes (e.g. result of activity should only be shown once)

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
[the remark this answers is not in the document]
shobhitagarwal1612: acked

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionTestData.kt
shobhitagarwal1612: Adding a dummy submission object to FakeData sounds better.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -34,21 +35,23 @@ import io.reactivex.processors.BehaviorProcessor
...
+  private val popups: Provider<EphemeralPopups>,
 ) : AbstractViewModel() {
   val submission: @Hot(replays = true) LiveData<Loadable<Submission>>
   val jobName: @Hot(replays = true) LiveData<String>
   val loiName: @Hot(replays = true) LiveData<String>
   /** "Continue" button clicks. */
   private val continueClicks: @Hot PublishProcessor<Nil> = PublishProcessor.create()
   /** Outcome of user clicking "Continue". */
-  val continueResults: Observable<String?>
+  private val continueResults: Observable<String?>
```
shobhitagarwal1612, line 54: Should we get rid of this observable as well? `onContinueClicked()` is already getting called from the layout directly. Not sure what's the purpose of this is now

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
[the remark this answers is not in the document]
JSunde: Actually I think we don't even need this anymore now that we're showing the toast in onContinueClicked.

Removed.

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionTestData.kt
JSunde: Sounds good, moved to FakeData

shobhitagarwal1612 (approved):

PR #1337, opened 2022-10-27
feat: map LOI clustering

scolsen:
Implements initial marker clustering support for LOIs of type Point. Currently, we just use the default clustering thresholds. We can tweak this as needed.

Note that we need to tweak a few things in the Map Fragment to make this work. First of all, we need to pass the clustermanager as the map's callback listener for certain events. We also implement a custom renderer for our marker stylings.

Screen Shot 2022-10-27 at 10 39 43 AM
Screen Shot 2022-10-27 at 10 39 54 AM
Screen Shot 2022-10-27 at 10 40 04 AM


scolsen:
Leaving this in draft state for the moment—need to make sure this behaves well w/ marker selections and clean up the code a little then I'll move this out of draft state.

JSunde:
Is this ready for review or is it till in a draft state?

scolsen:
> Is this ready for review or is it till in a draft state?

Yes, I think it's rfal now!

ground/src/main/java/com/google/android/ground/ui/map/gms/LocationOfInterestClusterItem.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.map.gms
+
+import com.google.android.gms.maps.model.LatLng
+import com.google.android.ground.model.geometry.Point
+import com.google.android.ground.model.locationofinterest.LocationOfInterest
+import com.google.maps.android.clustering.ClusterItem
+
+class LocationOfInterestClusterItem(
```
shobhitagarwal1612, line 23: Should we use a data class instead?

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -96,14 +94,16 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
    * References to Google Maps SDK Markers present on the map. Used to sync and update polylines
    * with current view and data state.
    */
-  private val markers: MutableMap<Marker, MapLocationOfInterest> = HashMap()
+  private val clusters: MutableMap<LocationOfInterestClusterItem, MapLocationOfInterest> = HashMap()
```
gino-m, line 97: Does the MapAdapter need to know about LOIs? Or should it operate directly on MapPin and other map features?

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -209,15 +211,16 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
     }
   }
 
-  private fun onMarkerClick(marker: Marker): Boolean =
-    if (getMap().uiSettings.isZoomGesturesEnabled) {
-      markers[marker]?.let { markerClicks.onNext(it) }
+  private fun onClusterItemClick(item: LocationOfInterestClusterItem): Boolean {
```
gino-m, line 214: Seems this is replacing `onMarkerClick, but we'll still need a way to handle these.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -209,15 +211,16 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
...
+    return if (getMap().uiSettings.isZoomGesturesEnabled) {
+      markerClicks.onNext(MapLocationOfInterest(item.locationOfInterest))
```
gino-m, line 216: Not sure I understand the relation between map pins and cluster items here.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -255,18 +258,6 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
   override fun moveCamera(point: Point, zoomLevel: Float) =
     getMap().moveCamera(CameraUpdateFactory.newLatLngZoom(point.toLatLng(), zoomLevel))
 
-  private fun addMapPin(mapLocationOfInterest: MapLocationOfInterest, point: Point) {
```
gino-m, line 258: We still need to be able to add pins to the map. I think the expectation is that map is smart enough to cluster pins and and polygons at the right zoom levels.

ground/src/main/java/com/google/android/ground/ui/map/gms/LocationOfInterestClusterManager.kt
```diff
@@ -0,0 +1,71 @@
...
+import android.content.Context
+import com.google.android.gms.maps.GoogleMap
+import com.google.android.ground.model.geometry.Point
+import com.google.android.ground.model.locationofinterest.LocationOfInterest
+import com.google.android.ground.ui.map.MapLocationOfInterest
+import com.google.android.ground.util.toImmutableSet
+import com.google.maps.android.clustering.ClusterManager
+import timber.log.Timber
+
+class LocationOfInterestClusterManager(context: Context?, map: GoogleMap) :
```
gino-m, line 27: Can we abstract LOI somehow as mentioned earlier so that the map adapter needs to know less about specific business logic?

gino-m (changes requested):
Some initial comments. Thanks for the PR!

ground/src/main/java/com/google/android/ground/ui/map/MapFragment.kt
```diff
@@ -88,11 +88,7 @@ interface MapFragment {
   /** Update the set of [MapLocationOfInterest]s rendered on the map. */
   fun renderLocationsOfInterest(mapLocationsOfInterest: ImmutableSet<MapLocationOfInterest>)
 
-  /**
-   * Refresh the set of rendered [MapLocationOfInterest], updating their size according to the map's
-   * current zoom level.
-   */
-  fun refreshRenderedLocationsOfInterest()
+  fun refresh()
```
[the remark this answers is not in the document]
scolsen: Added, thanks!

ground/src/main/java/com/google/android/ground/ui/map/gms/LocationOfInterestClusterItem.kt
scolsen: yes, done!

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
scolsen: This is handling marker clicks; the cluster manager becomes responsible for managing clicks on map markers. It's a bit confusing from the terminology, but this is equivalent to asking it to handle a single click on an individual marker ("LOI Item"). I can refactor the name back to "on marker click" if that's clearer.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
scolsen: This function is a no-op change. Basically, this is the same function we previously had, we just need to pass a different type since the cluster manager now handles marker interactions (It's a bit weird, but it's how the clustering API works).

ground/src/main/java/com/google/android/ground/ui/map/gms/LocationOfInterestClusterManager.kt
scolsen: We could in theory, but I'm a bit wary of introducing yet another abstraction. We can extend `MapLocationOfInterest` for this purpose and operate on those.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
scolsen: I can try renaming this, but the "ClusterItem" is an inheritor of a type the cluster manger needs to operate on. Instead of taking LOIs, I can have it operate on MapLocationOfInterest

gino-m: Sounds like I may have opened up a can of worms. Let's come back to this in a follow-up!

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: Ah ok, then we can leave it as-is for now. Maybe please just add a Javadoc explaining that this method will handle both normal ping and cluster marker clicks.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: Got it!

ground/src/main/java/com/google/android/ground/ui/map/gms/LocationOfInterestClusterManager.kt
gino-m: The enforce stricter separation of concerns, the MapAdapter would deal exclusively with geometries and map-like objects rathe rthan Ground-specific things like "LOIs". Let's discuss offline.

gino-m (changes requested):
Thanks! Needs unit tests, otherwise LGTM!

scolsen:
> Thanks! Needs unit tests, otherwise LGTM!

Added some basic tests, rfal whenever!

gino-m (approved):
LGTM!

PR #1338, opened 2022-10-27
Persist TaskData when user presses continue on final Task in Job

JSunde:
Towards #1146

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -93,8 +93,9 @@ internal constructor(
 
     taskCount =
       LiveDataReactiveStreams.fromPublisher(
-        submissionStream
-          .map { submission -> submission.value().map { it.job.tasks.size }.orElse(0) }
+        submissionStream.map { submission ->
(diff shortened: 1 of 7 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -98,11 +113,25 @@ internal constructor(
     val currentTask = taskViewModels.value!![currentPosition.value!!]
     val validationError = currentTask.validate()
     if (validationError == null) {
-      // TODO(#1146): Handle the scenario when the user clicks next on the last Task. This will
-      //  include persisting the list of responses to the database
+      responses[currentTask.task] = currentTask.taskData.value?.orElse(null)
+
+      // TODO(jsunde): Test this behavior
+      if (currentPosition.value!! == taskCount.value!! - 1) {
```
JSunde, line 119: For some reason `taskCount.value` is null in my DataCollectionFragmentTests.

Any ideas why that might be? I don't really understand it because `jobName.value` returns "Job" and the taskCount LiveData is set up in pretty much the same way

JSunde: Resolved offline

shobhitagarwal1612 (approved):

PR #1342, opened 2022-11-05
Update UX for LOI cards

### LINKED ISSUE #1153: [Map] Show job cards at bottom of map
- Each card represents one LOI.
- Cards show: Job name, LOI name and ID (if present), Completion status, "START" affordance to collect data
- If identical LOIs (e.g., same coordinates) exist in multiple jobs, one card is shown for each
- Swiping through cards sets focus to LOI 
- List updates after map settles after panning/zooming

@ngosue Please paste mocks once ready

shobhitagarwal1612:
Fixes #1153

@gino-m PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
@@ -22,6 +22,7 @@ import androidx.core.content.res.ResourcesCompat
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.ground.R
 import com.google.android.ground.databinding.LoiCardItemBinding
+import com.google.android.ground.model.job.Job
 import com.google.android.ground.model.locationofinterest.LocationOfInterest
 import com.google.android.ground.ui.home.mapcontainer.LoiCardAdapter.ViewHolder
 
@@ -31,9 +32,9 @@ import com.google.android.ground.ui.home.mapcontainer.LoiCardAdapter.ViewHolder
  */
 class LoiCardAdapter : RecyclerView.Adapter<ViewHolder>() {
 
-  private var selectedIndex: Int = -1
+  private var focusedIndex: Int = -1
   private val itemsList: MutableList<LocationOfInterest> = mutableListOf()
-  private lateinit var cardSelectedCallback: (LocationOfInterest?) -> Unit
+  private lateinit var cardFocusedCallback: (LocationOfInterest?) -> Unit
   private lateinit var collectDataCallback: (LocationOfInterest) -> Unit
   private lateinit var reviewDataCallback: (LocationOfInterest) -> Unit
 
@@ -53,7 +54,7 @@ class LoiCardAdapter : RecyclerView.Adapter<ViewHolder>() {
     holder.binding.wrapperView.background =
       ResourcesCompat.getDrawable(
         holder.itemView.context.resources,
-        if (selectedIndex == position) {
+        if (focusedIndex == position) {
           R.drawable.border
         } else {
           R.color.colorBackground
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
```diff
@@ -44,63 +45,58 @@ class LoiCardAdapter : RecyclerView.Adapter<ViewHolder>() {
...
 
-    // Handle action buttons.
-    holder.binding.loiCard.setOnClickListener { handleItemClicked(position) }
-    holder.binding.start.setOnClickListener { handleButtonClicked(position) }
-    holder.binding.review.setOnClickListener { TODO() }
-    holder.binding.markComplete.setOnClickListener { TODO() }
+    val isCaptionMissing = loi.caption.isNullOrEmpty()
+    holder.binding.loiIcon.visibility = if (isCaptionMissing) View.GONE else View.VISIBLE
+    holder.binding.loiName.visibility = if (isCaptionMissing) View.GONE else View.VISIBLE
+    holder.binding.loiCard.setOnClickListener { collectDataCallback.invoke(loi) }
```
gino-m, line 67: Is it possible to use data bindings for this instead of setting programmatically? If it's difficult or not possible please disregard.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
```diff
@@ -44,63 +45,58 @@ class LoiCardAdapter : RecyclerView.Adapter<ViewHolder>() {
...
+    focusedIndex = -1
     notifyDataSetChanged()
-    cardSelectedCallback.invoke(null)
+    cardFocusedCallback.invoke(null)
   }
 
-  fun setLoiCardSelectedCallback(callback: (LocationOfInterest?) -> Unit) {
-    this.cardSelectedCallback = callback
+  fun setLoiCardFocusedCallback(callback: (LocationOfInterest?) -> Unit) {
+    this.cardFocusedCallback = callback
```
gino-m, line 93: Should we call this `*Listener` for consistency with Android SDK callbacks?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
shobhitagarwal1612: We can do it by passing a boolean variable "isCaptionMissing" to the xml. I tried directly using `.isNullOrEmpty()` in data binding but it didn't work.

gino-m: It seems data binding still uses Java, so extension functions like `.isNullOrEmpty()` would need to be imported and called statically ([example](https://stackoverflow.com/a/59756207/9986466)). Ideally we'd model view states in the `ViewModel` class rather than expect the layout to compute them; in this case the states in the VM should already be non-null and reflect any defaults or overrides. This way we can unit test them more easily, and we have a singular place to look for view logic.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/LoiCardAdapter.kt
shobhitagarwal1612: Thanks for the tip! Done

ground/src/main/res/layout/loi_card_item.xml
```diff
@@ -57,6 +56,7 @@
         tools:text="Map the terrain" />
       <androidx.constraintlayout.utils.widget.ImageFilterView
         android:id="@+id/loiIcon"
+        visible="@{!TextUtils.isEmpty(loi.caption)}"
```
gino-m, line 59: Can we eventually move this logic to the ViewModel? That might require an extra class to contain the card view states, but it might be worth it for better testability.

gino-m (approved):
LGTM! Added one comment for future work.

ground/src/main/res/layout/loi_card_item.xml
shobhitagarwal1612: +1

shobhitagarwal1612: Filed as a separate bug

PR #1344, opened 2022-11-17
Pull out the Data Collection Task header into subcomponent

JSunde:
Added a new subcomponent view for the Data Collection Task header so it can be reused between different task fragments

Towards #1146 

@shobhitagarwal1612 PTAL

### DIFF
--- a/ground/src/main/res/layout/data_collection_header.xml
+++ b/ground/src/main/res/layout/data_collection_header.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  ~ Copyright 2022 Google LLC
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     https://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


JSunde:
Let me know if you have any ideas why the tests are failing now. I ran the app on my device and it seems to be working properly so I don't know why I'm getting these errors:

```
is displayed on the screen to the user' doesn't match the selected view.
Expected: is displayed on the screen to the user
     Got: "MaterialTextView{id=2131230879, res-name=data_collection_question, visibility=VISIBLE, width=6, height=65, has-focus=false, has-focusable=false, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=androidx.constraintlayout.widget.ConstraintLayout$LayoutParams@5a5f1e45, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=41.0, text=task 1, input-type=0, ime-target=false, has-links=false}"
```

shobhitagarwal1612:
All of the view assertions are failing. I'll try to take a look at it today.

scolsen:
@JSunde so, it's subtle, but a one-line change made it such that the view is not actually visible. Change the constraint on `data_collection_question` back to `app:layout_constraintTop_toBottomOf="@+id/data_collection_task_name" />`

JSunde:
Thanks @scolsen! I took a closer look and I think there were actually a couple things wrong with the constraints there, I ended up constraining the top to the bottom of `data_collection_step_number` and the bottom to the bottom of the parent and now it seems to be passing the test and looks correct on my device.

scolsen (approved):

PR #1345, opened 2022-11-23
Set up Mutliple Choice Data Collection Tasks

JSunde:
Towards #1146

Adds a new Task screen with RadioButtons for single selection Multiple Choice tasks and CheckBoxes for multi selection Multiple Choice tasks.

[Single Select](https://screenshot.googleplex.com/3iMz4De3xq8uM4k)
[Multi Select](https://screenshot.googleplex.com/9WY4XimT3XmKgkc)


JSunde:
`setHasFixedSize()` fixed the issue where checking an item at the bottom of the list would cause the RecyclerView to scroll back to the top of the list. However the fix only seems to have worked for the Select One RadioButton use case and not the Select Multiple CheckBox use case.

PTAL and let me know if you have any ideas or suggestions when you get a chance @shobhitagarwal1612, @gino-m!

ground/src/main/java/com/google/android/ground/ui/datacollection/MultipleChoiceDataCollectionFragment.kt
```diff
@@ -0,0 +1,68 @@
...
+    binding.lifecycleOwner = this
+    binding.setVariable(BR.viewModel, viewModel)
+
+    val multipleChoice = task.multipleChoice!!
+    if (multipleChoice.cardinality == MultipleChoice.Cardinality.SELECT_MULTIPLE) {
+      // TODO(jsunde): Figure out why the Multiple Choice RecyclerView still scrolls back to the top
+      //  then factor out any shared logic between these branches
+      val optionAdapter = SelectMultipleOptionAdapter(multipleChoice.options)
+      optionAdapter.setHasStableIds(true)
+      val optionListView = binding.root.findViewById<RecyclerView>(R.id.select_one_option_list)
```
scolsen, line 57: Should this be `select_multiple...`? That id appears elsewhere in the PR, or are we intentionally using the same layout for both?

scolsen:
Looking at the code, I'm not sure why the scrolling still happens. I'm guessing this is just an odd RecyclerView default behavior. Looks like there's a dedicated library for handling selections in RecyclerViews, maybe we should use that? https://developer.android.com/develop/ui/views/layout/recyclerview-custom#select

scolsen:
Note: I'm guessing this is probably the culprit of the scrolling: https://developer.android.com/develop/ui/views/layout/recyclerview-custom#animations

ground/src/main/java/com/google/android/ground/ui/datacollection/MultipleChoiceDataCollectionFragment.kt
JSunde: We're using the same layout for both, but I think I forgot to update the id of the list after I made that change. I'll update it to just select_option_list

PR #1355, opened 2022-12-14
Extract PhotoResult from EditSubmissionViewModel for reuse in Data Collection flow

JSunde:
Towards #1146

- Extracting the PhotoResult out of EditSubmissionViewModel for reuse from the upcoming PhotoDataCollectionTask
- Converting from Java @AutoValue to Kotlin data class

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoResult.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoResult.kt
@@ -1,0 +1,13 @@
+package com.google.android.ground.ui.editsubmission
+
+import android.graphics.Bitmap
+
+/** Contains the bitmap or path to the photo a user captured or selected. */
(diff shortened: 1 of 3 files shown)


JSunde:
@gino-m PTAL

ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoResult.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2020 Google LLC
```
gino-m: Note I updated this file to use the current year

gino-m (approved):
LGTM!

ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoResult.kt
JSunde: Thanks!

PR #1357, opened 2022-12-16
Add Photo Capture/Selection Data Collection Flow

JSunde:
Towards #1146


JSunde:
I'm still trying to figure out how to reduce the code duplication and share more of the logic between the Edit Submission flow and the Data Collection flow, but I'm currently running into an issue where the photo view doesn't get updated after the photo is captured.

It appears we are able to write the photo to the gallery and then updateResponse is called which should update the ViewModel's detailsTextFlowable which drives the uri flow, but for some reason the photo never shows up.

Let me know if anyone has any ideas!

JSunde:
Okay so it looks like the LiveData<Uri> is getting updated multiple times after the photo is captured, first with the image URI and then a second time with an empty string.

JSunde:
Alright I think I'm starting to understand this a little better. It looks like the LiveData is getting updated correctly once after the photo is captured, but then DataCollectionViewPagerAdapter#createFragment is getting called which causes us to instantiate a new PhotoTaskViewModel.

I think I might need to add a check in DataCollectionViewPagerAdapter#createFragment to see if there already is a ViewModel for the task in that position and reuse that ViewModel if so.

gino-m:
> Alright I think I'm starting to understand this a little better. It looks like the LiveData is getting updated correctly once after the photo is captured, but then DataCollectionViewPagerAdapter#createFragment is getting called which causes us to instantiate a new PhotoTaskViewModel.
> 
> I think I might need to add a check in DataCollectionViewPagerAdapter#createFragment to see if there already is a ViewModel for the task in that position and reuse that ViewModel if so.

Ideally the draft state would be stored somewhere that would survive not only configuration changes, but also stop/start. In that case we'd need to store the draft in the db, but that's probably out of scope of this PR.

JSunde:
> > Alright I think I'm starting to understand this a little better. It looks like the LiveData is getting updated correctly once after the photo is captured, but then DataCollectionViewPagerAdapter#createFragment is getting called which causes us to instantiate a new PhotoTaskViewModel.
> > I think I might need to add a check in DataCollectionViewPagerAdapter#createFragment to see if there already is a ViewModel for the task in that position and reuse that ViewModel if so.
> 
> Ideally the draft state would be stored somewhere that would survive not only configuration changes, but also stop/start. In that case we'd need to store the draft in the db, but that's probably out of scope of this PR.

Yea I think that we will probably want to support that eventually, but I was able to get this working for now by making sure we reuse the same ViewModels for the task.

JSunde:
Now I was able to get the image preview working, but for some reason it doesn't seem to be respecting its constraints. I believe it should be shown below the "Add photo" button, but it seems to be rendering over top of it.

Let me know if you have any ideas on that @gino-m or @scolsen! Thanks!
image

JSunde:
I ended up getting the constraint issue fixed simply by switching to LinearLayout as suggested by Gino, so everything is working as expected now.

ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoTaskViewModel.kt
```diff
@@ -24,16 +25,53 @@ import com.google.android.ground.model.submission.TextTaskData.Companion.fromStr
...
+  private val userMediaRepository: UserMediaRepository,
+  private val permissionsManager: PermissionsManager,
+  private val bitmapUtil: BitmapUtil,
+  resources: Resources
+) : AbstractTaskViewModel(resources) {
+
+  // TODO(jsunde): This logic is currently duplicated here and in EditSubmissionViewModel.
+  //  If it's here it can easily be used for the DataCollectionFragment, but is slightly more
+  //  complicated to reuse from the EditSubmissionFragment. I should explore some more to see if
+  //  it's possible to remove this logic from the EditSubmissionViewModel
```
JSunde, line 54: @gino-m Do you have any suggestions for how to dedupe some of this logic? I was hoping to be able to reuse the photo handling logic here and in the EditSubmissionFragment, but I was having a hard time seeing how I would do that when the individual ViewModels are not restored from the SavedInstanceState so I'm not sure how we would keep track of the task that was waiting for the photo in the EditSubmissionFragment.

ground/src/main/java/com/google/android/ground/ui/datacollection/PhotoDataCollectionFragment.kt
```diff
@@ -0,0 +1,171 @@
...
+import dagger.hilt.android.AndroidEntryPoint
+import io.reactivex.Completable
+import timber.log.Timber
+
+/** Fragment allowing the user to capture a photo to complete a task. */
+@AndroidEntryPoint
+class PhotoDataCollectionFragment
+constructor(
+  private val task: Task,
+  private val viewModel: PhotoTaskViewModel,
```
gino-m, line 50: Please sync name of task with owning fragment.

ground/src/main/java/com/google/android/ground/ui/editsubmission/PhotoTaskViewModel.kt
gino-m: Rather than managing view models in `getTaskViewModel`, why not binding them 1:1 to fragments and let Jetpack handle them for you as we do elsewhere? I believe the ViewModels should then survive the photo intent.

gino-m: Actually, I believe the app can be destroyed while in the photo intent, so we'd need to manage the persistent state in `SaveInstanceState`.

JSunde: Synced offline and we decided that we don't need to address this at this time since the EditSubmissionFragment will likely end up getting removed in favor of an edit mode for the data collection flow.

ground/src/main/java/com/google/android/ground/ui/datacollection/PhotoDataCollectionFragment.kt
JSunde: Done!

gino-m: Sorry, I was referring to the class name. `task` is find as a variable name and in comments since the type is clear from context. Re`PhotoDataCollectionFragment` vs `PhotoTaskViewModel`: `PhotoTask{Fragment|ViewModel}` would be sufficient. Open to other suggestions.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/PhotoDataCollectionFragment.kt
JSunde: Gotcha, I updated all the Data Collection Task names to {TaskType}TaskFragment.

gino-m (approved):
LGTM, nice work! 🏅

PR #1359, opened 2022-12-23
[Code health] Small tweaks to util methods

gino-m:
@scolsen PTAL?


ground/src/main/java/com/google/android/ground/util/StreamUtil.kt
```diff
@@ -22,26 +22,11 @@ import timber.log.Timber
 
 /** Methods for working with and manipulating [java8.util.stream.Stream]. */
 object StreamUtil {
```
shobhitagarwal1612, line 24: We should also rename this class to something more appropriate since `Stream` isn't used anymore

gino-m: What would be a better name? Hopefully when moving to coroutines this helper won't be needing anymore.

shobhitagarwal1612: The remaining method is only used in 1 class. We can even consider removing this util and making this a private method.

ground/src/main/java/com/google/android/ground/util/StreamUtil.kt
```diff
@@ -22,26 +22,11 @@ import timber.log.Timber
...
-  fun <R> logErrorsAndSkip(supplier: Supplier<R>): Stream<R> {
-    return try {
-      StreamSupport.stream(setOf(supplier.get()))
-    } catch (e: RuntimeException) {
-      Timber.e(e)
-      StreamSupport.stream(emptySet())
-    }
-  }
-
   fun <R> logErrorsAndSkipKt(supplier: Supplier<R>): Iterable<R> {
```
scolsen, line 25: sorry for the out-of-diff comment, but do we need the `Kt` suffix here?

scolsen (approved):

ground/src/main/java/com/google/android/ground/util/StreamUtil.kt
gino-m: Good catch. Fixed!

ground/src/main/java/com/google/android/ground/util/StreamUtil.kt
gino-m: Fixed!

scolsen (approved):

ground/src/main/java/com/google/android/ground/persistence/local/room/RoomLocalDataStore.kt
```diff
@@ -508,7 +506,7 @@ class RoomLocalDataStore @Inject internal constructor() : LocalDataStore {
    * Applies mutation to submission in database or creates a new one.
    *
    * @return A Completable that emits an error if mutation type is "UPDATE" but entity does not
-   * exist, or if type is "CREATE" and entity already exists.
+   *   exist, or if type is "CREATE" and entity already exists.
```
shobhitagarwal1612, line 509: This extra space is causing `ktfmtCheck` to fail. `ktfmtFormat` removes it while Android Studio's code formatter adds it. For now, let's just fix this manually to unblock this PR and tackle this issue separately.

gino-m: Funny, I thought I tried removing it and ktfmt still failed. Part of the problem is there's no line number or other hints reported by ktfmt, so hard to know where the actually error occurs. Trying this now.

PR #1360, opened 2022-12-23
Add worker and service for syncing survey in background

gino-m:
Towards #1373.

@shobhitagarwal1612 PTAL?

### DIFF
--- a/ground/build.gradle
+++ b/ground/build.gradle
@@ -41,6 +41,8 @@ project.ext {
     roomVersion = "2.3.0"
     rxBindingVersion = "2.2.0"
     workVersion = "2.5.0"
+    mockitoVersion = "4.4.0"
+    mockitoKotlinVersion = "4.1.0"
 }
 
 // Load secrets.properties
@@ -167,6 +169,7 @@ dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
     implementation project(':android-gmaps-addons')
     implementation 'androidx.recyclerview:recyclerview-selection:1.1.0'
+    implementation 'androidx.test.ext:junit-ktx:1.1.4'
     testImplementation project(':sharedTest')
     androidTestImplementation project(':sharedTest')
 
@@ -267,7 +270,7 @@ dependencies {
     // WorkManager
     implementation "androidx.work:work-runtime:$workVersion"
     implementation "androidx.work:work-rxjava2:$workVersion"
-    androidTestImplementation "androidx.work:work-testing:$workVersion"
+    testImplementation "androidx.work:work-testing:$workVersion"
 
     implementation "com.uber.autodispose:autodispose-android:$project.autoDisposeVersion"
     implementation "com.uber.autodispose:autodispose-android-archcomponents:$project.autoDisposeVersion"
@@ -286,10 +289,12 @@ dependencies {
     testImplementation "androidx.arch.core:core-testing:1.3.0"
 
     // Mockito
-    testImplementation 'org.mockito:mockito-inline:2.8.47'
-    testImplementation 'org.mockito:mockito-core:4.4.0'
-    androidTestImplementation 'org.mockito:mockito-core:4.4.0'
-    testImplementation "org.mockito.kotlin:mockito-kotlin:4.0.0"
+    testImplementation "org.mockito:mockito-inline:$mockitoVersion"
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
+    testImplementation "org.mockito:mockito-android:$mockitoVersion"
+    androidTestImplementation "org.mockito:mockito-core:$mockitoVersion"
+    androidTestImplementation "org.mockito:mockito-android:$mockitoVersion"
+    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
 
     androidTestImplementation('androidx.test.espresso:espresso-core:3.3.0', {
         exclude group: 'com.android.support', module: 'support-annotations'
--- a/ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
@@ -53,7 +53,7 @@ constructor(
   }
 
   companion object {
-    /**  The key in worker input data containing the id of the survey to be synced. */
+    /** The key in worker input data containing the id of the survey to be synced. */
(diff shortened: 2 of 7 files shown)


ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirebaseMessagingService.kt
```diff
@@ -0,0 +1,40 @@
...
+import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
+import timber.log.Timber
+
+@AndroidEntryPoint
+class FirebaseMessagingService : FirebaseMessagingService() {
+
+  @Inject lateinit var surveySyncService: SurveySyncService
+
+  override fun onMessageReceived(remoteMessage: RemoteMessage) {
```
gino-m, line 31: @shobhitagarwal1612 Discussed moving this logic to a delegate to facilitate testing.

ground/build.gradle
```diff
@@ -267,7 +270,7 @@ dependencies {
     // WorkManager
     implementation "androidx.work:work-runtime:$workVersion"
     implementation "androidx.work:work-rxjava2:$workVersion"
-    androidTestImplementation "androidx.work:work-testing:$workVersion"
```
shobhitagarwal1612, line 270: Are we certain that this dependency is no longer needed? Could you please verify this by running the instrumentation tests locally?

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
```diff
@@ -0,0 +1,64 @@
...
+/** Worker responsible for syncing latest surveys and LOIs from remote server to local db. */
+@HiltWorker
+class SurveySyncWorker
+@AssistedInject
+constructor(
+  @Assisted context: Context,
+  @Assisted params: WorkerParameters,
+  private val surveyRepository: SurveyRepository
+) : Worker(context, params) {
+  private val surveyId: String? = params.inputData.getString(SURVEY_ID_PARAM_KEY)!!
```
shobhitagarwal1612, line 38: Why is this variable marked as both nullable and the value asserted to be non-null(!!) ?

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
```diff
@@ -0,0 +1,64 @@
...
+
+  override fun doWork(): Result {
+    if (surveyId == null) {
+      Timber.e("Survey sync scheduled with null surveyId")
+      return Result.failure()
+    }
+
+    // It's ok to block here since WorkManager calls doWork() on a background thread.
+    Timber.d("Syncing survey $surveyId")
+    surveyRepository.syncSurveyWithRemote(surveyId).ignoreElement().blockingAwait()
```
shobhitagarwal1612, line 48: Out of curiosity, is there a single background thread for the entire application or does each worker spawn in a new thread?

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
```diff
@@ -0,0 +1,64 @@
...
+    // TODO: Handle failures - log and retry.
+    return Result.success()
+  }
+
+  companion object {
+    /** The key in worker input data containing the id of the survey to be synced. */
+    private const val SURVEY_ID_PARAM_KEY = "surveyId"
+
+    /** Returns a new work [Data] object containing the specified survey id. */
+    @JvmStatic
```
shobhitagarwal1612, line 60: Why is this annotation required? Isn't this meant to be used when the reference is from java code?

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
gino-m: The docs are conspicuously silent on this question:

> By default, WorkManager sets up an Executor for you—but you can also customize your own. 
https://developer.android.com/guide/background/persistent/threading/worker

How can we verify this easily?

ground/build.gradle
gino-m: I'm still unable to run tests via AS; 0/0 run when initiated via the UI. I also ran `gradlew connectedAndroidTest` which failed due to `com.android.tools.r8.internal.a: Cannot fit requested classes in a single dex file (# methods: 184732 > 65536)` during `:sharedTest:mergeExtDexDebugAndroidTest`.

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
gino-m: Good catch. This was probably due to copy/paste abuse, fixed.

ground/build.gradle
gino-m: After using a newer SDK image, I am able to run tests, but they're already failing at HEAD https://github.com/google/ground-android/issues/1410

We should be able to delete this dependency since we don't have integration tests for workers atm.

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
gino-m: It isn't.. removed!

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
gino-m: Found it:

```java
private @NonNull Executor createDefaultExecutor() {
  return Executors.newFixedThreadPool(
          // This value is the same as the core pool size for AsyncTask#THREAD_POOL_EXECUTOR.
          Math.max(2, Math.min(Runtime.getRuntime().availableProcessors() - 1, 4)));
}
```
https://android.googlesource.com/platform/frameworks/support.git/+/a9ac247af2afd4115c3eb6d16c05bc92737d6305/work/workmanager/src/main/java/androidx/work/Configuration.java#107

gino-m:
@shobhitagarwal1612 RFAL!

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
shobhitagarwal1612: Thanks for confirming

shobhitagarwal1612 (approved):

PR #1362, opened 2022-12-23
[WIP] Gps Data Collect Prototype

shobhitagarwal1612:
### 1. Pull shared code from MapContainerFragment and MapContainerViewModel to abstract classes

 * Create a new basemap_layout to contain shared map ui elements (basemap, gps lock button, map type button)
 * Refactor home_layout to use the basemap_layout
 * Create a new VM (BasemapViewModel) to control the basemap_layout
 * Refactor MapContainerViewModel by extending BasemapViewModel
 * Pull shared LiveData bindings from MapContainerFragment to AbstractMapViewerFragment

### 2. Create new fragment and VM implementations for GPS Data collection

 * Create new fragment and VM for collecting GPS Data
 * Add Gps data models
 * Integrate with DataCollectionFragment

## Next steps

1. Split the code refactoring into multiple PRs
2. Add support for GPS data collection

cc @JSunde

PR #1364, opened 2022-12-23
[GPS Data Collect]  Create shared map fragment and view model

### LINKED ISSUE #1363: [GPS Data Collect] Refactor MapContainerFragment and pull common code to abstract classes
Context: https://github.com/google/ground-android/pull/1362

shobhitagarwal1612:
Fixes #1363

Basic prototype: https://github.com/google/ground-android/pull/1362

This PR aims at creating shared fragment and view model classes for setting up basemap along with user interaction functionality. These shared classes would be used by MapContainerFragment and upcoming Data Collection fragments.

This includes:
1. Attach map fragment
2. Load last saved location
3. Load last saved map type
4. Map Controls (Gps Lock & Map Type)

Verified by installing the app locally and interacting with the map.

@gino-m PTAL?


ground/src/main/java/com/google/android/ground/ui/common/AbstractMapViewerFragment.kt
```diff
@@ -17,19 +17,108 @@ package com.google.android.ground.ui.common
...
+import com.google.android.ground.system.SettingsChangeRequestCanceled
+import com.google.android.ground.ui.home.mapcontainer.MapTypeDialogFragmentDirections
+import com.google.android.ground.ui.map.CameraPosition
 import com.google.android.ground.ui.map.MapFragment
 import javax.inject.Inject
+import kotlin.math.max
+import timber.log.Timber
 
-/** Injects a [MapFragment] in the container with id "map". */
+/** Injects a [MapFragment] in the container with id "map" and provides shared map functionality. */
```
gino-m, line 34: Thoughts on naming:

* Technically this is still a map container, so perhaps `AbstractMapContainerFragment` here?
* There will now be multiple map containers - should we rename `MapContainer*` to something unique, like `HomeScreenMapContainer` for consistency? Later we should find a better name for "home screen". @amegantz for discussion offline.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/MapContainerViewModel.kt
```diff
@@ -183,35 +171,17 @@ internal constructor(
...
-  }
-
   fun onMarkerClick(mapLocationOfInterest: MapLocationOfInterest) {
-    val locationOfInterest = mapLocationOfInterest.locationOfInterest
-    if (locationOfInterest != null) {
-      val geometry = locationOfInterest.geometry
-      if (geometry is Point) {
-        mapController.panAndZoomCamera(geometry)
-      }
+    val geometry = mapLocationOfInterest.locationOfInterest.geometry
```
gino-m, line 175: Change LG. Was any other logic changes besides this? Was moved/refactored code moved without changes? It's hard to tell from the diff for code that was moved/refactored.

gino-m (changes requested):
A few suggestions, LG!

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/MapContainerViewModel.kt
shobhitagarwal1612: This was flagged as "always true" by the IDE since LOI can't be null. This was the only logical change apart from the refactor

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapViewerFragment.kt
shobhitagarwal1612: Done. Thanks for the suggestion

shobhitagarwal1612:
@gino-m Resolved all comments, PTAL.

gino-m (approved):
LGTM. Thanks for the refactor!

PR #1371, opened 2022-12-27
[Code health] Rename classes to use correct names for various Firebase products

gino-m:
@scolsen PTAL?

PR #1382, opened 2022-12-29
[Code health] Fix naming of Firebase products

gino-m:
Some products were named incorrectly in our code base.

Moved `GeometryConverter` into the appropriate package as well.

@scolsen PTAL?


gino-m:
Merge of these commits already requested in #1371.

PR #1384, opened 2022-12-30
Add icon for location lock "not_fixed" state

### LINKED ISSUE #1378: [GPS Lock] Fix icon when lock is released
Currently, the lock icon is same for both lock enabled and disabled. Behaviour should be similar to how Google Maps work.

shobhitagarwal1612:
Fixes #1378

Switch location lock icon based on lock state.

@gino-m  PTAL?


gino-m (approved):
LG, thanks!!

PR #1394, opened 2023-01-06
[Code health] refactor: split localdatastore into subcomponents

scolsen:
Breaks down the monolithic datastore into a composition of a few dedicated data access components for each model object in the application. This  will enable us to update the implementations modularly or swap them out for alternatives. It should also help keep the diffs focused as we incrementally move to coroutines.


ground/src/main/java/com/google/android/ground/persistence/local/LocalDataStore.kt
```diff
@@ -48,76 +43,18 @@ import io.reactivex.Single
...
-    submissionId: String
-  ): @Cold Maybe<Submission>
-
-  /**
-   * Returns a long-lived stream that emits the full set of tiles on subscribe and continues to
-   * return the full set each time a tile is added/changed/removed.
-   */
-  val tileSetsOnceAndStream: @Cold(terminates = false) Flowable<ImmutableSet<TileSet>>
+  /** Provides access to [Survey] data in local storage. */
+  var surveyStore: SurveyStore
```
gino-m, line 47: If we're splitting up local data stores, for consistency would we then also split up remote ones? If so, should we prefix these with `Local` now to differentiate them conceptually from their eventual remote counterparts?

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
```diff
@@ -172,8 +174,8 @@ constructor(
     )
 
   private fun applyAndEnqueue(mutation: SubmissionMutation): @Cold Completable =
-    localDataStore
-      .applyAndEnqueue(mutation)
+    submissionStore
+      .commitThenEnqueue(mutation)
```
gino-m, line 178: Why the change from apply --> commit?

Also, because of the new files, git/GitHub aren't able to highlight which methods were renamed. Let's move the rename into its own PR or its own commit to make it clear?

gino-m (changes requested):
LG, two questions/comments.

ground/src/main/java/com/google/android/ground/repository/SubmissionRepository.kt
scolsen: Ah, sorry about that -- yeah I can revert the rename and separate it out. The motivation was that kotlin has a stdlib scope function called `apply` (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/apply.html) so using `apply` in kotlin code requires ugly backticking.

ground/src/main/java/com/google/android/ground/persistence/local/LocalDataStore.kt
scolsen: sgtm! Will do

scolsen:
@gino-m RFAL

PR #1397, opened 2023-01-08
Move xml logic for LoiCard to view model

### LINKED ISSUE #1346: Move xml logic in LoiCard to the ViewModel
Can we eventually move this logic to the ViewModel? That might require an extra class to contain the card view states, but it might be worth it for better testability.

_Originally posted by @gino-m in https://github.com/google/ground-android/pull/1342#discussion_r1044559245_

shobhitagarwal1612:
Fixes #1346

@gino-m  PTAL?


gino-m (approved):
Thanks for the cleanup!

Issue #1351, opened 2022-12-13
Integrate polygon drawing with data collection flow


shobhitagarwal1612:

shobhitagarwal1612:
Notes from UX meeting:

- [ ] Add a point counter at the bottom left screen denoting number of added points.

shobhitagarwal1612:
Filed https://github.com/google/ground-android/issues/1839 for the remaining AIs

PR #1398, opened 2023-01-08
[Cleanup] Remove usages of PolygonDrawingView and PolygonDrawingViewModel

shobhitagarwal1612:
Towards #1351

This PR aims at removing current usages of PolygonDrawingView and move reusable code to a new VM. This is in preparation for adding support to draw polygon with data collection flow

@JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
@@ -74,6 +74,7 @@ internal constructor(
   val showLocationOfInterestSelectorRequests: @Hot Subject<ImmutableList<LocationOfInterest>> =
     PublishSubject.create()
 
+  // TODO: Cleanup this method
   fun addLoi(job: Job, point: Point) {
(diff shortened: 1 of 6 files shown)


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -74,6 +74,7 @@ internal constructor(
   val showLocationOfInterestSelectorRequests: @Hot Subject<ImmutableList<LocationOfInterest>> =
     PublishSubject.create()
 
+  // TODO: Cleanup this method
```
scolsen, line 77: Do we have an associated issue # we can put here?

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
shobhitagarwal1612: Will file a bug separately and update this.

PR #1405, opened 2023-01-09
refactor: use coroutines for some UserRepository methods

scolsen:
A first experiment in a gradual migration to coroutines. We currently need to couple them with RX code, so we use runBlocking and RX methods to do so.

@gino-m @shobhitagarwal1612 @JSunde

### DIFF
--- a/ground/src/main/java/com/google/android/ground/MainViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/MainViewModel.kt
@@ -40,6 +40,8 @@ import io.reactivex.Observable
 import io.reactivex.disposables.Disposable
 import java8.util.Optional
 import javax.inject.Inject
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
 import timber.log.Timber
 
 /** Top-level view model representing state of the [MainActivity] shared by all fragments. */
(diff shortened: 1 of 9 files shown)


shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/MainViewModel.kt
```diff
@@ -131,18 +133,18 @@ constructor(
...
-          Observable.just(HomeScreenFragmentDirections.showHomeScreen())
-        } else {
-          termsOfServiceRepository.termsOfService
-            .map { SignInFragmentDirections.showTermsOfService().setTermsOfServiceText(it.text) }
-            .cast(NavDirections::class.java)
-            .switchIfEmpty(Maybe.just(HomeScreenFragmentDirections.showHomeScreen()))
-            .toObservable()
-        }
-      )
+    return runBlocking(Dispatchers.IO) {
```
gino-m, line 136: Iiuc this will block as soon as the method is called, not when the returned Observable is subscribed to. To reproduce the current behavior, should we instead wrap the coroutine to make its invocation lazy? Please cmiiw!

gino-m (changes requested):
Wicked! Cool to see how we can apply coroutines in practice. One question, otherwise LG.

ground/src/main/java/com/google/android/ground/MainViewModel.kt
scolsen: Yes, I think that's correct. Actually, I think this will block at two different times:

1. It will block immediately when this is called and execute `saveUser` and then return an Observable.
2. It will block again when the returned observable is subscribed to.

So you're right, I think for full parity of behavior we need laziness (to make saveUser also lazy).

JSunde (approved):

ground/src/main/java/com/google/android/ground/MainViewModel.kt
gino-m: Sg.. assigned to you for follow-up, please feel free to assign back and/or ping when ready.

scolsen:
Closing for now. We discussed this a bit further and decided the best pat h forward is likely to move to coroutines one layer at a time (until we get to the view/vm layer).

PR #1412, opened 2023-01-11
Persist location lock state across flows

### LINKED ISSUE #1377: Persist GPS Lock state
GPS Lock's state gets reset after every app restart. As discussed offline, this should remain consistent across screen changes and app restarts

shobhitagarwal1612:
Fixes #1377

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
@@ -45,11 +45,11 @@ class LocalValueStore @Inject constructor(private val preferences: SharedPrefere
       preferences.edit().putInt(MAP_TYPE, type).apply()
     }
 
-  /** State of the location lock. */
-  var isLocationLocked: Boolean
-    get() = preferences.getBoolean(LOCATION_LOCKED, false)
+  /** Whether location lock is enabled or not. */
+  var isLocationLockEnabled: Boolean
+    get() = preferences.getBoolean(LOCATION_LOCK_ENABLED, false)
     set(value) {
-      preferences.edit().putBoolean(LOCATION_LOCKED, value).apply()
+      preferences.edit().putBoolean(LOCATION_LOCK_ENABLED, value).apply()
     }
 
   /** Terms of service acceptance state for the currently signed in user. */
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
```diff
@@ -45,6 +45,13 @@ class LocalValueStore @Inject constructor(private val preferences: SharedPrefere
       preferences.edit().putInt(MAP_TYPE, type).apply()
     }
 
+  /** State of the gps lock. */
```
gino-m, line 48: Please replace "GPS" with "location" here and throughout, since Android fused location service derives location from other sources in addition to GPS.

ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
```diff
@@ -45,6 +45,13 @@ class LocalValueStore @Inject constructor(private val preferences: SharedPrefere
...
+  var isGpsLocked: Boolean
```
gino-m, line 49: isLocationLockEnabled

gino-m (changes requested):
One bit, otherwise LG.

ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/persistence/local/LocalValueStore.kt
```diff
@@ -45,11 +45,11 @@ class LocalValueStore @Inject constructor(private val preferences: SharedPrefere
       preferences.edit().putInt(MAP_TYPE, type).apply()
     }
 
-  /** State of the gps lock. */
-  var isGpsLocked: Boolean
-    get() = preferences.getBoolean(GPS_LOCKED, false)
+  /** State of the location lock. */
+  var isLocationLocked: Boolean
```
gino-m, line 49: "Location locked" implies the location cannot change; we're not locking the location, we're enabled the "location lock" feature, thus "location lock enabled/disabled" makes more sense here.

shobhitagarwal1612: Done

gino-m (approved):
🔒 📌 !

PR #1413, opened 2023-01-11
refactor: add map feature abstraction

scolsen:
Adds a new class, Feature, that represents rendered model objects on the map. Features accept a tag type that may hold arbitrary data associated with a model object, callers should check the type of the tag to determine how to appropriately interpret the map feature.

This CL also removes uses of MapLocationOfInterest in favor of Feature, except in the PolygonDrawing components, which will be updated in a separate PR.

@gino-m @shobhitagarwal1612


ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
```diff
@@ -0,0 +1,32 @@
...
+package com.google.android.ground.ui.map
+
+import com.google.android.ground.model.geometry.Geometry
+
+/** Represents an individual feature on a map with a given [Geometry]. */
+data class Feature(val tag: Tag, val geometry: Geometry) {
+  /** Denotes the kind of entity this map feature represents. */
+  sealed class Tag(open val id: String)
+  /** A [Feature.Tag] that indicates this map feature represents a location of interest. */
+  data class LocationOfInterestTag(
```
gino-m, line 25: Is it possible to abstract away LOIs completely from the map UI layer? Imo the map adapter should deal exclusively with map concepts like features, geometries, and UI events.

scolsen: We could just do `Feature` w/ `id, title, caption`... -- we'd presumably still need *some* kind of tag to indicate what kind of item it is. Note that we don't actually refer to LOIs directly at all outside of using the name in this tag type. (and the VM which converts LOis to Features)

scolsen: Though, I guess it's a YAGNI -- at the moment we really *only* have LOI features so we could just make this a data class with a few fields

gino-m: There are a few geometries besides LOIs that the map also needs to render in MVP:

- Point and polygon geometries submitted as part of data collection (Submissions)
- A LineString with handles on vertices representing the interim state while drawing polygons
- A pin at the current location when dropping a pin

In the future there will probably be more, but YAGNI for those.

scolsen: do we think something like:

```kotlin
data class Feature(val metadata: Metadata, val geometry: Geometry) {
  data class Metadata(
    val identifier: String
    val title: String?
    val modificationInfo: AuditInfo 
  )
}
```

Is generic enough to capture all of those cases? Otherwise we could also make the class generic and take various bundles for different types

gino-m: What does the UI layer need this Metadata for?  Iiuc a unique key (id and type), geometry, and some styling info should be sufficient.

scolsen: Ah you're right, I thought we needed the `title` and `modificationInfo` bits for cluster items but they are optional -- and we're just rendering a custom icon anyway. -- ok lemme refactor this a bit

scolsen:
@gino-m ready for round 2

scolsen:
er, ktfm strikes again

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -60,14 +64,30 @@ internal constructor(
...
 
   fun openNavDrawer() {
     openDrawerRequests.onNext(Nil.NIL)
   }
 
-  fun onMarkerClick(mapLocationOfInterest: MapLocationOfInterest) {
-    showBottomSheet(mapLocationOfInterest.locationOfInterest)
-  }
+  /** Intended for use as a callback for handling user clicks on rendered map features. */
+  fun onMarkerClick(feature: Feature) =
```
shobhitagarwal1612, line 78: nit: Let's name it in a more generic way since this would handle not just markers

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -97,23 +117,41 @@ internal constructor(
     navigator.navigate(HomeScreenFragmentDirections.actionHomeScreenFragmentToSettingsActivity())
   }
 
-  fun onLocationOfInterestClick(mapLocationsOfInterest: ImmutableList<MapLocationOfInterest>) {
+  fun onLocationOfInterestClick(features: ImmutableList<Feature>) {
```
shobhitagarwal1612, line 120: Just by the name, this method seems very similar to `onMarkerClick`. Is it possible to merge the two?

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
shobhitagarwal1612: Also, do we actually need this method? In the mocks, only the bottom LOI cards were clickable. @gino-m

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
scolsen: yes, I think merging these is a good idea -- let's bump it to a separate PR so that it's easier to validate behaviors. I'll open a follow-up after this one.

shobhitagarwal1612 (approved):
Please file follow up bugs if you are planning to resolve them separately

PR #1420, opened 2023-01-13
[Code Health] Remove MapLocationOfInterest

scolsen:
Remove the `MapLocationOfInterest` class. We now use map.Features to represent objects on the map instead. We no
longer need this class. I've updated the final users (polygon drawing) to use the new Feature class instead.


shobhitagarwal1612 (approved):
Lgtm

PR #1423, opened 2023-01-15
[Draw polygon] Render polygon

shobhitagarwal1612:
Towards #1351 

Only renders the polygon vertices as per figma mocks. 

@scolsen  PTAL?


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/PolygonDrawingViewModel.kt
```diff
@@ -154,43 +156,33 @@ class PolygonDrawingViewModel @Inject internal constructor(resources: Resources)
...
-
-      private fun createDrawingState(state: State, polygon: Polygon?): PolygonDrawingState {
-        return AutoValue_PolygonDrawingViewModel_PolygonDrawingState(state, polygon)
-      }
+      fun completed(polygon: Polygon): PolygonDrawingState =
+        PolygonDrawingState(State.COMPLETED, polygon)
     }
   }
 
   /** Returns a set of [Geometry] to be drawn on map for the given [Polygon]. */
```
scolsen, line 169: Update doc -- `Geometry -> Feature`

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/PolygonDrawingViewModel.kt
```diff
@@ -154,43 +156,33 @@ class PolygonDrawingViewModel @Inject internal constructor(resources: Resources)
...
       return ImmutableSet.of()
     }
 
-    // Include the given polygon and add 1 LOI with a Point for each of its vertex.
-    return ImmutableSet.builder<Geometry>().add(polygon).addAll(vertices.toList()).build()
+    return ImmutableSet.builder<Feature>()
+      .add(
+        Feature(
+          id = uuidGenerator.generateUuid(),
+          tag = Feature.Type.LOCATION_OF_INTEREST,
```
scolsen, line 181: Should we use a different tag type for these since they aren't really LOIs? (We are writing these to submission data right? They wouldn't end up in the LOI table, which are only locations specified by the survey owners)

shobhitagarwal1612: From what I understand, the renderer would remain the same for LOI or data collection feature. What are we planning to use it for?

gino-m: @scolsen I think what you're calling `tag` here is really the namespace for the `id`, and the id and namespace together are what Maps SDK call a "tag". Please cmiiw.

@shobhitagarwal1612 It's used to identify the target of click and other events. Maps SDK would returns this as a Tag object, which the application layer can then use to understand if the feature refers to an LOI, user data, KML overlay, or something else.

gino-m: @scolsen Perhaps id and namespace should be grouped together as a separate FeatureTag class, since they can be returned without the geometry?

Also, if we have `Feature.Type.LOCATION_OF_INTEREST` in the view layer we're again breaking the abstraction that the map doesn't know anything about the actual different types of geometries there can be. I'd suggest defining the namespace as a `String`, and moving the enum to the view layer.

scolsen: yes, @shobhitagarwal1612, as @gino-m is suggesting the `tag` is what tells us what DB table to look at. in the map we only have opaque geometries and ID values. we could interpret these as any model object in theory, so we need to carry some information that tells us what kind of ID we're looking at/representing on the map (does this geometry represent an LOI or some other kind of entity, like user submission data that we decided we should render?)

@gino-m sgtm

shobhitagarwal1612: As discussed, will take the refactoring in a followup PR.

gino-m (approved):

Issue #1373, opened 2022-12-27
[Survey sync] Implement survey load and background sync


gino-m:
As mentioned in #988, the current survey and LOI sync is bound to the view layer, and relies on Firestore SDK to report changes over time. 

Instead, I propose we move towards a more robust design:

1. User selects survey from survey list for the first time.
2. Survey and LOIs are loaded and stored in the local db.
3. App subscribes to [FCM](https://firebase.google.com/products/cloud-messaging) topic associated with survey ID.
4. [Firestore triggers](https://firebase.google.com/docs/functions/firestore-events) on survey and LOI write send a collapsible message (no data payload) to the related topic.
5. A custom messaging service in the app receives the ping and schedules a background worker to do a sync latest survey and LOIs to local db as soon as possible.
6. Background worker syncs latest survey and LOIs to local db.
8. When users reopen an already synced survey, remote fetch is not required since it's already done in the background.

@shobhitagarwal1612 Wdyt?

Background sync can initially refetch the survey and all LOIs. Production impl should also implement the following:

* Implement soft-deletion of LOIs (via web):
  * Mark with `state: 'deleted'` instead of deleting
  * Update `lastModified` audit info on delete
  * Periodically clean up deleted LOIs
* On sync, only fetch LOIs with `lastModified.serverTimestamp` > server timestamp at last sync. Returned LOIs marked `state: 'deleted'` should be removed from local db.
* If app has been inactive for longer than LOI TTY, IDs of all remote LOIs in survey should be fetched and missing ones deleted from local db.

@os-micmec Wdyt?

shobhitagarwal1612:
This trigger function would be written as a cloud function, right?

gino-m:
> This trigger function would be written as a cloud function, right?

Correct. WIP prototype in https://github.com/google/ground-platform/pull/1020/files.

gino-m:
Prototype is working, submitting actual implementation in parts as follows:

- [ ] Add worker to sync survey definitions (jobs, tasks, etc)
- [ ] Add service to enqueue work on new messages
- [ ] Retry sync on errors
- [ ] Add ability to sync LOIs
- [ ] Handle deletion of LOIs on remote server
- [ ] Handle deletion of survey on remote server

PR #1408, opened 2023-01-10
[Sync] Enqueue background survey sync on messages from FCM

gino-m:
Towards #1373. 

Depends on #1360.

PR #1428, opened 2023-01-17
[Load survey] Clean up survey activation flow using Coroutines

### LINKED ISSUE #1426: [Code health] Simplify survey activation process
Trigger  load/sync survey flow on demand rather than subscribing to `selectSurveyEvent` in `SurveyRepository` init block. 

The `Loadable` stream (`loadingOnceAndWrap`) can also be created in the caller rather than the repository.

gino-m:
Introduces coroutines for survey activation to make flow more readable (modulo interop cruft).

Fixes #1426
Towards #1373

Next steps:
* Fetch LOIs once when survey is first activated instead of subscribing in `MainViewModel.onUserSignedIn()`. (TBD)
* Sync survey and LOIs in background worker on change (#1429)


ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -79,29 +82,32 @@ constructor(
...
+    var survey = surveyStore.getSurveyById(surveyId).awaitSingleOrNull()
+    if (survey == null) {
+      survey = syncSurveyFromRemote(surveyId)
+    }
+    this.lastActiveSurveyId = surveyId
+    return attachJobPermissions(survey)
+  }
+
+  private suspend fun syncSurveyFromRemote(surveyId: String): Survey {
+    remoteDataStore.subscribeToSurveyUpdates(surveyId).await()
```
shobhitagarwal1612, line 109: Why aren't we subscribing to updates when the survey exists locally?

gino-m: Check out the Javadoc for `subscribeToSurveyUpdates` (with some corrections):

```
  /**
   * Listens for remote changes to the survey with the specified id. Implementations should handle
   * synchronization of new, changed, and deleted surveys and LOIs to the local db in the
   * background.
   */
  fun subscribeToSurveyUpdates(surveyId: String): @Cold Completable
```

The impl uses Firebase Cloud Messaging to listen for pings to reload the survey and LOIs when they're updated. Implemented in #1408.

ground/src/main/java/com/google/android/ground/coroutines/CoroutineDispatchersModule.kt
```diff
@@ -0,0 +1,48 @@
+/*
```
scolsen, line 1: Do we really need this? Why wouldn't we just write:

```kotlin
someCoroutineScope.launch(Dispatchers.IO) { 
 //...
}

// and
suspend fun foo() {
  // ...
  return withContext(Dispatchers.IO) {
      // other stuff
  }
}
```

Having this explicitly written in source seems clearer to me than relying on injection. Injection makes the code harder to follow and now we have more dagger magic to maintain in the codebase. While this gives us a little bit more modularity and flexibility, we're tied to the Android platform anyway and I think it might be a premature architectural optimization. I'd prefer to go with less code in this case and use the platform directly.

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -127,11 +98,34 @@ constructor(
...
+      return
+    }
+    // Clear survey if id is empty.
+    if (surveyId.isEmpty()) {
+      clearActiveSurvey()
+      return
+    }
 
-  fun clearActiveSurvey() = selectSurveyEvent.onNext("")
+    externalScope.launch {
```
scolsen, line 114: Pass the `Dispatcher` to `launch` directly? 

`launch(ioDispatcher) { ... }`

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -54,63 +59,29 @@ class SurveyRepository
 constructor(
   private val localDataStore: LocalDataStore,
   private val remoteDataStore: RemoteDataStore,
-  private val localValueStore: LocalValueStore
+  private val localValueStore: LocalValueStore,
+  @ApplicationScope private val externalScope: CoroutineScope,
```
scolsen, line 63: Here too, I personally find the use of injection more confusing and less readable than just using the platform and standard library directly.

Here, I need to recall what `externalScope` means in terms of CoroutineScopes, contrarily, I think just using Android's scopes directly is a lot clearer (https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope)

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
scolsen, line 114: I'm not convinced we should be launching anything in a repository. Shouldn't we write these as suspend functions an `launch` from a component that actually has a lifecycle? (viewmodel that calls the repo methods)

ground/src/main/java/com/google/android/ground/coroutines/CoroutineDispatchersModule.kt
gino-m: Yes! Please check out https://developer.android.com/kotlin/coroutines/coroutines-best-practices for motivations. tldr, we need to be able to replace these for tests.

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
gino-m: Same comment as above. Relevant note from the best practices guide:
image

ground/src/main/java/com/google/android/ground/coroutines/CoroutineDispatchersModule.kt
scolsen: Ah! Ok! sgtm then. 

> Note: While you might have seen hardcoded dispatchers in code snippets across this site and our codelabs, this is only to keep the sample code concise and simple. In your application, you should inject dispatchers.

Gotta love it.

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
scolsen: Yep, saw it. Like that they have more examples of the non-best practice in their docs and then have single, easy to miss note explaining that they are in fact not recommended practices 🙃

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
scolsen: Got it, that makes sense. Could we add some kotlindocs on the ApplicationScope explaining this? That its purpose is to launch coroutines that should have the same lifecycle as the app process?

gino-m: I know.. I have to say, coroutines and Hilt are a huge improvement , but there's still some complexity that needs to be managed.

Done!

gino-m:
@shobhitagarwal1612 @scolsen RTAL!

shobhitagarwal1612:
The checks are still failing. Is it ready to be reviewed?

gino-m:
> The checks are still failing. Is it ready to be reviewed?

Sorry about that, tests fixed. Ready for review!

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
```diff
@@ -16,46 +16,136 @@
...
+import org.mockito.Mockito.*
+import org.mockito.kotlin.any
 import org.robolectric.RobolectricTestRunner
 
+@OptIn(ExperimentalCoroutinesApi::class)
 @HiltAndroidTest
 @UninstallModules(LocalDataStoreModule::class)
 @RunWith(RobolectricTestRunner::class)
 class SurveyRepositoryTest : BaseHiltTest() {
   @BindValue @InjectMocks var mockLocalDataStore: LocalDataStore = RoomLocalDataStore()
```
shobhitagarwal1612, line 57: Can this be cleaned up in favor of mockSurveyStore already present in this test?

shobhitagarwal1612 (approved):
One minor nit. Thanks!

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
gino-m: Good catch, thanks. Fixed!

gino-m: Test broke after this change... I'll take a look.

shobhitagarwal1612: Oops, my bad. Please feel free to leave a todo to unblock the PR

gino-m: Unfortunately making that change will require additional refactors, since `LocalSurveyStore` is injected into both `SurveyRepository` and `LocalDataStore`. I'll punt on this for now, thanks! @scolsen FYI.

shobhitagarwal1612: :+1:

PR #1435, opened 2023-01-18
Fix submission data not uploading to firestore

shobhitagarwal1612:
@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -151,7 +151,9 @@ internal constructor(
           responses.forEach { (task, taskData) ->
             taskDataDeltas.add(TaskDataDelta(task.id, task.type, Optional.ofNullable(taskData)))
           }
-          submissionRepository.createOrUpdateSubmission(it, taskDataDeltas.build(), isNew = true)
+          submissionRepository
```
gino-m, line 154: If the returned subscription is disposed of before the update completes, won't the update process be interrupted? Can we instead introduce an `externalScope` as defined [here](https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad), and call `submissionRepository.createOrUpdateSubmission(..).blockingFoo()` from a coroutine?

It's also not clear why this method returns a Single at all. Could this be simplified?

shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m PTAL.

gino-m (changes requested):
LG, just one nit. Thanks!

gino-m (approved):
LGTM!

PR #1441, opened 2023-01-19
feat: job counts on markers

scolsen:
- Updates the feature map object to hold an addition boolean, bundles tag data
- display job data collection counts on cluster markers.

Screenshot 2023-01-19 at 4 22 34 PM
Screenshot 2023-01-19 at 4 22 47 PM


scolsen:
@gino-m @shobhitagarwal1612

gino-m:
@scolsen Nicely done! One q - why do we see pins and clusters at the same zoom level in one of the images above? Were those markers intentionally marked as "do not cluster" to test?

scolsen:
> @scolsen Nicely done! One q - why do we see both pins and clusters at the same zoom level in one of the images above? Were those markers intentionally marked as "do not cluster" to test?

The second image is the cluster of 8 split up as we increase the zoom level. We're using the default clustering algorithm provided by the clustering lib right know. I believe it is a mixture of zoom threshold + distance, but I'd have to double check. Basically, I think it looks at the current zoom level then looks at the distance between markers. It generally seems to cluster markers from greater distances at lower zoom levels (max zoom out), while markers that have lower distances between each other are popped out as the zoom level increases. In the second image you're seeing some markers that were unclustered (maybe they're absolute lat lngs are ~10 units apart or so) while markers that are closer to each other in terms of absolute position are still clustered at that zoom threshold (e.g. maybe their absolute lat lngs are only 1 unit apart or they overlap).

Finally, at a high enough zoom the markers won't cluster at all. I can throw in a gif if that helps

scolsen:
https://user-images.githubusercontent.com/11237600/213797105-fde48113-3689-4c2f-82c6-e736285f8dc3.mov

@gino-m here's a short recording that illustrates the clustering behavior a bit better -- we can change this if we want, we'll just have to write our own clustering algorithm.

ground/src/main/java/com/google/android/ground/model/job/Job.kt
```diff
@@ -32,4 +32,6 @@ constructor(
     get() = tasks.values.sortedBy { it.id }.toImmutableList()
 
   fun getTask(id: String): Optional<Task> = Optional.ofNullable(tasks[id])
+
+  fun hasData(): Boolean = !this.tasks.isEmpty()
```
shobhitagarwal1612, line 36: nit: `this` keyword isn't necessary

ground/src/main/java/com/google/android/ground/ui/MarkerIconFactory.kt
```diff
@@ -69,4 +67,38 @@ class MarkerIconFactory @Inject constructor(@ApplicationContext private val cont
     // TODO: Cache rendered bitmaps.
     return BitmapDescriptorFactory.fromBitmap(bitmap)
   }
+
+  fun getClusterIcon(
```
shobhitagarwal1612, line 71: This method has a lot of common code with `getMarkerBitmap()` method. Can we please refactor it?

ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
```diff
@@ -17,11 +17,28 @@ package com.google.android.ground.ui.map
...
-  /** Denotes the kind of entity this map feature represents. */
+/** Represents an individual feature on a map with a given [Geometry] and [Tag]. */
+data class Feature(val tag: Tag, val geometry: Geometry) {
+  /**
+   * Denotes the kind of entity this map feature represents and contains any additional data it
+   * carries.
+   */
+  data class Tag(
+    /** A unique identifier for the model object that this feature represents. */
+    val id: String = "",
```
shobhitagarwal1612, line 28: Why does it have a default value? Shouldn't this always be non-null & non-empty?

ground/src/main/java/com/google/android/ground/ui/map/gms/FeatureClusterRenderer.kt
```diff
@@ -48,10 +50,42 @@ class FeatureClusterRenderer(
...
+      )
+    return icon
+  }
+
+  override fun onBeforeClusterRendered(
+    cluster: Cluster<FeatureClusterItem>,
+    markerOptions: MarkerOptions
+  ) {
+    super.onBeforeClusterRendered(cluster, markerOptions)
+
```
shobhitagarwal1612, line 83: nit: extra line

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DropAPinTaskViewModel.kt
```diff
@@ -41,5 +41,5 @@ constructor(resources: Resources, private val uuidGenerator: OfflineUuidGenerato
   }
 
   private fun createFeature(point: Point): Feature =
-    Feature(uuidGenerator.generateUuid(), Feature.Type.LOCATION_OF_INTEREST, point)
+    Feature(Feature.Tag(uuidGenerator.generateUuid(), Feature.Type.LOCATION_FEATURE), point)
```
gino-m, line 44: Why do map features have different IDs than the point LOIs they represent? In that case we'd need to maintain a mapping between the two, which should be unnecessary.

ground/src/main/java/com/google/android/ground/ui/datacollection/PolygonDrawingViewModel.kt
```diff
@@ -176,8 +176,7 @@ internal constructor(private val uuidGenerator: OfflineUuidGenerator, resources:
     return ImmutableSet.builder<Feature>()
       .add(
         Feature(
-          id = uuidGenerator.generateUuid(),
-          tag = Feature.Type.LOCATION_OF_INTEREST,
+          Feature.Tag(id = uuidGenerator.generateUuid(), type = Feature.Type.LOCATION_FEATURE),
```
gino-m, line 179: Same comment for polygons.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -105,7 +105,7 @@ internal constructor(
   /** Intended for use as a callback for handling user clicks on rendered map features. */
   fun onFeatureClick(features: ImmutableList<Feature>) {
     val loiFeatureIds =
```
gino-m, line 107: Same here.. why not just loiIds?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -99,7 +99,12 @@ internal constructor(
   ): ImmutableSet<Feature> {
     // TODO: Add support for polylines similar to mapPins.
     return locationsOfInterest
-      .map { Feature(id = it.id, tag = Feature.Type.LOCATION_OF_INTEREST, geometry = it.geometry) }
+      .map {
+        Feature(
+          Feature.Tag(id = it.id, type = Feature.Type.LOCATION_FEATURE, flag = it.job.hasData()),
```
gino-m, line 104: Can we roll the two tag args into the constructor of Feature for convenience?

ground/src/main/java/com/google/android/ground/ui/datacollection/DropAPinTaskViewModel.kt
scolsen: Generally they don't -- this code is in the DropAPin task, so I think @shobhitagarwal1612 is generating a new feature that represents the pin, not an LOI. Likely this should have a different Feature Type to indicate as much. The code only made it into this PR due to the Feature LOI type name refactor.

ground/src/main/java/com/google/android/ground/ui/datacollection/PolygonDrawingViewModel.kt
scolsen: Yes, I think the problem here is the Feature Type being used.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
scolsen: At that point I would wonder why we need the additional class wrapping them in the first place and why we wouldn't just store these as direct properties of `Feature`. Rolling them into the primary ctor would make it awkward to do things like:

```kt
val tag = FeatureTag(...)
val feat = Feature(tag, ...)
```

since instead you'd have to do this:

```kt
val tag = FeatureTag(...)
val feat = Feature(tag.x, tag.y, ...)
```

which would seem to make the `Tag` superfluous.

Maybe bundling them up was a premature optimization? (if we go back on this again this will be the second or third time we've changed our minds about this -- the original impl used this approach, I unrolled the fields into the Feature, then you suggested rebundling them into a Tag...now we're considering unrolling them again....)

tbh all we need probably is one flat data class. there's not a huge benefit to having a wrapping class IMO unless we think we have a need to manipulate tags separately from Features.

ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
scolsen: My thinking was that there may have been cases in which we don't need an ID right away (e.g. user-drawn polygons), but you're right that we probably do just need one in all cases.

ground/src/main/java/com/google/android/ground/ui/MarkerIconFactory.kt
scolsen: Refactored out the shared bitmap generation bits.

gino-m:
> > @scolsen Nicely done! One q - why do we see both pins and clusters at the same zoom level in one of the images above? Were those markers intentionally marked as "do not cluster" to test?
> 
> The second image is the cluster of 8 split up as we increase the zoom level. We're using the default clustering algorithm provided by the clustering lib right know. I believe it is a mixture of zoom threshold + distance, but I'd have to double check. Basically, I think it looks at the current zoom level then looks at the distance between markers. It generally seems to cluster markers from greater distances at lower zoom levels (max zoom out), while markers that have lower distances between each other are popped out as the zoom level increases. In the second image you're seeing some markers that were unclustered (maybe they're absolute lat lngs are ~10 units apart or so) while markers that are closer to each other in terms of absolute position are still clustered at that zoom threshold (e.g. maybe their absolute lat lngs are only 1 unit apart or they overlap).
> 
> Finally, at a high enough zoom the markers won't cluster at all. I can throw in a gif if that helps

I see.. my understanding was that we would always cluster when below some particular zoom level. This is especially helpful when viewing polygons, which will be small to see at lower zoom levels, but also for consistency across the map. Is it possible to tweak this behavior? Perhaps in a follow-up PR if it's non-trivial?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: Tag is still needed as an object to be returned by events; there's no need to fetch the geometry from the map and convert it back to a `Feature`.

Expanding the tag args in Feature constructor was suggested as syntactic sugar to aid readability.

ground/src/main/java/com/google/android/ground/ui/datacollection/DropAPinTaskViewModel.kt
gino-m: I see.. In that case shouldn't the pin ID be fixed? Definitely shouldn't be marked as an LOI if it's being used as a placeholder/cursor for the task UI.

scolsen: Yep, I updated the types for both this and polygon drawing in the latest commit

gino-m:
> > > > @scolsen Nicely done! One q - why do we see both pins and clusters at the same zoom level in one of the images above? Were those markers intentionally marked as "do not cluster" to test?
> > > 
> > > 
> > > The second image is the cluster of 8 split up as we increase the zoom level. We're using the default clustering algorithm provided by the clustering lib right know. I believe it is a mixture of zoom threshold + distance, but I'd have to double check. Basically, I think it looks at the current zoom level then looks at the distance between markers. It generally seems to cluster markers from greater distances at lower zoom levels (max zoom out), while markers that have lower distances between each other are popped out as the zoom level increases. In the second image you're seeing some markers that were unclustered (maybe they're absolute lat lngs are ~10 units apart or so) while markers that are closer to each other in terms of absolute position are still clustered at that zoom threshold (e.g. maybe their absolute lat lngs are only 1 unit apart or they overlap).
> > > Finally, at a high enough zoom the markers won't cluster at all. I can throw in a gif if that helps
> > 
> > 
> > I see.. my understanding was that we would always cluster when below some particular zoom level. This is especially helpful when viewing polygons, which will be small to see at lower zoom levels, but also for consistency across the map. Is it possible to tweak this behavior? Perhaps in a follow-up PR if it's non-trivial?
> 
> yep, we can definitely tweak it -- I think it's pretty flexible. We can implement custom logic, but hopefully we'll be able to just set a few parameters to get the job done. Let's do it in a separate PR

SGTM! Filed #1444 to track.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
scolsen: Gotcha, sg. I'll add a second ctor for convenience

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/MarkerIconFactory.kt
```diff
@@ -30,30 +31,43 @@ import javax.inject.Singleton
...
+  private fun createBitmap(
+    drawable: Drawable,
+    zoomLevel: Float,
+    isSelected: Boolean = false
+  ): Bitmap {
     // TODO: Adjust size based on selection state.
     var scale = ResourcesCompat.getFloat(context.resources, R.dimen.marker_bitmap_default_scale)
-    if (currentZoomLevel >= HomeScreenMapContainerViewModel.ZOOM_LEVEL_THRESHOLD) {
+
+    if (zoomLevel >= HomeScreenMapContainerViewModel.ZOOM_LEVEL_THRESHOLD) {
```
gino-m, line 43: Can we move this constant into this file or pass it in as an arg to avoid circular references and enforce separation of concerns here? 

Also could you please add a comment somewhere explaining what this does?

ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
```diff
@@ -38,7 +45,14 @@ data class Feature(val tag: Tag, val geometry: Geometry) {
   enum class Type {
     /** This feature represents an unknown kind of object. */
     UNKNOWN,
-    /** This feature represents a single, discreet location on the map. */
+    /**
```
gino-m, line 48: Does `Feature` need to have a reference to the domain model at all? (e.g., location vs user polygon)? Can we abstract this to main better SoC between the map layer and business logic?

gino-m (changes requested):
Small design suggestions, otherwise LG! :100:

ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
gino-m: @scolsen Gentle ping!

ground/src/main/java/com/google/android/ground/ui/MarkerIconFactory.kt
scolsen: Done and done.

ground/src/main/java/com/google/android/ground/ui/map/Feature.kt
gino-m: `type` could be an `Int` or a `String` which would only have meaning for the caller. Alternatively, callers could prepend a namespace to the string `id`, though this is less desirable. Since the map layer doesn't know about LOIs or temporary map features for drawing tools, it would be up to the business logic and calling UI logic to set and interpret these fields. Does that make sense?

scolsen: Gotcha, yes I that makes sense -- so we'll just set it as `Int` and let callers decided the interpretation?

Only slight downside I see to that is that it does allow us to accidentally write code that misinterprets Features. For example, we could unintentionally pass a Feature representing a user-drawn polygon to LOI processing code and re-interpret it as an LOI if they happen to use the same convention (e.g. interpret 0 as my model type).

On the plus side this would give us much better separation between the layers.

gino-m: If can assume there will only be a single enum on the callers side with int values, the values will be compile checked. If we wanted we could provide the enum class as a generic to the map adapter, but this is likely overkill.

scolsen: Ahh, I see what you mean now -- so we'd have a single model-layer enum for distinguishing these. sgtm

scolsen: @gino-m is it closer to what you were thinking now?

ground/src/main/java/com/google/android/ground/model/geometry/ModelTypeTag.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.geometry
+
+/**
+ * Indicates the type of a geometric model object. Used to interpret other objects (e.g. map
+ * features) back into model objects.
+ */
+enum class ModelTypeTag {
```
gino-m, line 22: Nice! Should this class maybe live in the ui layer, since it's specific to UI interactions?

Maybe `FeatureType` would make sense here? Note that tag := id + type.

gino-m (approved):
Yes, that's it! One small suggestion, otherwise LG!

ground/src/main/java/com/google/android/ground/model/geometry/ModelTypeTag.kt
scolsen: done and done!

gino-m (approved):
🗺️ 🌐 ⭐

Issue #1449, opened 2023-01-25
[Data collection] Show camera inline when taking photo


JSunde:
Currently the Photo Capture task intents to the camera app for the user to capture a photo, but the mocks show an embedded camera fragment allowing the user to capture a photo without leaving the app.

I'm thinking this should be p2 and not block the current prototype. What are your thoughts @gino-m?

shobhitagarwal1612:
The latest mocks do not contain inline camera widget. Can we close this?

PR #1465, opened 2023-01-26
Draft change attempting to persist Data Collection Fragment position

JSunde:
Towards #1449


JSunde:
I think I got the SavedStateHandle set up properly to persist the LiveData storing the current position in the Data Collection Flow, but for some reason when navigating back after capturing a photo for the Photo Task, the DataCollectionViewPagerAdapter is still being passed 0, causing the DataCollectionFragment to navigate back to the first task in the flow.

Any ideas @shobhitagarwal1612, @gino-m?

JSunde:
> I think I got the SavedStateHandle set up properly to persist the LiveData storing the current position in the Data Collection Flow, but for some reason when navigating back after capturing a photo for the Photo Task, the DataCollectionViewPagerAdapter is still being passed 0, causing the DataCollectionFragment to navigate back to the first task in the flow.
> 
> Any ideas @shobhitagarwal1612, @gino-m?

I ended up figuring out the issue. It seems that the DataCollectionViewPagerAdapter was being recreated on resume due to the submission observer. I was able to resolve the issue by only creating a new DataCollectionViewPagerAdapter if there was no adapter set or if the tasklist was changed.

JSunde:
Ready for review, PTAL @gino-m, @shobhitagarwal1612

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -39,15 +40,20 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
...
+    dataCollectionViewModelFactory.create(savedStateHandle)
+  }
+
   private lateinit var viewPager: ViewPager2
 
-  override fun onCreate(savedInstanceState: Bundle?) {
-    super.onCreate(savedInstanceState)
-    viewModel = getViewModel(DataCollectionViewModel::class.java)
-  }
+  //  override fun onResume() {
```
gino-m, line 52: Nit: Remove commented out code.

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionViewModelTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+class DataCollectionViewModelTest : BaseHiltTest() {
+  @Inject lateinit var dataCollectionViewModelFactory: DataCollectionViewModel.Factory
+
+  @Test
+  fun savedState_containsCurrentPosition_returnsCorrectCurrentPosition() {
```
gino-m, line 33: Can you add tests for the null case and for the case when the task list changes?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
JSunde: Oops 😅 

Removed!

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionViewModelTest.kt
JSunde: Hmm... do you have suggestions for how to emit a new value for the submission in the test to trigger the DataCollectionViewPager recreation?

gino-m: Ah nvm, that happens in the fragment, not the vm. We should eventually add instrumented tests to cover that case.

gino-m (approved):

PR #1469, opened 2023-01-28
[Code health] Fix ktfmt style

gino-m:
Removes "Google (internal)" style so that Android Studio ktfmt behavior matches Gradle ktfmt `googleStyle()`.

@shobhitagarwal1612 PTAL?

@scolsen @JSunde  FYI..


shobhitagarwal1612:
After this change, behavior of `ktfmtFormat` and `ktfmtCheck` remains the same. But, AS code formatter changes the indent size from 2 to 4 which then fails `ktfmtCheck`.

gino-m:
> After this change, behaviour of gradle tasks `ktfmtFormat` and `ktfmtCheck` remain the same. But, AndroidStudio's code formatter now changes the code indentation size from 2 to 4 which then fails `ktfmtCheck`.

Hmm.. not sure why I didn't catch this. Ok, closing this PR, thanks!

PR #1480, opened 2023-01-30
Switch from Cbor serialization to ProtoBuf serialization

### LINKED ISSUE #1461: [Offline support] Verify that Kotlin binary serialization allows for additive changes to classes
@JSunde Are we able to add new fields? I assume they'd need to be nullable to be able to support objects persisted with older versions of the app? Any other constraints to be aware of?

JSunde:
Fixes #1461

@gino-m PTAL?


gino-m (approved):
Thanks! At some point I wonder if it makes sense to use protobufs directly across the stack.. that could eventually be returned by some backend service. Food for future thought.

PR #1481, opened 2023-01-30
Replace Point with Coordinate in MapFragment

### LINKED ISSUE #1467: [Code health] Replace `Point` with `Coordinate` in `MapFragment` where appropriate
_Points_ ≠ _Coordinates_:

_Point_ represents the geometry of an actual feature on the map. _Coordinate_ is simply a pair of X, Y coordinates. The difference is analogous to `Marker` and `LatLng` in Maps SDK.

@JSunde Could you PTAL when you have a sec?

JSunde:
Fixes #1467

@gino-m PTAL?


gino-m (approved):
Thanks for the cleanup!

PR #1484, opened 2023-01-31
[Code health] Clean up repositories

### LINKED ISSUE #1482: [Code health] Stop keeping state in repositories
States currently stored in repositories should be either made transient, or moved to a single source-of-truth in a persistent data store or in the UI layer:

- [ ] SurveyRepository

https://github.com/google/ground-android/blob/e8806ca772d20930b1f520641c897106fe21300f/ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt#L67-L79

- [ ] MapsRepository

https://github.com/google/ground-android/blob/e8806ca772d20930b1f520641c897106fe21300f/ground/src/main/java/com/google/android/ground/repository/MapsRepository.kt#L28

gino-m:
- [x] Use stores as single source of truth (fixes #1482)
- [x] Several renames to clarify scope
- [x] Remove deprecated survey load progress dialog
- [x] Other smal cleanups and refactors

@shobhitagarwal1612 PTAL?


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
```diff
@@ -269,20 +256,6 @@ class HomeScreenFragment :
     bottomSheetBehavior.state = BottomSheetBehavior.STATE_HIDDEN
   }
 
-  private fun showSurveyLoadingDialog() {
-    if (progressDialog == null) {
-      progressDialog = ProgressDialogs.modalSpinner(context, R.string.survey_loading_please_wait)
-      progressDialog?.show()
```
shobhitagarwal1612, line 275: Can we also cleanup the class variable?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -51,12 +52,16 @@ class HomeScreenMapContainerViewModel
...
-  private val surveyRepository: SurveyRepository,
+  private val mapStateRepository: MapStateRepository,
   private val locationOfInterestRepository: LocationOfInterestRepository,
   private val locationController: LocationController,
   private val mapController: MapController,
+  surveyRepository: SurveyRepository,
   offlineAreaRepository: OfflineAreaRepository
 ) : BaseMapViewModel(locationController, mapController) {
+
+  private var activeSurveyId: String = ""
```
shobhitagarwal1612, line 63: Why are we maintaining this state locally? Shouldn't we fetch the value from our source of truth (i.e. LocalValueStore)?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -194,8 +199,9 @@ internal constructor(
     // into the repo
     // LOIs that are persisted to the local and remote dbs.
     val loiStream =
-      surveyRepository.activeSurvey.switchMap { activeProject ->
-        getLocationsOfInterestStream(activeProject)
+      surveyRepository.activeSurvey.switchMap { survey ->
+        activeSurveyId = survey.map { it.id }.orElse("")
+        getLocationsOfInterestStream(survey)
```
shobhitagarwal1612, line 204: We already have a class which provides this stream (LoiCardSource). Maybe we can rename that to something more generic and reuse it here?

shobhitagarwal1612 (changes requested):
Some minor suggestions. Looks good otherwise!

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
gino-m: I meant to do this but forgot, thanks for the heads up!

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: UI state may lag behind the source of truth in the local datastore; in this case, we're using the UIs understanding of the current survey instead of the local stores since that is what the camera position is referring not. This race condition is unlikely cause issues in practice, but the UI self-reporting the context for camera moves seems to be more correct.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: This is going away in an upcoming PR; LOIs will be synced in a background worker instead. Added a TODO to clarify.

shobhitagarwal1612 (approved):

shobhitagarwal1612:
Some tests are still failing

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
shobhitagarwal1612: Thanks for clarifying.

shobhitagarwal1612 (approved):
LGTM!

PR #1485, opened 2023-01-31
Fix multiple choice task responses not being persisted properly

JSunde:
When the user completes the data collection flow, we get the TaskDatas from each of the TaskViewModels and persist them in the SubmissionRepository, however there was an issue with Multiple Choice tasks where the TaskData was null because we weren't updating the ViewModel when the user made a selection.


JSunde:
PTAL @gino-m

gino-m (approved):
Thanks for the fix

gino-m:
Needs ktfmt :/

Issue #1471, opened 2023-01-29
[Code health] Replace mocks with fakes


gino-m:
Specifically:

- [ ] LocalDataStore
- [ ] RemoteDataStore
- [ ] WorkManager

..others?

Related to #1470

PR #1498, opened 2023-02-03
Simplify SurveyRepository tests using in memory local db instead of mocks

gino-m:
Also pushes launch of `activateSurvey()` up to view models so that exceptions can be propagated and handled there.

Towards #1471.

@shobhitagarwal1612 Could you PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
@@ -36,7 +36,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.rx2.await
 import kotlinx.coroutines.rx2.awaitSingleOrNull
-import kotlinx.coroutines.withContext
 import timber.log.Timber
 
 private const val LOAD_REMOTE_SURVEY_TIMEOUT_SECS: Long = 15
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -109,15 +108,9 @@ constructor(
       return
     }
 
-    externalScope.launch {
```
shobhitagarwal1612, line 112: Please cleanup the injected dependency as well.

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
```diff
@@ -17,106 +17,88 @@ package com.google.android.ground.repository
...
   @Inject lateinit var localValueStore: LocalValueStore
-
   @DefaultDispatcher @Inject lateinit var testDispatcher: CoroutineDispatcher
 
   @Before
   override fun setUp() {
     super.setUp()
-    `when`(mockLocalSurveyStore.insertOrUpdateSurvey(any())).thenReturn(Completable.complete())
+    // Reset local test db between tests.
+    localDatabase.clearAllTables()
```
shobhitagarwal1612, line 56: Isn't this implicit?

shobhitagarwal1612 (changes requested):

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
gino-m: I suppose it is. Each test gets a different instance of `LocalDatabase`, which I assume has a different SQLite behind it. Thanks for pointing this out.

shobhitagarwal1612 (approved):

shobhitagarwal1612 (approved):
LGTM!

Issue #1490, opened 2023-02-02
[Survey selector] Show progress spinner while survey loads


gino-m:
`SurveySelectorViewModel.activateSurvey` navigates to the home screen without waiting for the survey to load. Instead, some a loading spinner should be shown while the survey and LOIs are loading. 

@vittorino For recommended UXD.

@shobhitagarwal1612 PTAL for eng?

gino-m:
Can be shown as modal dialog.

PR #1503, opened 2023-02-03
[Code health] Cleanup survey activation

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/1490

LocalValueStore already sets the last active survey id during init. So, there is no need to explicitly invoke survey activation in HomeScreenFragment anymore.

Verified by running the app locally with app data cleaned.

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -65,9 +65,11 @@ constructor(
       else getOfflineSurvey(it).map { s -> Optional.of(s) }
     }
 
-  var lastActiveSurveyId: String = ""
+  var lastActiveSurveyId: String
```
gino-m, line 68: While you're at it can you please rename this to `activeSurveyId`, since it refers to the current state, not previous one?

gino-m (changes requested):
Thanks for the cleanup! When does reactivation of the last survey occur? On the first subscriber to `activeSurvey`?

Would you be able to verify that the survey is not reloaded for each subscriber? If it is we may need to add `shareReplay()` to `activeSurvey`

shobhitagarwal1612:
"Survey activation" in terms of code refers to `activeSurveyIdFlowable` returning the last `survey_id` value (in case of first load) and latest value (whenever survey gets switched). 

Since the underlying implementation of this flowable is BehaviorProcessor, it automatically replays the last value when subscribed

gino-m (approved):

PR #1515, opened 2023-02-04
[Survey Selector] Add back arrow and overflow menu for deleting offline surveys

### LINKED ISSUE #504: [Survey selection] Rationalize and simplify "join survey" flow
Currently, we show a simple list of surveys shared with the user.

Is a browse experience the way we think users should discover and join surveys? 

Future entry points could include via a shared link, by entering a project ID and pin code, or by scanning a QR code. More work is needed here. @ngosue re our earlier discussion today.

shobhitagarwal1612:
Fixes #504

Things done:
 * Improve look by adjusting padding & margin 
 * Add a toolbar with a back arrow
 * Fix behavior of back press based on origin
 * Create a new overflow menu for deleting offline surveys

@gino-m @JSunde  PTAL?


ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -132,4 +132,12 @@ constructor(
   ): @Cold(terminates = false) Flowable<List<Mutation>> {
     return localDataStore.getMutationsOnceAndStream(survey)
   }
+
+  suspend fun deleteSurvey(surveyId: String) {
```
gino-m, line 136: Could we perhaps name this `removeOfflineSurvey()` or similar? 

From the repository client's standpoint, surveys are either available offline, or not, and the fact that the survey is deleted from the local db is an implementation detail.

Also, this would make it clear that we're not actually deleting the source-of-truth survey on the server.

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
```diff
@@ -80,4 +80,39 @@ class SurveyRepositoryTest : BaseHiltTest() {
...
+    runTest(testDispatcher) {
+      surveyStore.insertOrUpdateSurvey(SURVEY).await()
+      surveyRepository.activateSurvey(SURVEY.id)
+      advanceUntilIdle()
+
+      surveyRepository.deleteSurvey(SURVEY.id)
+      advanceUntilIdle()
+
+      // Verify survey is deleted
+      surveyStore.surveys.test().assertValues(listOf())
```
gino-m, line 95: Can we use public methods of `SurveyRepository` here and below to verify rather than relying on the backing implementation? It could make the tests more readable and less brittle.

gino-m (changes requested):
Thanks for the edits!

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
gino-m: @shobhitagarwal1612 Could you do this for the tests you added? No need to retroactively apply this comment.

shobhitagarwal1612: I applied the suggestion to this class. Did I miss any occurrence?

shobhitagarwal1612: Done

ground/src/test/java/com/google/android/ground/repository/SurveyRepositoryTest.kt
```diff
@@ -72,7 +70,9 @@ class SurveyRepositoryTest : BaseHiltTest() {
   @Test
   fun activateSurvey_alreadyAvailableOffline() =
     runTest(testDispatcher) {
-      surveyStore.insertOrUpdateSurvey(SURVEY).await()
+      fakeRemoteDataStore.surveys = listOf(SURVEY)
+      surveyRepository.syncSurveyWithRemote(SURVEY.id).await()
+      advanceUntilIdle()
```
gino-m, line 75: I think one `advanceUntilIdle()` is sufficient.

Would moving the setup to `@Before` break other tests?

shobhitagarwal1612: Yes, we don't want to sync everything in @Before. Some tests want to invoke that by activating the survey manually.

gino-m (approved):
Some tests are failing, otherwise changes LGTM.

Issue #1491, opened 2023-02-02
[Code health] Simplify `LocalDataStore`


gino-m:
Proposals:

- [x] Minor renames in https://github.com/google/ground-android/issues/1492
- [ ] Move mutation methods into new repository, `MutationRepository`
- [ ] Remove `LocalDataStore`, inject individual stores directly, simplifying how stores are injected.
- [ ] Remove `LocalStore` and `LocalMutationStore` base classes: stores don't always implement the same operations, and we may want to migrated methods to coroutines incrementally.
- [ ] Move `room` under `stores` since it's an impl of stores.
- [ ] Move classes in `room.stores` up one level to `room`
- [ ] Remove domain object name from methods since clear from context
- [ ] Split up `LocalDataStoreTests`
- [ ] Test via stores instead of daos in `LocalDataStoreTests`
- [ ] Stop returning `Entity` classes (eg  `LocationOfInterestMutationEntity` in `LocalLocationOfInterestMutationStore`
- [ ] Shorten names where appropriate (eg `LocalLocationOfInterestMutationStore.getLocationOfInterestMutationsByLocationOfInterestIdOnceAndStream`)

@scolsen Wdyt?

gino-m:
One challenge I encountered while doing the cleanup is that we uninstall the entire LocalDataStoreModule module in tests. Since we don't inject a fake for each individual datastore, we're required to mock any uses of injected substores. I'll do that temporarily if there aren't too many instances, then return to cleanup #1471 to simplify usage in tests.

gino-m:
Most work done in #1521. Closing this for now.

PR #1521, opened 2023-02-05
Refactor LocalDataStore

gino-m:
- [x] Move mutation coordination methods into `MutationRepository`
- [x] Inject local data stores directly instead of via `LocalDataStore`
- [x] Delete `LocalDataStore`
- [x] Preserve client timestamps passed through `LocationOfInterestMutation`

Towards #1491. See also #1559, #1561, #1562.

@shobhitagarwal1612 and/or @scolsen PTAL?


ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
```diff
@@ -66,7 +66,7 @@ data class LocationOfInterest(
       surveyId = surveyId,
       locationOfInterestId = id,
       userId = userId,
-      clientTimestamp = Date(),
+      clientTimestamp = lastModified.clientTimestamp,
```
shobhitagarwal1612, line 69: Do we update `lastModified` before creating a mutation? I thought the purpose of using `Date()` was to ensure that the timestamp always matches when the mutation is created. Can you please check?

shobhitagarwal1612 (approved):
Approving the PR to unblock but please address the above comment before merging. Thanks!

ground/src/main/java/com/google/android/ground/model/locationofinterest/LocationOfInterest.kt
gino-m: I don't think this belongs in the model.. It's business logic that needs to be managed at the appropriate layer in the code, likely in the repository. This codepath is currently unused, but should be set the right thing once we allow adding LOIs in the data collection flow.

gino-m:
> Approving the PR to unblock but please address the above comment before merging. Thanks!

Thanks for the review, Shobhit!

PR #1531, opened 2023-02-07
[Feature] Tapping LOI should highlight the respective card with zoom/pan

### LINKED ISSUE #1530: Tapping an LOI should focus its respective job card, not show the old bottom sheet and zoom/pan

shobhitagarwal1612:
Fixes #1530

Things done:
 * Clicking LOI doesn't open the bottom sheet anymore. The code related to bottom sheet isn't removed yet as there are some usages still left.
 * Removes local cache of LOIs from HomeScreenViewModel as it is no longer needed.
 * If there are more than 1 feature available at the target, then the first one gets selected
 * The map zooms and pans to the selected target position and the bottom card list gets updated accordingly.
 * Also changes the behavior of GoogleMapsFragment to **smoothly transition at the target rather than snapping directly.**

@gino-m @JSunde PTAL?


shobhitagarwal1612:
Gentle ping

gino-m (approved):
Lgtm, thanks for the fix.

PR #1535, opened 2023-02-07
Cleanup test data collection flow entrypoint

shobhitagarwal1612:
Now that the actual data collection flow has been integrated, we don't need this.

This is a no-op change. Doesn't require any extra validation steps.

@JSunde PTAL?


JSunde:
Thanks Shobhit!

JSunde (approved):

PR #1537, opened 2023-02-07
Remove parameters from Task Fragment constructors

JSunde:
We were previously passing in data via the TaskFragment constructors but that was causing a crash on orientation changes because Android system couldn't find the default constructor.

This change removes those parameters, now we're still getting a failure on orientation change because the ViewModels aren't persisted, but I will address that in a follow up change.


ground/src/main/java/com/google/android/ground/ui/datacollection/TaskFragment.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.datacollection
+
+import com.google.android.ground.ui.editsubmission.AbstractTaskViewModel
+
+interface TaskFragment {
+  var viewModel: AbstractTaskViewModel
+}
```
shobhitagarwal1612, line 22: Can we add generic to the interface so that the implementation can provide a type for the ViewModel instead of having to cast before using it?

JSunde: I went ahead and added the generic to TaskFragment. Now the fragment's don't need to cast the viewModel, but we still currently need to cast the ViewModels in the DataCollectionViewPagerAdapter.

Although I will be fixing this in the near future because I'm going to be making the fragments instantiate their own ViewModels.

shobhitagarwal1612 (approved):

PR #1538, opened 2023-02-08
Remove GeometryWrapper reference from LocationOfInterestMutation

### LINKED ISSUE #1529: [Code health] GeometryWrapper in Room impl leaked to domain model
`GeometryWrapper` appears in `LocationOfInterestMutation`, which lives in the domain model.

~Perhaps `GeometryWrapper` should be moved there, or to the geometry model as well?~

Correction: `GeometryWrapper` isn't needed in `LocationOfInterestMutation`, only in the local db Room impl. Instead, we should just be able to have `geometry: Geometry` in `LocationOfInterestMutation`.

@JSunde PTAL?

JSunde:
Fixes #1529

@gino-m PTAL?


gino-m (changes requested):
Thanks for the fix. There are a few build errors left:

e: file:///workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt:157:16 Type mismatch: inferred type is Geometry? but GeometryWrapper? was expected
e: file:///workspace/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt:167:19 Type mismatch: inferred type is Geometry? but GeometryWrapper? was expected
e: file:///workspace/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiMutationConverter.kt:42:45 Unresolved reference: getGeometry

gino-m (approved):

PR #1551, opened 2023-02-10
Rework locationState tracking using StateFlow

### LINKED ISSUE #1540: [Map] Don’t show “Cannot show location without permission” toast when prompting for permission.

JSunde:
Fixes #1540


ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -79,6 +86,40 @@ constructor(
...
+          }
+        )
+        .andThen(settingsManager.enableLocationSettings(FINE_LOCATION_UPDATES_REQUEST))
+        .andThen(requestLocationUpdates())
+        .toSingle {
+          lock()
+          Result.success(true)
+        }
+        .onErrorReturn { Result.failure(it) }
+        .subscribe()
```
JSunde, line 110: I feel like I'm missing something here, I'm guessing maybe I should be returning this result, rather than subscribing here? But I don't fully understand what would be consuming the result.

As is this correctly launches the permissions dialog, but it seems that the onPermissionRequestResult isn't called afterwards, I'm guessing because there are no longer any subscribers?

@gino-m do you know what I should be doing with the result here  and how I can make sure the onPermissionRequestResult is called?

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -79,6 +86,40 @@ constructor(
       .onErrorReturn { Result.failure(it) }
   }
 
+  fun toggleLocationLock() {
+    if (locationLockState.value.getOrDefault(false)) {
+      unlock()
```
gino-m, line 91: Nit: `disableLocationLock()` might be more appropriate, since we've been talking about "location lock" as a feature that can be turned on and off, not about "locking the location".

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
gino-m: Right.. if you turn this into a suspend function you can call `await()` instead of subscribing.

gino-m: You could also get rid of the chained calls to `andThen()`, instead calling `await()` on each step in the flow.

JSunde: Resolved after switching to a suspend fun

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -44,43 +44,70 @@ private val FINE_LOCATION_UPDATES_REQUEST =
...
-  private val locationClient: RxFusedLocationProviderClient
+  private val locationClient: RxFusedLocationProviderClient,
+  private val mapStateRepository: MapStateRepository
 ) {
 
-  private val locationUpdates: @Hot(replays = true) Subject<Location> = BehaviorSubject.create()
-  private val locationUpdateCallback: RxLocationCallback = RxLocationCallback(locationUpdates)
+  val locationLockState: MutableStateFlow<Result<Boolean>> =
+    MutableStateFlow(Result.success(mapStateRepository.isLocationLockEnabled))
+  private val locationUpdatesStateFlow: MutableStateFlow<Location?> = MutableStateFlow(null)
```
gino-m, line 56: Does this stream contain location updates (i.e., events) or the current location (i.e., states)? If it's the former, this should probably be modeled as a Channel.

If they're states, I'd name them accordingly. In that case we can drop the `StateFlow` suffix for better readability, imo.

gino-m (changes requested):
This is cool!

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
JSunde: I'm not sure I fully understand the distinction, between location updates vs current location. I think I'm leaning towards the latter because it receives updates to the location from the StateFlowCallback, but it stores the current location as state, which is then combined with the locationLockState in the locationLockAwareLocationUpdates (poorly named I think) to provide the current location if the location lock is enabled, and null if not.

I'll go ahead and update the names of both the locationUpdatesStateFlow and locationLockAwareLocationUpdates to currentLocationState and locationLockAwareCurrentLocation.

Let me know if that makes sense, or if you still think we should use a channel instead of a StateFlow.

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -44,43 +42,21 @@ private val FINE_LOCATION_UPDATES_REQUEST =
...
+   * will only see the latest location. Returns null if location lock is disabled
    */
-  fun getLocationUpdates(): Flowable<Location> =
-    locationUpdates
-      // There sometimes noticeable latency between when location update request succeeds and when
-      // the first location update is received. Requesting the last know location is usually
-      // immediate, so we merge into the stream to reduce perceived latency.
-      .startWith(locationClient.lastLocation.toObservable())
-      .toFlowable(BackpressureStrategy.LATEST)
+  fun getLocation(): Flow<Location?> = locationUpdates.receiveAsFlow()
```
gino-m, line 56: Why not make `locationUpdates` read-only field and expose it as a `Channel`?

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -44,43 +44,70 @@ private val FINE_LOCATION_UPDATES_REQUEST =
...
    * up will only see the latest location.
    */
-  fun getLocationUpdates(): Flowable<Location> =
-    locationUpdates
-      // There sometimes noticeable latency between when location update request succeeds and when
-      // the first location update is received. Requesting the last know location is usually
-      // immediate, so we merge into the stream to reduce perceived latency.
-      .startWith(locationClient.lastLocation.toObservable())
-      .toFlowable(BackpressureStrategy.LATEST)
+  fun getLocationUpdates(): StateFlow<Location?> = locationLockAwareLocationUpdates
```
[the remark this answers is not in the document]
JSunde: Done! Moved the locationLockState tracking to the BaseMapViewModel and now the LocationManager has a Channel for providing the locationUpdates.

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
JSunde: SG, removed unnecessary qualifications and we're now exposing the location updates through a Channel.

JSunde:
PTAL @gino-m

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -15,26 +15,24 @@
...
 import kotlinx.coroutines.flow.*
-import kotlinx.coroutines.rx2.await
 import timber.log.Timber
 
 private const val UPDATE_INTERVAL: Long = 1000 /* 1 sec */
 
 private const val FASTEST_INTERVAL: Long = 250 /* 250 ms */
 
-private val FINE_LOCATION_UPDATES_REQUEST =
+val FINE_LOCATION_UPDATES_REQUEST: LocationRequest =
```
gino-m, line 35: Can type be inferred here?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -44,43 +42,21 @@ private val FINE_LOCATION_UPDATES_REQUEST =
...
-  private val permissionsManager: PermissionsManager,
-  private val settingsManager: SettingsManager,
-  private val locationClient: RxFusedLocationProviderClient
+  @ApplicationScope private val externalScope: CoroutineScope,
+  private val locationClient: RxFusedLocationProviderClient,
 ) {
 
-  private val locationUpdates: @Hot(replays = true) Subject<Location> = BehaviorSubject.create()
-  private val locationUpdateCallback: RxLocationCallback = RxLocationCallback(locationUpdates)
+  private val locationUpdates = Channel<Location>()
```
gino-m, line 49: We spoke:

```kt
  private val _locationUpdates = MutableSharedFlow<Location>()
  val locationUpdates: SharedFlow<Location>
    get() = _locationUpdates
  ```

File: ground/src/main/java/com/google/android/ground/system/LocationManager.kt (hunk 6 above)
Comment (JSunde) [reply ground/src/main/java/com/google/android/ground/system/LocationManager.kt:34]: I got this warning when the type was inferred previously "Declaration has type inferred from a platform call, which can lead to unchecked nullability issues"

File: ground/src/main/java/com/google/android/ground/system/LocationManager.kt [hunk 8]
```diff
@@ -45,69 +43,20 @@ class LocationManager
...
-      }
-      .stateIn(externalScope, SharingStarted.Lazily, null)
+  private val locationUpdates = Channel<Location>()
+  private val locationCallback = ChannelLocationCallback(locationUpdates, externalScope)
 
   /**
-   * Returns the location update stream. New subscribers and downstream subscribers that can't keep
-   * up will only see the latest location.
+   * Returns the location StateFlow. New subscribers and downstream subscribers that can't keep up
+   * will only see the latest location. Returns null if location lock is disabled
```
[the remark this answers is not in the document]
JSunde: No longer relevant.

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
```diff
@@ -44,43 +42,21 @@ private val FINE_LOCATION_UPDATES_REQUEST =
...
 
-  private val locationUpdates: @Hot(replays = true) Subject<Location> = BehaviorSubject.create()
-  private val locationUpdateCallback: RxLocationCallback = RxLocationCallback(locationUpdates)
+  private val locationUpdates = Channel<Location>()
+  private val locationCallback = ChannelLocationCallback(locationUpdates, externalScope)
 
   /**
-   * Returns the location update stream. New subscribers and downstream subscribers that can't keep
-   * up will only see the latest location.
+   * Returns the location StateFlow. New subscribers and downstream subscribers that can't keep up
```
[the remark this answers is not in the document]
JSunde: Returning the SharedFlow as discussed offline.

ground/src/main/java/com/google/android/ground/system/LocationManager.kt
gino-m: ah, ok!

gino-m (approved):
Really cool, great improvement!

Issue #1348, opened 2022-12-09
[Main Map] Don’t show a card for clustered LOIs (ie below certain zoom level)


shobhitagarwal1612:
* Add a dependency on current zoom level when showing the cards

shobhitagarwal1612:
Fixed by https://github.com/google/ground-android/pull/1553

PR #1553, opened 2023-02-11
[Main Map] Only display job cards if below certain zoom level

shobhitagarwal1612:
Towards #1348 

@scolsen  @gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -194,23 +150,15 @@ internal constructor(
...
-
-    val savedMapLocationsOfInterest =
-      Flowable.combineLatest(
-        loiStream.map { locationsOfInterest -> toLocationOfInterestFeatures(locationsOfInterest) },
-        selectedLocationOfInterest
-      ) { locationsOfInterest, selectedLocationOfInterest ->
-        updateSelectedLocationOfInterest(locationsOfInterest, selectedLocationOfInterest)
-      }
 
+    // TODO: Should we only render the LOIs that are present within the map bounds?
```
gino-m, line 154: This is a bigger project imo and not actionable here. Can you move this to an issue instead of a TODO?

ground/src/main/java/com/google/android/ground/ui/map/LoiController.kt
```diff
@@ -30,27 +28,23 @@ import io.reactivex.Flowable
...
 internal constructor(
   private val surveyRepository: SurveyRepository,
   private val locationOfInterestRepository: LocationOfInterestRepository,
 ) {
 
   private val cameraBoundsSubject: @Hot Subject<LatLngBounds> = PublishSubject.create()
-  val locationsOfInterest: LiveData<List<LocationOfInterest>>
 
-  init {
+  fun loisWithinMapBounds(): Flowable<List<LocationOfInterest>> {
```
gino-m, line 43: Can this be moved into the loiRep as `getWithinBounds(LatLngBounds)`? The camera bounds should probably live in the `MapStateRepository` or `BaseMapViewModel`, then this class can be deleted.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -127,4 +124,33 @@ constructor(
...
+    surveyRepository.activeSurvey
+      .switchMap { survey ->
+        survey
+          .map { localLoiStore.getLocationsOfInterestOnceAndStream(it) }
+          .orElse(Flowable.just(setOf()))
+      }
+      .distinctUntilChanged()
+
+  /** Returns a flowable of all [LocationOfInterest] within the map bounds (viewport). */
+  fun getWithinBoundsOnceAndStream(
```
gino-m, line 139: Camera bounds sounds like something that belongs in the view or in MapStateRepository. Can we pass that in as an arg here instead of maintaining it in this class?

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -127,4 +124,33 @@ constructor(
...
+    cameraBoundUpdates: Flowable<LatLngBounds>
+  ): Flowable<List<LocationOfInterest>> {
+    val loiStream = getAllLocationsOfInterestOnceAndStream()
+    return cameraBoundUpdates
+      .flatMap { bounds -> loiStream.map { it.toLoiCardsWithinBounds(bounds) } }
+      .distinctUntilChanged()
+  }
+
+  /** Filters all [LocationOfInterest] within [bounds]. */
+  private fun Set<LocationOfInterest>.toLoiCardsWithinBounds(
```
gino-m, line 149: Cards are also a UI element that is out of scope of the repository layer. Maybe it's just a naming issue, and this should be `getLoisWithinBounds()`?

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -48,6 +55,13 @@ constructor(
   mapController: MapController
 ) : AbstractViewModel() {
 
+  private val cameraZoomSubject: @Hot Subject<Float> = PublishSubject.create()
+  val cameraZoomUpdates: Flowable<Float> = cameraZoomSubject.toFlowable(BackpressureStrategy.LATEST)
```
gino-m, line 59: Consider using `SharedFlow` for new event streams. Feel free to change in a follow-up if that causes too much churn in this PR.

gino-m (changes requested):
Great! Thanks for the updates.

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
shobhitagarwal1612: `LatLngBounds` flowable is already passed via an arg. Are you referring to something else?

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
shobhitagarwal1612: Removed mentions of cards from repository. Thanks!

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: I'll file a bug to do it separately

shobhitagarwal1612:
@gino-m Gentle ping!

scolsen (approved):

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
gino-m: Sorry, I missed that the bounds were being passed in as an arg. Please disregard.

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -127,4 +124,33 @@ constructor(
       MutationEntitySyncStatus.IN_PROGRESS,
       MutationEntitySyncStatus.FAILED
     )
+
+  /** Returns a flowable of all [LocationOfInterest] for the currently active [Survey]. */
+  fun getAllLocationsOfInterestOnceAndStream(): Flowable<Set<LocationOfInterest>> =
```
gino-m, line 127: Can we update this to only look at the offline LOIs? I'm working on code that will update those in the background, so we no longer need to subscribe to updates from Firestore here; instead we can subscribe to changes in the local Room store.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
shobhitagarwal1612: This does return only offline LOIs. Should I rename it to something else to make it more explicit?

gino-m: It's ok for now, I'll update it in my upcoming PR. Thanks!

gino-m (approved):
🔍 ↕️

PR #1582, opened 2023-02-21
[WIP] Sync LOIs in background worker on FCM message

### LINKED ISSUE #1373: [Survey sync] Implement survey load and background sync

gino-m:
Fixes #1373.


ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -121,15 +101,9 @@ constructor(
...
   /** Returns a flowable of all [LocationOfInterest] for the currently active [Survey]. */
-  fun getAllLocationsOfInterestOnceAndStream(): Flowable<Set<LocationOfInterest>> =
-    surveyRepository.activeSurvey
-      .switchMap { survey ->
-        survey
-          .map { localLoiStore.getLocationsOfInterestOnceAndStream(it) }
-          .orElse(Flowable.just(setOf()))
-      }
-      .distinctUntilChanged()
+  fun getAllLocationsOfInterestOnceAndStream(): Flowable<Set<LocationOfInterest>> = Flowable.never()
```
gino-m, line 106: @scolsen  @shobhitagarwal1612 @JSunde After this PR, `SurveyRepository` depends on `LocationOfInterestRepository` to sync LOIs during survey activation, so the latter can't also depend on the former to get the active survey and build a `Flowable`. We could move that logic to the view, but it does sound like a job for the Repository. How do you think we should represent this? Some ideas:

* Introduce use cases, eg `ActivateSurveyUseCase`, `UpdateSurveyUseCase`,  `ViewJobCardsUseCase`, and ``
* Create a super-repo that coordinates `SurveyRepository` and `LocationOfInteresteRepository`.
* Move this logic to the views (:vomiting_face:).

Thoughts?

scolsen: Maybe this one?

> Create a super-repo that coordinates SurveyRepository and LocationOfInteresteRepository.

I'd be curious to hear more about the UseCases idea.

Survey is feeling more and more to me like actual global state that we're trying desperately to pretend is not global state...

scolsen: For example, could we just model the active Survey as a singleton? Then give it internal logic to keep it up to date?

Then, rather than having to coordinate, dependents can just point to the singleton which is always in the latest state.

gino-m: "Active Survey" is already a persisted global state managed by the repository.

"Use Cases" are an optional but recommended component of Android arch best practices: http://go/android-dev/topic/architecture/domain-layer

gino-m: That sounds a lot like a repository dedicated to the active survey; repositories are singletons, so if we don't want to use UseCases maybe that's the right fit?

scolsen: Looking at the use cases doc, that seems like a clean way to do it. Yeah that's more or less what I'm proposing. My original thinking was that we could eliminate the dependency problem by factoring the active survey out of the repository. Basically we'd have an ActiveSurvey singleton, the survey repo would update it as a side effect and the LOI repo would read from it...but that would probably also not be so great in certain respects.

On switching a survey, the survey repo would update ActiveSurvey asynchronously, the LOI repo would then be pointing at this Survey and, on change, load up the LOIs.

Thinking about it in general, I feel like we previously had a bit of extra complexity insofar as we needed to pass a survey to fetch LOIs. Unless we do cacheing on device of LOI data, there's really no reason to have to pass a survey for an LOI fetch, since the user can only interact with LOIs associated with the active survey. If we only update things lazily, in theory, LOI fetches would never have to care about any survey other than the active one. I suppose the issue is we *must* be eager, since users may need to switch surveys while in the field, and at that point we'd need to rely on cached data.

This is a verbose way of saying that I think either a dedicated ActiveSurveyRepo (or some other dep breaker) or the UseCase pattern would work. The latter seems pretty nice, actually. Maybe we try that? (I do think the UseCase terminology is a bit unclear though).

gino-m: That makes sense, modulo that we always need to get LOIs for the active survey; the background sync will continue to sync data for _all_ surveys, not just the active one. 

Re use cases , according to the doc:

> To keep these classes simple and lightweight, each use case should only have responsibility over a single functionality, and they should not contain mutable data. You should instead handle mutable data in your UI or data layers.

So in this case, the "use cases" describe user-facing functionality exposed via the UI. I assume they can be decomposed as much as we like, but that each use case is a single operation. I can give that a try to see how it looks?

scolsen: sgtm!

gino-m:
Superseded by #1614.

Issue #1580, opened 2023-02-19
[Code health] Enforce more kotlin detekt rules


shobhitagarwal1612:
Draft PR: https://github.com/google/ground-android/pull/1570

scolsen:
@shobhitagarwal1612 assigning to you for now. Adding to RC1.

PR #1583, opened 2023-02-21
[Code health] Detect functions that can be converted to expression functions

shobhitagarwal1612:
Towards #1580

Split from https://github.com/google/ground-android/pull/1570

Enables the rule to detect functions that can be converted to expression function. For now, I've excluded all `*Module.kt` files from this rule due to noisy output.

@JSunde @gino-m @scolsen PTAL?


JSunde (approved):
Wow nice! Thanks Shobhit!

PR #1586, opened 2023-02-21
Switch DataCollectionViewModel to be @HiltViewModel

JSunde:
This should allow the DataCollectionViewModel to be injected and allow us to inject the SavedStateHandle.


gino-m (approved):
LGTM!

PR #1587, opened 2023-02-22
[Code Health] Fix class ordering

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/1580

Split from https://github.com/google/ground-android/pull/1570

https://detekt.dev/docs/rules/style/#classordering

Things done:
 * Converts variables to functions where necessary and vice versa
 * Moves variables to the top of the class
 * Moves init block after variables
 * Moves companion block at the end of the class

@scolsen PTAL?


ground/src/main/java/com/google/android/ground/persistence/sync/PhotoSyncWorkManager.kt
```diff
@@ -34,7 +34,7 @@ constructor(
   override val workerClass: Class<PhotoSyncWorker>
     get() = PhotoSyncWorker::class.java
 
-  override fun preferredNetworkType(): NetworkType =
+  override val preferredNetworkType: NetworkType =
```
scolsen, line 37: This will be frozen to UNMETERED or CONNECTED depending on the value of the localValueStore when the worker is first instantiated--is that what we want?

If not, use a `get()` so that this is instead computed every time we reference the field (then we'd always fetch the latest value from the localValueStore whenever this is referenced).

ground/src/main/java/com/google/android/ground/persistence/sync/TileSetDownloadWorkManager.kt
```diff
@@ -29,7 +29,7 @@ constructor(private val workManager: WorkManager, private val localValueStore: L
   override val workerClass: Class<TileSetDownloadWorker>
     get() = TileSetDownloadWorker::class.java
 
-  override fun preferredNetworkType(): NetworkType =
+  override val preferredNetworkType: NetworkType =
```
scolsen, line 32: Same thing here. This will be fixed on instantiation. We might want to use a `get()` so that we can still reference this as a property but it will always have the latest value from the localValueStore.

scolsen (changes requested):

ground/src/main/java/com/google/android/ground/persistence/sync/PhotoSyncWorkManager.kt
shobhitagarwal1612: thanks for catching this! reverted this change

ground/src/main/java/com/google/android/ground/persistence/sync/TileSetDownloadWorkManager.kt
shobhitagarwal1612: also reverted. thanks

scolsen:
Looks like this is failing because some properties changed to functions:

```
e: file:///workspace/ground/src/main/java/com/google/android/ground/ui/offlinebasemap/selector/OfflineAreaSelectorFragment.kt:87:17 Function invocation 'cameraMovedEvents()' expected
e: file:///workspace/ground/src/main/java/com/google/android/ground/ui/offlinebasemap/selector/OfflineAreaSelectorViewModel.kt:83:81 Function invocation 'tileSets()' expected
```

scolsen (approved):

PR #1588, opened 2023-02-23
[Data Collection] Show the final confirmation screen after collecting data

### LINKED ISSUE #1546: [Data Collection] Show the final confirmation screen after collecting data

shobhitagarwal1612:
Fixes #1546

Adds a confirmation dialog fragment after data collection is complete.

Alternative approach considered
------------
Another way to achieve this was to add an argument to the `HomeScreenFragment` and set the boolean to true when returning. Then in `HomeScreenFragments` onResume we can inflate the dialog by checking the arguments. This was causing an issue since arguments gets persisted until the home screen gets destroyed. So, pausing and resuming was also triggering the dialog. Hence the cleaner solution is to trigger it from within DataCollection flow and let navigator handle it.

Screenshot
--------------

@JSunde PTAL?


shobhitagarwal1612:
Gentle ping!

JSunde (approved):

PR #1589, opened 2023-02-23
Persist task position in TaskFragments, allowing them to instantiate their own ViewModel

### LINKED ISSUE #1458: [DataCollectionFlow] App crashes when changing screen orientation
Steps to reproduce:
1. Open any job
2. Rotate the screen

JSunde:
Fixes #1458

@scolsen PTAL?


scolsen:
Looks like it's failing with:  `The parameters in the factory method must match the @Assisted parameters in com.google.android.ground.ui.datacollection.DataCollectionViewPagerAdapter.`

JSunde:
> Looks like it's failing with: `The parameters in the factory method must match the @Assisted parameters in com.google.android.ground.ui.datacollection.DataCollectionViewPagerAdapter.`

I think that should be fixed now.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -123,9 +123,11 @@ internal constructor(
 
   fun loadSubmissionDetails(args: DataCollectionFragmentArgs) = argsProcessor.onNext(args)
 
-  fun getTaskViewModel(position: Int, task: Task): AbstractTaskViewModel {
+  fun getTaskViewModel(position: Int): AbstractTaskViewModel {
     val viewModels = taskViewModels.value
     require(viewModels != null)
+    val tasks = submission.value!!.job.tasksSorted
```
scolsen, line 129: How much work would it be to do this without the `!!`? 

If it's null we should probably handle it with some kind of error toast or screen. Maybe let's add a `require` now and TODO for handling this case?

ground/src/main/java/com/google/android/ground/ui/datacollection/DropAPinTaskFragment.kt
```diff
@@ -30,8 +30,9 @@ import javax.inject.Inject
 
 @AndroidEntryPoint
 class DropAPinTaskFragment : AbstractMapContainerFragment(), TaskFragment<DropAPinTaskViewModel> {
-
   override lateinit var viewModel: DropAPinTaskViewModel
+  // Not needed for DropAPinTaskFragment
+  override var position = -1
```
scolsen, line 35: It's probably overkill, but we could also implement this out as a concretization over the TaskFragment, e.g. some PositionlessTaskFragment<T>. Probably not worth it for now, but if we wind up with 3+ classes like this I'd consider it.

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DropAPinTaskFragment.kt
JSunde: Yea I think that sounds good, if we end up adding many more that don't need a reference to the position I think we could revisit at that point.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: Updated to requireNotNull. I think it would probably be a decent amount of work do without that.

PR #1614, opened 2023-03-06
Sync LOIs on survey activation and remote change

### LINKED ISSUE #1373: [Survey sync] Implement survey load and background sync

gino-m:
Fixes #1373.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,6 +91,7 @@ detekt {
     source = files("ground/src", "sharedTest/src")
     allRules = true
     config = files("${project.rootDir}/config/detekt/detekt.yml")
+    baseline = file("${project.rootDir}/config/detekt/baseline.xml")
     buildUponDefaultConfig = true
     ignoreFailures = false
 
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@
 
 buildscript {
     ext {
-        hiltVersion = "2.44"
+        hiltVersion = "2.45"
         kotlinVersion = "1.8.10"
         navigationVersion = "2.5.3"
     }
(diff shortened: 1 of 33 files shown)


gino-m:
Squashed commits and fixed bad base change in #1616

PR #1618, opened 2023-03-08
Use navGraph scoped ViewModel to share DataCollectionViewModel

### LINKED ISSUE #1608: App crashes after opening job with different tasks
**Describe the bug**
App crashes if a task of different type is opened

**To Reproduce**
1. Open a job containing "text" task. 
2. Finish all responses and click continue
3. Open another job containing "number" task
4. App crashes

**Expected behavior**
Correct tasks should get loaded

**Actual behavior**
App crashes as the fragment tries to cast `TextTaskViewModel` to `NumberTaskViewModel`

The issue seems to be that DataCollectionViewModel doesn't reset the state after data collection is completed or canceled.

@JSunde This seems related to the recent addition of persisting VMs across config changes. Could please take a look and fix this?

JSunde:
Fixes #1608

Previously we used an ActivityScoped view model to share the DataCollectionViewModel between the DataCollectionFragment and the Task fragments, but this caused issues when the user completed one job and opened another because the DataCollectionViewModel would not be cleared. Now the DataCollectionViewModel is scoped to a new nav graph which includes only the Data Collection fragments so we properly create a new DataCollectionViewModel instance for each new job.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -7,6 +7,7 @@
         <option name="TAB_SIZE" value="2" />
       </value>
     </option>
+    <option name="ENABLE_SECOND_REFORMAT" value="true" />
     <AndroidXmlCodeStyleSettings>
       <option name="LAYOUT_SETTINGS">
         <value>
--- a/ground/build.gradle
+++ b/ground/build.gradle
@@ -269,6 +269,8 @@ dependencies {
     androidTestImplementation 'com.squareup.rx.idler:rx2-idler:0.11.0'
     testImplementation 'com.squareup.rx.idler:rx2-idler:0.11.0'
     testImplementation "org.jetbrains.kotlin:kotlin-test:$project.kotlinVersion"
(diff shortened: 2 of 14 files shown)


JSunde:
Looks like something went wrong with my merges. Let me see if I can resolve that.

JSunde:
I seem to get the same problem after resetting to my original commit and then rebasing onto master.

JSunde:
Okay I think I might have gotten it resolved now.

JSunde:
@scolsen Any thoughts on how to fix DataCollectionFragmentTest? I didn't see any references to NavController in our tests, so I'm not sure how I might need to modify the setup to ensure there is a NavController set up.

```
Fragment DataCollectionFragment{1ff50d8d} (51e23762-ac76-49b8-ad6a-56662365ab78 id=0x1020002 tag=) does not have a NavController set
```

JSunde:
I tried using the test controller like you suggested but I'm still getting the same error. I wonder if the issue is that the fragment is created and the nav controller is accessed before we can actually set the navcontroller for the view.

Issue #1621, opened 2023-03-12
[Data Collection] Revamp UI as per latest mocks


shobhitagarwal1612:
As per [figma mocks](https://www.figma.com/file/r2uGtyOiWoDa29za7DNuUc/Ground-Acacia-UX?node-id=1566%3A47266&t=dG3k5Uq9xfJLrC3s-0), tasks can have different action buttons with varying themes. This is a tracking bug to apply this feature to all tasks in a generic way.

[Design Doc](https://docs.google.com/document/d/15CGT-9LvgarzBvrC8kPPAEHlrGuZCqVL0i9lpmya2dU/r/0-OLrHylzybpz6_mkioEm7og/edit#)

PR #1622, opened 2023-03-12
[Do not merge] Prototype for adding custom action buttons to tasks

shobhitagarwal1612:
Related to #1621

The purpose of this PR is to prototype the feature for all tasks end-to-end to get a sense of current requirements. A lot of things were identified as a part of this:

1. Separating the base template for tasks and the inset task view
2. Dynamically adding the buttons instead of hardcoding in the xml
3. Button factory which accepts some parameters (type/theme)
4. Using a single base class for all tasks. Currently, map and non-map type tasks were extending different base classes.
5. Separating out the ui components from the business logic to keep the classes lightweight.

Next steps:
1. Prepare a design doc with the above findings and share with the team before freezing the design.

@JSunde @scolsen  PTAL?


shobhitagarwal1612:
Closing this in favour of https://github.com/google/ground-android/pull/1638

Issue #1541, opened 2023-02-09
[Data Collection] Suggest new LOI + collect data


JSunde:
We want to have a special task type which is a variant of the DropAPinTask allowing the user to suggest a new LOI.

JSunde:
The premise of this feature is that we would have some entry point to initiate a job without an associated LOI and the first task of that job would be to "Drop a pin" or "Draw an area" which would create the LOI that the submission would be associated with.

scolsen:
@JSunde is most of the way there on this; we need to clarify some product requirements but will proceed with an initial implementation.

gino-m:
We spoke, done!

PR #1623, opened 2023-03-13
Add MapCardUiData class to be shared between Loi cards and Suggest Loi Cards

JSunde:
Towards #1541


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -68,11 +68,27 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
...
+      adapter.updateData(
+        lois.map {
+          MapCardUiData(
+            name = it.caption ?: it.type.name,
+            subtitle = it.job.name,
+            surveyId = it.surveyId,
+            // TODO(#1483): Add submission count
+            submissionText = "No submissions",
+            loiId = it.id
+          )
```
shobhitagarwal1612, line 89: Can we move this logic to the VM?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -68,11 +68,27 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
       .`as`(RxAutoDispose.autoDisposable(this))
       .subscribe { onZoomThresholdCrossed() }
 
-    adapter = LoiCardAdapter()
-    adapter.setLoiCardFocusedListener { mapFragment.setActiveLocationOfInterest(it) }
+    adapter = MapCardAdapter()
+    adapter.setLoiCardFocusedListener {
+      if (it?.loiId != null) {
+        mapFragment.setActiveLocationOfInterest(it.loiId)
+      }
```
shobhitagarwal1612, line 75: Maybe use `?.let{}`

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/MapCardUiData.kt
```diff
@@ -0,0 +1,10 @@
+package com.google.android.ground.ui.home.mapcontainer
+
+/** Data used to populate the Map cards (either an Loi card, or a Suggest Loi card). */
+data class MapCardUiData(
```
shobhitagarwal1612, line 4: Should this also contain a enum type field instead of using `loiId` being null as a proxy for suggest loi card?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
JSunde: Updated this now that I decided to use a sealed class for MapCardUiData

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/MapCardUiData.kt
JSunde: I actually ended up deciding to make MapCardUiData a sealed interface and then add a new LoiCardUiData which wraps the loi for the loi cards.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
JSunde: Done! I ended up reverting my changes to the LoiCardViewModel and updating the MapCardUiData to a sealed interface and wrapped the loi in an LoiCardUiData. Take a look and let me know what you think!

shobhitagarwal1612 (approved):
LGTM.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
shobhitagarwal1612: Yes, this looks much better.

shobhitagarwal1612:
@JSunde One issue is still left. 

Expected license not found or incorrect in the file: /workspace/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/MapCardUiData.kt. [AbsentOrWrongFileLicense]

PR #1625, opened 2023-03-14
Reactivate last survey on sign in

### LINKED ISSUE #1612: Regression: Last active survey not restored after sign-in

gino-m:
Fixes #1612.

@scolsen and/or @shobhitagarwal1612 PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/MainViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/MainViewModel.kt
@@ -18,6 +18,7 @@ package com.google.android.ground
 import androidx.core.view.WindowInsetsCompat
 import androidx.lifecycle.MutableLiveData
 import androidx.navigation.NavDirections
+import com.google.android.ground.coroutines.DefaultDispatcher
 import com.google.android.ground.domain.usecases.survey.ReactivateLastSurveyUseCase
 import com.google.android.ground.model.User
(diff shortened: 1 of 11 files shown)


ground/src/main/java/com/google/android/ground/MainViewModel.kt
```diff
@@ -74,7 +80,8 @@ constructor(
     return signInState.result.fold(
       {
         when (signInState.state) {
-          SignInState.State.SIGNED_IN -> onUserSignedIn(it!!)
+          SignInState.State.SIGNED_IN ->
+            rxObservable(defaultDispatcher) { send(onUserSignedIn(it!!)) }
```
shobhitagarwal1612, line 84: What is the purpose of the added constructs? How do we know when to use them?

shobhitagarwal1612: The default ktdoc is sufficient. I was mostly concerned about how do we ensure that we apply these changes as we migrate more code to coroutines.

gino-m: Makes sense.. Tests won't work correctly without setting the dispatcher, and nothing will be emitted without `send`, so correctness elsewhere can be enforced by unit and manual testing. Lmk if you have another idea!

shobhitagarwal1612 (approved):
💯

PR #1628, opened 2023-03-16
[Revamped Data Collection UI] Add layouts for TaskButton and TaskView

### LINKED ISSUE #1621: [Data Collection] Revamp UI as per latest mocks

shobhitagarwal1612:
Fixes #1621

This is a no-op change. Adds the xml layouts, styles and helper classes for TaskView and TaskButton.

These will be used in the upcoming PRs for refactoring task fragments.

Full prototype: https://github.com/google/ground-android/pull/1622

[Design doc](https://docs.google.com/document/d/15CGT-9LvgarzBvrC8kPPAEHlrGuZCqVL0i9lpmya2dU/r/0-OLrHylzybpz6_mkioEm7og/edit#)

@JSunde @scolsen  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.datacollection.components
+
+/** Defines a unique action that can be bound to a [TaskButton], along with the UI styling. */
+enum class ButtonAction(val type: Type, val theme: Theme) {
+  // All tasks
+  CONTINUE(Type.TEXT, Theme.DARK_GREEN),
+  SKIP(Type.TEXT, Theme.LIGHT_GREEN),
+  UNDO(Type.ICON, Theme.LIGHT_GREEN),
+
+  // Drop a pin task
+  DROP_PIN(Type.TEXT, Theme.TRANSPARENT),
+
+  // Draw a polygon task
+  ADD_PIN(Type.TEXT_ICON, Theme.TRANSPARENT),
+  COMPLETE(Type.TEXT, Theme.TRANSPARENT);
+
+  enum class Type {
+    TEXT,
+    ICON,
+    TEXT_ICON,
+  }
+
+  enum class Theme {
+    DARK_GREEN,
+    LIGHT_GREEN,
+    TRANSPARENT,
+  }
+}
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
@@ -23,11 +23,11 @@ enum class ButtonAction(val type: Type, val theme: Theme) {
   UNDO(Type.ICON, Theme.LIGHT_GREEN),
 
   // Drop a pin task
-  DROP_PIN(Type.TEXT, Theme.TRANSPARENT),
+  DROP_PIN(Type.TEXT, Theme.OUTLINED),
 
   // Draw a polygon task
-  ADD_PIN(Type.TEXT_ICON, Theme.TRANSPARENT),
-  COMPLETE(Type.TEXT, Theme.TRANSPARENT);
+  ADD_PIN(Type.TEXT_ICON, Theme.OUTLINED),
+  COMPLETE(Type.TEXT, Theme.OUTLINED);
 
   enum class Type {
     TEXT,
@@ -38,6 +38,6 @@ enum class ButtonAction(val type: Type, val theme: Theme) {
   enum class Theme {
     DARK_GREEN,
     LIGHT_GREEN,
-    TRANSPARENT,
+    OUTLINED,
   }
 }
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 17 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
```diff
@@ -0,0 +1,43 @@
...
+  enum class Type {
+    TEXT,
+    ICON,
+    TEXT_ICON,
+  }
+
+  enum class Theme {
+    DARK_GREEN,
+    LIGHT_GREEN,
+    TRANSPARENT,
```
JSunde, line 41: Do we need a Theme for the Outlined buttons as well?

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
```diff
@@ -0,0 +1,104 @@
...
+        }.apply {
+          id = View.generateViewId()
+          // Default state
+          isEnabled = false
+        }
+      )
+  }
+}
+
+// TODO(Shobhit): Figure out a way to create styled buttons without using XML.
```
JSunde, line 70: We could use Jetpack Compose

ground/src/main/res/drawable/task_chip_image_button.xml
```diff
@@ -0,0 +1,26 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+  android:shape="rectangle">
+  <size
+    android:width="20dp"
+    android:height="20dp" />
+  <solid android:color="#FFFFFF" />
```
JSunde, line 24: Should this be a reference to a color resource?

ground/src/main/res/layout/task_chip_button_with_text_and_icon_transparent.xml
```diff
@@ -0,0 +1,28 @@
...
+  ~ limitations under the License.
+  -->
+
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools">
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/button"
+    style="@style/TaskChipButton.Transparent.TextAndIcon"
+    app:iconTint="#216C2E"
```
JSunde: Color resource?

ground/src/main/res/values/colors.xml
```diff
@@ -37,4 +37,14 @@
...
   <color name="colorDataSubmissionConfirmationGreen">#216C2E</color>
+
+  <!-- Task UI Components -->
+  <color name="taskBottomControlsBackground">#FCFDF7</color>
+  <color name="taskBtnDarkGreen">#47854B</color>
+  <color name="taskBtnTextDarkGreen">#FFFFFF</color>
+  <color name="taskBtnLightGreen">#D5E8D0</color>
+  <color name="taskBtnTextLightGreen">#000000</color>
+  <color name="taskBtnTransparentStrokeColor">#747775</color>
+  <color name="taskBtnTextTransparent">#216C2E</color>
```
JSunde, line 48: A couple of these are the same as the colorDataSubmissionConfirmationGreen and colorJobCardCollectButton. I wonder if it would make sense to define a new color with a semantic name like colorSecondary or colorPrimaryLight that could be used for all of those colors. That we if we update it in the future all the corresponding UI would be updated automatically.

JSunde (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
shobhitagarwal1612: Can you share some link which has more info about it?

ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
JSunde: Yea I think maybe OUTLINED as the name would be more clear

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
JSunde: It's the more modern way to build UIs in Android, it's a bit of a steep learning curve, but I think it would make it easier to do what you want here.

https://developer.android.com/jetpack/compose?gclid=CjwKCAjw_MqgBhAGEiwAnYOAeiTR-73YT5sPVtYV0cZOEUq0ZGhdVcfyYaagJaqZB0-munJvxYO5TRoCpmsQAvD_BwE&gclsrc=aw.ds

ground/src/main/java/com/google/android/ground/ui/datacollection/components/ButtonAction.kt
shobhitagarwal1612: Done

ground/src/main/res/drawable/task_chip_image_button.xml
shobhitagarwal1612: This color is for background. I don't feel too strongly about moving this to colors.xml as it is independent of which button theme is being used (dark green, light green or outlined)

ground/src/main/res/values/colors.xml
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
shobhitagarwal1612: Discussed offline. Added a note to the list of tasks. Needs to be taken up separately.

ground/src/main/res/values/colors.xml
shobhitagarwal1612: Now done.

ground/src/main/res/drawable/task_chip_image_button.xml
JSunde: Well I think it's probably worth having a colorBackground resource in case we end up changing the background color at some point. Maybe we will eventually want it to be very faintly tinted with the primary color and then we would only have to update the one resource to do that.

ground/src/main/res/values/colors.xml
JSunde: Thanks!

ground/src/main/res/drawable/task_chip_image_button.xml
shobhitagarwal1612: Done

PR #1629, opened 2023-03-16
Draft change getting Jetpack Compose working in Ground

JSunde:


shobhitagarwal1612:
Thanks @JSunde! Can we keep this PR around until I get to integrating this with the task buttons?

gino-m:
@JSunde Is this PR still valid? If not, can we close it for now?

JSunde:
I think it's still valid, and might be useful as a reference if we decide we want to start using Jetpack Compose for any new UIs.

We can always re-open it if we close it, right? If so then I think we could close it and can dig it up for reference later.

gino-m:
Sg, I'll close it for now. I've created a bug to tracked and linked to this PR so we can find it.

Issue #1581, opened 2023-02-21
[Code health] Migrate data layers to coroutines


gino-m:
- [ ] Local DAOs
- [ ] Local data store
- [x] Remote data store
- [ ] Manager / system services
- [ ] Proceed per use case / feature
- [ ] Worker

Be sure to follow [best practices](https://developer.android.com/kotlin/coroutines/coroutines-best-practices) and ensure suspend functions are main-safe.

gino-m:
@scolsen @shobhitagarwal1612 @JSunde Shall we split this work up somehow? Maybe we'll be more likely to get it done in a reasonable time frame and it will accelerate other work?

shobhitagarwal1612:
Just curious, how would it speed up other things?

scolsen:
I'm happy to take at least some of the DAOs and data store as I had already experimented with converting some of them: https://github.com/google/ground-android/pull/1405.

gino-m:
Closing in favor of #1593.

PR #1633, opened 2023-03-17
Migrate fetch of surveys from local db to Coroutines Flow

gino-m:
First step towards migrating local db layer to Coroutines (#1581).

@JSunde PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/local/LocalDatabaseModule.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/LocalDatabaseModule.kt
@@ -37,11 +37,10 @@ object LocalDatabaseModule {
   fun localDatabase(
     @ApplicationContext context: Context,
     @IoDispatcher ioDispatcher: CoroutineDispatcher
-  ): LocalDatabase {
-    return Room.databaseBuilder(context, LocalDatabase::class.java, Config.DB_NAME)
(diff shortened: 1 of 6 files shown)


ground/src/main/java/com/google/android/ground/persistence/local/LocalDatabaseModule.kt
```diff
@@ -18,22 +18,30 @@ package com.google.android.ground.persistence.local
...
+  fun localDatabase(
+    @ApplicationContext context: Context,
+    @IoDispatcher ioDispatcher: CoroutineDispatcher
+  ): LocalDatabase {
     return Room.databaseBuilder(context, LocalDatabase::class.java, Config.DB_NAME)
       .fallbackToDestructiveMigration() // TODO(#128): Disable before official release.
+      // Run queries and transactions on background I/O thread.
+      .setQueryExecutor(ioDispatcher.asExecutor())
       .build()
   }
```
JSunde, line 46: nit: this could be a single-expression function

JSunde (approved):

ground/src/main/java/com/google/android/ground/persistence/local/LocalDatabaseModule.kt
gino-m: Fixed, ty.

PR #1637, opened 2023-03-21
Add Suggest Loi Card View

JSunde:
Towards #1541

Adds a new card view for Suggest Loi cards and manages them alongside the Loi cards in the MapCardAdapter

PR #1639, opened 2023-03-22
Load Suggest Loi Tasks and render cards for them

JSunde:
Towards #1637


JSunde:
For some reason there don't seem to be any jobs loaded in the SuggestLoiJobs LiveData.

I already added a new "Suggest LOI" job to the survey with the suggestTaskType of "point", but there don't seem to be any jobs loaded with that suggestionTaskType.

@shobhitagarwal1612 @scolsen do either of you have any ideas? I'm guessing I maybe missed something in the creation of Job objects, but I'm not sure.

shobhitagarwal1612:
Is the new object getting saved to local data store? LoiCards data is read from cached local storage and not directly from firebase.

JSunde:
Are jobs not persisted to the local data store?

shobhitagarwal1612:
The model is serialised and deserialized when storing locally. So, you might have missed adding new fields in the existing job model

JSunde:
Do you have any idea  where that would be? I'm not as familiar with the local data store, and I thought I updated all of the places I needed to:
- JobConverter
- Job
- JobNestedObject

shobhitagarwal1612:
That would be JobEntity class

JSunde:
Ah okay, thanks! It looks like I was missing the suggestLoiTaskType in the JobEntity and some logic in ConvertExt.

It looks like when the app first loads up the Suggest LOI job is present, but then it is missing on subsequent loads, any ideas what I might be missing now?

shobhitagarwal1612:
> It looks like when the app first loads up the Suggest LOI job is present, but then it is missing on subsequent loads, any ideas what I might be missing now?

I'll take a look at the code. My first guess would be to look at the filtering logic and ensure we are not implicitly removing the suggest LOIs.

shobhitagarwal1612:
Is this PR ready for a review?

PR #1641, opened 2023-03-29
Loi Card and Suggest Loi Card UX Tweaks

JSunde:
Updates to cards to bring in line with latest mocks.

image
image


shobhitagarwal1612:
QQ: In the above screenshot, the suggest LOI card isn't highlighted. Is that intentional?

JSunde:
> QQ: In the above screenshot, the Suggest LOI card isn't highlighted. Is that intentional?

Not entirely, it just hasn't been implemented yet. I believe the LOI cards are focused based on the active Location of Interest, which won't work for the Suggest LOI cards, so we'll have to manage the focused state differently for those.

shobhitagarwal1612 (approved):

Issue #1513, opened 2023-02-04
[Code health] Convert instrumentation tests into unit tests


shobhitagarwal1612:
```
java.lang.RuntimeException: Unable to create service com.google.android.ground.persistence.remote.firebase.FirebaseMessagingService: java.lang.IllegalStateException: The component was not created. Check that you have added the HiltAndroidRule.
     FATAL EXCEPTION: main
Process: com.google.android.ground, PID: 14552
java.lang.RuntimeException: Unable to create service com.google.android.ground.persistence.remote.firebase.FirebaseMessagingService: java.lang.IllegalStateException: The component was not created. Check that you have added the HiltAndroidRule.
	at android.app.ActivityThread.handleCreateService(ActivityThread.java:4198)
	at android.app.ActivityThread.access$1500(ActivityThread.java:237)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1932)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7664)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
Caused by: java.lang.IllegalStateException: The component was not created. Check that you have added the HiltAndroidRule.
	at dagger.hilt.internal.Preconditions.checkState(Preconditions.java:83)
	at dagger.hilt.android.internal.testing.TestApplicationComponentManager.generatedComponent(TestApplicationComponentManager.java:96)
	at dagger.hilt.android.testing.HiltTestApplication.generatedComponent(HiltTestApplication.java:50)
	at dagger.hilt.EntryPoints.get(EntryPoints.java:59)
	at dagger.hilt.android.internal.managers.ServiceComponentManager.createComponent(ServiceComponentManager.java:70)
	at dagger.hilt.android.internal.managers.ServiceComponentManager.generatedComponent(ServiceComponentManager.java:58)
	at com.google.android.ground.persistence.remote.firebase.Hilt_FirebaseMessagingService.generatedComponent(Hilt_FirebaseMessagingService.java:56)
	at com.google.android.ground.persistence.remote.firebase.Hilt_FirebaseMessagingService.inject(Hilt_FirebaseMessagingService.java:50)
[...]

PR #1643, opened 2023-04-01
Converts MapTypeDialogFragmentTest instrumentation test to unit test

shobhitagarwal1612:
Towards #1513

@scolsen  PTAL?


ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -149,11 +143,6 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
     watermark.layoutParams = params
   }
 
-  override fun onCreate(savedInstanceState: Bundle?) {
-    super.onCreate(savedInstanceState)
-    customCap = CustomCap(bitmapUtil.bitmapDescriptorFromVector(R.drawable.ic_endpoint))
```
scolsen, line 149: Is the removal of the custom drawable here meant to be in this PR? Is this part of the revamp work for the Polygon UI?

ground/src/test/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragmentTest.kt
```diff
@@ -0,0 +1,89 @@
...
+    hiltActivityScenario()
+      .launchFragment<HomeScreenMapContainerFragment>(
+        argsBundle,
+        preTransactionAction = {
+          fragment = this as HomeScreenMapContainerFragment
+          this.also {
+            it.viewLifecycleOwnerLiveData.observeForever { viewLifecycleOwner ->
+              if (viewLifecycleOwner != null) {
+                // Bind the controller after the view is created but before onViewCreated is called.
+                Navigation.setViewNavController(fragment.requireView(), navController)
```
scolsen, line 82: No action to take but note to self: since we also use this in some of the data collection fragment tests and might need it in the future, maybe we should wrap this pattern into a function, something like `launchFragementWithNavigator(FragmentType, NavController)`

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
shobhitagarwal1612: This code is unused at the moment and was also preventing the unit test from running. While working on Polygon UI, I'll have to figure out how to implement it correctly.

ground/src/test/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragmentTest.kt
shobhitagarwal1612: Ack!

PR #1646, opened 2023-04-05
Defer Submission creation in DataCollectionViewModel

JSunde:
Towards #1541

This will help in the addition of the Suggest LOI Data Collection Flow because now we don't need the LOI to already exist before starting the flow, now the Submission is created when the user presses continue on the last task in the job.

There are still a few steps left I will address in follow up PRs and I added some TODOs to keep track of that.


scolsen:
> Any ideas why I'd be getting this LocalDataConsistencyException in the DataCollectionFragmentTest?
> 
> ```
> com.google.android.ground.persistence.local.LocalDataConsistencyException: Unknown jobId LocationOfInterestEntity(id=loi id, surveyId=SURVEY, jobId=JOB, state=DEFAULT, created=AuditInfoEntity(user=UserDetails{id=user_id, email=user@gmail.com, displayName=User}, clientTimestamp=1680730632952, serverTimestamp=null), lastModified=AuditInfoEntity(user=UserDetails{id=user_id, email=user@gmail.com, displayName=User}, clientTimestamp=1680730632953, serverTimestamp=null), geometry=GeometryWrapper(point=Point(coordinate=Coordinate(x=0.0, y=0.0)), polygon=null, multiPolygon=null)).jobId in location of interest LocationOfInterestEntity(id=loi id, surveyId=SURVEY, jobId=JOB, state=DEFAULT, created=AuditInfoEntity(user=UserDetails{id=user_id, email=user@gmail.com, displayName=User}, clientTimestamp=1680730632952, serverTimestamp=null), lastModified=AuditInfoEntity(user=UserDetails{id=user_id, email=user@gmail.com, displayName=User}, clientTimestamp=1680730632953, serverTimestamp=null), geometry=GeometryWrapper(point=Point(coordinate=Coordinate(x=0.0, y=0.0)), polygon=null, multiPolygon=null)).id
> ```

Looks like you solved this but now there's an issue with the view hierarchy?

JSunde:
What view hierarchy issue are you seeing? I think I'm still seeing the unknown jobId issue.

JSunde:
I think I was able to get to the bottom of the issue. The job id was being overridden in the job field stored in the submission.

The tests are passing now.

JSunde:
Oh oops, turns out I was only running the one test locally and broken something else in DataCollectionFragmentTest.

JSunde:
Okay, now I think it should be fixed.

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -67,41 +69,44 @@ import org.robolectric.shadows.ShadowToast
...
 
+  @Inject lateinit var activateSurvey: ActivateSurveyUseCase
+  @Inject lateinit var fakeRemoteDataStore: FakeRemoteDataStore
   @BindValue @Mock lateinit var submissionRepository: SubmissionRepository
   @Captor lateinit var taskDataDeltaCaptor: ArgumentCaptor<List<TaskDataDelta>>
   lateinit var fragment: DataCollectionFragment
 
   @Before
   override fun setUp() {
     super.setUp()
```
shobhitagarwal1612, line 80: this is now just calling super(). can be removed

JSunde: Done

JSunde:
It seems like one of the tests is failing now because the Loi caption is null, but the only 2 LOIs used in FakeData both have a caption being set.

Any ideas what might be going on there @shobhitagarwal1612, @scolsen?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -48,33 +49,63 @@ import kotlin.collections.component2
...
 ) : AbstractViewModel() {
 
-  val submission: @Hot(replays = true) LiveData<Submission>
-  val jobName: @Hot(replays = true) LiveData<String>
-  val loiName: @Hot(replays = true) LiveData<String>
+  // TODO(#1541): Set loiId once Suggest LOI task is completed.
+  private val loiId: StateFlow<String?> = MutableStateFlow(savedStateHandle["locationOfInterestId"])
+  private val activeSurvey: Survey = requireNotNull(surveyRepository.activeSurvey)
+  private val job: Job =
+    activeSurvey.getJob(requireNotNull(savedStateHandle["jobId"])).orElseThrow()
```
shobhitagarwal1612, line 79: I feel that the previous approach of using generated keys instead of using hard-coded string values is less error prone. Is there a way to share the key values between here and the nav_graph?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -48,33 +49,63 @@ import kotlin.collections.component2
...
+
+  val surveyId: String = surveyRepository.lastActiveSurveyId
```
shobhitagarwal1612, line 81: Should we pass the surveyId as well similar to loiId and jobId?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -48,33 +49,63 @@ import kotlin.collections.component2
...
+    MutableStateFlow(job.name ?: "").stateIn(viewModelScope, SharingStarted.Lazily, "")
+  val loiName: StateFlow<String> =
+    loiId
+      .flatMapLatest { id ->
+        if (id == null) flowOf("")
+        else
+          locationOfInterestRepository
+            .getOfflineLocationOfInterest(surveyId, id)
+            .toFlowable()
+            .asFlow()
```
shobhitagarwal1612, line 100: IMO, we should consider moving this to the repository so that we can incrementally remove usages from other places as well. Same comment for the above usage of `.asFlow()` as well.

ground/src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
```diff
@@ -286,34 +283,48 @@ class DataCollectionFragmentTest : BaseHiltTest() {
...
-
-    verify(submissionRepository)
-      .createOrUpdateSubmission(any(), capture(taskDataDeltaCaptor), eq(true))
-    assertThat(taskDataDeltaCaptor.value[0]).isEqualTo(expectedTaskDataDeltas)
+    //    onView(allOf(withText("Continue"), isDisplayed())).perform(click())
+    //    advanceUntilIdle()
+    //
+    //    verify(submissionRepository)
+    //      .createOrUpdateSubmission(any(), capture(taskDataDeltaCaptor), eq(true))
+    //    assertThat(taskDataDeltaCaptor.value[0]).isEqualTo(expectedTaskDataDeltas)
```
shobhitagarwal1612, line 291: Please cleanup

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -70,12 +72,17 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
...
+    lifecycleScope.launch {
+      viewModel.submission.filterNotNull().collect { submission: Submission ->
+        loadTasks(submission.job.tasksSorted)
+      }
+    }
     viewModel.currentPosition.observe(viewLifecycleOwner) { onTaskChanged(it) }
     viewModel.currentTaskDataLiveData.observe(viewLifecycleOwner) { onTaskDataUpdated(it) }
   }
 
+
```
shobhitagarwal1612, line 85: nit: extra line

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: I don't have a strong opinion here. The only issue I had is that we are coupling the active survey with data collection fragment. In theory, DataCollectionFragment shouldn't worry about what's currently loaded or not. I'll defer to your judgement here. Thanks!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: I don't have any experience with using `SavedStateHandle`. Will defer to your final call here as well.

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: I think since we need to get the activeSurvey from the SurveyRepository anyway I'll stick with getting the survey id from the SurveyRepository,

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
JSunde: ACK, will leave as is and we can revisit if it becomes an issue.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
JSunde: Removed

PR #1647, opened 2023-04-08
[Data Collection] Update UX for polygon drawing as per latest mocks

### LINKED ISSUE #1351: Integrate polygon drawing with data collection flow

shobhitagarwal1612:
Towards #1621 
Towards #1351 

Until now, we used to create partial polygon and use helper methods like isComplete() to ensure that the geometry is correct. This worked fine when the rendered UI was same for partial and complete drawn polygon. But with the updated UX, the rendering needs to change as well. So, the proposed change in this PR is to create geometry object LineString, LinearRing or Polygon based on current state of drawn area.

 * LineString - List of points
 * LinearRing - A LineString whose first and last point is same
 * Polygon - A shell (LinearRing) along with a list of holes within the shell ring

Additional idea: Introduce a GeometryValidator to ensure the well-formedness of the geometry. Then we won't have to check after the model has been created whether the geometry actually follows its definition or not. LinearRing would validate itself during creation (in the constructor).

When the user starts drawing, a LineString would be generated. When the first and last coordinates are same, then LineString would be converted to a LinearRing. After user presses "Complete", then the LinearRing would be converted into a Polygon without any holes.

Rendering: 
* LineString & LinearRing would be rendered in a similar way using `Polyline`. The support for adding/removing polylines needs to be added first.
* Polygon rendering is already supported.

Things done in this PR:
 * Adds TRANSPARENT theme for action buttons
 * Adjusts theme for all buttons based on latest mocks
 * Adds support for rendering polylines
 * Update "draw an area" task
 * Adds validator for geometry models

[Design Doc](https://docs.google.com/document/d/15CGT-9LvgarzBvrC8kPPAEHlrGuZCqVL0i9lpmya2dU/r/0-OLrHylzybpz6_mkioEm7og/edit#)

https://user-images.githubusercontent.com/8918023/230763087-430d34ee-a7ce-4389-8ec5-072bc2f0cfb5.mp4

@JSunde @scolsen  PTAL?


ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /** Returns true if the current geometry is closed. */
+  fun Geometry?.isClosedGeometry(): Boolean = this is Polygon || this is LinearRing
+
+  /** Returns true of the current list of vertices can generate a polygon. */
+  fun List<Coordinate>.isComplete(): Boolean {
+    if (size < 4) return false
+    return isFirstAndLastSame()
+  }
```
JSunde, line 39: nit: could be simplified slightly

```
isComplete() = if (size < 4) false else isFirstAndLastSame()
```

shobhitagarwal1612: Done. Thanks

ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.ground.model.geometry
+
+object GeometryValidator {
+  private fun List<Any>.isFirstAndLastSame(): Boolean = firstOrNull() == lastOrNull()
+
+  /** Validates that the current [LinearRing] is well-formed. */
+  fun LinearRing.validate() {
```
scolsen, line 23: Since we'll probably use variations on this for different types of Geometries, I think we can define this as a function in the Geometry interface. A trivial default implementation would just be the no-op Unit function (a naked `return`).

ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.ground.model.geometry
+
+object GeometryValidator {
+  private fun List<Any>.isFirstAndLastSame(): Boolean = firstOrNull() == lastOrNull()
```
scolsen, line 20: I don't think we need this for now. I'd just write the comparison directly in the impl of `isComplete`.

ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+    if (coordinates.isEmpty()) {
+      return
+    }
+    if (!coordinates.isFirstAndLastSame()) {
+      error("Invalid linear ring")
+    }
+  }
+
+  /** Returns true if the current geometry is closed. */
+  fun Geometry?.isClosedGeometry(): Boolean = this is Polygon || this is LinearRing
```
scolsen, line 33: Let's combine this condition with the `isComplete`:

```kotlin
fun Geometry.isClosed: Boolean = 
  when(this) {
    is Polygon, LinearRing -> size >= 4 && coordinates.firstOrNull() == coordinates.lastOrNull()
    else -> false
  }
```

This way, in the fragment etc. you can just write:

```kotlin
getButton(ButtonAction.ADD_POINT).showIfTrue(!isClosed)
```

shobhitagarwal1612: The idea behind validation was that it will fail the creation of geometry. So, just the type of geometry is enough to assert whether it is a closed or open geometry. The above `isClosed()` is for `Geometry` and `isComplete()` is a helper method on `List<Coordinate>`.

ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/model/geometry/GeometryValidator.kt
shobhitagarwal1612: Done

scolsen (approved):

PR #1661, opened 2023-04-28
Fix multiselect task

### LINKED ISSUE #1658: Response for MultipleChoice task with MULTI_SELECT cardinality doesn't work

shobhitagarwal1612:
Fixes #1658

@scolsen @JSunde  PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/SelectMultipleOptionAdapter.kt
```diff
@@ -42,26 +42,25 @@ class SelectMultipleOptionAdapter(
...
-  override fun handleItemStateChanged(position: Int, selected: Boolean) {
+  private fun handleItemStateChanged(position: Int) {
     if (position in selectedPositions) {
       selectedPositions.remove(position)
     } else {
       selectedPositions.add(position)
     }
 
-    viewModel.updateResponse(
+    handleOptionSelected.invoke(
```
scolsen, line 57: Should be able to just call it directly instead of using `invoke`: `handleOptionSelected(options.filterIndexed {}...)`

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/SelectOneOptionAdapter.kt
```diff
@@ -47,24 +47,26 @@ class SelectOneOptionAdapter(
...
-      selectedIndex = holder.adapterPosition
-      viewModel.updateResponse(listOf(options[selectedIndex]))
+      handleItemStateChange(it, holder.adapterPosition)
+    }
+  }
+
+  private fun handleItemStateChange(view: View, position: Int) {
+    val oldPosition = selectedIndex
+    selectedIndex = position
+    handleOptionSelected.invoke(listOf(options[selectedIndex]))
```
scolsen, line 58: Should be able to just call it like a function `handleOptionSelected(listOf(options[selectedIndex]))`

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/SelectOneOptionAdapter.kt
```diff
@@ -29,7 +29,7 @@ import com.google.android.ground.ui.datacollection.tasks.multiplechoice.SelectOn
  */
 class SelectOneOptionAdapter(
   private val options: List<Option>,
-  private val viewModel: MultipleChoiceTaskViewModel
+  private val handleOptionSelected: (List<Option>) -> Unit
```
scolsen, line 32: Since this only operates on one value, change the signature to `Option -> Unit`. We can still use the same function in the viewmodel, but I'd change the signature here for the sake of correctness, and this way the adapter does not need to know about the VM impl detail; e.g. in the VM, we can do the list wrapping:

```kotlin
updateResponse(opts: List<Option> ...)

updateResponse(option: Option ...) {
  updateResponse(listOf(option))
}
```

scolsen (approved):
some optional suggestions

PR #1672, opened 2023-05-04
Use non-gms objects outside of gms package

### LINKED ISSUE #1671: [Code health] Remove dependency of "com.google.android.gms.maps" from map-provider agnostic code
e.g. LocationOfInterestRepository uses `com.google.android.gms.maps.model.LatLngBounds` instead of `com.google.android.ground.ui.map` which makes the overall code tightly coupled with GMS APIs. This leaves us in a state where half of the logic is written in GMS models and remaining in app specific models. 

It'd be nice if we can remove dependency of these models from our core package and introduce a layer of abstraction before using these APIs

shobhitagarwal1612:
Fixes #1671

Verified that the changes don't introduce a regression by running the app locally and testing following scenarios:
1. Fresh install
2. Moving the center, closing the app and then reopening the app
3. Bottom LOI cards are correctly shown when inside bounds at correct zoom level.

@scolsen @gino-m  PTAL?


shobhitagarwal1612:
@scolsen Gentle ping!

ground/src/main/java/com/google/android/ground/model/geometry/Coordinate.kt
```diff
@@ -19,6 +19,7 @@ import kotlinx.serialization.Serializable
 
 /** A coordinate on a two-dimensional Cartesian plane. */
 @Serializable
-data class Coordinate(val x: Double, val y: Double) : Comparable<Coordinate> {
-  override fun compareTo(other: Coordinate): Int = compareValuesBy(this, other, { it.x }, { it.y })
+data class Coordinate(val latitude: Double, val longitude: Double) : Comparable<Coordinate> {
```
scolsen, line 22: I would update the doc comment as well to state that we explicitly mean a coordinate in ref to the Earth and using https://en.wikipedia.org/wiki/Decimal_degrees to represent lat/lng (note that originally this was just a plain Real-numbered cartesian coordinate, so the concept is different (and better!) now).

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
```diff
@@ -92,8 +87,8 @@ data class LinearRing(val coordinates: List<Coordinate>) : Geometry {
    * Returns a *synthetic* coordinate containing the maximum x and y coordinate values of this ring.
    */
   private fun maximum(): Coordinate {
-    val maximumX = this.coordinates.maxOfOrNull { it.x }
-    val maximumY = this.coordinates.maxOfOrNull { it.y }
+    val maximumX = this.coordinates.maxOfOrNull { it.latitude }
```
scolsen, line 90: I would change these to use the new terminology lat/lng. Also, X should map to Lng and Y to Lat.

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
```diff
@@ -102,8 +97,8 @@ data class LinearRing(val coordinates: List<Coordinate>) : Geometry {
    * Returns a *synthetic* coordinate containing the minimum x and y coordinate values of this ring.
    */
   private fun minimum(): Coordinate {
-    val minimumX = this.coordinates.minOfOrNull { it.x }
-    val minimumY = this.coordinates.minOfOrNull { it.y }
+    val minimumX = this.coordinates.minOfOrNull { it.latitude }
```
scolsen, line 100: Same as above.

ground/src/main/java/com/google/android/ground/ui/offlinebasemap/viewer/OfflineAreaViewerFragment.kt
```diff
@@ -63,7 +62,7 @@ class OfflineAreaViewerFragment @Inject constructor() : Hilt_OfflineAreaViewerFr
   override fun getMapViewModel(): BaseMapViewModel = viewModel
 
   private fun panMap(offlineArea: OfflineArea) {
-    mapFragment.viewport = offlineArea.bounds.toModelObject()
+    mapFragment.viewport = offlineArea.bounds
```
scolsen, line 65: I might be missing something, but where do we actually perform the conversion back to GMS objects in the Map Fragment? Seems like we currently don't do this anywhere?

scolsen (commented):
This is so a major improvement, thank you! Just a few comments.

ground/src/main/java/com/google/android/ground/ui/offlinebasemap/viewer/OfflineAreaViewerFragment.kt
shobhitagarwal1612: Only when we are setting the camera position or checking for points within a bound.

scolsen (approved):

PR #1681, opened 2023-05-16
sync: retry on failure

### LINKED ISSUE #1468: [Background sync] Report failures and retry

scolsen:
Updates the survey sync worker and tileset download worker to retry sync
on certain failures. In addition, adds unit tests to verify the workers
return the expected results.

Fixes #1468


ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
```diff
@@ -44,20 +44,23 @@ constructor(
...
-      syncSurvey(surveyId)
+    try {
+      // It's ok to block here since WorkManager calls doWork() on a background thread.
+      runBlocking {
+        Timber.d("Syncing survey $surveyId")
+        syncSurvey(surveyId)
+      }
+    } catch (e: Throwable) {
+      Timber.e("error syncing survey in the background", e)
+      return Result.retry()
```
shobhitagarwal1612, line 55: How many times will this retry the work request?

ground/src/test/java/com/google/android/ground/persistence/sync/SurveySyncWorkerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+          inputData = workDataOf(Pair(SURVEY_ID_PARAM_KEY, SURVEY.id)),
+        )
+        .setWorkerFactory(factory)
+        .build()
+    val result = worker.doWork()
+    assertThat(result).isEqualTo(Result.success())
+  }
+
+  @Test
+  fun doWork_RetriesOnFailure() = runWithTestDispatcher {
```
shobhitagarwal1612, line 100: Will this retry the failed work indefinitely?

ground/src/test/java/com/google/android/ground/persistence/sync/SurveySyncWorkerTest.kt
```diff
@@ -0,0 +1,114 @@
...
+  fun doWork_FailsOnNoInput() {
+    val worker =
+      TestWorkerBuilder<SurveySyncWorker>(
+          context,
+          executor,
+          inputData = workDataOf(Pair(SURVEY_ID_PARAM_KEY, null)),
+        )
+        .setWorkerFactory(factory)
+        .build()
+    val result = worker.doWork()
```
shobhitagarwal1612, line 79: Can we create an abstract class for creating the worker along with the dependencies and calling `.doWork()`? 

Something like:

```
fun createWorkerAndDoWork<T>(args: Pair) {}
```

shobhitagarwal1612 (changes requested):

ground/src/test/java/com/google/android/ground/persistence/sync/SurveySyncWorkerTest.kt
scolsen: If we wind up needing more of these that sounds like a good idea; for now I don't think the extra indirection is worth it yet.

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
shobhitagarwal1612: Thanks

shobhitagarwal1612 (approved):

PR #1685, opened 2023-05-20
Fix photo upload flow

### LINKED ISSUE #1684: PhotoSyncWorker fails to upload the photos
LocalMutationSyncWorker schedules a PhotoSyncWorker by getting the response from mutation and getting the path from it. The problem is that it generates the path using TaskData.tostring(), assuming the underlying classes have correctly overridden it.

shobhitagarwal1612:
Also add valid filename check to add more valuable error messages in case an invalid path is passed to the UserMediaRepository

Fixes #1684

@JSunde  @scolsen  PTAL?


scolsen (approved):
Nice work!

PR #1699, opened 2023-06-08
Initial pass at switching to use new theme colors

JSunde:


shobhitagarwal1612:
Are there any major changes in the UI?

JSunde:
image
image
image
image

shobhitagarwal1612:
I believe there used to be a light green background behind the action buttons.

shobhitagarwal1612:
Can you please check the latest figma mocks?

JSunde:
> I believe there used to be a light green background behind the action buttons.

Which action buttons are you referring to?

JSunde:
> > I believe there used to be a light green background behind the action buttons.
> 
> Which action buttons are you referring to?

I think that the changes I've made so far have been swapping colors for other colors that are very close, so I think if there's an issue with a component being the wrong color we should maybe follow up on that separately so I think I'll go ahead with this CL as is.

shobhitagarwal1612 (approved):

PR #1700, opened 2023-06-09
Switch more colors to new theme

JSunde:


JSunde:
image
image
image

JSunde:
> From the above screenshot, it appears that background color is now light green. Is that correct?

Yes I believe that's correct, I'm using `md_theme_surface` from the new theme which is defined as F9FAF4 which is lightly tinted green.

JSunde:
And colorSurface is generally used for the default background in Android apps.

shobhitagarwal1612 (approved):

PR #1709, opened 2023-06-14
Update LICENSE

### LINKED ISSUE #1708: [Documentation] Improve language of APACHE License by adding
I am a student from OSU and was hoping to contribute to this project. When looking through your documentation I noticed that the language used in the license file could use some clarification relating to the use of third party trademarks/copyrighted materials. This could help new developers (like myself) avoid making mistakes when utilizing these materials to make contributions.

OSURandle:
Fixes #1708

This is the update I propose as an appendix to the existing license documentation.

@... PTAL?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,6 +186,21 @@
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
+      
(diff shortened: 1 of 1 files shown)


gino-m:
See comment on #1708.

PR #1711, opened 2023-06-16
Improve map clustering UX

### LINKED ISSUE #1602: LOIs clustered even when zoomed in
LOIs shouldn't be clustered above the zoom level when job cards are shown

Screenshot_20230227_150600

@scolsen PTAL?

shobhitagarwal1612:
Fixes #1602 

 * Ensures markers and clusters aren't shown at the same zoom level
 * Clicking a cluster zooms to the LOI bounds contained within the cluster
 * When a survey is loaded for the first time, map camera pans to all LOIs

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
@@ -125,6 +125,9 @@ constructor(
   fun getLocationsOfInterestOnceAndStream(survey: Survey): Flowable<Set<LocationOfInterest>> =
     localLoiStore.getLocationsOfInterestOnceAndStream(survey)
 
+  suspend fun getAllGeometries(survey: Survey): List<Geometry> =
+    remoteDataStore.loadLocationsOfInterest(survey).map { it.geometry }
+
   /** Returns a flowable of all [LocationOfInterest] within the map bounds (viewport). */
   fun getWithinBoundsOnceAndStream(
     survey: Survey,
--- a/ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
@@ -38,6 +38,7 @@ import io.reactivex.Flowable
 import io.reactivex.Single
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.reactive.awaitFirst
 
 /**
  * Coordinates persistence and retrieval of [LocationOfInterest] instances from remote, local, and
@@ -121,12 +122,13 @@ constructor(
       MutationEntitySyncStatus.FAILED
     )
 
-  /** Returns a flowable of all [LocationOfInterest] for the currently active [Survey]. */
+  /** Returns a flowable of all [LocationOfInterest] for the given [Survey]. */
   fun getLocationsOfInterestOnceAndStream(survey: Survey): Flowable<Set<LocationOfInterest>> =
     localLoiStore.getLocationsOfInterestOnceAndStream(survey)
 
+  /** Returns a list of geometries associated with the given [Survey]. */
   suspend fun getAllGeometries(survey: Survey): List<Geometry> =
-    remoteDataStore.loadLocationsOfInterest(survey).map { it.geometry }
+    getLocationsOfInterestOnceAndStream(survey).awaitFirst().map { it.geometry }
 
   /** Returns a flowable of all [LocationOfInterest] within the map bounds (viewport). */
   fun getWithinBoundsOnceAndStream(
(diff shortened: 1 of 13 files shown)


ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -125,6 +125,9 @@ constructor(
   fun getLocationsOfInterestOnceAndStream(survey: Survey): Flowable<Set<LocationOfInterest>> =
     localLoiStore.getLocationsOfInterestOnceAndStream(survey)
 
+  suspend fun getAllGeometries(survey: Survey): List<Geometry> =
```
gino-m, line 128: Don't we already load all the features when opening a survey? Would we be doing it twice instead of once after this chance?

ground/src/main/java/com/google/android/ground/ui/map/CameraPosition.kt
```diff
@@ -18,16 +18,16 @@ package com.google.android.ground.ui.map
 import com.google.android.ground.model.geometry.Coordinate
 
 data class CameraPosition(
-  val target: Coordinate,
+  val target: Coordinate? = null,
   val zoomLevel: Float? = null,
   val isAllowZoomOut: Boolean = false,
   val bounds: Bounds? = null
```
gino-m, line 24: I didn't realize we have bounds here; essentially (`target`, `zoomLevel`) and `bounds` model the same info. Also, `isAllowZoomOut` is an arg to an event, not to the current camera state. Can you file a bug (against me?) to clean this up and add a TODO here?

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
```diff
@@ -96,14 +96,23 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
...
-      map.moveCamera(newPosition.target, zoomLevel)
+    val bounds = newPosition.bounds
+    val target = newPosition.target
+    var zoomLevel = newPosition.zoomLevel
+
+    if (target != null && zoomLevel != null && !newPosition.isAllowZoomOut) {
+      zoomLevel = max(zoomLevel, map.currentZoomLevel)
+    }
+
+    // TODO(Shobhit): Ensure that the order of preference is correct.
```
gino-m, line 107: See comment on `CameraPosition` to refactor so that it actually represents a non-duplicated state rather than an action with sparse params.

ground/src/main/java/com/google/android/ground/ui/map/gms/GmsExt.kt
```diff
@@ -33,5 +34,16 @@ object GmsExt {
 
   fun Bounds.center(): Coordinate = toGoogleMapsObject().center.toModelObject()
 
+  fun List<Geometry>.toBounds(): Bounds? {
+    val coordinates = this.map { it.vertices.first().coordinate }
```
gino-m, line 38: Is `vertices` is well-defined? e.g. what does it return for multipolygons with holes, for example?

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -194,7 +195,14 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), MapFragment {
         map.cameraPosition.zoom,
         featureColor
       )
-    clusterManager.setOnClusterItemClickListener(this::onClusterItemClick)
+    clusterManager.setOnClusterClickListener { cluster ->
```
gino-m, line 198: Since this lambda go a little more complex can we refactor into a fn?

gino-m (changes requested):
Thanks for the changes! A few suggestions and questions.

ground/src/main/java/com/google/android/ground/ui/map/CameraPosition.kt
shobhitagarwal1612: https://github.com/google/ground-android/issues/1712

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
shobhitagarwal1612: Added a TODO to fix this once the CameraPosition class is cleaned up

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
shobhitagarwal1612: Thanks for flagging this! Fixed to read the LOIs from local database.

ground/src/main/java/com/google/android/ground/ui/map/gms/GmsExt.kt
shobhitagarwal1612: For polygon with holes, we only return the list of vertices of shell.

gino-m (approved):
LGTM! 🚀

PR #1713, opened 2023-06-20
Clear Multiple Choice selection when user skips the task

JSunde:
I also made some UX tweaks to bring the Task buttons in line with the mocks.

image

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskFragment.kt
@@ -32,14 +32,21 @@ import dagger.hilt.android.AndroidEntryPoint
  */
 @AndroidEntryPoint(AbstractTaskFragment::class)
 class MultipleChoiceTaskFragment : Hilt_MultipleChoiceTaskFragment<MultipleChoiceTaskViewModel>() {
+  private lateinit var binding: MultipleChoiceTaskFragBinding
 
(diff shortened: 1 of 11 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskFragment.kt
```diff
@@ -32,16 +32,23 @@ import dagger.hilt.android.AndroidEntryPoint
...
-    val binding = MultipleChoiceTaskFragBinding.inflate(inflater)
+    binding = MultipleChoiceTaskFragBinding.inflate(inflater)
     setupMultipleChoice(binding.selectOptionList)
     return binding.root
   }
 
+  override fun onSkip() {
+    super.onSkip()
+
+    setupMultipleChoice(binding.selectOptionList)
```
shobhitagarwal1612, line 49: Instead of reinitializing the adapter, shouldn't we just update the list in the adapter?

shobhitagarwal1612 (approved):
Nice!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskFragment.kt
JSunde: There doesn't appear to be an API for that and I think it would be a bit more complicated if we added that API to the two MultipleChoice adapters because then we would either need to cast the RecyclerView adapter to the MultipleChoiceAdapter types, or a perhaps a new MutlipleChoiceAdapter interface. Either way I think it's simpler as is.

PR #1716, opened 2023-06-21
Add confirmation dialog when collected data will be discarded

JSunde:


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -120,7 +121,20 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
 
   protected fun addSkipButton() =
     addButton(ButtonAction.SKIP)
-      .setOnClickListener { onSkip() }
+      .setOnClickListener {
```
gino-m, line 124: Can we extract this event handler to a fun?

gino-m (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
JSunde: Resolved, went with onSkip()

JSunde:
> @JSunde I'm unable to test this feature on HEAD. Can you please double check that the dialog is visible for you when attempting to skip? Also, can we add one unit test for it?

Did you try skipping after data had been input? 

And yea I can look into adding a unit test for this.

shobhitagarwal1612:
Yes, I tried doing it with select type tasks

JSunde:
Oh huh, you're right. I think I had tested with the multiple choice task previously but not after changing the implementation slightly. It seems to work properly with the Text input task.

I'm guessing it's an issue with the way I'm checking the response text to see if the user has input any data.

PR #1724, opened 2023-06-24
Improve UX for survey selector screen in dark theme

shobhitagarwal1612:
Before | After
-------|------
before |  before

@JSunde  PTAL?


gino-m (approved):
Thanks for the fix!

Issue #1736, opened 2023-07-05
Replace uses of `Geometry.vertices` with appropriate fields on each subclass


gino-m:
- [ ] Ensure polygons, multipolygons, and points are rendering

PR #1737, opened 2023-07-05
Add method center() to geometry instead of always using the first vertex

shobhitagarwal1612:
Towards #1736

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
+++ b/ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
@@ -15,6 +15,8 @@
  */
 package com.google.android.ground.model.geometry
 
+import com.google.android.ground.ui.map.gms.GmsExt.center
+import com.google.android.ground.ui.map.gms.GmsExt.toBounds
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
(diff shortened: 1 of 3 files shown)


ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
```diff
@@ -120,6 +138,10 @@ data class LinearRing(val coordinates: List<Coordinate>) : Geometry {
     this.maximum() >= other.maximum() && this.minimum() <= other.minimum()
 }
 
+/** Returns the center coordinate of the bounding box from the given list of coordinates. */
+private fun List<Coordinate>?.centerOrError(): Coordinate =
+  this?.map { Point(it) }?.toBounds()?.center() ?: error("missing vertices")
```
gino-m, line 143: I think this could work as a temporary fix, but using the center of the bounds will break for oddly shaped polygons. That said, implementation is non-trivial ([example](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/algorithm/Centroid.java)), so we may want to save that for future work.

gino-m (approved):
LGTM + one comment for future work. Thanks!

ground/src/main/java/com/google/android/ground/model/geometry/Geometry.kt
shobhitagarwal1612: Added a note. Thanks!

PR #1738, opened 2023-07-05
UX tweaks to DropAPin task

JSunde:
-Added new type of Task view where the header is combined with the title bar
-Reordered bottom buttons to align with UX spec

Follow up changes will add the help text under the center marker and move the progress bar from the below the title bar, to above the bottom bar.

image


ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskViewFactory.kt
```diff
@@ -43,4 +44,21 @@ object TaskViewFactory {
     labelResId?.let { binding.headerLabel.setText(labelResId) }
     return TaskViewWithoutHeader(binding)
   }
+
+  /**
+   * creates a TaskView with a header that is an extension of the app bar
```
shobhitagarwal1612, line 49: nit: First letter of ktdoc should be capital and the sentence must end with a full stop.

shobhitagarwal1612 (approved):

Issue #1715, opened 2023-06-21
Migrate to COGs for offline map imagery


gino-m:
- [x] Show live overlay in offline area selector map.
- [x] Use downloader to download images to local file system.
- [x] Show images in local filesystem if available.
- [ ] #1820
- [ ] #1788 
- [x] Delete remnants of mbtiles downloader and viewer.

For this phase we can reuse the basecmap url specified in the survey

gino-m:
Closing in favor of fine-grained tracking issues #1820 and #1788

PR #1740, opened 2023-07-07
Refactor: Abstract offline map imagery MOG data sources

gino-m:
Towards #1715.


gino-m:
@shobhitagarwal1612 PTAL?

PR #1744, opened 2023-07-10
[Rx -> Coroutine Migration] : Use coroutine for get/set user in local db

shobhitagarwal1612:
Towards #1581 

Verified by running the app locally and verifying that user details are correctly shown in the navigation drawer.

@gino-m @JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/local/room/dao/UserDao.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/room/dao/UserDao.kt
@@ -23,7 +23,14 @@ import io.reactivex.Maybe
 
 @Dao
 interface UserDao : BaseDao<UserEntity> {
+
   @Transaction
   @Query("SELECT * FROM user WHERE id = :id")
+  @Deprecated("Use findByIdSuspend")
   fun findById(id: String): Maybe<UserEntity>
+
+  @Transaction
+  @Query("SELECT * FROM user WHERE id = :id")
+  // TODO(#1581): Rename to findById once all existing usages are migrated to coroutine.
+  suspend fun findByIdSuspend(id: String): UserEntity?
 }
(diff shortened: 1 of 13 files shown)


ground/src/main/java/com/google/android/ground/persistence/local/room/dao/UserDao.kt
```diff
@@ -23,7 +23,14 @@ import io.reactivex.Maybe
 
 @Dao
 interface UserDao : BaseDao<UserEntity> {
+
   @Transaction
   @Query("SELECT * FROM user WHERE id = :id")
+  @Deprecated("Use findByIdSuspend")
   fun findById(id: String): Maybe<UserEntity>
```
JSunde, line 30: Would it make sense to name this findByIdSync instead and then the non-deprecated function can have the simpler findById name?

Same goes for other functions in this PR

JSunde (changes requested):

ground/src/main/java/com/google/android/ground/persistence/local/room/dao/UserDao.kt
shobhitagarwal1612: I've already marked it as deprecated. So, the plan is to iteratively remove it soon. Calling it sync would be inconsistent as rest of the codebase is still not using that naming convention. Thoughts?

JSunde: Yea I was just thinking about the case where it takes some time for us to get around to removing the deprecated method, then it would be nice if the new function had the cleaner name, but that's a good point that it may be confusing since the rest of the codebase is not using that convention.

SGTM as is

shobhitagarwal1612: Thanks! Can you please stamp approve?

JSunde: Oops, I thought I did

JSunde (approved):

PR #1751, opened 2023-07-19
[DRAFT] Synchronously download tiles to local device

gino-m:
Towards #1715.


gino-m:
@shobhitagarwal1612 @JSunde PTAL?

ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
```diff
@@ -203,4 +209,49 @@ constructor(
...
+    // TODO(#1754): Create a factory and inject rather than instantiating here. Add tests.
+    return MogClient(mogCollection)
+  }
+
+  /**
+   * Returns the URL of the first tile source in the current survey, or throws an error if no survey
+   * is active or if no tile sources are defined.
+   */
+  private suspend fun getFirstTileSourceUrl(): String {
+    val survey = surveyRepository.activeSurveyFlowable.asFlow().firstOrNull()?.orElse(null)
```
shobhitagarwal1612, line 253: Is there any particular reason for using `.activeSurveyFlowable` instead of `.activeSurvey`?

ground/src/main/java/com/google/android/ground/ui/map/gms/mog/MogTileDownloader.kt
```diff
@@ -27,19 +27,22 @@ import java.io.File
...
       val (x, y, zoom) = tile.metadata.tileCoordinates
       val path = File(outputBasePath, "$zoom/$x")
       path.mkdirs()
       val gmsTile = tile.toGmsTile()
-      File(path, "$y.jpg").writeBytes(gmsTile.data!!)
+      val outFile = File(path, "$y.jpg")
+      val data = gmsTile.data!!
+      outFile.writeBytes(data)
+      Timber.d("Saved ${data.size} bytes to ${outFile.path}")
+      emit(data.size)
```
shobhitagarwal1612, line 45: Should we emit the path of the downloaded tile as well? Is the downloaded size being used at the caller place?

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
gino-m: Good catch, resolved.

Issue #1753, opened 2023-07-20
Update map type selector UI based on mocks


gino-m:
![image](https://github.com/google/ground-android/assets/228050/6f528ede-0bee-49ed-8175-dce0435ce5db)

@shobhitagarwal1612 Is this in your wheelhouse?

@vittorino FYI.

shobhitagarwal1612:
Yes, please assign this to me

shobhitagarwal1612:
Last remaining task: adjust the width of the items according to screen size to remove the horizontal scrolling

shobhitagarwal1612:
@gino-m Are we targeting this bug for the upcoming code freeze too? I thought we decided to punt on this to "Field testing" milestone

gino-m:
I created two follow up bugs with the appropriate milestones, should be ok to close this one now.

PR #1759, opened 2023-07-23
Update map type selector UI

shobhitagarwal1612:
Towards #1753

map selector

@gino-m  PTAL?


ground/src/main/res/layout/map_type_dialog_item.xml
```diff
@@ -21,24 +21,24 @@
...
-        android:id="@+id/imageview"
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        tools:src="@drawable/ic_type_roadmap" />
-    </androidx.cardview.widget.CardView>
-  </FrameLayout>
+  <com.google.android.material.card.MaterialCardView
+    android:id="@+id/card"
+    android:layout_width="116dp"
+    android:layout_height="116dp"
```
gino-m, line 31: This looks amazing! The only visual glitch I see is it appears the right side of the satellite/hybrid thumbnail is clipped a bit. Is it possible to allow the cards to grown/shrink based on available container width, up to some max size?

shobhitagarwal1612: It's a scrollable list at the moment. I can try replacing it with a static view which shrinks/fills up the available container width with a max size of 116dp.

gino-m: I think that's preferable. The number of options will be fixed (3) for the foreseeable future, so no need for the user to scroll to catch the extra bit of thumbnail. Thanks!

gino-m: Hi Shobhit, I'm going to approve this PR in case you want to remove scrolling in a follow-up PR. HTH!

gino-m (approved):

ground/src/main/res/layout/map_type_dialog_item.xml
shobhitagarwal1612: I tried creating a static view but it requires a lot more changes than I anticipated (basically a rewrite).

PR #1772, opened 2023-08-01
Refresh user profile in remote db on sign in

### LINKED ISSUE #1766: User info (image, name) not shown for new users
https://github.com/google/ground-platform/pull/1104 introduced a callable Cloud Function that should be invoked on sign-in to create or update the user's profile in the db. That PR implements the call from the web app; the same must be added to the Android app to ensure the name and profile appear for mobile-only users who never log into the web app.

gino-m:
Fixes #1766


gino-m:
@JSunde PTAL?

JSunde (approved):

Issue #1691, opened 2023-05-31
Once multi-user login is supported, avoid clearing local db data. This is


shobhitagarwal1612:
currently being done to prevent one user's data to be submitted as another user after

re\-login\.

https://github.com/google/ground-android/blob/68a2f58c4d8bf8c96acb2cbd4e99d13fc29311b9/app/src/main/java/org/groundplatform/android/repository/UserRepository.kt#L89-L92

```kotlin


  override suspend fun getUser(userId: String): User = localUserStore.getUser(userId)

  override suspend fun clearUserData() {
    clearUserPreferences()
    // TODO: Once multi-user login is supported, avoid clearing local db data. This is
    //  currently being done to prevent one user's data to be submitted as another user after
    //  re-login.
    // Issue URL: https://github.com/google/ground-android/issues/1691
    localDatabase.clearAllTables()
  }

  private fun clearUserPreferences() {
    val deferredDeeplinkConsumed = localValueStore.isDeferredDeeplinkConsumed
    localValueStore.clear()
    if (deferredDeeplinkConsumed) {
      // A deferred deep link is spent once per install, so signing out must not make an already
      // consumed one eligible to be consumed again.
      localValueStore.isDeferredDeeplinkConsumed = true
    }
  }

  override suspend fun canUserSubmitData(): Boolean {
    if (

```

gino-m:
Some states to handle:

## Sign in when not in passlist

* Show modal with eg "Contact survey organizer to obtain access":
  * "Close app" or "Sign out"
  * On "Sign out": Warn that "all unsynced data will be lost, are you sure?"

## Mutation in local db but user not in passlist

* If sync of a mutation fails, stop retrying, update mutation state
* Eventually (not now) we can add notification

shobhitagarwal1612:
I'll start with converting the remaining remote firestore calls to corountine and then will handle the above cases at the implementation place rather than how it is being currently handled.

gino-m:
> I'll start with converting the remaining remote firestore calls to corountine and then will handle the above cases at the implementation place rather than how it is being currently handled.

Should we generally propagate them down to the consumer of the Flow or coroutine rather than in the implementation? My motivation for this is two-fold:

1) if coroutines only handle their own errors some of the time, it's safer to try to catch them in the caller to avoid bugs due to misunderstood or changing function contracts. If we always need to handle them downstream, swallowing them upstream provides little benefit.

[...]

PR #1808, opened 2023-08-19
Remote firestore refactor

shobhitagarwal1612:
Towards #1691 
Towards #1581

Migrates loading remote submissions and terms of service from Rx to coroutine

Tested for regressions by:
 *  Opening the TOS page 
 * Creating and verifying submission count

@gino-m @JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
@@ -39,6 +39,7 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.rx2.asFlowable
 import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.withTimeoutOrNull
 import timber.log.Timber
 
 private const val LOAD_REMOTE_SURVEY_TIMEOUT_MILLS: Long = 15 * 1000
@@ -106,15 +107,11 @@ constructor(
    */
   suspend fun loadAndSyncSurveyWithRemote(id: String): Survey? {
     Timber.d("Loading survey $id")
-    val survey =
-      withTimeout(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
-        ?: return null
-
-    localSurveyStore.insertOrUpdateSurvey(survey)
-    survey.tileSources.forEach {
-      tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(survey.id))
-    }
-    return survey
+    return withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
+      ?.apply {
+        localSurveyStore.insertOrUpdateSurvey(this)
+        tileSources.forEach { tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(id)) }
+      }
   }
 
   fun clearActiveSurvey() {
--- a/ground/src/main/java/com/google/android/ground/repository/TermsOfServiceRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/TermsOfServiceRepository.kt
@@ -19,12 +19,12 @@ package com.google.android.ground.repository
 import com.google.android.ground.model.TermsOfService
 import com.google.android.ground.persistence.local.LocalValueStore
 import com.google.android.ground.persistence.remote.RemoteDataStore
-import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Singleton
-import kotlinx.coroutines.rx2.awaitSingleOrNull
+import kotlinx.coroutines.withTimeoutOrNull
+import timber.log.Timber
 
-private const val LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_SECS: Long = 30
+private const val LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_MILLIS: Long = 30 * 1000
 
 @Singleton
 class TermsOfServiceRepository
@@ -37,12 +37,13 @@ constructor(
   var isTermsOfServiceAccepted: Boolean by localValueStore::isTermsOfServiceAccepted
 
   suspend fun getTermsOfService(): TermsOfService? =
-    runCatching {
-        // TODO(#1691): Maybe parse the exception and display to the user.
-        remoteDataStore
-          .loadTermsOfService()
-          .timeout(LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_SECS, TimeUnit.SECONDS)
-          .awaitSingleOrNull()
+    // TODO(#1691): Maybe parse the exception and display to the user.
+    withTimeoutOrNull(LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_MILLIS) {
+      try {
+        remoteDataStore.loadTermsOfService()
+      } catch (e: Error) {
+        Timber.e(e)
+        return@withTimeoutOrNull null
       }
-      .getOrNull()
+    }
 }
--- a/ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
+++ b/ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
@@ -75,9 +75,12 @@ public class SubmissionListViewModel extends AbstractViewModel {
       // TODO(#354): Show message or special treatment for layer with no task.
       return Single.just(List.of());
     }
-    return submissionRepository
-        .getSubmissions(req.surveyId, req.locationOfInterestId, req.taskId.get())
-        .onErrorResumeNext(this::onGetSubmissionsError);
+
+    //    return submissionRepository
+    //        .getSubmissions(req.surveyId, req.locationOfInterestId, req.taskId.get())
+    //        .onErrorResumeNext(this::onGetSubmissionsError);
+    // TODO(#1691): Replace the above call with coroutine once migrated to kotlin
+    throw new RuntimeException("Please convert to kotlin and use appropriate method");
   }
 
   private Single<List<Submission>> onGetSubmissionsError(Throwable t) {
--- a/config/detekt/baseline.xml
+++ b/config/detekt/baseline.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" ?>
 <SmellBaseline>
-  <ManuallySuppressedIssues />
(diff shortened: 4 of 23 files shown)


ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -106,15 +107,11 @@ constructor(
...
-        ?: return null
-
-    localSurveyStore.insertOrUpdateSurvey(survey)
-    survey.tileSources.forEach {
-      tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(survey.id))
-    }
-    return survey
+    return withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
+      ?.apply {
+        localSurveyStore.insertOrUpdateSurvey(this)
```
gino-m, line 112: Can we write this as sequential operations to decrease nesting here?

eg 
```kotlin
val survey = await withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) } ?: return null

localSurveyStore.insertOrUpdateSurvey(this)
tileSources.forEach { tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(id)) }
```

Also, one silly question - what happens here if you don't `await` on `insertOrUpdateSurvey` and `insertOrUpdateSuspend`? Are Coroutines smart enough to run them in separate threads, and the wait for all the complete before returning from this suspend function? Or does something else happen?

ground/src/main/java/com/google/android/ground/repository/TermsOfServiceRepository.kt
```diff
@@ -37,12 +37,13 @@ constructor(
...
   suspend fun getTermsOfService(): TermsOfService? =
-    runCatching {
-        // TODO(#1691): Maybe parse the exception and display to the user.
-        remoteDataStore
-          .loadTermsOfService()
-          .timeout(LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_SECS, TimeUnit.SECONDS)
-          .awaitSingleOrNull()
+    // TODO(#1691): Maybe parse the exception and display to the user.
+    withTimeoutOrNull(LOAD_REMOTE_SURVEY_TERMS_OF_SERVICE_TIMEOUT_MILLIS) {
+      try {
```
gino-m, line 42: Can we handle this downstream? See comment https://github.com/google/ground-android/issues/1691#issuecomment-1686372000

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
```diff
@@ -75,9 +75,12 @@ private Single<List<Submission>> getSubmissions(SubmissionListRequest req) {
       // TODO(#354): Show message or special treatment for layer with no task.
       return Single.just(List.of());
     }
-    return submissionRepository
-        .getSubmissions(req.surveyId, req.locationOfInterestId, req.taskId.get())
-        .onErrorResumeNext(this::onGetSubmissionsError);
+
+    //    return submissionRepository
```
gino-m, line 79: I think you mean to delete this?

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
```diff
@@ -75,9 +75,12 @@ private Single<List<Submission>> getSubmissions(SubmissionListRequest req) {
...
+    //        .getSubmissions(req.surveyId, req.locationOfInterestId, req.taskId.get())
+    //        .onErrorResumeNext(this::onGetSubmissionsError);
+    // TODO(#1691): Replace the above call with coroutine once migrated to kotlin
+    throw new RuntimeException("Please convert to kotlin and use appropriate method");
```
gino-m, line 83: In Kotlin you can also do:
```kotlin
TODO("Convert to Kotlin and use appropriate method")
```

gino-m (changes requested):
LG! A few small suggestions.

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -106,15 +107,11 @@ constructor(
...
-    val survey =
-      withTimeout(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
-        ?: return null
-
-    localSurveyStore.insertOrUpdateSurvey(survey)
-    survey.tileSources.forEach {
-      tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(survey.id))
-    }
-    return survey
+    return withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
```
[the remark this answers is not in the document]
shobhitagarwal1612: That's a good question. We might want to consider showing a "retry" button in case something times out. In that scenario, we shouldn't return null and handle the timeout exception at the caller place. Wdyt?

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
shobhitagarwal1612: Thanks for the suggestion! SG to rewrite as a sequential operation.

> Also, one silly question - what happens here if you don't await on insertOrUpdateSurvey and insertOrUpdateSuspend? Are Coroutines smart enough to run them in separate threads, and the wait for all the complete before returning from this suspend function? Or does something else happen?

`await` is required only when we are using a Rx object inside a coroutine. When calling a `suspend` function from within a `suspend` function, it doesn't switch threads if we haven't explicitly mentioned the dispatcher.

ground/src/main/java/com/google/android/ground/repository/TermsOfServiceRepository.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
shobhitagarwal1612: This is java codebase. So we can't use this.

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
shobhitagarwal1612: I considered deleting it, but we might need this class for displaying the submission list in the future. So, I've commented it out and left a TODO to replace the code with appropriate kotlin coroutine in the future. 

Please let me know if you feel too strongly about deleting this code.

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
gino-m: Sorry I missed that. In that case `UnsupportedOperationException` may be appropriate?

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
gino-m: Ah sorry I missed it. Maybe more the TODO up so that the example can be considered part of the TODO?

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/home/locationofinterestdetails/SubmissionListViewModel.java
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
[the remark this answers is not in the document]
shobhitagarwal1612: This is a "sync" operation. So if it times out, we just abort the task and don't retry.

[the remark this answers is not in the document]
gino-m: I see. May be worth clarifying w/a comment

ground/src/main/java/com/google/android/ground/MainViewModel.kt
```diff
@@ -115,6 +116,15 @@ constructor(
       onUserSignInError(e)
     }
 
+  private suspend fun getTos(): TermsOfService? =
+    try {
+      termsOfServiceRepository.getTermsOfService()
+    } catch (e: Error) {
+      // Request timed out or failed due to some error (e.g. PERMISSION_DENIED)
+      Timber.e(e)
+      null
```
gino-m, line 125: Wouldn't this mean users can skip accepting the ToS if they fail to load for any reason?

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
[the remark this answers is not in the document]
shobhitagarwal1612: done

ground/src/main/java/com/google/android/ground/MainViewModel.kt
shobhitagarwal1612: This was the behavior so far. Do you think we should just error out if the request fails?

ground/src/main/java/com/google/android/ground/repository/SurveyRepository.kt
```diff
@@ -107,11 +103,11 @@ constructor(
    */
   suspend fun loadAndSyncSurveyWithRemote(id: String): Survey? {
     Timber.d("Loading survey $id")
-    return withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(id) }
-      ?.apply {
-        localSurveyStore.insertOrUpdateSurvey(this)
-        tileSources.forEach { tileSourceDao.insertOrUpdateSuspend(it.toLocalDataStoreObject(id)) }
-      }
+    val survey =
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done

gino-m (approved):
LG, thanks for the edits!

ground/src/main/java/com/google/android/ground/MainViewModel.kt
shobhitagarwal1612: Fixed

PR #1809, opened 2023-08-19
#2 Gracefully handle permission denied error during signin

shobhitagarwal1612:
Towards #1691

### Current behavior
If the user isn't in the allowlist, then the app would directly finish with a toast message. One needs to manually clear to app data in order to reset the app and attempt login again.

### Proposed behavior
A dialog is shown with 2 options : Sign out and Close app

### Other details 
Note that this PR also changes the behavior of sign out action by clearing all local tables as well.

@gino-m  PTAL?

PR #1811, opened 2023-08-20
#4 Remove global error handling during mutations sync

shobhitagarwal1612:
Towards #1809
Towards https://github.com/google/ground-android/issues/1691

Remove global exception handling when applying mutations. 

This is a no-op change as DatastoreException isn't thrown during that process anyways.

In a follow-up PR, I'll add some error handling within the mutation sync worker and handle it locally to prevent infinite retries.

@JSunde @gino-m  PTAL?


shobhitagarwal1612:
@gino-m @JSunde ping!

JSunde (approved):

PR #1817, opened 2023-08-29
Use NODATA masks when provided in offline map imagery

### LINKED ISSUE #1816: Mask out areas of world TIFF where hi-res imagery is not available
Showing all imagery at lower zoom levels causes confusion about which areas have offline imagery. This would avoid potential confusion.

gino-m:
This makes it clearer to the user that there is imagery only at the masked locations.

Fixes #1816.

[Screen_recording_20230829_171533.webm](https://github.com/google/ground-android/assets/228050/a76f01fc-409c-4ada-abec-15b34de9af00)


ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
```diff
@@ -218,7 +218,10 @@ constructor(
     val baseUrl = getFirstTileSourceUrl()
     val mogCollection =
       MogCollection(
-        listOf(MogSource("${baseUrl}/world.tif", 0..7), MogSource("${baseUrl}/{x}/{y}.tif", 8..14))
+        listOf(
+          MogSource("${baseUrl}/world-masked.tif", 0..7),
+          MogSource("${baseUrl}/{x}/{y}.tif", 8..14)
+        )
```
shobhitagarwal1612, line 224: Should we consider creating a cons field for these params and moving to the central Config file?

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -406,7 +406,7 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), Map {
     // TODO(#1730): Make sub-paths configurable and stop hardcoding here.
     val mogCollection =
       MogCollection(
-        listOf(MogSource("${url}/world.tif", 0..7), MogSource("${url}/{x}/{y}.tif", 8..14))
+        listOf(MogSource("${url}/world-masked.tif", 0..7), MogSource("${url}/{x}/{y}.tif", 8..14))
```
shobhitagarwal1612, line 409: Same as above.

shobhitagarwal1612 (changes requested):
Some minor comments, rest looks good.

Can you please check if we can add/update the unit tests too?

ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
gino-m: I'd much rather move them directly into the survey offline map sources, but I can move to a single constant for now to make the deps clearer.

shobhitagarwal1612 (approved):

PR #1829, opened 2023-09-01
Prevent multiple re-rendering of map features and clear existing markers on attach

### LINKED ISSUE #1594: [Regression] Existing clustered geometries aren't cleared when switching survey
**Describe the bug**
Existing geometries should be cleared from the home screen after switching to a new survey

**To Reproduce**
1. Create 2 surveys, one which contains LOIs and another which doesn't
2. Select survey 1 
3. Observe that the clusters are visible for the LOIs
4. Switch to survey 2

**Expected behavior**
Nothing should be visible on the home screen

**Actual behavior**
Existing clusters are still visible

shobhitagarwal1612:
Fixes #1594

During initialisation, we always create a new instance of `FeatureClusterManager` and `FeatureClusterRenderer` instead of reusing the previous one. This results in creation of a new marker manager (internally) which now doesn't have any reference to the previously added markers to the map. Therefore, clearing the items from the FeatureClusterMananger doesn't work.

So, the initial approach was to check whether that can be reused. However on further debugging, I found that we re-render the features on every map attach regardless of whether they are already shown or not. In this PR, I've added a new helper method in `Map` interface which clears everything when the map is attached to the fragment. This resolves the issue of previous map features not getting cleared.

As a separate performance enhancement task, we can explore whether the renderers and managers can be preserved as long as the `GoogleMapFragment` instance is alive to prevent recreation of objects and even prevent clearing all previously rendered features. (https://github.com/google/ground-android/issues/1830)

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -68,7 +72,7 @@ internal constructor(
     surveyRepository
   ) {
 
-  val mapLocationOfInterestFeatures: LiveData<Set<Feature>>
+  val mapLocationOfInterestFeatures: Flow<Set<Feature>>
```
gino-m, line 75: Since this represents UI state, shouldn't it be modeled as a `StateFlow` so that the last value is replayed to late subscribers?

gino-m (changes requested):
One comment/question. Great work!

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
shobhitagarwal1612: Thanks! Done

gino-m (approved):

PR #1837, opened 2023-09-04
[Rx -> Coroutine] Convert Flowable to Flow in MapController

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/1581

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -163,7 +163,7 @@ internal constructor(
     val geometry = features[0].geometry
 
     if (geometry is Point) {
-      mapController.panAndZoomCamera(geometry.coordinates)
+      panAndZoomCamera(geometry)
```
gino-m, line 166: I think `Coordinates` is the correct type to send here, no `Geometry` (i.e. `geometry.coordinates`)

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
```diff
@@ -43,50 +50,51 @@ constructor(
   private val locationOfInterestRepository: LocationOfInterestRepository,
   private val surveyRepository: SurveyRepository,
   private val mapStateRepository: MapStateRepository,
-  @DefaultDispatcher private val defaultDispatcher: CoroutineDispatcher,
+  @ApplicationScope private val externalScope: CoroutineScope
 ) {
 
-  private val cameraUpdatesSubject: @Hot Subject<CameraPosition> = PublishSubject.create()
+  private val _cameraUpdates = MutableStateFlow<CameraPosition?>(null)
```
gino-m, line 56: Are we intending to treat this stream as a state container, or an event bus? If we're using `StateFlow` here, I'd assume the former, and that the field would be named something like `_cameraPosition`. If this is intended to model an event, then I assume we wouldn't want replay and a `MutableSharedFlow` would be more appropriate. Please cmiiw.

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
```diff
@@ -43,50 +50,51 @@ constructor(
...
-    val locationUpdates = locationManager.locationUpdates.asFlowable().map { it.toCoordinates() }
+  private fun getCameraUpdatesFromLocationChanges(): Flow<CameraPosition> {
+    val locationUpdates = locationManager.locationUpdates.map { it.toCoordinates() }
     // The first update pans and zooms the camera to the appropriate zoom level;
     // subsequent ones only pan the map.
     return locationUpdates
       .take(1)
       .map { CameraPosition(it, DEFAULT_LOI_ZOOM_LEVEL) }
-      .concatWith(locationUpdates.map { CameraPosition(it) }.skip(1))
+      .onCompletion { emitAll(locationUpdates.map { CameraPosition(it) }.drop(1)) }
```
gino-m, line 75: Can we simplify this flow by using `withIndex`? In that case we'd update the zoom level only when the index is 0.

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
```diff
@@ -43,50 +50,51 @@ constructor(
...
-      .toFlowable(BackpressureStrategy.LATEST)
-      .mergeWith(getCameraUpdatesFromLocationChanges())
-      .mergeWith(getCameraUpdatedFromSurveyChanges())
+  fun getCameraUpdates(): SharedFlow<CameraPosition> =
+    merge(
+        _cameraUpdates.filterNotNull(),
+        getCameraUpdatesFromLocationChanges(),
+        getCameraUpdatedFromSurveyChanges(),
+      )
+      .shareIn(externalScope, replay = 1, started = SharingStarted.Eagerly)
```
gino-m, line 65: Can we simplify this by having the ViewModel collect values from each of these streams and calling the appropriate event method on the Map`panAndZoomCamera`? Possibly out of scope of this PR, but I'd also consider moving this logic into the base ViewModel, since a "Controller" is hard to conceptualize in MVVM architecture model. That might help further reduce the complexity of the map logic.

gino-m (changes requested):
Hi Shobhit, this is a huge improvement over the Rx-based logic we had before, thank you! I added some suggestions to further simplify the impl which could help us better understand why tests aren't passing. Lmkwyt1

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
shobhitagarwal1612: Thanks for the tip! I didn't know of this method. Makes it so much better.

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
shobhitagarwal1612: In that case, won't the flow be destroyed and recreated on each screen change?

gino-m: ViewModels survive configuration changes, so as long as we're retrieving the existing one in `onCreate` rather than creating a new one the state should be preserved.

ground/src/test/java/com/google/android/ground/ui/map/MapControllerTest.kt
```diff
@@ -16,18 +16,20 @@
...
+import app.cash.turbine.test
 import com.google.android.ground.BaseHiltTest
+import com.google.android.ground.coroutines.ApplicationScope
 import com.google.android.ground.model.geometry.Coordinates
 import com.google.android.ground.model.geometry.Point
 import com.google.android.ground.repository.LocationOfInterestRepository
 import com.google.android.ground.repository.MapStateRepository
 import com.google.android.ground.repository.SurveyRepository
 import com.google.android.ground.system.LocationManager
+import com.google.common.truth.Truth.assertThat
```
gino-m, line 28: Are you still having trouble with tests? If so, lmk and I can take a look.

ground/src/main/java/com/google/android/ground/ui/map/MapController.kt
gino-m: Ah, I see. If we're being consistent, view state and behavior should be models in VM. Those states can come from repositories and managers (in our design, the repositories for system states). If we need state to persist across views, we would want to move it upstream to specialized singletons. I'd also consider modeling each usage as a UseCase class if it makes sense. Wdyt?

gino-m (changes requested):

ground/src/test/java/com/google/android/ground/ui/map/MapControllerTest.kt
shobhitagarwal1612: I've removed the tests for now as the map controller class has been disintegrated and moved into the VM

gino-m (approved):
This is much more understandable, thanks for all the refactors!

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -175,6 +187,60 @@ constructor(
     }
   }
 
+  /** Emits a stream of camera update requests. */
+  fun getCameraUpdates(): Flow<CameraPosition> = _cameraPosition.filterNotNull()
```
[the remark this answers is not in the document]
shobhitagarwal1612: +1

PR #1846, opened 2023-09-08
Estimate offline map imagery download size and disable download for larger areas

gino-m:
Towards #1822. 

[Screen_recording_20230908_184423.webm](https://github.com/google/ground-android/assets/228050/13c6e71b-d78e-490a-88b5-8d1d2468670e)

@amegantz @vittorino FYI

@shobhitagarwal1612 PTAL?


shobhitagarwal1612:
Should the download button be disabled at high zoom levels?

shobhitagarwal1612:
Another thing I noticed in the screencast, should we add a back navigation button on the toolbar?

gino-m:
> Another thing I noticed in the screencast, should we add a back navigation button on the toolbar?

This isn't in the mocks so I removed it. @vittorino for comment.

gino-m:
> Should the download button be disabled at high zoom levels?

+1. Can I do this in a follow-up PR? Filed https://github.com/google/ground-android/issues/1855 to track.

gino-m:
@shobhitagarwal1612 for final approval if it LG. Thanks!

shobhitagarwal1612 (approved):

shobhitagarwal1612:
Looks like the dialog can be dismissed but the tile download still works in the background. We should either make the dialog non-dismissable along with a cancel button or display download progress in the UI or notification bar. Thoughts?

gino-m:
Hi Shobhit, it should be cancelable as in the mocks. Would you mind filing
a bug against me when you have a sec?

PR #1852, opened 2023-09-09
Clip offline tiles to downloaded bounds

### LINKED ISSUE #1756: Stop accumulating tile overlays on map
Tile overlays are currently added to the map but never removed. When tileoverlays change, we can simply remove all existing ones and readd the new set.

### LINKED ISSUE #1850: Crop offline imagery to areas selected by user
We currently see tiles of varying sizes as we zoom from higher to lower zoom levels. This is confusing to the user - we should instead only render imagery in the area selected by the user, either based on what's available on the device, or by the clipping at all zoom levels to the specific bounds the user selected.

gino-m:
Includes some small refactors of tile overlay and MOG impl.

Fixes #1850
Fixes #1756

[Screen_recording_20230910_174722.webm](https://github.com/google/ground-android/assets/228050/29c3cfce-7c61-41f9-8d50-0b1ede31571f)

@shobhitagarwal1612 PTAL?


ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -353,14 +353,16 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), Map {
...
+    addTileOverlay(
+      ClippingTileProvider(TemplateUrlTileProvider(url), bounds.map { it.toGoogleMapsObject() })
+    )
   }
 
   override fun addTileOverlay(tileSource: TileSource) =
     when (tileSource.type) {
       MOG_COLLECTION -> addMogCollectionTileOverlay(tileSource.url)
-      TILED_WEB_MAP -> addTemplateUrlTileOverlay(tileSource.url)
+      TILED_WEB_MAP -> addOfflineTileOverlay(tileSource.url, tileSource.clipBounds!!)
```
shobhitagarwal1612, line 365: Should `clipBounds` be nullable?

ground/src/main/java/com/google/android/ground/ui/offlinebasemap/selector/OfflineAreaSelectorViewModel.kt
```diff
@@ -107,7 +108,8 @@ internal constructor(
   }
 
   fun onMapReady(map: Map) {
-    tileSources.forEach { map.addTileOverlay(it) }
+    map.mapType = MapType.TERRAIN
```
shobhitagarwal1612, line 111: This line can be removed now. It was addressed separately using MapConfig

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: In the current impl, no, but replacing nullable w/default value upstream anyway to prevent regressions here.

gino-m:
@shobhitagarwal1612 Thanks for reviewing! RFAL

shobhitagarwal1612 (approved):

PR #1874, opened 2023-09-12
Use correct Button and remove redundant styles

### LINKED ISSUE #1873: Stop using Chip where we need MaterialButton
If we replace `com.google.android.material.chip.Chip` -> `com.google.android.material.button.MaterialButton` styling "just works"; no need to override enabled/disabled states, set primary background and text color, etc.

@JSunde @shobhitagarwal1612 @vittorino FYI

gino-m:
Fixes #1873 

@JSunde PTAL?


JSunde (approved):

JSunde (changes requested):
Wait, should we be using MaterialButton instead of just Button?

gino-m:
> Wait, should we be using MaterialButton instead of just Button?

Apparently they're the same thing:
https://stackoverflow.com/a/57926083/9986466

Issue #1889, opened 2023-09-14
Refactor new `MapViewModel` out of `MapsView` and simplify view/VM class hierarchy


gino-m:
@shobhitagarwal1612 FYI

PR #1890, opened 2023-09-14
Replace `Flowable` with `SharedFlow` for map events

gino-m:
Also does a few minor renames, including Map -> MapView (towards #1889)

@JSunde @shobhitagarwal1612  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
@@ -28,17 +28,17 @@ import com.google.android.ground.system.PermissionDeniedException
 import com.google.android.ground.system.SettingsChangeRequestCanceled
 import com.google.android.ground.ui.home.mapcontainer.MapTypeDialogFragmentDirections
 import com.google.android.ground.ui.map.CameraPosition
-import com.google.android.ground.ui.map.MapView
+import com.google.android.ground.ui.map.MapUi
 import javax.inject.Inject
 import kotlin.math.max
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
-/** Injects a [MapView] in the container with id "map" and provides shared map functionality. */
+/** Injects a [MapUi] in the container with id "map" and provides shared map functionality. */
 abstract class AbstractMapContainerFragment : AbstractFragment() {
 
-  @Inject lateinit var map: MapView
+  @Inject lateinit var map: MapUi
   @Inject lateinit var navigator: Navigator
   @Inject lateinit var geocodingManager: GeocodingManager
   @Inject @DefaultDispatcher lateinit var defaultDispatcher: CoroutineDispatcher
@@ -51,7 +51,7 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
     lifecycleScope.launch { repeatOnLifecycle(Lifecycle.State.STARTED) { fn.invoke() } }
   }
 
-  private fun onMapAttached(map: MapView) {
+  private fun onMapAttached(map: MapUi) {
     val viewModel = getMapViewModel()
 
     // Removes all markers, overlays, polylines and polygons from the map.
@@ -70,7 +70,7 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
     onMapReady(map)
   }
 
-  private fun applyMapConfig(map: MapView) {
+  private fun applyMapConfig(map: MapUi) {
     val viewModel = getMapViewModel()
     val config = viewModel.mapConfig
 
@@ -103,7 +103,7 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
     navigator.navigate(MapTypeDialogFragmentDirections.showMapTypeDialogFragment(types))
   }
 
-  private fun onLocationLockStateChange(result: Result<Boolean>, map: MapView) {
+  private fun onLocationLockStateChange(result: Result<Boolean>, map: MapUi) {
     result.fold(
       onSuccess = {
         Timber.d("Location lock: $it")
@@ -125,7 +125,7 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
     Toast.makeText(context, messageId, Toast.LENGTH_LONG).show()
   }
 
-  private fun onCameraUpdateRequest(newPosition: CameraPosition, map: MapView) {
+  private fun onCameraUpdateRequest(newPosition: CameraPosition, map: MapUi) {
     Timber.v("Update camera: %s", newPosition)
     val bounds = newPosition.bounds
     val target = newPosition.target
@@ -148,7 +148,7 @@ abstract class AbstractMapContainerFragment : AbstractFragment() {
   }
 
   /** Called when the map is attached to the fragment. */
-  protected open fun onMapReady(map: MapView) {}
+  protected open fun onMapReady(map: MapUi) {}
 
   /** Provides an implementation of [BaseMapViewModel]. */
   protected abstract fun getMapViewModel(): BaseMapViewModel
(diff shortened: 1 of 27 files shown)


ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
```diff
@@ -18,62 +18,60 @@ package com.google.android.ground.ui.common
...
-      .subscribe { viewModel.onMapDragged() }
-
-    lifecycleScope.launch {
-      getMapViewModel().locationLock.collect { onLocationLockStateChange(it, map) }
-    }
-    lifecycleScope.launch {
+    launchWhenStarted { map.cameraMovedEvents.collect { viewModel.onMapCameraMoved(it) } }
+    launchWhenStarted { map.startDragEvents.collect { viewModel.onMapDragged() } }
+    launchWhenStarted { viewModel.locationLock.collect { onLocationLockStateChange(it, map) } }
+    launchWhenStarted {
```
shobhitagarwal1612, line 63: What would happen when the map is attached again (e.g. when pausing and then resuming the fragment)? Would the previous coroutine be cancelled?

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
```diff
@@ -18,62 +18,60 @@ package com.google.android.ground.ui.common
...
       viewModel.getCameraUpdateRequests().collect { onCameraUpdateRequest(it, map) }
     }
+    viewModel.mapType.observe(viewLifecycleOwner) { map.mapType = it }
```
shobhitagarwal1612, line 66: Please remove this line. It is already happening inside `applyMapConfig`.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropAPinMapFragment.kt
```diff
@@ -99,19 +99,19 @@ class DropAPinMapFragment(private val viewModel: DropAPinTaskViewModel) :
...
+  override fun onMapReady(map: MapView) {
     viewModel.features.observe(this) { map.renderFeatures(it) }
   }
 
   override fun getMapViewModel(): BaseMapViewModel = mapViewModel
 
-  override fun onMapCameraMoved(position: CameraPosition) {
-    super.onMapCameraMoved(position)
+  fun onMapCameraMoved(position: CameraPosition) {
+    super.getMapViewModel().onMapCameraMoved(position)
```
shobhitagarwal1612, line 109: I believe this won't work. Earlier we were overriding the `onMapCameraMoved` which was getting called when the map moved. But now, it is just a method with no callers.

Also, the `super` keyword can be omitted too.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/PolygonDrawingMapFragment.kt
```diff
@@ -83,8 +83,8 @@ class PolygonDrawingMapFragment(private val viewModel: PolygonDrawingViewModel)
 
   override fun getMapViewModel(): BaseMapViewModel = mapViewModel
 
-  override fun onMapCameraMoved(position: CameraPosition) {
-    super.onMapCameraMoved(position)
+  fun onMapCameraMoved(position: CameraPosition) {
+    super.getMapViewModel().onMapCameraMoved(position)
```
shobhitagarwal1612, line 87: Same as above

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -75,19 +71,17 @@ const val MARKER_Z = 3f
  * on window insets.
  */
 @AndroidEntryPoint(SupportMapFragment::class)
-class GoogleMapsFragment : Hilt_GoogleMapsFragment(), Map {
+class GoogleMapsFragment : Hilt_GoogleMapsFragment(), MapView {
+  // TODO: Inject dispatchers.
+  private val defaultDispatcher = Dispatchers.Default
```
shobhitagarwal1612, line 76: Just curious why we haven't injected it? Since this class extends `AndroidEntryPoint`, we can simply inject it :)

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -287,7 +251,21 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), Map {
     return checkNotNull(customCap)
   }
 
-  private fun onMapClick(latLng: LatLng) = handleAmbiguity(latLng)
+  private fun onMapClick(latLng: LatLng) =
+    lifecycleScope.launch(defaultDispatcher) {
+      val clickedPolygons = getPolygonFeaturesContaining(latLng)
+      if (clickedPolygons.isNotEmpty()) {
+        featureClicks.emit(clickedPolygons)
```
shobhitagarwal1612, line 258: Just curious: Why have we used `tryEmit` in some places and `emit` in others? Is there a benefit of using one over the other?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapContainerFragment.kt
gino-m: Yes, `repeatOnLifecycle` does this for us. I just learned we can now move these initializations to the constructor and replace our `launchWhenStarted` helper with `whenStarted`, but out of scope of this PR.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: I thought that since we're instantiating it ourselves in `MapProviderModule` that injection wouldn't work. But I guess that since it's an entry point it gets applied after init? I'll give it a try.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: `tryEmit()` doesn't suspend, and require a coroutine. It can theoretically overflow the buffer, but in practice in these cases there would be few items and would be unlikely. In cases where we were already in a coroutine, I used `emit()` since it's slightly safer.

gino-m:
@shobhitagarwal1612 I ended up using `emit` with `launch` rather than `tryEmit`, since `tryEmit` was dropping values. To solve that we'd have to set a buffer and/or strategy on the flow, which I'd rather not need to remember to do. See https://github.com/Kotlin/kotlinx.coroutines/issues/2387 for context.

JSunde (approved):

ground/src/main/java/com/google/android/ground/ui/map/MapView.kt
```diff
@@ -20,15 +20,12 @@ import androidx.annotation.IdRes
...
 import com.google.android.ground.ui.common.AbstractFragment
-import io.reactivex.Flowable
-import io.reactivex.Observable
 import java8.util.function.Consumer
+import kotlinx.coroutines.flow.SharedFlow
 
-/** Interface for a Fragment that renders a map view. */
-interface Map {
+/** Generalization of a map view. */
+interface MapView {
```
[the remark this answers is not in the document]
gino-m: It's in the "view" layer, but I can see how it can be confusing. It's really meant to be on the Fragment. How about `MapFragment`?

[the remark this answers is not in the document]
gino-m: We chatted and agreed on `MapUi` as a generic name.

JSunde (approved):

gino-m:
@shobhitagarwal1612 Gentle ping on restamp. Tx!

shobhitagarwal1612 (approved):

PR #1894, opened 2023-09-15
Show "no data" message when area with no imagery is selected

### LINKED ISSUE #1872: Disable download and show message when there's no hi-res imagery for selected area
Zooming in on area with no hi res imagery currently allows user to download masked lo res imagery even though preview is not show. Instead, we should show message "No imagery to download in the selected area" (as per @amegantz in https://github.com/google/ground-android/issues/1761#issuecomment-1679285449)

gino-m:
Fixes #1872 

@JSunde PTAL?

Screenshot_20230915_105924

@amegantz Any advice on the messsage shown at the bottom of the screen? I was thinking:

> Imagery unavailable. Select another area to download to your device

Wdyt?


JSunde (approved):

amegantz:
"No imagery for this area. Select an area with imagery to download to your device."

PR #1897, opened 2023-09-15
Tweak "no imagery" text

gino-m:
Follow-up to #1872. 

@amegantz PTAL and approve?

Screenshot_20230915_150244

### DIFF
--- a/ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt
@@ -147,6 +147,12 @@ constructor(
     surveyRepository.activeSurvey?.tileSources?.firstOrNull()?.url
       ?: error("Survey has no tile sources")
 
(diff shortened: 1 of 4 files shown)


amegantz:
Need a period after the second sentence, otherwise LGTM.

UXCD rule of thumb for content like this: If one sentence shorter than four lines, no period necessary. If two sentences, period after each.

gino-m:
> Need a period after the second sentence, otherwise LGTM.
> 
> UXCD rule of thumb for content like this: If one sentence shorter than four lines, no period necessary. If two sentences, period after each.

Thanks! What if we're showing other messages in the same space for which the "no period" rule applied? Currently that space can be filled with one of:

# No imagery for this area. Select an area with imagery to download to your device.</string>
# The selected area may take up to _n_ MB of space on your device
# Zoom in and select a smaller area to download to your device

@amegantz Is it correct that the (1) has the "." while (2) and (3) do not?

amegantz:
Yup, that's correct!

amegantz (approved):

Issue #1904, opened 2023-09-17
App crashes on missing task definition


gino-m:
This occurs when zooming into an LOI or POI.

```
15:04:06.332  E  FATAL EXCEPTION: main
                 Process: com.google.android.ground, PID: 21309
                 java.lang.IllegalStateException: Unknown task id BALlgDD4Y0lMlGdkK5sA
                 	at com.google.android.ground.model.job.Job.getTask(Job.kt:34)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionConverter.putResponse(SubmissionConverter.kt:89)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionConverter.toResponseMap(SubmissionConverter.kt:75)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionConverter.toSubmission(SubmissionConverter.kt:60)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionCollectionReference.submissionsByLocationOfInterestId$lambda$0(SubmissionCollectionReference.kt:35)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionCollectionReference.$r8$lambda$tKBwySqVOnNIY7mflxiC54292EE(Unknown Source:0)
                 	at com.google.android.ground.persistence.remote.firebase.schema.SubmissionCollectionReference$$ExternalSyntheticLambda0.apply(Unknown Source:4)
                 	at com.google.android.ground.persistence.remote.firebase.base.FluentCollectionReference.runQuery(FluentCollectionReference.kt:48)
                 	at com.google.android.ground.persistence.remote.firebase.base.FluentCollectionReference$runQuery$1.invokeSuspend(Unknown Source:15)
                 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
                 	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
                 	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
                 	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
                 	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
[...]

Issue #1907, opened 2023-09-17
Refactor Maps *Manager and *Renderer classes


gino-m:
Some initial thoughts:

* Rather than manage our own Maps to keep track of feature, we should instead extend `MapObjectManager` and manage them there.
* `PolygonRenderer` and `PolylineRenderer` implement `FeatureRenderer` don't render anything; they're pseudo-manager / adapter classes responsible for keeping track of features and adding and removing them for the map. This is probably the job better scoped and served by new concrete `PolygonFeatureManager`, `PolylineFeatureManager`, and `PointFeatureManager` which would extend `PolygonManager`, `PolylineManager`, and `MarkerManager`, respectively.
* Renderers index use the entire Feature as key; this is not necessary; we now have a unique Tag we can use to identify and update individual features.
* Marker rendering logic currently only lives in the `FeatureClusterRenderer`, which makes it impossible to correctly handle non-clustered markers. This logic should probably live in its own `PointFeaturerRenderer` class.
* Introducing singlular MapControlsFragment and MapControlsViewModel with dials for the various configurations rather than using both static config and subclasses to customize. The parent fragment could attach the map controls to the map fragment and configure.This would disentangle a lot of our map logic currently spread across the class hierarchy and fragments

PR #1908, opened 2023-09-18
Trick cluster manager into clustering polygons as well as points

### LINKED ISSUE #1895: Polygon LOIs not clustered
This makes polygons hard to see when zoomed out:

Screenshot_20230915_124325

@scolsen I know you did some background research on how we'd do this with the ClusterManager.. do you recall any other details?

gino-m:
**Please merge #1904 and PR to `master` before merging.**

Closes #1895

@JSunde PTAL? @shobhitagarwal1612 @scolsen FYI

[demo.webm](https://github.com/google/ground-android/assets/228050/9ae2d019-02b2-4d07-9e69-e247b4d9a49b)

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/map/gms/FeatureClusterRenderer.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/map/gms/FeatureClusterRenderer.kt
@@ -20,6 +20,7 @@ import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.model.BitmapDescriptor
 import com.google.android.gms.maps.model.Marker
 import com.google.android.gms.maps.model.MarkerOptions
+import com.google.android.ground.model.geometry.Point
 import com.google.android.ground.ui.MarkerIconFactory
 import com.google.maps.android.clustering.Cluster
 import com.google.maps.android.clustering.view.DefaultClusterRenderer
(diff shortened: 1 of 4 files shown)


gino-m:
Known issues:

- [ ] "Drop a pin" shows a cluster bubble instead of a pin
- [ ] need to test other flows as well

gino-m:
There's currently no way to force "Drop a pin" and "Capture location" flows to render a marker that is never clustered. Will follow up in a separate PR for this. Files #1910 to track this but, and #1907 to track the larger cleanup.

ground/src/main/java/com/google/android/ground/ui/map/gms/FeatureClusterRenderer.kt
```diff
@@ -90,6 +124,11 @@ class FeatureClusterRenderer(
     cluster: Cluster<FeatureClusterItem>,
     markerOptions: MarkerOptions
   ) {
+    // Remove non-points before rendering cluster.
+    cluster.items
+      .map { it.feature }
+      .filter { it.geometry !is Point }
```
shobhitagarwal1612, line 130: Since we have geometry types other that Point, Polygon and MultiPolygon, would it be safer to check for equality rather than the other way around?

```
.filter { it.geometry is Polygon || it.geometry is Multipolygon }
```

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/map/gms/FeatureClusterRenderer.kt
gino-m: I think it's safer to only allow `Point` since that's the only type the clusters supports. If we do it the other way around the list would need to be exhaustive, and would need to be updated if we ever add a new geometry type. Wdyt?

PR #1911, opened 2023-09-18
Prevent app crash when attempting to parse corrupt remote data

shobhitagarwal1612:
As discussed offline

@gino-m  PTAL?


gino-m (approved):
LGTM, thanks!

PR #1928, opened 2023-09-24
Fix overlapping of survey title with offline icon in SurveySelectorList

shobhitagarwal1612:
Fixes #

@gino-m @JSunde  PTAL?


gino-m (approved):
LGTM

PR #1930, opened 2023-09-25
Fix extra Suggest LOI tasks being inserted

### LINKED ISSUE #1920: Suggest LOI "Drop a pin" flow triggered on normal LOI data collection
Cliciking "Collect data" on an LOI job card prompts the user to drop a pin. That flow should only be triggered on jobs with `suggestLoiTaskType` set, and only on "suggest" cards.

@shobhitagarwal1612 @JSunde FYI

JSunde:
Only insert Suggest LOI tasks if there is a `suggestLoiTaskType` AND there is no associated LOI

Fixes #1920


JSunde:
PTAL @gino-m

gino-m (approved):
LGTM!

PR #1932, opened 2023-09-26
Increase stroke width of selected polygon

### LINKED ISSUE #1912: Style of polygons not updated when selected
Outline should become thicker. @JSunde PTAL?

JSunde:
Fixes #1912


ground/src/main/java/com/google/android/ground/ui/map/gms/renderer/PolygonRenderer.kt
```diff
@@ -60,9 +61,10 @@ class PolygonRenderer(
     polygon.holes.forEach { options.addHole(it.coordinates.toLatLngList()) }
 
     val mapsPolygon = map.addPolygon(options)
+    val strokeScale = if (isSelected) 2f else 1f
```
gino-m, line 64: We should define thisin a resource file or constant, but we can come back to this later.

gino-m (approved):
LGTM! 1️⃣ ⭕ 🔲 %!

Issue #1910, opened 2023-09-18
Pins turn into cluster baloons when zooming out during Drop a Pin and Capture Location flows


gino-m:
Related to #1907 - marker rendering is currently only implemented inside the cluster manager.

PR #1940, opened 2023-09-27
Improve cluster marker styling

gino-m:
Includes some minor cleanup and refactors. Towards #1910.

cluster

@JSunde @shobhitagarwal1612 PTAL? @vittorino FYI.

### DIFF
--- a/ground/src/main/res/drawable/cluster_marker.xml
+++ b/ground/src/main/res/drawable/cluster_marker.xml
@@ -18,7 +18,7 @@
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
   <item>
     <shape android:shape="oval">
-      <solid android:color="#4B6DDD81" />
+      <solid android:color="@color/clusterBalloonStrokeColor" />
       <size
         android:width="68dp"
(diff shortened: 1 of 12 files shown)


ground/src/main/res/drawable/cluster_marker.xml
```diff
@@ -15,14 +15,23 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+  <item>
+    <shape android:shape="oval">
+      <solid android:color="#4B6DDD81" />
```
JSunde, line 21: Should we add a comment indicating this is ~29% opacity clusterColor?

JSunde (approved):

ground/src/main/res/drawable/cluster_marker.xml
gino-m: Sure. I'll do you one better; since the color of the cluster balloons doesn't change, I now moved the color codes to `colors.xml` and make them static so that we don't need to tint them at runtime.

PR #1943, opened 2023-09-30
Display customId in job card name, if available

### LINKED ISSUE #1926: Job cards should show ID and/or name of the LOI they represent
The second line of the card should should the human-readable LOI label, originally outlined in https://docs.google.com/document/d/1kM7dxF0OFXjN-Yhc44nH95nYIx0xvhWQbKvX3KVzpNE, taken from the geometry's imported key-value properties. 

Rather than using heuristics to find the property names, we can simplify this design by looking for the key "id" and/or "name", ignoring case. Depending on what values are defined/null, the human-readable label would be:

- `$name ($id)`
- `$name`
- `$geometryType $id` where $geometryType is one of "Point" or "Area"
- "Unnamed area" or "Unnamed point"

@shobhitagarwal1612 Would you like to take this one since it has to do with job cards?

shobhitagarwal1612:
Fixes #1926 

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
@@ -15,9 +15,9 @@
  */
 package com.google.android.ground.ui.home.mapcontainer.cards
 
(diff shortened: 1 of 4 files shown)


gino-m (approved):
Nice work!

PR #1944, opened 2023-09-30
Only animate the camera when device location is updated or user clicks

### LINKED ISSUE #1869: Don't animate map when restoring viewport / zoom
@shobhitagarwal1612 in case you have a free cycle at some point.

Animate should be used only in these cases:

- Tap on cluster to zoom
- On device location update when lock enabled

In all other cases map should move directly to target.

shobhitagarwal1612:
Fixes #1869

@gino-m @JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
@@ -40,6 +40,7 @@ import com.google.android.ground.system.PermissionDeniedException
 import com.google.android.ground.system.PermissionsManager
 import com.google.android.ground.system.SettingsManager
 import com.google.android.ground.ui.map.CameraPosition
+import com.google.android.ground.ui.map.CameraUpdateRequest
 import com.google.android.ground.ui.map.MapType
 import com.google.android.ground.ui.map.gms.GmsExt.toBounds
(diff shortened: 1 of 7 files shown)


ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -236,17 +237,17 @@ constructor(
   }
 
   /** Requests moving the map camera to [coordinates]. */
-  private fun panCamera(coordinates: Coordinates) {
-    updateMapCamera(CameraPosition(coordinates))
+  private fun panCamera(coordinates: Coordinates, shouldAnimate: Boolean) {
```
gino-m, line 240: "pan" implies animation, and the boolean arg makes these calls a little harder to grok. Perhaps it would be clearer if we had `panCamera` and `panAndZoomCamera` which animate, and `setCameraPosition` which does not?

gino-m (approved):
LG, one small suggestion. Thanks!

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: Thanks for the suggestion!

PR #1946, opened 2023-10-01
Move app initialization to onResume() to prevent blank screen

### LINKED ISSUE #901: [Crash] App freezes with white screen after device is unlocked
1. Open app, leave homescreen open with active project
2. Lock device
3. Unlock device
4. Screen appears blank/white or shows splash screen

More details to follow.

### LINKED ISSUE #1881: App sometimes freezes on splash screen at startup

shobhitagarwal1612:
Fixes #901
Fixes #1881

The app gets stuck with a blank screen as the initialization flow is called in `onAttach()` but this method is only called once during creation and not when resuming the fragment from background.

Verified by running the below steps manually:

1. Turn off the device screen
2. Install the app using AS
3. Wake the device

@gino-m @JSunde  PTAL?


gino-m (approved):
Good catch and quick fix, kudos on the Coroutines refactor as well! 🙏 🥇

PR #1951, opened 2023-10-02
Use job color in geometry tasks and refactor feature renderers

### LINKED ISSUE #1907: Refactor Maps *Manager and *Renderer classes

### LINKED ISSUE #1952: Job color not used by geometry collection tasks

gino-m:
Fixes #1907 
Fixes #1952 

@JSunde @shobhitagarwal1612 PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropAPinTaskViewModel.kt
```diff
@@ -32,9 +36,15 @@ class DropAPinTaskViewModel
 constructor(resources: Resources, private val uuidGenerator: OfflineUuidGenerator) :
   AbstractTaskViewModel(resources) {
 
+  private var pinColor: Int = 0
```
shobhitagarwal1612, line 39: nit: Should we make it `lateinit var` instead of assigning a default?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/PolygonDrawingViewModel.kt
```diff
@@ -61,6 +65,13 @@ internal constructor(private val uuidGenerator: OfflineUuidGenerator, resources:
   /** Represents whether the user has completed drawing the polygon or not. */
   private var isMarkedComplete: Boolean = false
 
+  private var strokeColor: Int = 0
```
shobhitagarwal1612, line 68: Same as above

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropAPinTaskViewModel.kt
```diff
@@ -55,6 +65,8 @@ constructor(resources: Resources, private val uuidGenerator: OfflineUuidGenerato
       id = uuidGenerator.generateUuid(),
       type = FeatureType.USER_POINT.ordinal,
       geometry = point,
+      // TODO: Set correct pin color.
+      style = Feature.Style(pinColor),
```
shobhitagarwal1612, line 69: As this is the only usage, should we consider inlining the code? Then we'd have to set the job to a class var in the super class. Thoughts?

```
style = Feature.Style(job.getDefaultColor())
```

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -169,4 +175,21 @@ internal constructor(
   }
 
   fun getZoomThresholdCrossed(): Observable<Nil> = zoomThresholdCrossed
+
+  private fun getLocationOfInterestFeatures(survey: Survey): Flow<Set<Feature>> =
+    loiRepository.getLocationsOfInterest(survey).map {
+      it.map { loi -> loi.toFeature() }.toPersistentSet()
+    }
```
shobhitagarwal1612, line 182: nit: Can we get rid of the nesting by using `flatMap`?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -169,4 +175,21 @@ internal constructor(
...
+
+  private suspend fun LocationOfInterest.toFeature() =
+    Feature(
+      id = id,
+      type = FeatureType.LOCATION_OF_INTEREST.ordinal,
+      flag =
+        submissionCount + submissionRepository.getPendingCreateCount(id) -
+          submissionRepository.getPendingDeleteCount(id) > 0,
```
shobhitagarwal1612, line 190: Can we move these to separate lines for better readability?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropAPinTaskViewModel.kt
gino-m: Unfortunately 'lateinit' modifier is not allowed on properties of primitive types

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/PolygonDrawingViewModel.kt
gino-m: Same here

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropAPinTaskViewModel.kt
gino-m: The superclass is scoped more widely that geometry tasks, and each geometry task may require slightly different styling treatments. Also, I'd prefer to avoid too much indirection to keep things locally comprehensible.

Sorry I missed this TODO, removing it.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: `flatMap` pulls the elements of a list or flow up a level. That would change the output to `Flow<Feature>`, which is not what we want here; we want a full snapshot of all the displayed features on each emission.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: Sure, coming right up

gino-m:
@shobhitagarwal1612 PTAL?

shobhitagarwal1612 (approved):

PR #1957, opened 2023-10-04
Allow zooming in to the current location when toggling the lock 2nd time

### LINKED ISSUE #1956: [Bug] Location lock is not zooming in when activating it for 2nd time

shobhitagarwal1612:
Fixes #1956

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -72,6 +72,9 @@ constructor(
   @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
 ) : AbstractViewModel() {
 
+  /** This is set to true as soon as the location lock is disabled. */
+  private var isLocationLockReset = true
```
gino-m, line 76: In that case should this be named for the positive case, `locationLockEnabled`?

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -206,9 +210,9 @@ constructor(
   private suspend fun updateCameraPositionOnLocationChange() {
     getLocationUpdates()
       .map { it.toCoordinates() }
-      .withIndex()
-      .collect { (index, coordinates) ->
-        if (index == 0) {
+      .collect { coordinates ->
+        if (isLocationLockReset) {
+          isLocationLockReset = false
```
gino-m, line 215: The way it's written now kit's not immediately clear what we're trying to do something different on the first emission. Would it be more idiomatic to still use `withIndex()` and to reset the stream each time to location updates are enabled (maybe via flatMapLatest?)

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: The intention is to mark the field as false as soon as we consume it. `locationLockEnabled` would mean that we can't set the bit to false after it's first use.

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: I wasn't sure how to reset the stream. Will check once again.

gino-m:
```kt
  locationLockEnabled
    .flatMapLatest { enabled -> getLocationUpdates().filter { enabled } }
    .withIndex()
    .collect { ...
    }
```

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
gino-m: I see. In any case I think the functional approach described below is easier to understand than managing an additional state and relying on side-effects.

shobhitagarwal1612: This is now reverted

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: Thanks for the suggestion! This also gets rid of the boolean bit

gino-m (approved):
🚀 !!!

Issue #1966, opened 2023-10-07
Top bug fixes for live demo


gino-m:
P0-P1

- [ ] Can't select surveys below the fold (#1924)
- [x] App crashes when job created with new UI due to missing defaultStyle (#1967)
- [ ] #1984
- [ ] #1985
- [x] Take a picture -> Says I don't have permissions after allowing permissions (@JSunde)
- [ ] #1986
- [ ] #1987
- [ ] #1975
- [ ] #816

@JSunde @shobhitagarwal1612 Do you think you could take on any of these P0s before Tuesday? If so please add your username above and optionally in a tracking bug!

P2

- [ ] #1973
- [ ] #1988
- [ ] #1989
- [ ] #1871
- [ ] #1859

P3

- [ ] #1990

gino-m:
Split off into separate bugs

PR #1970, opened 2023-10-09
Fix photo permissions regression

JSunde:
Towards #1966


gino-m:
> Towards #1966

Does this close #1966?

gino-m (approved):
Thanks!

PR #1992, opened 2023-10-15
Prevent users from getting signed out when opening the app in offline mode

### LINKED ISSUE #1985: Users logged out when offline
Can repro by reopening the app in Airplane mode

shobhitagarwal1612:
Fixes #1985

The issue was happening because 2 remote requests were failing. 
* Refresh user profile data
* Terms of service 

Verified the changes by opening the app in airplane mode after first signin.

@gino-m @JSunde  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
@@ -25,9 +25,9 @@ import com.google.android.ground.model.mutation.Mutation
 import com.google.android.ground.model.mutation.SubmissionMutation
 import com.google.android.ground.model.submission.Submission
 import com.google.android.ground.persistence.remote.RemoteDataStore
+import com.google.android.ground.system.NetworkManager
 import com.google.firebase.firestore.WriteBatch
 import com.google.firebase.functions.FirebaseFunctions
-import com.google.firebase.functions.FirebaseFunctionsException
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.messaging.ktx.messaging
 import javax.inject.Inject
@@ -45,7 +45,8 @@ class FirestoreDataStore
 internal constructor(
   private val firebaseFunctions: FirebaseFunctions,
   private val groundFirestoreProvider: GroundFirestoreProvider,
-  @IoDispatcher private val ioDispatcher: CoroutineDispatcher
+  @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+  private val networkManager: NetworkManager
 ) : RemoteDataStore {
 
   private suspend fun db() = groundFirestoreProvider.get()
@@ -76,13 +77,16 @@ internal constructor(
     Firebase.messaging.subscribeToTopic(surveyId).await()
   }
 
-  /** Calls Cloud Function to refresh the current user's profile info in the remote database. */
+  /**
+   * Calls Cloud Function to refresh the current user's profile info in the remote database if
+   * network is available.
+   */
   override suspend fun refreshUserProfile() {
-    try {
-      firebaseFunctions.getHttpsCallable(PROFILE_REFRESH_CLOUD_FUNCTION_NAME).call().await()
-    } catch (e: FirebaseFunctionsException) {
-      Timber.w(e, "Failed to refresh user data")
+    if (!networkManager.isNetworkAvailable()) {
+      Timber.d("Skipped refreshing user profile as device is offline.")
+      return
     }
+    firebaseFunctions.getHttpsCallable(PROFILE_REFRESH_CLOUD_FUNCTION_NAME).call().await()
   }
 
   override suspend fun applyMutations(mutations: List<Mutation>, user: User) {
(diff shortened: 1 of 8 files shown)


ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
```diff
@@ -77,7 +78,11 @@ internal constructor(
 
   /** Calls Cloud Function to refresh the current user's profile info in the remote database. */
   override suspend fun refreshUserProfile() {
-    firebaseFunctions.getHttpsCallable(PROFILE_REFRESH_CLOUD_FUNCTION_NAME).call().await()
+    try {
+      firebaseFunctions.getHttpsCallable(PROFILE_REFRESH_CLOUD_FUNCTION_NAME).call().await()
+    } catch (e: FirebaseFunctionsException) {
+      Timber.w(e, "Failed to refresh user data")
```
gino-m, line 84: This will be a valid error when the user is actually online. Instead of eating the exception, can we handle if downstream in the caller? 

A broader problem is that we shouldn't be trying to load ToS and updating user profile when offline; waiting for timeout takes time, which makes the app appear to load slowly. Can we skip those steps when offline and the user is already signed in?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
shobhitagarwal1612: Makes sense! Thanks for the suggestion. I'll update this PR.

shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m PTAL

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
```diff
@@ -75,8 +77,15 @@ internal constructor(
     Firebase.messaging.subscribeToTopic(surveyId).await()
   }
 
-  /** Calls Cloud Function to refresh the current user's profile info in the remote database. */
+  /**
+   * Calls Cloud Function to refresh the current user's profile info in the remote database if
+   * network is available.
+   */
   override suspend fun refreshUserProfile() {
+    if (!networkManager.isNetworkAvailable()) {
```
gino-m, line 85: The RemoteDataStore just abstracts calls to the remote server, in this case Firestore. Deciding whether to do this or not (i.e., if online) should be up to the calling repository imo, not the data store.

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
```diff
@@ -18,22 +18,26 @@ package com.google.android.ground.system
...
+  fun isNetworkAvailable(): Boolean {
     val cm = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
     val networkInfo = cm.activeNetworkInfo
     return networkInfo?.isConnected ?: false
   }
 
   /** Throws an error if network isn't available. */
-  fun requireActiveNetwork(context: Context) {
-    if (!isNetworkAvailable(context)) {
+  fun requireActiveNetwork() {
```
gino-m, line 39: Same here, but `requireNetworkConnection`?

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
```diff
@@ -18,22 +18,26 @@ package com.google.android.ground.system
...
 
 /** Abstracts access to network state. */
-object NetworkManager {
+@Singleton
+class NetworkManager @Inject constructor(@ApplicationContext private val context: Context) {
 
   /** Returns true iff the device has internet connectivity, false otherwise. */
   @RequiresPermission("android.permission.ACCESS_NETWORK_STATE")
-  private fun isNetworkAvailable(context: Context): Boolean {
+  fun isNetworkAvailable(): Boolean {
```
gino-m, line 32: Can we also rename to `isNetworkConnected` for consistency with Android SDK's network info nomenclature?

gino-m (changes requested):
A few nits, otherwise LG!

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirestoreDataStore.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
shobhitagarwal1612: Done

gino-m (approved):

PR #1994, opened 2023-10-18
Fix bug in capture location task

### LINKED ISSUE #1986: Capture location flow broken
- [ ] Doesn't force on location lock
- [ ] Clicking "Capture" has no effect

shobhitagarwal1612:
Fixes #1986

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
@@ -222,6 +222,9 @@ constructor(
       .collect { (index, coordinates) ->
         if (index == 0) {
           panAndZoomCamera(coordinates)
+          // TODO(#1889): Track the zoom level in a VM associated with the MapFragment and use it in
+          //  panCamera().
+
           // Set a small delay before emitting another value to allow previous zoom animation to
           // finish. Otherwise, the map camera stops at some other zoom level.
           delay(3000)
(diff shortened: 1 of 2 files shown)


ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -219,6 +222,9 @@ constructor(
       .collect { (index, coordinates) ->
         if (index == 0) {
           panAndZoomCamera(coordinates)
+          // Set a small delay before emitting another value to allow previous zoom animation to
+          // finish. Otherwise, the map camera stops at some other zoom level.
+          delay(3000)
```
gino-m, line 227: I think the root problem is that we don't have a clear source of truth of the map state at the moment. In most cases, we're relying on the GoogleMap to tell us its current state, in others it's in the adapter, and yet others in the VM base class or concrete subclass. Ideally we would track the zoom level in a VM associated with the MapFragment and use it in `panCamera()`. Would you be able to add a `TODO(#1889)` for this?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
shobhitagarwal1612: Done

gino-m (approved):
Thanks!

PR #2003, opened 2023-10-21
Include pending submissions as well when displaying the bottom loi cards

### LINKED ISSUE #1998: Submission count not updated on card after submitting data
Card still says "No submissions". @shobhitagarwal1612 PTAL?

shobhitagarwal1612:
Fixes #1998

When the device is offline, then the loi entity in the table doesn't get updated. Therefore, the submission count in the bottom loi cards doesn't include the pending submissions when offline.

@gino-m  PTAL?


gino-m (approved):
LGTM!

PR #2008, opened 2023-10-22
Change default map type to terrain

### LINKED ISSUE #1989: Set Terrain as default map instead of Satellite
This helps make it clear for where offline imagery has been downloaded w/o the user switching map types

shobhitagarwal1612:
Fixes #1989

@gino-m  PTAL?


gino-m (approved):
🗺️

PR #2010, opened 2023-10-24
Allow Android app to use local emulator for development

gino-m:
This introduces two new variants `dev` and `local` to run against live dev Firebase and local FIrebase emulators, respectively. New debug build targets: `devDebug` and `localDebug`. Note that the latter currently only works from Android emulator. 

Created to facilitate creation of test data for debugging https://github.com/google/ground-platform/issues/1250.

@shobhitagarwal1612 PTAL?

@os-micmec FYI


ground/src/main/java/com/google/android/ground/repository/UserRepository.kt
```diff
@@ -54,7 +55,9 @@ constructor(
       Timber.d("Skipped refreshing user profile as device is offline.")
       return
     }
-    remoteDataStore.refreshUserProfile()
+    if (!USE_EMULATORS) {
+      remoteDataStore.refreshUserProfile()
+    }
```
shobhitagarwal1612, line 60: Instead of exposing USE_EMULATORS to the repositories, can we instead create a custom annotation (something like `@IsAnonymousSession`), provide it in the authentication module and use it here and everywhere else?

ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
```diff
@@ -91,11 +98,24 @@ constructor(
 
   override fun signIn() {
     signInState.onNext(SignInState.signingIn())
+    if (USE_EMULATORS) {
```
shobhitagarwal1612, line 101: Same as above

ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
```diff
@@ -105,7 +125,15 @@ constructor(
...
   }
 
-  private suspend fun getFirebaseAuth() = withContext(ioDispatcher) { FirebaseAuth.getInstance() }
+  private suspend fun getFirebaseAuth() =
+    withContext(ioDispatcher) {
+      val auth = FirebaseAuth.getInstance()
+      if (USE_EMULATORS) {
+        // Use the auth emulator so we can sign-in anonymously during dev.
+        auth.useEmulator(EMULATOR_HOST, AUTH_EMULATOR_PORT)
+      }
```
shobhitagarwal1612, line 134: Can we move this to the module instead?

ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
```diff
@@ -135,7 +163,8 @@ constructor(
   private fun getFirebaseAuthCredential(googleAccount: GoogleSignInAccount): AuthCredential =
     GoogleAuthProvider.getCredential(googleAccount.idToken, null)
 
-  private suspend fun getFirebaseUser(): User? = getFirebaseAuth().currentUser?.toUser()
+  private suspend fun getFirebaseUser(): User? =
+    if (USE_EMULATORS) anonymousUser else getFirebaseAuth().currentUser?.toUser()
```
shobhitagarwal1612, line 167: Same as above

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/repository/UserRepository.kt
gino-m: How would the annotation be used? Is there a reason not to reference the global BuildConfig in repositories? Testability?

shobhitagarwal1612: The main intention behind the suggestion is to enhance readability. Right now some logic around what to do is hidden behind this boolean. The repositories shouldn't care whether the app is running in an emulator or not. All it cares about is whether the session is anonymous or not. 

Then we can add some test coverage around it easily as well.

gino-m: Still not clear how the annotation helps, since they are static bindings, while user sessions are transient. In this case, perhaps just checking whether the current user is anonymous here is sufficient?

ground/src/main/java/com/google/android/ground/system/auth/GoogleAuthenticationManager.kt
gino-m: Unless we expose this as a Provider, it would require us to call getInstance() earlier, during DI initialization, rather than on demand. I can try.

gino-m:
Hi @shobhitagarwal1612 I moved build variant logic into modules and refactored the test auth manager into its own class. PTAL?

shobhitagarwal1612 (approved):
LGTM!

PR #2011, opened 2023-10-25
Immediately update location when location lock is tapped

### LINKED ISSUE #1999: Location does not update immediately when location lock is tapped
We should move to the last known location, not wait for a new location to be emitted. See:
https://developer.android.com/training/location/retrieve-current

shobhitagarwal1612:
Fixes #1999

Also adds some unit test coverage

@gino-m  PTAL?


gino-m (approved):
🧭

PR #2012, opened 2023-10-25
Remove unused preference for offline download

### LINKED ISSUE #1964: Network preference for downloading offline tiles doesn't do anything
We have a preference in settings page which is currently not used anywhere. We should consider cleaning it up or actually using it

shobhitagarwal1612:
Fixes #1964

@gino-m  PTAL?


gino-m (approved):
✂️!

PR #2016, opened 2023-10-26
Use coroutine instead of Rx when fetching offlineArea from local db

shobhitagarwal1612:
Towards #1581

Verified the changes for regression manually:
 * Opening the offline selector fragment
 * Ensuring that the size preview is shown correctly
 * Ensuring that I'm able to download offline area, view it in basemap on homescreen and delete it

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/offlineareas/viewer/OfflineAreaViewerViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/offlineareas/viewer/OfflineAreaViewerViewModel.kt
@@ -36,7 +36,6 @@ import com.google.android.ground.util.toMbString
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.rx2.await
 import timber.log.Timber
 
 /**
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/ui/offlineareas/viewer/OfflineAreaViewerViewModel.kt
```diff
@@ -87,7 +86,7 @@ constructor(
   fun initialize(args: OfflineAreaViewerFragmentArgs) {
     offlineAreaId = args.offlineAreaId
     viewModelScope.launch(ioDispatcher) {
-      val thisArea = offlineAreaRepository.getOfflineArea(offlineAreaId).await()
+      val thisArea = offlineAreaRepository.getOfflineArea(offlineAreaId)!!
```
gino-m, line 89: Previously, await would return `null` if the `Maybe` was empty. Should we be doing the same here?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/offlineareas/viewer/OfflineAreaViewerViewModel.kt
shobhitagarwal1612: It was actually `Single`, so await would throw an exception

gino-m: Sorry, you're right!

gino-m (approved):
LGTM!

PR #2020, opened 2023-10-27
Use coroutine in Navigator

shobhitagarwal1612:
Towards #1581

Verified the changes for regression by opening various screens

@gino-m  PTAL?


gino-m (approved):
Nice work! Go with the flow!

PR #2021, opened 2023-10-28
[Rx -> Coroutine] Migrate Offline Area Data layer

shobhitagarwal1612:
Towards #1581 

Migrates usages of flowable to flow in offline area screens

Verified for regressions by running the app locally, downloading offline areas and then removing them.

@gino-m @scolsen  PTAL?


gino-m (approved):
flow++ 👏🏻

PR #2023, opened 2023-10-29
[Data Collection] Update "Next" to "Done" when on last task

### LINKED ISSUE #1982: Data collection task next button labels incorrect
- [x] Label should be "Next" and not "Continue"
- [ ] On last task, button label should be "Done" <--  @amegantz Is that correct?

shobhitagarwal1612:
Fixes #1982

[done_button.webm](https://github.com/google/ground-android/assets/8918023/25d8cd3e-5060-45fe-88b9-dc7c0783e501)

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
@@ -145,15 +145,26 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
 
   protected fun addButton(action: ButtonAction): TaskButton {
     check(!buttons.contains(action)) { "Button $action already bound" }
+    val updatedAction = maybeOverrideButton(action)
     val button =
       TaskButtonFactory.createAndAttachButton(
-        action,
+        updatedAction,
         taskView.actionButtonsContainer,
         layoutInflater
       )
-    buttons[action] = button
+    buttons[updatedAction] = button
     return button
   }
+
+  /**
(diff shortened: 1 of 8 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -145,16 +149,27 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
...
       )
-    buttons[action] = button
+    buttons[updatedAction] = button
     return button
   }
 
+  /**
+   * Changes the button from "Next" to "Done" if the current task fragment is last in it's position.
+   */
+  private fun maybeOverrideButton(action: ButtonAction): ButtonAction =
```
gino-m, line 166: Can this be done reactively / declaratively using data binding rather than programmatically? e.g. by adding LiveData<Boolean> `isLastTask` to the ViewModel?

gino-m (changes requested):
One suggestions, otherwise LG!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
shobhitagarwal1612: I don't think we can do it reactively or using data binding. We create the layout of the task fragment dynamically

gino-m: I see. Sgtm!

gino-m (approved):

PR #2024, opened 2023-10-29
[Test] Update tests to also check for button order

shobhitagarwal1612:
Adds another map to store index of the added buttons for verification.

This change is no op. 

@gino-m  PTAL?

### DIFF
--- a/ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
+++ b/ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
@@ -35,6 +35,7 @@ import com.google.android.ground.ui.common.ViewModelFactory
 import com.google.android.ground.ui.datacollection.DataCollectionViewModel
 import com.google.android.ground.ui.datacollection.components.ButtonAction
 import com.google.common.truth.Truth.assertThat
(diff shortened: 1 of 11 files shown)


ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
```diff
@@ -73,8 +74,14 @@ abstract class BaseTaskFragmentTest<F : AbstractTaskFragment<VM>, VM : AbstractT
   }
 
   protected fun hasButtons(vararg buttonActions: ButtonAction) {
-    // TODO: Also check for order of action buttons.
+    // Verify the list matches (in any order)
```
gino-m, line 77: Nit: Move to method doc, add final "."

gino-m (changes requested):
LG. General nit: Methods which assert things should reflect this in their names, e.g. `assertThingsDontExplode`.

ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
shobhitagarwal1612: Done

gino-m (approved):
🧪 🧪 🧪

PR #2027, opened 2023-10-30
[WIP] Reactive survey list

### LINKED ISSUE #1973: Survey list doesn't update on remote changes (not reactive)
@shobhitagarwal1612 FYI

Proposed logic: stream of `List<Survey>` would be derived from (pseudocode):
`onlineStatus.flatMapLatest { if (online) remoteSurveysFlow else localSurveysFlow }`

shobhitagarwal1612:
Fixes #1973 

@... PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/RemoteDataStore.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/RemoteDataStore.kt
@@ -21,6 +21,7 @@ import com.google.android.ground.model.User
 import com.google.android.ground.model.locationofinterest.LocationOfInterest
 import com.google.android.ground.model.mutation.Mutation
 import com.google.android.ground.model.submission.Submission
+import com.google.firebase.firestore.ListenerRegistration
 import kotlinx.coroutines.flow.Flow
 
 /**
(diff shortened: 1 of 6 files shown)


ground/src/main/java/com/google/android/ground/persistence/remote/RemoteDataStore.kt
```diff
@@ -21,13 +21,18 @@ import com.google.android.ground.model.User
...
 
 /**
  * Defines API for accessing data in a remote data store. Implementations must ensure all
  * subscriptions are run in a background thread (i.e., not the Android main thread).
  */
 interface RemoteDataStore {
-  suspend fun loadSurveySummaries(user: User): List<Survey>
+  suspend fun loadSurveySummaries(
+    user: User,
+    cancelRegistrationCallback: (listenerRegistration: ListenerRegistration) -> Unit
```
gino-m, line 34: You shouldn't need to cancel explicitly if you're using `flatMapLatest`

shobhitagarwal1612: If we don't cancel it manually, then the app continues to listen to the changes even after switching to a new screen. I've added toast message to prove the same.

To observe it, 
1. Run the app
2. Observe the toast appearing for initial load
3. Now create a new project or rename an existing one using the same user in web app
4. Observe another toast message and the list should be updated as well.
5. Select any survey and move to next screen
6. Make another edit in the survey using web app
7. Observe that the toast is still shown

shobhitagarwal1612:
Closing in favor of https://github.com/google/ground-android/pull/2030/

PR #2037, opened 2023-11-03
Add "Previous" button to data collection tasks

### LINKED ISSUE #1981: "Previous" button not shown on data collection tasks
See mocks for details

shobhitagarwal1612:
Fixes #1981

[previous_button.webm](https://github.com/google/ground-android/assets/8918023/b6b5412b-388a-43d6-a542-f5f378fac050)

@gino-m PTAL?

### DIFF
--- a/ground/build.gradle
+++ b/ground/build.gradle
@@ -146,6 +146,9 @@ android {
     testOptions {
         unitTests {
             includeAndroidResources = true
+            all {
+                jvmArgs '-Xmx4g' // '-XX:+HeapDumpOnOutOfMemoryError'
+            }
         }
         animationsDisabled = true
     }
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -158,6 +158,22 @@ internal constructor(
   }
 
   /**
+   * Validates the user's input and displays an error if the user input was invalid. Moves back to
+   * the previous Data Collection screen if the user input was valid.
+   */
+  fun onPreviousClicked(position: Int) {
+    check(position != 0)
+
+    val validationError = currentTaskViewModel?.validate()
+    if (validationError != null) {
+      popups.get().showError(validationError)
+      return
+    }
+
+    updateCurrentPosition(position - 1)
+  }
+
+  /**
    * Validates the user's input and displays an error if the user input was invalid. Progresses to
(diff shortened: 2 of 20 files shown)


gino-m:
> The build is failing due to
> 
> ```
>     android.view.InflateException
>         Caused by: android.view.InflateException
>             Caused by: java.lang.reflect.InvocationTargetException at SurveySelectorFragmentTest.kt:246
>                 Caused by: java.lang.OutOfMemoryError
> ```
> 
> @gino-m Any ideas if it could be related to any recent changes?

Trying this now, but it seems strange since `SurveySelectorFragmentTest` only has 238 lines.

gino-m (approved):

shobhitagarwal1612:
I think the issue is that GCB uses `N1_HIGHCPU_8` machine which has max 8 GB of ram.

gino-m:
> I think the issue is that GCB uses `N1_HIGHCPU_8` machine which has max 7.2 GB of ram.
> 
> https://cloud.google.com/compute/docs/general-purpose-machines#n1_machine_types

Do these tests complete when run locally? Is it normal for Robolectric tests to require so much RAM, or could something else be going on here?

shobhitagarwal1612:
The tests pass in the local environment only after increasing the memory limits. It seems unusual for unit tests to consume so much memory. I do remember seeing a bunch of warnings related to memory leakage when running the tests at HEAD as well.

shobhitagarwal1612:
The issue is happening even if I remove the newly added unit tests and only keep the "previous" button related changes. So, there is no way to check-in the changes without breaking the build. I'll try understanding how to resolve the memory leakage issue which is causing the heap size to grow so big.

gino-m:
> The issue is happening even if I remove the newly added unit tests and only keep the "previous" button related changes. So, there is no way to check-in the changes without breaking the build. I'll try understanding how to resolve the memory leakage issue which is causing the heap size to grow so big.

I see the same when running locally. The error only occurs when running multiple tests, running testing in `SurveySelectorFragmentTest` pass. Also seeing this error in unrelated tests like `com.google.android.ground.ui.signin.SignInFragmentTest`.

shobhitagarwal1612:
I was able to generate the hprof file containing the heap dump. Analyzing it now.

shobhitagarwal1612:
I also found that gradle.properties options aren't applied to unit tests. So, slightly increasing the heap size for unit tests should fix the problem. Tried looking at heap but wasn't able to narrow down to a specific issue. In general, a major chuck of memory was taken by `shadowVMRuntime` which is used by robolectric.

PR #2045, opened 2023-11-06
Fix TileSourceEntity id generation.

### LINKED ISSUE #2044: [Persistence] Crash on UNIQUE constraint failure
After starting up the app post-hiatus I encountered the following crash:

```
java.lang.Throwable: UNIQUE constraint failed: offline_base_map_source.id (code 1555 SQLITE_CONSTRAINT_PRIMARYKEY)
```

That's reasonable enough, but we probably shouldn't crash on this, esp. for offline basemaps, and instead handle the error gracefully. Filing this bug to track that.

scolsen:
Room treats non-null values assigned to boxed types (e.g. Int?, Integer) as explicit values and will not auto-generate IDs when these values are provided. This was causing UNIQUNESS constraint failures when inserting/updating tilesource entites, as we were using `0` as a default.

Converting to a non-boxed primitive type `Int` results in the correct behavior (Room treats `0` as an indicator that a value should be generated).

Fixes #2044

Note: We can also fix this by setting the default to `null` and retaining the type, so if we need to keep this nullable for some reason that's possible too.

### DIFF
--- a/ground/src/main/java/com/google/android/ground/Config.kt
+++ b/ground/src/main/java/com/google/android/ground/Config.kt
@@ -26,7 +26,7 @@ object Config {
 
   // Local db settings.
   // TODO(#128): Reset version to 1 before releasing.
-  const val DB_VERSION = 108
+  const val DB_VERSION = 109
   const val DB_NAME = "ground.db"
 
   // Firebase Cloud Firestore settings.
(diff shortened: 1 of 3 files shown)


shobhitagarwal1612 (approved):

shobhitagarwal1612:
If possible, can we also add a basic unit test?

scolsen:
Added!

PR #2050, opened 2023-11-07
[WORK IN PROGRESS] Use Protobuf serialization for LOI properties in local db

gino-m:


gino-m:
@rachaprince Let's punt on this for now. Filed #2054 to track.

Issue #2049, opened 2023-11-07
[Code health] Introduce protos as structure for checking data in Firestore at compile-time


gino-m:
See also https://github.com/google/ground-platform/issues/1277

@os-micmec FYI tracking bug also here

Proposed micro-design (read):

1. Use [`DocumentSnapshot.getData()`](https://firebase.google.com/docs/reference/kotlin/com/google/firebase/firestore/DocumentSnapshot#getData(com.google.firebase.firestore.DocumentSnapshot.ServerTimestampBehavior)) to fetch data as a `Map<String!, Any!>`
2. Tweak data in `Map` when needed. We should try to minimize special logic in this layer.
3. Recursively wrap `Map` in Gson `JsonElement` and related classes.
4. Use Gson to populate classes generated with [`json-kotlin-schema-codegen`](https://github.com/pwall567/json-kotlin-schema-codegen)

We would do the reverse for writing:

1. Convert object to JsonElement using Gson
1. Convert JsonElement to Map
2. Tweak Map as needed.
3. Pass `Map` to update methods in Firestore libs.

We can preserve our current representation of nested arrays, using indexed maps instead.

Alternatives considered:

* kpropmap libraries requires conversion to its own `PropertyMap` interface. It's also not clear how we'd implement polymorphism or custom (de)serializers.
* ~kotlinx.serialization Properties requires `@Serialization` annotations on data classes ([issue](https://github.com/Kotlin/kotlinx.serialization/issues/1808#issuecomment-1001512246)). Our data classes are generated, so we can't do that.~ Also not correct - you can [set custom annotations](https://github.com/pwall567/json-kotlin-schema-codegen/blob/main/CONFIG.md#annotations)
* ~Gson and kotlinx.serialization Json require conversion to/from JSON strings, which can be _slow_.~ <-- not true - we can also recursively wrap our object graph to `JsonObject` and `JsonElement` as per [this example](https://medium.com/@iliamsharipov_56660/handling-polymorphism-with-gson-f4a702014ffe)
* `DocumentSnapshot.toObject` doesn't handle polymorphism, and doesn't offer much flexibility in customizing mappings.

@sufyanAbbasi FYI

Comments which follow are from early research and are out of date.

gino-m:
Perhaps something like JSON Schema could also work?

Code gen:

* https://github.com/pwall567/json-kotlin-schema-codegen
* https://github.com/jimblackler/jsonschematypes/tree/master/codegen

[...]

PR #2051, opened 2023-11-07
[WORK IN PROGRESS] protobuf<>Firestore proof of concept

gino-m:
Towards #2049


gino-m:
Also, protobuf might not be the right tool due to the reasons mentioned here:
https://groups.google.com/g/firebase-talk/c/0y8nihK3XxE

gino-m:

PR #2053, opened 2023-11-08
DRAFT: Photo Sync Status

### LINKED ISSUE #877: [Sync status] Include photo upload status in sync status

scolsen:
NOT DONE YET. Sharing early for feedback. @gino-m, @shobhitagarwal1612, @rachaprince FYI.

Fixes #877

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/Photo.kt
+++ b/ground/src/main/java/com/google/android/ground/model/Photo.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model
+
+import com.google.android.ground.model.job.Job
+import com.google.android.ground.model.locationofinterest.LocationOfInterest
+
+/** Represents a single photo associated with a given [LocationOfInterest] and [Job]. */
+data class Photo(
+  val id: String,
+  val surveyId: String,
+  val locationOfInterestId: String,
+  val jobId: String,
+  val created: AuditInfo,
+  val lastModified: AuditInfo,
+  val name: String,
+)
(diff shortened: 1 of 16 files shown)


ground/src/main/java/com/google/android/ground/model/Photo.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model
+
+import com.google.android.ground.model.job.Job
+import com.google.android.ground.model.locationofinterest.LocationOfInterest
+
+/** Represents a single photo associated with a given [LocationOfInterest] and [Job]. */
```
gino-m, line 21: As with other tasks, the data collected by the user is associated with a taskId and modeled by a `Value` subclass. This class should implement `Value` and be associated with the Task by virtue of its containing in `SubmissionData`.

scolsen: ahhhh, ok yeah that makes sense -- I noticed we didn't have one for Photos, I'll make the change!

ground/src/main/java/com/google/android/ground/model/mutation/PhotoMutation.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.mutation
+
+/** Represents a modification of a photo in remote storage. */
+data class PhotoMutation(val submissionMutationId: String, val remotePath: String) : Mutation()
```
gino-m, line 19: If Photo is just a special Value class, wouldn't SubmissionMutation already cover this? Or is that too coarse grained to capture the photo mutation? I think the easiest way of doing this with the current model would be to make SubmissionMutation more fine-grained (per value rather than per submission).

ground/src/main/java/com/google/android/ground/persistence/local/room/entity/PhotoMutationEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+      ForeignKey(
+        entity = SubmissionMutationEntity::class,
+        parentColumns = ["id"],
+        childColumns = ["submission_mutation_id"],
+        onDelete = ForeignKey.CASCADE
+      ),
+    ],
+  indices = [Index("submission_mutation_id")]
+)
+data class PhotoMutationEntity(
```
gino-m, line 42: If we can refactor the submission mutation logic as mentioned above we can avoid a lot of complexity - the new DAO, entity, mutation, conversion logic, etc.

ground/src/main/java/com/google/android/ground/model/mutation/PhotoMutation.kt
scolsen: Yep, that approach sounds way better. We can just handle this as part of the submission mutation.

scolsen:
Closing this one since it's no longer relevant.

PR #2057, opened 2023-11-10
Remove view/edit submission fragments

shobhitagarwal1612:
Removes fragments related to viewing and editing submissions. These fragments are currently not used anywhere and there is no plan to use them in future as well.

Verified that this doesn't cause any regressions by running the app locally.

@gino-m PTAL?


gino-m (approved):
🪓 🪓 🪓 !

PR #2069, opened 2023-11-15
Cleanup obsolete flowables

shobhitagarwal1612:
Remove unused code and update unit tests

@scolsen  PTAL?

### DIFF
--- a/ground/src/test/java/com/google/android/ground/repository/LocationOfInterestRepositoryTest.kt
+++ b/ground/src/test/java/com/google/android/ground/repository/LocationOfInterestRepositoryTest.kt
@@ -54,6 +54,7 @@ class LocationOfInterestRepositoryTest : BaseHiltTest() {
   @Inject lateinit var fakeAuthenticationManager: FakeAuthenticationManager
   @Inject lateinit var fakeRemoteDataStore: FakeRemoteDataStore
   @Inject lateinit var locationOfInterestRepository: LocationOfInterestRepository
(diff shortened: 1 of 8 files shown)


ground/src/test/java/com/google/android/ground/repository/LocationOfInterestRepositoryTest.kt
```diff
@@ -93,11 +94,9 @@ class LocationOfInterestRepositoryTest : BaseHiltTest() {
   fun testApplyAndEnqueue_enqueuesLoiMutation() = runWithTestDispatcher {
     locationOfInterestRepository.applyAndEnqueue(mutation)
 
-    locationOfInterestRepository
-      .getIncompleteLoiMutationsOnceAndStream(LOCATION_OF_INTEREST.id)
```
scolsen, line 97: The semantics here are a bit difference than getSurveyMutations -- if this was the only use of this method, I think that's fine, just calling it out in case.

scolsen (approved):

ground/src/test/java/com/google/android/ground/repository/LocationOfInterestRepositoryTest.kt
shobhitagarwal1612: I wanted to deprecate `getIncompleteLoiMutationsOnceAndStream` without removing this test case. So, used `getSurveyMutations` as a fallback.

PR #2070, opened 2023-11-19
Use SharedFlow for handling openDrawer requests

shobhitagarwal1612:
Replaces `PublishProcessor` with `SharedFlow` for handling navigation drawer requests.

Tested by running the app locally.

@scolsen  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
@@ -17,16 +17,19 @@ package com.google.android.ground.ui.home
 
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.asLiveData
+import androidx.lifecycle.viewModelScope
 import com.google.android.ground.repository.SurveyRepository
 import com.google.android.ground.rx.Nil
-import com.google.android.ground.rx.annotations.Hot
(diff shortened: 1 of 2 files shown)


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -36,12 +39,15 @@ internal constructor(
   private val surveyRepository: SurveyRepository
 ) : AbstractViewModel() {
 
-  val openDrawerRequests: @Hot FlowableProcessor<Nil> = PublishProcessor.create()
+  private val _openDrawerRequests: MutableSharedFlow<Nil> = MutableSharedFlow()
+  val openDrawerRequestsFlow: SharedFlow<Nil> =
+    _openDrawerRequests.shareIn(viewModelScope, SharingStarted.Lazily, replay = 0)
```
scolsen, line 44: I think here you can just use `asSharedFlow` instead. I believe emissions will automatically occur in the ViewModelScope because of the `launch` call below.

scolsen (approved):

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
shobhitagarwal1612: Thanks! Done

PR #2106, opened 2023-12-11
Directly bind the terms of service text to view

### LINKED ISSUE #1478: [Code health] Use MutableLiveData in TermsOfServiceViewModel for binding view text

shobhitagarwal1612:
Fixes #1478 

Since terms of service text is passed to the fragment from previous screen, it will always remain constant and there is no need to actually store a reference to it in the view model. 

Verified the changes by opening the affected screens on a physical device and checking for regressions.

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/res/layout/fragment_terms_service.xml
+++ b/ground/src/main/res/layout/fragment_terms_service.xml
@@ -21,9 +21,6 @@
 
   <data>
     <variable
-      name="termsOfServiceText"
-      type="String" />
-    <variable
       name="viewModel"
       type="com.google.android.ground.ui.tos.TermsOfServiceViewModel" />
   </data>
@@ -59,7 +56,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_margin="8dp"
-        android:text="@{termsOfServiceText}"
+        android:text="@{viewModel.getTermsOfServiceText()}"
         android:visibility="visible" />
     </ScrollView>
     <Button
--- a/ground/src/main/res/layout/fragment_terms_service.xml
+++ b/ground/src/main/res/layout/fragment_terms_service.xml
@@ -21,6 +21,9 @@
 
   <data>
     <variable
+      name="termsOfServiceText"
+      type="String" />
+    <variable
       name="viewModel"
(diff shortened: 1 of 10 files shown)


gino-m (commented):
Fragment args are meant to send only small amounts of data, usually IDs, between views. The actual text is loaded from the DB and it represents view state, which is why it's modeled as a ViewModel stat. From docs:

> In general, you should strongly prefer passing only the minimal amount of data between destinations. For example, you should pass a key to retrieve an object rather than passing the object itself, as the total space for all saved states is limited on Android.

https://developer.android.com/guide/navigation/use-graph/pass-data

shobhitagarwal1612:
Should we fetch the ToS again from the repository in TermsOfServiceViewModel?

gino-m:
> Should we fetch the ToS again from the repository in TermsOfServiceViewModel?

Yes, I think that's the right thing to do.

shobhitagarwal1612:
> > Should we fetch the ToS again from the repository in TermsOfServiceViewModel?
> 
> Yes, I think that's the right thing to do.

Done

ground/src/main/res/layout/fragment_terms_service.xml
```diff
@@ -56,7 +56,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_margin="8dp"
-        android:text="@{viewModel.termsOfServiceText}"
+        android:text="@{viewModel.getTermsOfServiceText()}"
```
gino-m, line 59: Why not expose this as a LiveData<String> field `termsOfServiceText`? I think it would look something like:

`termsOfServiceText = liveData { emit(termsOfServiceRepository.getTermsOfService()?.text) }`

gino-m (changes requested):
Sorry, one more suggestion

ground/src/main/res/layout/fragment_terms_service.xml
shobhitagarwal1612: This is cool! I didn't know we could do this.

gino-m (approved):
LGTM, thanks!

PR #2108, opened 2023-12-11
Replace suggestLoiTaskType with DataCollectionStrategy

### LINKED ISSUE #2097: Replace `suggestLoiTaskType` with special LOI task
Sync with https://github.com/google/ground-platform/issues/1213

@rfontanarosa FYI

gino-m:
Fixes google/ground-android#2097


gino-m:
This is now working, modulo:

- [ ] https://github.com/google/ground-platform/issues/1349
- [ ] https://github.com/google/ground-platform/issues/1213

Will mark Ready for review once those are fixed and/or out for review. @rfontanarosa FYI.

gino-m:
@scolsen PTAL

ground/src/main/java/com/google/android/ground/model/job/Job.kt
```diff
@@ -19,21 +19,28 @@ import android.graphics.Color
...
- * @param suggestLoiTaskType the type of task used to suggest the LOI for this Job. Null if the job
- * is already associated with an LOI.
- */
 data class Job(
   val id: String,
   val style: Style? = null,
   val name: String? = null,
   val tasks: Map<String, Task> = mapOf(),
-  val suggestLoiTaskType: Task.Type? = null,
+  val strategy: DataCollectionStrategy = DataCollectionStrategy.MIXED
```
scolsen, line 27: Should this have a default? Don't we need this to be defined by survey organizers? Shouldn't it be a required argument, and shouldn't we error out if it's not available/provided?

ground/src/main/java/com/google/android/ground/model/job/Job.kt
```diff
@@ -19,21 +19,28 @@ import android.graphics.Color
...
 ) {
+  enum class DataCollectionStrategy {
+    PREDEFINED,
```
scolsen, line 30: So far, it's been helpful to have `UNKNOWN` for cases in when the web app evolves before the android app/field names change. Not sure if you want to add one here.

ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
```diff
@@ -112,7 +113,7 @@ fun JobEntityAndRelations.toModelObject(): Job {
     jobEntity.style?.toModelObject(),
     jobEntity.name,
     taskMap.toPersistentMap(),
-    jobEntity.suggestLoiTaskType?.let { Task.Type.valueOf(it) }
+    jobEntity.strategy.let { DataCollectionStrategy.valueOf(it) }
```
scolsen, line 116: Why not just `DataCollectionStrategy.valueOf(jobEntity.strategy)`?

Or. like other entities, I might expect to see `jobEntity.strategy.toModelObject()`

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/JobNestedObject.kt
```diff
@@ -24,5 +25,5 @@ data class JobNestedObject(
   val defaultStyle: StyleNestedObject? = null,
   val name: String? = null,
   val tasks: Map<String, TaskNestedObject>? = null,
-  val suggestLoiTaskType: String? = null
+  val strategy: String = MIXED.toString()
```
scolsen, line 28: I worry that this default value and the default value for the Job model object could get out of sync. I think removing a default from the model might be one way to avoid that.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -86,12 +83,8 @@ internal constructor(
...
   private val job: Job =
     activeSurvey.getJob(requireNotNull(jobId)) ?: error("couldn't retrieve job for $jobId")
-  val tasks: List<Task> = buildList {
-    if (job.suggestLoiTaskType != null && loiId == null) {
-      add(createSuggestLoiTask(job.suggestLoiTaskType))
-    }
-    addAll(job.tasksSorted)
-  }
+  // LOI creation task is included only on "new data collection site" flow..
+  val tasks: List<Task> = job.tasksSorted.filter { loiId == null || !it.isAddLoiTask }
```
scolsen, line 87: I find this filter a little tricky to understand. Seems like we're getting tasks that either have no LOI Id or that are explicitly _not_ add LOI tasks. The latter makes sense given the comment, but I don't understand why we filter for tasks with `loiId == null`

scolsen (approved):
couple of small comments but lgtm

ground/src/main/java/com/google/android/ground/model/job/Job.kt
gino-m: The default value was mainly for tests, but that's probably the wrong reason. Updating now.

ground/src/main/java/com/google/android/ground/model/job/Job.kt
gino-m: I think UNKNOWN is useful for serialized representations like protos, but for model objects less so; if the strategy is invalid in remote or local data store, by the time to deserialize we should either have chosen a fall-back value, or thrown an error.

ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
gino-m: > Why not just `DataCollectionStrategy.valueOf(jobEntity.strategy)`?

Because `strategy` might be null, and Enum.valueOf() doesn't accept nulls
> 
> Or. like other entities, I might expect to see `jobEntity.strategy.toModelObject()`

Because in `jobEntity` strategy is a String, and defining `fun String.?toModelObject()` would be too broad.

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/JobNestedObject.kt
gino-m: Good idea. Done!

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
gino-m: We're including all tasks when loiId = null. I can rewrite this, lmkwyt.

ground/src/main/java/com/google/android/ground/model/job/Job.kt
gino-m: Doing this turned out to be more disruptive than I expected; basically we end up needing to defined it in all tests even where it's not relevant. Ideally we'd have a custom constructor with default values for tests, but since we don't have yet that we can apply UNKNOWN as a default based on your next suggestion

Issue #2122, opened 2023-12-15
Hide "Collect data" button for jobs with no tasks


gino-m:
Don't show button if "add LOI" is the only task, since there's no additional data to capture.

PR #2124, opened 2023-12-15
Handle unsupported task types.

### LINKED ISSUE #2107: Survey list crashes app with fatal exception
When using dev db due to invalid data:

```
FATAL EXCEPTION: main
                 Process: com.google.android.ground, PID: 8010
                 java.lang.IllegalArgumentException: No enum constant com.google.android.ground.model.task.Task.Type.DRAW_POLYGON
                 	at java.lang.Enum.valueOf(Enum.java:257)
                 	at com.google.android.ground.model.task.Task$Type.valueOf(Unknown Source:2)
                 	at com.google.android.ground.persistence.local.room.converter.ConverterExtKt.toModelObject(ConverterExt.kt:115)
                 	at com.google.android.ground.persistence.local.room.converter.ConverterExtKt.toModelObject(ConverterExt.kt:366)
                 	at com.google.android.ground.persistence.local.room.stores.RoomSurveyStore$special$$inlined$map$1$2.emit(Emitters.kt:224)
                 	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
                 	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
                 	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:87)
                 	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:37)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Unknown Source:14)
                 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
                 	at android.os.Handler.handleCallback(Handler.java:883)
                 	at android.os.Looper.loop(Looper.java:214)
                 	at android.app.ActivityThread.main(ActivityThread.java:7356)
                 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
                 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
                 	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@b7159bb, Dispatchers.Main.immediate]
```

Instead, a warning should be logged and the task ignored.

scolsen:
We now handle unsupported task types when these are received from remote data sources. When the Job has other, valid tasks, we simply ignore the invalid task type. For the edge case in which the Job's one and only task is invalid, we render a special message to the user, urging them to contact the survey organizer.

Screenshot 2023-12-15 at 12 24 36 PM

Fixes #2107

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
@@ -162,8 +162,7 @@ fun LocationOfInterestEntity.toModelObject(survey: Survey): LocationOfInterest =
       geometry = geometry.getGeometry(),
       submissionCount = submissionCount,
       properties = properties,
-      job =
-        survey.getJob(jobId = jobId)
+      job = survey.getJob(jobId = jobId)
           ?: throw LocalDataConsistencyException(
             "Unknown jobId ${this.jobId} in location of interest ${this.id}"
           )
--- a/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
@@ -112,7 +112,15 @@ fun JobEntityAndRelations.toModelObject(): Job {
     id = jobEntity.id,
     style = jobEntity.style?.toModelObject(),
     name = jobEntity.name,
-    strategy = jobEntity.strategy.let { DataCollectionStrategy.valueOf(it) },
+    strategy =
+      jobEntity.strategy.let {
+        try {
+          DataCollectionStrategy.valueOf(it)
+        } catch (e: IllegalArgumentException) {
+          Timber.e("unknown data collection strategy $it")
+          DataCollectionStrategy.UNKNOWN
+        }
+      },
     tasks = taskMap.toPersistentMap()
   )
 }
@@ -154,7 +162,8 @@ fun LocationOfInterestEntity.toModelObject(survey: Survey): LocationOfInterest =
       geometry = geometry.getGeometry(),
       submissionCount = submissionCount,
       properties = properties,
-      job = survey.getJob(jobId = jobId)
+      job =
+        survey.getJob(jobId = jobId)
           ?: throw LocalDataConsistencyException(
             "Unknown jobId ${this.jobId} in location of interest ${this.id}"
           )
--- a/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
@@ -112,7 +112,14 @@ fun JobEntityAndRelations.toModelObject(): Job {
     jobEntity.style?.toModelObject(),
     jobEntity.name,
     taskMap.toPersistentMap(),
-    jobEntity.suggestLoiTaskType?.let { Task.Type.valueOf(it) }
+    jobEntity.suggestLoiTaskType?.let {
+      try {
+        Task.Type.valueOf(it)
+      } catch (e: Exception) {
+        Timber.e("unknown task type: $it")
+        Task.Type.UNKNOWN
+      }
+    }
   )
 }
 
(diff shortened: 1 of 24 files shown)


shobhitagarwal1612:
Should we also show a toast message to the user saying that some tasks could not be loaded?

shobhitagarwal1612:
I'd also suggest adding some test coverage for this edge case. It's easy to miss this for validation when testing manually.

gino-m:
Related to #2122.

scolsen:
Removed the tasks not defined hint, since we'll tackle this by implementing #2122 instead. @gino-m rdy for a review

ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
```diff
@@ -112,7 +112,14 @@ fun JobEntityAndRelations.toModelObject(): Job {
     jobEntity.style?.toModelObject(),
     jobEntity.name,
     taskMap.toPersistentMap(),
-    jobEntity.suggestLoiTaskType?.let { Task.Type.valueOf(it) }
+    jobEntity.suggestLoiTaskType?.let {
+      try {
+        Task.Type.valueOf(it)
+      } catch (e: Exception) {
+        Timber.e("unknown task type: $it")
+        Task.Type.UNKNOWN
```
gino-m, line 120: I think it would be more correct here to skip the task rather than create a task with an "UNKNOWN" task type, since those tasks will be inactionable downstream.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -75,18 +75,20 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
     view.doOnAttach {
-      @Suppress("UNCHECKED_CAST")
-      viewModel = dataCollectionViewModel.getTaskViewModel(position) as T
+      @Suppress("UNCHECKED_CAST") val vm = dataCollectionViewModel.getTaskViewModel(position) as? T
 
-      taskView.bind(this, viewModel)
-      taskView.addTaskView(onCreateTaskBody(layoutInflater))
+      if (vm != null) {
```
gino-m, line 80: Let's use short circuit to avoid unnecessary nesting.  I think you can just do:

```kt
 val vm = dataCollectionViewModel.getTaskViewModel(position) as? T ?: return
```

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/local/room/converter/ConverterExt.kt
scolsen: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
scolsen: I could in theory, but it would need to be a labelled return, which our static analyzer disallows.

gino-m: Can you override `onAttach` instead of `onViewCreated`? Then you also won't need the labelled return and `view.doOnAttach`

scolsen: cool, I'll do that!

scolsen: switching to onAttach breaks a bunch of tests--something about a navcontroller not being attached. I think(?) it's because we call a bunch of functions that assume the view is available at that point

gino-m (approved):

PR #2125, opened 2023-12-15
Render polygons opaquely when Jobs lack a default style.

### LINKED ISSUE #2092: Imported polygons not appearing when zooming into cluster
@shobhitagarwal1612 PTAL?

Cluster markers are shown when zoomed out, polygons not shown when zoomed in.

scolsen:
Previously, if a Job lacked a default style value, we rendered features with the color value 0. We use ARGB color format, in which this value corresponds to a completely transparent color. In result, polygons and points in a survey would not be rendered at all if a survey's job lacked a proper default style.

We now instead render features with a default, opaque black color when the job lacks default styling. This is the same behavior as the web app.

Screenshot 2023-12-15 at 1 23 51 PM

Fixes #2092


shobhitagarwal1612:
Can we also add a unit test for this?

gino-m (approved):

PR #2127, opened 2023-12-15
SurveySelector - Only render the latest list of surveys

### LINKED ISSUE #2109: Survey list jitters after tap
This occurs when scrolling down and selecting a survey at the bottom of the list.

Ideally, rather than refreshing the list, we would show a modal spinner or separate loading screen while the survey is loading.

@shobhitagarwal1612 PTAL?

[Screen_recording_20231212_191324.webm](https://github.com/google/ground-android/assets/228050/fc11b619-8fc7-4df5-90f6-cfe2c94a61a5)

scolsen:
Fixes some UI jitter by restricting our rendering of surveys in the selector to the most recent list at the time the user opens the selector. This means the list isn't fully dynamic (if an organizer adds them to a survey while the selector is open, they'll need to close then re-open the selector). However, it does fix the rendering and UX issues that resulted from our previous dynamism, and this seems like an acceptable tradeoff.

We do re-render the list in the case that the user removes a survey from offline use.

https://github.com/google/ground-android/assets/11237600/dcef1e7a-674e-4e0a-bf7c-fb934879829d

Fixes #2109


gino-m:
Hi @scolsen! The reactive list was actually by design - it was updated from non-reactive, so this change would be a regression. Instead, can we unsubscribe to changes and show a spinner while loading the survey?

scolsen:
Ah sure thing — we can just cancel on survey selection instead, lemme make the change.

gino-m:
Thanks!

scolsen:
kk, @gino-m we now handle this via coroutine cancellation -- the effect is the same. On selection we briefly show a spinner then immediately take the user to the home screen without re-rendering the list.

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -96,6 +98,11 @@ class SurveySelectorFragment : Hilt_SurveySelectorFragment(), BackPressListener
     }
   }
 
+  fun activateSurvey(surveyId: String) {
```
gino-m, line 102: This would work, but it introduces more view logic into the Fragment. To keep this in the VM, can we just stop `viewModel.getSurveyList()` from sending updates after click?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
scolsen: imo this is view logic and this is the simplest possible approach. We're saying we don't want to update the view anymore when a certain condition occurs (user clicks card). Doing it on the VM side requires something similar to what I had in this PR before, we need to convolute our flow such that we have fine grained control over emissions based on VM state. Flows are declarative, and it's best imo when you can write them as such without weird emit/don't emit logic. The simplest way I can think of to keep this in the VM would require something like this:

```kotlin
val shouldEmit: Bool = true 

surveyRepository
      .getSurveyList(authManager.getAuthenticatedUser())
      .onStart { setLoading() }
      .map { surveys -> surveys.sortedBy { it.title }.sortedByDescending { it.availableOffline } }
      .onEach {
        if (it.isEmpty()) {
          setNotFound()
        } else {
          setLoaded()
        }
      }.transform {
        if (shouldEmit) { emit(it) }
      }

fun activateSurvey() {
  ...
  shouldEmit = false
}
```

Which, admittedly isn't too bad, but I think it's slightly simpler to just have the view say "I want updates and now I don't" than to have some mechanism whereby the VM state change fundamentally alters emissions logic. It would make the flow more difficult to reason about. The view already controls the subscription and I think your suggestion of simply stopping the subscription on click makes more sense than contorting our flow -- if we really want this to be fully reactive, that is. If we don't actually care about real time dynamic reactivity in response to db changes during this interaction there are much simpler ways to do it.

iI'm fine with it either approach. the fragment driven approach seems a little simpler to me and also makes more sense to me from a "who controls what" perspective. Our underlying data model isn't *really* changing. We just want the view to stop responding to it. At the same time, I see the benefit to keeping most logic in the VM. I'm will ing to make the swap if you think the transform approach is OK.

gino-m: > imo this is view logic and this is the simplest possible approach. We're saying we don't want to update the view anymore when a certain condition occurs (user clicks card). Doing it on the VM side requires something similar to what I had in this PR before, we need to convolute our flow such that we have fine grained control over emissions based on VM state. Flows are declarative, and it's best imo when you can write them as such without weird emit/don't emit logic. The simplest way I can think of to keep this in the VM would require something like this:

After we spoke, I realized the reason the list order isn't deterministic. In `SurveySelectorViewModel:68`:

```kt
      .map { surveys -> surveys.sortedBy { it.title }.sortedByDescending { it.availableOffline } }
```

I believe the call to `sortedByDescending` doesn't present the order introduced by the first call `sortedBy`! Instead, I think we'd want something like:

```kt
    .map { surveys -> surveys.sortedWith(compareBy({ !it.availableOffline }, { it.title })) )
```

We probably don't even need `distinctUntilChanged()` after `map()` at that point, since in the reactive model state delta optimizations should be done in the view if needed.

Would you like to make this change in this PR? We can tackle the cancellation issue in a follow-up is necessary.

scolsen: Aha! Indeed. `SortedByDescending` uses the natural sort order, which would be whatever order  in which we happened to have them in the list, good catch!

scolsen: So, I applied the sort and...the jitter still happens when we only use the sort, but adding a distinctUntilChanged works! Since that's the simplest fix, let's do that. However, we still ultimately have a data replication issue upstream stemming from the way room's flow/coroutine APIs work. If you log the data coming back from the DB on the survey list calls you'll note that we get an emission like 5ish times on a single fetch:

- Room re-emits data in a flow any time the *table* changes: https://medium.com/androiddevelopers/room-flow-273acffe5b57#:~:text=Because%20the%20barking,in%20the%20database%3A
- Our survey list is not just a survey, but is actually `SurveyEntityAndRelations`. My current hypothesis is that some other stream/logic that we execute upon fetching survey data causes a subsequent update to one of the related entities (Job or tile source) which in turn, due to the foreign keys in the relation, indicates to room that the `SurveyEntityAndRelations` data changed, thus re-triggering the flow emission. This seems to happen roughly 5 or so times whenever we get the list of all surveys, as in the selector view model. 

I'd have to investigate further to be certain but this seems like the main cause

gino-m (approved):
👏 Nice, LGTM!

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
gino-m: Good find, nice sleuthing! I'm surprised the order would change if we're sorting by a stable attribute, in that case was the jank due to list items appearing/disappearing? `distinctUntilChanged` sounds like a reasonable solution, not sure what you're describing from Roome is a bug, sounds like it's WAI?

scolsen: Yeah, I think it is WAI and it looks like `distinctUntilChanged` is the recommend solution when you only want updates for a subset, so I think it's all good. And yes, I think the order should be stable if we keep a stable sort—without one I believe it just changes because Room doesn't make any promises about ordering of returned elements (I think).

gino-m: Makes sense!

PR #2146, opened 2023-12-27
Simplify "draw an area" impl

### LINKED ISSUE #2143: ["Draw area" task] Simplify impl - Polygon and LinearRing not needed while drawing
https://github.com/google/ground-android/blob/260e797d7010dddc7e834aac5822d41807d0b4e1/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/PolygonDrawingViewModel.kt#L177-L186

While drawing, all possible states can be represented as a `LineString`, so we shouldn't need this additional logic.

@shobhitagarwal1612 Can you please confirm?

gino-m:
Fixes #2143

@shobhitagarwal1612 PTAL?


shobhitagarwal1612:
LG. Some tests are still failing.

gino-m:
@shobhitagarwal1612 I wasn't able to resolve one of the failing tests so I was forced to remove it. I've filed #2147 to simplify this test impl.

shobhitagarwal1612 (approved):

PR #2157, opened 2024-01-05
Refactor map feature managers

### LINKED ISSUE #2155: [Code health] Clean up feature manager implementations
- [x] Turn `FeatureManager` in an interface and implement in `FeatureClusterRenderer`
- [x] Remove `removeAll*` cases in `GoogleMapsFragment`
- [x] `isSelected` state can be moved into `Style`
- [x] The name `render()` is misleading; those methods add features to be managed and rendered by Maps SDK

gino-m:
Closes #2155

@scolsen @shobhitagarwal1612 PTAL?

@justinkimtang FYI


ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -138,6 +140,11 @@ internal constructor(
         }
   }
 
+  private fun applySelectedState(features: Set<Feature>, selectedLoiId: String?): Set<Feature> =
```
scolsen, line 143: Nit feel free to ignore—this name feels a bit general since we only want this for LOIs, but we don't encode that fact in the signature (since we take a string). Maybe a more specific name would help readability ever so slightly? Or are we intending this to be more general in the future (selections working for other features)?

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Updates the current set of features managed by the manager, adding and removing items from the
+   * map as needed to sync the map state with the provided collection.
+   */
+  fun setFeatures(updatedFeatures: Collection<Feature>) {
+    // remove stale
+    val removedOrChanged = features - updatedFeatures.toSet()
+    removedOrChanged.forEach(this::remove)
```
scolsen, line 77: Nit use trailing lambda or method reference consistently (we use a trailing lambda below).

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
```diff
@@ -0,0 +1,117 @@
...
+  fun getIntersectingPolygons(latLng: LatLng): Set<Feature> =
+    mapsItemManager.getIntersectingPolygonTags(latLng).mapNotNull { featuresByTag[it] }.toSet()
+
+  /**
+   * Adds a feature to the map, cluster, and to this class' internal index. Clusterable features are
+   * initialized as hidden so that the clusterer can determine whether they should be shown based on
+   * zoom level.
+   */
+  private fun add(feature: Feature) =
+    with(feature) {
```
scolsen, line 99: I don't think the `with` adds much value here, and it makes me have to pause and think ever so slightly more as a reader. this is exactly equivalent to just:

```kotlin
private fun add(feature: Feature) {
  features.add(feature)
  featuresByTag[tag] = feature
  if (feature.clusterable) clusterManager.addFeature(feature)
  mapsItemManager.put(feature, visible != feature.clusterable)
}
```

Which is slightly more verbose, but I like that I don't have to think about why we're using `with`. Unless I'm missing something here, I'm not sure the `this` binding value is totally there. In fact, I actually find it require a little bit more work to read implicit property access like an unqualified `clusterable`, but that might just be me.

To me, `with` makes most sense when you're doing a bunch of state mutations, i.e. you have lines like:

```kotlin
with(feature) {
  clusterable = false
  // ....
}.foo(//do some stuff with the mutated thing)
```

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.google.maps.android.PolyUtil.*
+
+/** Manages [Feature]s displayed on the map as Maps SDK items (marker, polyline, etc). */
+class MapsItemManager(
+  private val map: GoogleMap,
+  private val pointRenderer: PointRenderer,
+  private val polygonRenderer: PolygonRenderer,
+  private val lineStringRenderer: LineStringRenderer
+) {
+  private val itemsByTag = mutableMapOf<Feature.Tag, List<Any>>()
```
scolsen, line 38: Nit: kind of a shame to use `Any` and abandon all type specificity here imo -- can we target a supertype instead/add one? Feel free to ignore, as this is a bit pedantic (but it might catch a future bug!).

Feels like these types are amenable to a sealed class, which would also give use exhaustive compile-time `when` checking

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Adds one or more items to the map representing the specified [Feature], replacing and existing
+   * items associated with same tag are replaced.
+   */
+  fun put(feature: Feature, visible: Boolean) =
+    with(feature) {
```
scolsen, line 45: Does `with` really add much here?

scolsen (approved):
NIce cleanup!

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -146,46 +128,36 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), MapFragment {
     layoutInflater: LayoutInflater,
     viewGroup: ViewGroup?,
     bundle: Bundle?
-  ): View =
-    super.onCreateView(layoutInflater, viewGroup, bundle).apply {
+  ): View {
+    Timber.v("Lifecyle event: onCreateView()")
```
shobhitagarwal1612, line 132: Please cleanup debug only logs.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -146,46 +128,36 @@ class GoogleMapsFragment : Hilt_GoogleMapsFragment(), MapFragment {
...
     containerFragment.replaceFragment(containerId, this)
     getMapAsync { googleMap: GoogleMap ->
       onMapReady(googleMap)
-      onMapReadyCallback.accept(this)
+      onMapReadyCallback(this)
     }
   }
 
   private fun onMapReady(map: GoogleMap) {
+    Timber.v("Map event: onMapReady()")
```
shobhitagarwal1612, line 153: Same as above.

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
gino-m: Unfortunately Maps doesn't offer a common supertype. I tried several variations of generic and refactorings, but because of the lack of a common supertype they all involved checking types at runtime someone. This was the simplest impl I could find.

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: All lifecycle events are logged across the app. Ok to keep here in that case?

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
scolsen: Dang! lgtm then, thanks for looking into it!

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
gino-m: Good points. How about this:

* `applySelectedState` -> `updatedLoiSelectedStates`
* Move `selected` from `Style` to `Feature`
* Replace `withSelected(..)` with `select()` and `deselect()` on `Feature`

gino-m: Note that this function also needs to deselect LOIs which are no longer selected.

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
gino-m: I'm on the fence as to which is more readable here, but I agree they should be consistent. Using method ref for now. Some valuable insight here: https://stackoverflow.com/a/24493905/9986466

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
gino-m: I see two reasons why `with` might make this more readable:

* Everything we're doing here is with a `feature` instance; the `with` makes that context explicit at the top level of the function due to single-expression function syntax.
* We don't need to repeat `feature` multiple times, making the code more concise within that well-defined context.

Note in your above example, we'd need to quality feature one additional time for `tag`:

```kt
  private fun add(feature: Feature) {
    features.add(feature)
    featuresByTag[feature.tag] = feature
    if (feature.clusterable) clusterManager.addFeature(feature)
    mapsItemManager.put(feature, visible = !feature.clusterable)
  }
```

vs now:

```kt
  private fun add(feature: Feature) =
    with(feature) {
      features.add(this)
      featuresByTag[tag] = this
      if (clusterable) clusterManager.addFeature(this)
      mapsItemManager.put(this, visible = !clusterable)
    }
```

Change my mind! :)

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
gino-m: Same comment as above. As with natural language, using a preposition can help clarity. Was `with` actually designed for use only with state mutations, or is that just a bias since it's a common pattern?

ground/src/main/java/com/google/android/ground/ui/map/gms/GoogleMapsFragment.kt
gino-m: See previous comment.

gino-m:
@shobhitagarwal1612 @scolsen Applied a few suggestions, replied to several others. PTAL?

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
scolsen: It's a fair point, I've just only ever seen `with` used mostly for the "configuration pattern" (as we use it in this PR for maps styling configuration in the renderers). 

BUT I don't *really* have a good argument against it in this case, so I think it's fine!

ground/src/main/java/com/google/android/ground/ui/map/gms/features/MapsItemManager.kt
scolsen: Mainly pattern, bias I think. You've convinced me that it's fine here since I can't think of any actual problems it induces in this usage.

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
scolsen: Same here. I don't really care which we use so long as it's (locally) consistent in the body of this function. I have a slight preference for method reference in this case since we're doing nothing other than transparently passing through a unary argument.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
scolsen: Sounds like a great plan to me!

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
gino-m: I was looking around but couldn't find anything either way. The docs just have this to say:
https://kotlinlang.org/docs/scope-functions.html#with

ground/src/main/java/com/google/android/ground/ui/map/gms/features/FeatureManager.kt
gino-m: +1, same preference here.

shobhitagarwal1612 (approved):

PR #2158, opened 2024-01-07
Hide "Undo" button in area task when drawing

### LINKED ISSUE #2102: "Add point" button in "Draw polygon" task appears incorrectly
When the task index is > 0, an additional button "previous" is added to the task for navigating to the previous task. This results in "Add point" button to look incorrectly due to insufficient space. 

bug-app-point

shobhitagarwal1612:
Fixes #2102

Default behavior of "Undo" button is to display it only when the response is empty. In draw area task, we were only persisting the response when the drawing was complete. So, we also had to override the default "undo" behavior which wasn't working as expected. 

Proposed solution in this PR:
1. Save incomplete "linestring" to response map and remove overridden "undo" behavior.
2. Add validation to prevent navigating previous/next when the current response is incomplete

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
@@ -25,7 +25,6 @@ import com.google.android.ground.model.submission.Value
 import com.google.android.ground.model.submission.isNullOrEmpty
 import com.google.android.ground.model.task.Task
 import com.google.android.ground.ui.common.AbstractViewModel
-import com.google.android.ground.ui.datacollection.tasks.polygon.DrawAreaTaskIncompleteResult
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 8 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
```diff
@@ -67,10 +68,19 @@ open class AbstractTaskViewModel internal constructor(private val resources: Res
...
-      Optional.of(resources.getString(R.string.required_task))
-    else Optional.empty()
+  private fun validate(task: Task, value: Value?): Optional<String> {
+    // Empty response for a required task.
+    if (task.isRequired && (value == null || value.isEmpty())) {
+      return Optional.of(resources.getString(R.string.required_task))
+    }
+
+    // Invalid response for draw area task.
+    if (task.type == Task.Type.DRAW_AREA && value is DrawAreaTaskIncompleteResult) {
```
gino-m, line 78: To avoid leaking subclass details into the parent here, can we instead make `validate` protected and override in `DrawAreaTaskViewModel`?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
```diff
@@ -67,10 +68,19 @@ open class AbstractTaskViewModel internal constructor(private val resources: Res
   }
 
   // TODO: Check valid values.
-  private fun validate(task: Task, value: Value?): Optional<String> =
-    if (task.isRequired && (value == null || value.isEmpty()))
-      Optional.of(resources.getString(R.string.required_task))
-    else Optional.empty()
+  private fun validate(task: Task, value: Value?): Optional<String> {
```
gino-m, line 71: Can we replace return value with String? while we're at it?

gino-m (changes requested):
A few small suggestions

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
shobhitagarwal1612: Done

gino-m (approved):

PR #2172, opened 2024-01-10
MediaUploadWorker: Filter on PhotoResponse.

scolsen:
- Make `path` in PhotoResponse public so that we can use it directly when uploading media.
- Filter submission deltas by response value instead of task type.
- Minor stylistic tweaks in trailing lambdas.

@gino-m PTAL

Note that at this point we already show media upload statuses in the sync history fragment (though it may happen fast enough that one really only sees the "COMPLETED" state reported). If we want to report on the status of photos separately from their associated mutations, there's quite a bit more work to do.


gino-m (approved):
Nice cleanup and steps towards media sync status!

gino-m:
@scolsen Gentle ping, this is approved but some checks are failing.

gino-m:
@scolsen Gentle ping. Could you please fix checks so we can merge this?

Also, is this a pure code health fix, or does it change user-visible behavior? When you can please link to an open issue for posterity.

scolsen:

PR #2175, opened 2024-01-11
Don't reset viewport when resuming from data collection flow

### LINKED ISSUE #1892: Reuse viewport across map views
In hindsight, the map bouncing from one viewport to another has a jarring effect and bucks expectation

### LINKED ISSUE #2163: [Map] Viewport not maintained after "drop a pin" task
After a "drop a pin" task (and likely other geometry tasks?) and submitting data, the viewport reverts to its original position. The last viewport in the geometry tasks should be maintained.

@shobhitagarwal1612 PTAL?

### LINKED ISSUE #2173: [Map] Viewport reset to fully zoomed out after data collection flow
After collecting data for an existing LOI, viewport is reset to fully zoomed out. @shobhitagarwal1612 PTAL?

shobhitagarwal1612:
Fixes #2173
Fixes #1892
Fixes #2163

We were only persisting the current camera position to shared prefs only for home screen. This PR extends this for all map based screens (data collection tasks, offline area screen, home screen)

Verified by moving the map within a task and navigating back to the home screen and confirming that the viewport is not reset.

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
@@ -252,6 +252,7 @@ constructor(
     Timber.d("Camera moved : ${newCameraPosition.target}")
     lastCameraPosition = currentCameraPosition.value
     currentCameraPosition.value = newCameraPosition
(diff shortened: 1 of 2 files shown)


gino-m (approved):
Smooth fix! Lgtm!

Issue #2179, opened 2024-01-12
Error when sync: Unknown task id


rfontanarosa:
Sometimes, when "History and sync status" is selected. After that the app crashes.

```
13:03:31.689  E  FATAL EXCEPTION: main
                 Process: com.google.android.ground, PID: 5611
                 java.lang.IllegalStateException: Unknown task id BSfYEdtkHN94U9XcsVY0
                 	at com.google.android.ground.model.job.Job.getTask(Job.kt:36)
                 	at com.google.android.ground.persistence.local.room.converter.SubmissionDeltasConverter.fromString(SubmissionDeltasConverter.kt:67)
                 	at com.google.android.ground.persistence.local.room.converter.ConverterExtKt.toModelObject(ConverterExt.kt:336)
                 	at com.google.android.ground.persistence.local.room.stores.RoomSubmissionStore$getAllSurveyMutationsFlow$$inlined$map$1$2.emit(Emitters.kt:224)
                 	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
                 	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
                 	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:87)
                 	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:37)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
                 	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Unknown Source:14)
                 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
                 	at android.os.Handler.handleCallback(Handler.java:958)
                 	at android.os.Handler.dispatchMessage(Handler.java:99)
                 	at android.os.Looper.loopOnce(Looper.java:205)
                 	at android.os.Looper.loop(Looper.java:294)
[...]

PR #2182, opened 2024-01-13
Add "active survey id" to all the crash logs

shobhitagarwal1612:
This PR adds a custom key-value for the selected survey id to each crash log. 

This would be helpful for debugging https://github.com/google/ground-android/issues/2179 and other issues in general 

We can also log other useful app states (like user id, login state, last UI action, etc.) to add more context to the exceptions.

@gino-m  PTAL?


gino-m (approved):
Great idea! LGTM!

PR #2184, opened 2024-01-13
[WORK IN PROGRESS] Add shared multiplatform lib module

gino-m:
Towards #2049


shared/build.gradle.kts
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+  kotlin("multiplatform")
+  id("com.android.library")
```
jamesward, line 19: You might want to try the `com.android.kotlin.multiplatform.library` plugin as it works better in this use case.  Below the `androidTarget` becomes `androidLibrary` and the config in the top-level `android` moves into the `androidLibrary` config.  Unfortunately not a lot of docs on that approach yet as it is quite new.

gino-m: Thanks @jamesward, sorry just saw this now. I'll apply that chance when we get back to this.

gino-m:

PR #2193, opened 2024-01-19
Add dummy config for local emu

### LINKED ISSUE #2191: [DevX] `localDebug` variant not buildable on fresh git clone
..due to missing `google-services.json`. Ideally this file wouldn't be required to test against local emulator.

gino-m:
Fixes #2191

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 .*
+**/google-services.json
+!ground/src/debug/local/google-services.json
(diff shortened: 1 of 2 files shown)


gino-m:
A google-services.json file is required, even when using local Firebase emulator. I created and deleted the project used to generate this test file.

shobhitagarwal1612 (approved):

Issue #1675, opened 2023-05-10
[Code health] Coverage for classes annotated with @AndroidEntryPoint is 0%


shobhitagarwal1612:
Context: 
* https://github.com/google/dagger/issues/1982
* https://b.corp.google.com/issues/161300933?pli=1

shobhitagarwal1612:
Workaround : https://github.com/google/dagger/issues/1982#issuecomment-710964631

PR #2197, opened 2024-01-20
Remove reference to hilt generated classes

shobhitagarwal1612:
This was added to fix jacoco test coverage issue where all classes with `@AndroidEntryPoint` annotation were reporting 0% coverage. It appears that the underlying issue has been resolved.

Context: https://github.com/google/ground-android/issues/1675

@gino-m  PTAL?


gino-m (approved):
Brilliant! Glad we don't need the old extends syntax anymore.

PR #2203, opened 2024-01-22
Add ID formatting to job cards to parse out insignificant decimal bits

### LINKED ISSUE #2137: [Job cards] Additional ".0" added to numeric LOI ID s
IDs are stored as numeric LOI properties in `terrabio` test project.
image

justinkimtang:
Fixes #2137  

This adds formatting to job cards so that IDs have insignificant decimals removed.

1) Upload a LOI with an id that can be represented as a float or a double
2) Go to the survey in the app with the given jobId

Before:
Screenshot 2024-01-22 at 2 23 35 PM
After:
Screenshot 2024-01-22 at 2 22 51 PM

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
@@ -23,15 +23,17 @@ import com.google.android.ground.model.geometry.Point
 import com.google.android.ground.model.geometry.Polygon
 import com.google.android.ground.model.locationofinterest.LocationOfInterest
 import com.google.android.ground.util.isNotNullOrEmpty
+import kotlin.math.floor
 
 /** Helper class for creating user-visible text. */
 object LoiCardUtil {
 
   fun getDisplayLoiName(context: Context, loi: LocationOfInterest): String {
     val loiId =
-      if (loi.customId.isNotNullOrEmpty()) loi.customId else loi.properties?.get("id")?.toString()
+      if (loi.customId.isNotNullOrEmpty()) loi.customId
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612 (commented):
Can we please add some unit tests as well?

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
```diff
@@ -43,6 +45,23 @@ object LoiCardUtil {
...
+  /**
+   * Converts float/double IDs into ints if the decimal portion is 0; ignores all other id types.
+   * * ex. "1234.00" -> "1234",
+   * * "1234.01" -> "1234.01",
+   * * "abcdefg" -> "abcdefg",
+   */
+  private fun String.formatAsId(): String {
+    return when {
+      toDoubleOrNull() != null && toDouble() - floor(toDouble()) == 0.0 -> {
+        String.format("%.0f", toDouble())
```
gino-m, line 57: I don't think we need this; IDs will rarely, if ever, have decimals, and will rarely be numeric. I think we carried forward numeric IDs from GeoJSON in case users want them to also be numbers in the exported CSV. Please feel free to just convert to a string if the id is a number. Thanks!

gino-m (changes requested):
Thanks for the fix!

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/cards/LoiCardUtil.kt
justinkimtang: Going to drop this PR in that case! It sounds like the original issue was a one off with unexpected inputs.

PR #2204, opened 2024-01-22
Change the color of loi card items to white except for in the add loi…

### LINKED ISSUE #2169: [Job cards] The color of cards doesn't match designs
Non "add LOI" cards should be white. We should also confirm the color of "add LOI card".

image

justinkimtang:
… case

Fixes #2169

Updates the color of loi_card_item to have a default background color of white. This does not update add_loi_card_item.xml which maintains its greenish tint.

Open a survey and verify that the color of the cards are as expected.

Before:
Screenshot 2024-01-22 at 6 07 35 PM
After:
Screenshot 2024-01-22 at 6 08 06 PM

@gino-m PTAL?


gino-m (approved):
LGTM, thanks! 🎨

PR #2205, opened 2024-01-23
[DRAFT WORK IN PROGRESS] Prototype of JSON schema codegen

gino-m:


gino-m:

FYI @jcqli Just closed two other similar draft PRs. We can restore this if/when we get back to them.

PR #2208, opened 2024-01-24
Fix regression when attempting to update camera position on survey activation

### LINKED ISSUE #2198: Map doesn't point to default bounds after activating survey
When activating the survey for the first time, the map should move to the default bounds which covers all LOIs. Instead it remains at 0,0 position.

shobhitagarwal1612:
Fixes #2198

Verified the change by opening the app locally and activating a survey.

@gino-m  PTAL?


gino-m (approved):
🏆

PR #2214, opened 2024-02-01
Job: Handle unknown tasks

scolsen:
This commit adds a specific exception type for the case in which a caller attempts to retrieve a task from a Job that does not contain a task with the designated ID.

Updates callsites to properly handle and log this exception (in all cases we just ignore the data).

Fixes #2179

PR #2218, opened 2024-02-02
Job: Return null on unknown tasks

### LINKED ISSUE #2216: [Code health] Revisit use of Exceptions and Nullable Types
There are a few places in the code base where we currently throw exceptions that we might be able to convert to nullable return values or Kotlin's `Result` type instead, for example, [in the SubmissionConverter](https://github.com/google/ground-android/blob/6ea86989fd8d9e2e366412a2c139b55306679de0/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/SubmissionConverter.kt#L87).

We should do a pass to see if there are any cases in which we're overusing exceptions, or not using exceptions for cases that really are exceptional.

See https://github.com/google/ground-android/pull/2214#discussion_r1475224703 for additional context.

scolsen:
Return null when attempting to retrieve unknown tasks from a Job (instead of throwing an exception). Updates callers accordingly.

NOTE: Depends on https://github.com/google/ground-android/pull/2214

Fixes #2216

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/job/Job.kt
+++ b/ground/src/main/java/com/google/android/ground/model/job/Job.kt
@@ -33,17 +33,13 @@ data class Job(
     UNKNOWN
   }
 
-  class TaskNotFoundException(taskId: String) : Throwable(message = "unknown task $taskId")
-
   val canDataCollectorsAddLois: Boolean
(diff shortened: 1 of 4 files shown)


scolsen:

PR #2220, opened 2024-02-03
Reduce "Collect data" tap target to button

### LINKED ISSUE #2112: Reduce "Collect data" tap target to button
It's currently too easy to accidentally enter the data collection flow. We should only start collecting data when the button is pressed.

shobhitagarwal1612:
Fixes #2112

@gino-m  PTAL?


gino-m (approved):
🔘 👍

Issue #2235, opened 2024-02-13
[Code health] Fixes and cleanup of `LocalMutationSyncWorker`


gino-m:
- [ ] Warn and ignore mutations not belonging to current signed in user
- [ ] Move `getPendingOrEligibleFailedMutations()` into repository.
- [ ] Extract business logic into `UserDataUploadUseCase`

scolsen:
https://github.com/google/ground-android/pull/2237 will take care of:

>  Warn and ignore mutations not belonging to current signed in user
>  Move getPendingOrEligibleFailedMutations() into repository

PR #2237, opened 2024-02-14
Mutations Sync: Refactor and filter on current user

scolsen:
Two small bits of cleanup in the LocalMutationSyncWorker:

  1. Move fetching of mutations eligible for retry into the mutations repository
  2. Only sync mutations associated with the current authenticated user.

Also adds a new test to validate (2.). Progress toward issue #2235.

@gino-m


ground/src/main/java/com/google/android/ground/persistence/sync/MediaUploadWorker.kt
```diff
@@ -65,9 +67,11 @@ constructor(
     check(loiId.isNotEmpty()) { "work was queued for an empty location of interest ID" }
     Timber.d("Starting media upload for LOI: $loiId")
 
+    val user = userRepository.getAuthenticatedUser()
     val mutations =
       mutationRepository.getSubmissionMutations(
```
gino-m, line 72: We don't need to filter by user since only mutations for the current user should be in the db. Instead, we can continue to fetch all and `Timber.w` and ignore those that don't belong to the current user.

scolsen: gotcha. since we're still processing everything in a list at this point, and not one by one, unless I'm missing something, I think the only way to do this is with a filter, it's just a matter of where we want to apply said filter.

I've moved it into the worker instead.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -59,61 +58,33 @@ constructor(
...
-  }
-
   /**
    * Applies mutations to remote data store. Once successful, removes them from the local db.
    *
    * @return `true` if the mutations were successfully synced with [RemoteDataStore].
    */
   private suspend fun processMutations(mutations: List<Mutation>, user: User): Boolean {
-    check(mutations.isNotEmpty()) { "List of mutations is empty" }
+    // TODO(#2235): Process and update each mutation individually.
```
gino-m, line 86: I think we realized yesterday we may need to process them together so we can build a batch. Ok to remove this comment.

gino-m (approved):
Nice work! One nit, and some checks are failing, otherwise LGTM!

gino-m:
Looks like there are detekt errors to correct.

gino-m:
@scolsen Gentle ping.. could you please resolve failing checks? @jcqli for visibility.

scolsen:
@gino-m FYI should be good to go, holding off on merging just to be sure we manage these correctly before cutting a release

scolsen:

Issue #876, opened 2021-06-08
[Sync status] Improve contents and layout of sync status


gino-m:
@coreyleamon I hacked together a sync status screen, but it's just the barebones of what's needed to check the sync status for features and observations. Would you like to have a go at a proper design for this interface? The known follow-ups at the moment are:

- [x] Show icon for each status (PENDING, IN_PROGRESS, COMPLETE, FAILED)
- [x] Show related feature caption/name on each line if available (@scolsen)
- [x] Add date/time completed/failed
- [ ] Add number of retries
- [ ] i18n: Make text translatable
- [ ] Create custom layout instead of using multiline string

Screenshot of current UI:

![Screenshot_1623177689](https://user-images.githubusercontent.com/228050/121240160-091d4800-c868-11eb-9cb4-f5799fedf1ce.png)

@kdyson FYI

coreyleamon:
Is this a status screen that conveys the status of uploading data once connected to a network? If so, I have a few questions:

1. Is it important for these data lines to be viewed together in a screen, which disassociates them from the geographical representation of the point or polygon?
2. Does the user need to manually re-initiate when there is a failure, or could the system make x attempts before being considered "failed"?
3. If a data point continues failing to upload, what action can a user take? Will they need to re-submit/recreate the data point?
4. Do uploaded data points need to be archived in any way? How long and in what contexts might the user need to maintain visibility of the status even if successful?

gino-m:
Hi @coreyleamon sorry I missed these:
 
> Is this a status screen that conveys the status of uploading data once connected to a network? If so, I have a few questions:

Yes, that's correct.

> 1. Is it important for these data lines to be viewed together in a screen, which disassociates them from the geographical representation of the point or polygon?

I think both may be useful. This list is useful because data collectors want to know which changes were recently made without needing to search for possibly modified features on the map.

> 2. Does the user need to manually re-initiate when there is a failure, or could the system make x attempts before being considered "failed"?

It already does this, we just don't show it yet in the UI. But after X attempts we don't have a well-defined story for how the user either retries or abandons the change.

> 3. If a data point continues failing to upload, what action can a user take? Will they need to re-submit/recreate the data point?

Currently none; see previous point.

> 4. Do uploaded data points need to be archived in any way? How long and in what contexts might the user need to maintain visibility of the status even if successful?

Good q - one option would be a CTA to clear completed items from the list. Wdyt?

gino-m:
@ngosue FYI

[...]

PR #2242, opened 2024-02-15
Sync Status: Update UI

scolsen:
Modify the sync status/history UI to better reflect the latests designs. Also refactors our approach to passing data over to the list adapter slightly, and adds some new drawable resources.

Screenshot 2024-02-15 at 3 54 37 PM

@gino-m FYI - but I think we still need a few tweaks on the UI itself?

### DIFF
--- a/ground/src/main/res/values/strings.xml
+++ b/ground/src/main/res/values/strings.xml
@@ -69,7 +69,7 @@
   <string name="date">Date</string>
   <string name="time">Time</string>
   <!-- Other strings (to be organized) -->
-  <string name="sync_status">History and sync status</string>
+  <string name="sync_status">History</string>
   <string name="layer_label_format">Job: %s</string>
 
   <string name="add_point">Add point</string>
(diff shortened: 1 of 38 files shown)


scolsen:
See #876

ground/src/main/res/values/strings.xml
```diff
@@ -69,7 +69,7 @@
   <string name="date">Date</string>
   <string name="time">Time</string>
   <!-- Other strings (to be organized) -->
-  <string name="sync_status">History and sync status</string>
+  <string name="sync_status">History</string>
```
gino-m, line 72: The title in the nav menu should still be "History and sync status". @vittorino  Was the different labels between that and the status screen header intentional?

gino-m (approved):
Nice! I assume the correct colors and layout will come later, so LGTM for this change, modulo @vittorino 's guidance on the screen title and failing checks.

vittorino:
I think we should keep the entire "History and sync status" if that's what's in the nav bar. In very small screens the header may break, so I'd also recommend we shorten this title if we can. Maybe just "history and sync"?

gino-m:
> I think we should keep the entire "History and sync status" if that's what's in the nav bar. In very small screens the header may break, so I'd also recommend we shorten this title if we can. Maybe just "history and sync"?

Good point. @amegantz Can we just say "Sync status" in the screen title and side menu? History is implied.

PR #2271, opened 2024-02-28
Remove dependency on survey.tileSources, instead use default Firebase storage imagery path

### LINKED ISSUE #2115: Access to offline imagery doesn't authenticate using user's credentials
This requires the bucket to allow public access. To prevent misuse, we should authenticate using the current user's credentials. Proposed design with @sufyanAbbasi: 

- [ ] Move data from GCS into Firebase Storage
- [ ] Replace custom `tileSources` config in survey document with "configuration by convention": 
  - [ ] Look for tiles under `/offline-imagery/default/{z}/{x}/{y}.tif`, where `{z}` is the zoom level at which hi-res tilesets are sliced. The hi-res tilesets have tiles starting at zoom level _z_
  - [ ] Assume the overview image will be in `/offline-imagery/default/{z}/overview.tif`. The overview will have zoom levels  0.._z_-1. This will replace the current hard-coded `world-masked.tif`.
  - [ ] Read zoom level extents of `overviewtiff` and hi-res imagery directly from GeoTIFFs
- [ ] Update `MogClient` to get `InputStream` from Firebase SDK instead of `UrlInputStream`. This should handle auth for us.
- [ ] Update Firebase Storage Rules to only allow access to `/offline-imagery` if user is in passlist (https://github.com/google/ground-platform/issues/1373)

sufyanAbbasi:
Fixes #2115 

Removes dependencies on `survey.tileSources` to show the "Offline Imagery" button in the nav bar, and will always attempt to find imagery in the default path, `/offline-imagery/default`.

- [x] Removes invocations of `survey.tileSources` field, uses a default TileSource instead.
- [ ] Does not remove tileSources from the local storage DAO.

- Removed `tilesources` from the survey, was able to successfully download data
- Changed the default path to `/offline-imagery/default_NOT_HERE` to simulate a missing repo, shows "no imagery found" message after failing gracefully.

Screenshot 2024-02-28 at 12 31 30 AM

PTAL @gino-m


shobhitagarwal1612 (approved):

gino-m:
Hey @sufyanAbbasi thanks for the fix! Looks like the following tests will also need to updated/deleted:

```
com.google.android.ground.ui.home.HomeScreenFragmentTest > offlineMapImageryMenuIsDisabledWhenActiveSurveyHasNoBasemap FAILED
    androidx.test.espresso.base.AssertionErrorHandler$AssertionFailedWithCauseError at HomeScreenFragmentTest.kt:144
        Caused by: junit.framework.AssertionFailedError at HomeScreenFragmentTest.kt:144

com.google.android.ground.ui.home.NavigationDrawerItemClickTest > clickDrawerMenuItem[Clicking 'offline map imagery' when survey doesn't have offline tiles should do nothing] FAILED
    app.cash.turbine.TurbineAssertionError at HomeScreenFragmentTest.kt:178
```

ground/src/main/java/com/google/android/ground/model/Survey.kt
```diff
@@ -24,6 +24,7 @@ data class Survey(
   val title: String,
   val description: String,
   val jobMap: Map<String, Job>,
+  // TODO(#1730): Allow tile source configuration from a non-survey accessible source.
```
gino-m, line 27: It's not immediately clear what we mean by "a non-survey accessible source". Perhaps the TODO here is to clean up / remove `tileSources` from the Survey altogether?

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -34,14 +34,15 @@ class HomeScreenViewModel
...
   private val navigator: Navigator,
-  private val surveyRepository: SurveyRepository,
+  surveyRepository: SurveyRepository,
 ) : AbstractViewModel() {
 
   private val _openDrawerRequests: MutableSharedFlow<Unit> = MutableSharedFlow()
   val openDrawerRequestsFlow: SharedFlow<Unit> = _openDrawerRequests.asSharedFlow()
 
+  // TODO(#1730): Allow tile source configuration from a non-survey accessible source.
   val showOfflineAreaMenuItem: LiveData<Boolean> =
```
gino-m, line 44: I think you can just use `= MutableLiveData(true)` here instead of the flow at this point.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/model/Survey.kt
sufyanAbbasi: Updated TODO.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
sufyanAbbasi: Thank you! Done.

ground/src/main/java/com/google/android/ground/Config.kt
```diff
@@ -55,10 +55,11 @@ object Config {
   const val MAX_MEDIA_UPLOAD_RETRY_COUNT = 5
 
   // TODO(#1730): Make sub-paths configurable and stop hardcoding here.
+  const val DEFAULT_MOG_TILE_URL = "/offline-imagery/default/"
```
gino-m, line 58: Iirc we may be adding the trailing `/` twice - let's remove it here? Also, not this is no longer a URL - Can we rename to `DEFAULT_MOG_TILE_LOCATION`?

gino-m (approved):
Lgtm! Added one suggestion, feel free to apply now to ignore and merge.

gino-m (commented):
> [ktfmt] Found 1 files that are not properly formatted:
  src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt

gino-m (approved):
LGTM! 🥳

gino-m:
@sufyanAbbasi 

> Execution failed for task ':workspace:ground:ktfmtCheckMain'.
> [ktfmt] Found 2 files that are not properly formatted:
  src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
  src/main/java/com/google/android/ground/repository/OfflineAreaRepository.kt

gino-m:
ktfmt still failing, running now

PR #2280, opened 2024-03-01
Add a small delay after the page scroll is idle to calculate the progress bar position.

### LINKED ISSUE #1975: Data collection progress indicator appears over tasks
Green horizontal line appears across screen.

Trying on Pixel 4 API 32, photo was not first task in job.

image

sufyanAbbasi:
Towards #1975 

Waits a little bit after the task view has changed in order to recalculate the progress bar position, which is based on the top of the action buttons container. This is is in order to allow the action buttons to settle into their position after the keyboard has moved out of the way, since the callbacks are called too quickly.

Also replaces deprecated Java `systemWindowInsetTop` property (sorry, I used Bard/Gemini for that).

- [x] Waits a ~100ms delay before attempting to set the progress bar position.
- [x] Replaces deprecated  `systemWindowInsetTop` with `WindowInsetsCompat` API.

- Clicked on a text task, wait for the keyboard to show up, then moved to the next and previous tasks while the action buttons are pushed up.

Screenshot 2024-02-29 at 7 12 06 PM

@gino-m  PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
@@ -17,11 +17,14 @@ package com.google.android.ground.ui.datacollection
 
 import android.animation.ValueAnimator
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ProgressBar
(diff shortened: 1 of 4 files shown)


shobhitagarwal1612 (approved):

gino-m:
Thanks @sufyanAbbasi ! Note that the progress bar still isn't visible when the keyboard is active. Also, ideally we wouldn't need the delay. I've experimented with embedding the progress bar in the task view in https://github.com/google/ground-android/pull/2287; the layout works well, but the progress state isn't set correcting on each page change.

Issue #2286, opened 2024-03-01
Data collection snackbar hint doesn't go away when zooming in


gino-m:
@sufyanAbbasi's repro steps:
1. Zoom in enough to be able to start collecting points.
1. Switch to a different menu item (e.g. sync menu).
1. Switch back to map view, see that the "Zoom in" message is triggered.

![Screenshot 2024-03-12 at 6 53 01 PM](https://github.com/google/ground-android/assets/12646706/f8916ff7-2640-4ad1-afda-f5e7ed723694)
![Screenshot_20240301_154314](https://github.com/google/ground-android/assets/228050/d36a7f23-fd72-4b7f-a1e6-6c9c35d24210)

gino-m:
@scolsen PTAL?

sufyanAbbasi:
I'll take a stab at this one, but @scolsen, feel free to snipe it if you can get it done faster.

Issue #2285, opened 2024-03-01
Data collection hint snackbar obscures location lock button


gino-m:
There should be a way to dismiss the snackbar, and the location lock should be repositioned so that it isn't obscured.

Expected:

![image](https://github.com/google/ground-android/assets/228050/851769e9-57a9-485a-9df9-964b5960399d)

Actual:

![Screenshot_20240301_154235](https://github.com/google/ground-android/assets/228050/223634f5-a9fc-4083-8be9-e6b1b0475b17)

gino-m:
@scolsen PTAL?

jcqli:
@scolsen Could you take a look at this?

jcqli:
Closing since https://github.com/google/ground-android/pull/2304 is merged

PR #2288, opened 2024-03-01
Workaround undismissable snackbar data collection hint by setting to short duration

gino-m:
Towards #2286 

@scolsen FYI


scolsen:
the snackbar should be dismissible by swiping on it--at least this works on a Nexus 5X. Maybe we should debug further? What device did dismisal fail on?

gino-m:
> the snackbar should be dismissible by swiping on it--at least this works on a Nexus 5X. Maybe we should debug further? What device did dismisal fail on?

Hi @scolsen,. I didn't realize it was swipeable! Don't recall whether I tested on physical Pixel 7 or emu Nexus 5 API 24. Do you have cycles to look into a few small improvements in #2297, #2286, and #2285?

PR #2293, opened 2024-03-02
Generate code coverage reports only on merge

### LINKED ISSUE #2289: [Code health] Generate code coverage reports only on merge
Doing this for every check adds a lot of wait time, slowing down velocity. @shobhitagarwal1612 Would it be possible to trigger this step only on merge with `master`?

shobhitagarwal1612:
Fixes #2289

@gino-m  PTAL?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -220,10 +220,11 @@ steps:
     args:
       - '-c'
       - |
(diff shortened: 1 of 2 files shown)


gino-m (approved):
Awesome, thanks!

PR #2296, opened 2024-03-03
Reuse the logic for displaying the label for an LOI

### LINKED ISSUE #2046: Incorrect LocationOfInterest label in DataCollectionFragment
LocationOfInterest label in DataCollectionFragment should use the format described in:
https://github.com/google/ground-platform/issues/1242

### LINKED ISSUE #2291: [Data collection flow] Header doesn't show full label of LOI, only ID
For example, an LOI with only ID 5 shows as "5" rather than "Point 5" as on the card.

@shobhitagarwal1612 PTAL?

shobhitagarwal1612:
Fixes #2046
Fixes #2291

Reuse the logic for displaying label for the LOI everywhere.

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/common/LocationOfInterestHelper.kt
```diff
@@ -17,46 +17,58 @@ package com.google.android.ground.ui.common
...
-    return caption.ifEmpty { getLocationOfInterestType(loi) }
+    return if (loiName.isNotEmpty() && loiId.isNotEmpty()) {
+      "$loiName ($loiId)"
+    } else if (loiName.isNotEmpty()) {
+      loiName
+    } else if (loiId.isNotEmpty()) {
+      "${loi.geometry.toType()} ($loiId)"
+    } else {
+      loi.geometry.toDefaultName()
+    }
```
anandwana001, line 42: how about going with `when`, the nesting of if else will keep growing in the future.

WDYT?

ground/src/main/java/com/google/android/ground/ui/common/LocationOfInterestHelper.kt
```diff
@@ -17,46 +17,58 @@ package com.google.android.ground.ui.common
...
-      is LineString -> "LineString"
-      is LinearRing -> "LinearRing"
-      is MultiPolygon -> "MultiPolygon"
+  fun getJobName(loi: LocationOfInterest): String? = loi.job.name
+
+  fun getSubmissionsText(count: Int): String =
+    when (count) {
+      0 -> "No submissions"
+      1 -> "$count submission"
+      else -> "$count submissions"
```
anandwana001, line 51: how about bringing these hardcoded values from strings.xml, this way it will be more easy to modify in future.

shobhitagarwal1612: Thanks for the suggestion! Done

ground/src/main/java/com/google/android/ground/ui/common/LocationOfInterestHelper.kt
shobhitagarwal1612: I'm not sure how to achieve that since `when` expects a statement or an expression.

anandwana001:
```
when {
        loiName.isNotEmpty() -> ...
        else -> ...
    }
```

This way we can use when, it's doable.

shobhitagarwal1612: Thanks! Done

anandwana001 (approved):

gino-m (approved):
Thanks for the fix!

PR #2304, opened 2024-03-06
Home: Adjust location lock position for snackbar.

scolsen:
Updates our snackbar popup to support two enhancements:

1. The location lock button moves dynamically so that it is not obscured by the snackbar
2. The snackbar has a dismiss action which, on tap, dismisses the snackbar.

Towards #2285

Screenshot 2024-03-06 at 12 49 52 PM

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/EphemeralPopups.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/EphemeralPopups.kt
@@ -64,7 +64,7 @@ class EphemeralPopups @Inject constructor(private val context: Application) {
   inner class InfoPopup(view: View, @StringRes messageId: Int, duration: PopupDuration) {
     private var snackbar: Snackbar =
       Snackbar.make(view, context.resources.getString(messageId), durationToSnackDuration(duration))
-        .setAction("Dismiss") { this.dismiss() }
+        .setAction(context.getString(R.string.dismiss_info_popup)) { this.dismiss() }
     val isShown: Boolean
       get() = this.snackbar.isShown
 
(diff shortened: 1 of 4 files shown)


scolsen:
@gino-m I realize the mocks have an "X" button instead of "Dismiss" text. Action text is supported out of the box for snackbars, but to use a drawable we'll prob have to create our own custom snackbar view, which I don't have time for this week. don't know if you want to move ahead with this as a first impl and switch to the button later, or do something else

gino-m:
> @gino-m I realize the mocks have an "X" button instead of "Dismiss" text. Action text is supported out of the box for snackbars, but to use a drawable we'll prob have to create our own custom snackbar view, which I don't have time for this week. don't know if you want to move ahead with this as a first impl and switch to the button later, or do something else

This is great, thanks Scott!

ground/src/main/java/com/google/android/ground/ui/common/EphemeralPopups.kt
```diff
@@ -61,24 +61,29 @@ class EphemeralPopups @Inject constructor(private val context: Application) {
...
-      @StringRes messageId: Int,
-      duration: PopupDuration = PopupDuration.INDEFINITE,
-    ) {
-      val msg = context.resources.getString(messageId)
-      showSnackbar(view, msg, duration)
-    }
+  inner class InfoPopup(view: View, @StringRes messageId: Int, duration: PopupDuration) {
+    private var snackbar: Snackbar =
+      Snackbar.make(view, context.resources.getString(messageId), durationToSnackDuration(duration))
+        .setAction("Dismiss") { this.dismiss() }
```
gino-m, line 67: Please move to strings.xml

gino-m (changes requested):

scolsen:
@gino-m done and done!

gino-m (approved):
Nice!

Issue #708, opened 2021-01-18
[Data collection] Auto-save draft while collecting data


gino-m:
Store locally to ensure no data is lost even if app is restarted. Restore previous state when app is reloaded.

Remaining tasks

- [ ] Unit tests
- [ ] Load draw area task from draft

scolsen:
@JSunde do we support this yet in the data collection flow? Feel free to reassign to me if needed!

JSunde:
No we don't support this yet, Gino and I have discussed it a little offline and were thinking that we should add a DraftRepository which stores the draft state of the current Job.

I'm not sure I have bandwidth for this at the moment, so reassigning back to you for now, but we can revisit in a few weeks and see if I have bandwidth then!

JSunde:
Oh I just realized I think this is at least partially blocking #1542 since addressing this will allow the Data Collection flow to load in existing data that corresponds to the current job. Then when we add the edit flow I think we could potentially load the existing submission as a draft.

Raising the priority on this to p1.

gino-m:
This is tied up with other TBD features: How do you get back to an unsaved draft? Submissions list? Job card? Sync status/history screen? What do those look like and how do they interact with this feature? @vittorino for discussion.

gino-m:
> Still need some product input.
> 
> At minimum, if app closes (e.g. in background) while collecting data we should resume the data collection flow from before the app closed

@shobhitagarwal1612 and I just chatted briefly. If the app quits for any reason while the user is in the data collection flow, the flow should resume with the data in tact. We can't tell whether the user or OS killed the app, so we will do the same for both.

shobhitagarwal1612:
I have a basic prototype of this working. Now, I need to handle the restoring state in the UI for each task type individually.

Prototype:
[auto-save-draft.webm](https://github.com/google/ground-android/assets/8918023/121cdde4-e20e-45bc-9f5e-ca5727e83442)

shobhitagarwal1612:
One task is remaining, will keep this bug open until everything is done.

shobhitagarwal1612:
https://github.com/google/ground-android/issues/2373

PR #2305, opened 2024-03-06
Add unit tests for saving and loading drafts in data collection fragment

shobhitagarwal1612:
Towards #708

Adds unit tests for following behaviors:
 * Draft is saved when clicking next or previous button
 * Tasks are pre-loaded when data collection fragment is opened with draft submission

@gino-m  PTAL?


gino-m (approved):
🧪 🧪 !

PR #2306, opened 2024-03-06
Make DataCollectionFragment tests more idiomatic 

shobhitagarwal1612:
Related to #708

This is an exploratory PR aiming to make the data collection tests easier to write and understand. I felt that the tests in DataCollectionFragment are harder to read as there is a lot of syntax involved for interacting with the UI or validating state of the UI. To resolve it, I created a wrapper class which has a bunch of helper methods that can be called sequentially without having to worry about waiting for any async task to be finished. 

Please feel free to provide suggestions / feedback.

@gino-m @scolsen @sufyanAbbasi  PTAL?


shobhitagarwal1612:
@gino-m Gentle ping!

gino-m (approved):
MUCH better, thank you @shobhitagarwal1612 !

PR #2308, opened 2024-03-07
Fix broken test due to commit 1ced50ec1354fb32babb55cce8cc0c1efa063c94

sufyanAbbasi:
- `Sync Status` --> `Sync status`: https://github.com/google/ground-android/commit/1ced50ec1354fb32babb55cce8cc0c1efa063c94

@gino-m PTAL?

### DIFF
--- a/ground/src/test/java/com/google/android/ground/ui/home/HomeScreenFragmentTest.kt
+++ b/ground/src/test/java/com/google/android/ground/ui/home/HomeScreenFragmentTest.kt
@@ -194,7 +194,7 @@ class NavigationDrawerItemClickTest(
           "Clicking 'change survey' should navigate to fragment"
         ),
         arrayOf(
-          "Sync Status",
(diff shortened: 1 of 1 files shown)


gino-m:
Shame on me for pushing this seemingly harmless change without review.. and thank you @sufyanAbbasi for fixing my mess!  😳

gino-m (approved):

sufyanAbbasi:
No problem! I think the test is a bit brittle, I'd recommend we generalize those strings to prevent this from occurring if possible (or if we try to unit tests in a non-English language!).

gino-m:
> No problem! I think the test is a bit brittle, I'd recommend we generalize those strings to prevent this from occurring if possible (or if we try to unit tests in a non-English language!).

I wish! I've tried, but I wasn't able to get match by ID working; there's no idiom in the testing framework for it! :\

PR #2321, opened 2024-03-09
Add a dialog to input LOI name after LOI step.

### LINKED ISSUE #2267: [Add LOI task] Allow user to name LOIs when adding
image

@rawbzz FYI

sufyanAbbasi:
Fixes #2267

Uses `Composables` to trigger a dialog after the user clicks "Next" on an LOI task that will prompt them to enter a name to save for the LOI. Handles draft submissions and save states.

- [x] Extends a stylized `AlertDialog` to display a text field for collecting the LOI name.
- [x] Triggers the dialog after the user hits "Next" after the LOI selection.
- [x] Saves the LOI name in the LOI/Mutation `name` in `properties`.
- [x] Updates the `properties` field on remote storage.
- [x] Handles save state.
- [x] Handles pulling from drafts.
- [ ] ~Spams `Invalid resource ID 0x00000005.` error in the logs.~ <-- It's happening on the main branch, worth investigating.

- [x] Created an ad-hoc survey, added new LOI with names. Verified that it saved to remote storage.
- [x] Verified that it shows the name on the submission card.
- [x] Tested that load from draft mode preserves the LOI name.

Screenshot 2024-03-08 at 10 18 44 PM
Screenshot 2024-03-08 at 8 21 14 PM
Screenshot 2024-03-08 at 10 14 49 PM

PTAL @gino-m and @shobhitagarwal1612 !


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -64,6 +64,7 @@ internal constructor(
       navigator.navigate(
         HomeScreenFragmentDirections.actionHomeScreenFragmentToDataCollectionFragment(
           draft.loiId,
+          draft.loiName ?: "",
```
shobhitagarwal1612, line 67: For consistency with other args, we should use `null` as the default value.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -214,6 +214,7 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
         navigator.navigate(
           HomeScreenFragmentDirections.actionHomeScreenFragmentToDataCollectionFragment(
             cardUiData.loi.id,
+            cardUiData.loi.properties["name"] as? String ?: "",
```
shobhitagarwal1612, line 217: Same as above

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -223,6 +224,7 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
         navigator.navigate(
           HomeScreenFragmentDirections.actionHomeScreenFragmentToDataCollectionFragment(
             null,
+            "",
```
shobhitagarwal1612, line 227: Same as above

ground/src/main/res/navigation/nav_graph.xml
```diff
@@ -86,6 +86,10 @@
       android:name="locationOfInterestId"
       type="string"
       app:nullable="true" />
+      <argument
+      android:name="locationOfInterestName"
+      type="string"
+      app:nullable="false" />
```
shobhitagarwal1612, line 92: Same as above. This can be nullable.

ground/src/main/res/navigation/nav_graph.xml
```diff
@@ -106,6 +110,10 @@
         android:name="locationOfInterestId"
         type="string"
         app:nullable="true" />
+      <argument
+        android:name="locationOfInterestName"
+        type="string"
+        app:nullable="false" />
```
shobhitagarwal1612, line 116: Same as above.

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/repository/LocationOfInterestRepository.kt
```diff
@@ -87,6 +88,7 @@ constructor(
       created = auditInfo,
       lastModified = auditInfo,
       ownerEmail = user.email,
+      properties = loiName?.let { mapOf("name" to it) } ?: mapOf(),
```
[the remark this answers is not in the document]
sufyanAbbasi: Good call, added a new property to `LocationOfInterest` and also added a helper method `generateProperties(loiName=null)` method to standardize property generation for future properties.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -197,6 +237,122 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
...
+        focusedContainerColor = getColor(R.color.md_theme_text_field_container),
+        unfocusedContainerColor = getColor(R.color.md_theme_text_field_container),
+        cursorColor = primaryColor,
+      )
+
+    return Triple(saveButtonColors, cancelButtonColors, textFieldColors)
+  }
+
+  @Composable
+  private fun LoiNameDialog(
```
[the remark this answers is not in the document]
sufyanAbbasi: Good call, done. Had to inject the `fragment` into the dialog to retrieve the context etc.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -122,6 +125,10 @@ internal constructor(
 
   lateinit var submissionId: String
 
+  fun setLoiName(name: String) {
+    savedStateHandle[TASK_LOI_NAME_KEY] = name
+  }
```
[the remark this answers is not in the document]
sufyanAbbasi: Great call! Thank you for reminding me of this language feature. Done.

ground/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
sufyanAbbasi: Done.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
sufyanAbbasi: I was getting a null pointer exception, which I tried to solve upstream here, but I realized that I should have been doing it downstream instead. Done.

ground/src/main/res/navigation/nav_graph.xml
sufyanAbbasi: Done.

ground/src/main/res/navigation/nav_graph.xml
sufyanAbbasi: Done.

sufyanAbbasi:
Updated to use `OutlinedButton` and `colorPrimary` attribute, similar to what was done in the `DrawAreaTaskFragment` code.

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
```diff
@@ -0,0 +1,129 @@
...
+      unfocusedContainerColor = textFieldColor,
+      cursorColor = primaryColor,
+    )
+
+  return Triple(saveButtonColors, cancelButtonColor, textFieldColors)
+}
+
+@Composable
+fun LoiNameDialog(
+  fragment: AbstractFragment,
```
anandwana001, line 75: Doesn't passing a fragment reference, make this function dependent?
How about we use the State-Hoisting approach and let all the values come from a specific state, no matter which fragment or activity or anyone call this composable function.

WDYT?

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
```diff
@@ -0,0 +1,129 @@
...
+  val (saveButtonColors, cancelButtonColor, textFieldColors) = fragment.getElementColors()
+  AlertDialog(
+    onDismissRequest = onDismissRequest,
+    icon = {},
+    title = {
+      Column(modifier = Modifier.fillMaxWidth()) {
+        Text(
+          text = fragment.getString(R.string.loi_name_dialog_title),
+          fontSize = 5.em,
+          textAlign = TextAlign.Start,
```
anandwana001, line 90: isn't the start align a default value of the Text function itself?

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
```diff
@@ -0,0 +1,129 @@
...
+        )
+      }
+    },
+    text = {
+      Column {
+        Text(
+          text = fragment.getString(R.string.loi_name_dialog_body),
+          modifier = Modifier.padding(0.dp, 0.dp, 0.dp, 16.dp),
```
anandwana001, line 98: rather than going for a modifier, we can use a separate component called `Spacer` to give a height of 16.dp to bring space between these two components.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -158,10 +162,28 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     lifecycleScope.launch { dataCollectionViewModel.onPreviousClicked(viewModel) }
   }
 
-  fun moveToNext() {
+  private fun moveToNext() {
     lifecycleScope.launch { dataCollectionViewModel.onNextClicked(viewModel) }
   }
 
+  fun handleNext() {
+    if (getTask().isAddLoiTask) {
```
anandwana001, line 170: how about using `when`, so that if in future, checks increase, we don't have to do lot of nesting.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -197,6 +219,31 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
 
   @TestOnly fun getButtonsIndex() = buttonsIndex
 
+  private fun showLoiNameDialog(initialTextValue: String) {
+    (view as ViewGroup).addView(
+      ComposeView(requireContext()).apply {
+        setContent {
+          val openAlertDialog = remember { mutableStateOf(Pair(initialTextValue, true)) }
+          when {
```
anandwana001, line 227: The state-Hoisting approach can be very helpful, if, in future checks or logic increases, we have to move from Pair to other options, which will require more testing and more time. 

Also, currently, it's not much readable, as from `true`, it's hard to understand what it refers to.

ground/src/main/res/values/strings.xml
```diff
@@ -144,4 +145,6 @@
     <item quantity="one">%d submission</item>
     <item quantity="other">%d submissions</item>
   </plurals>
+  <string name="loi_name_dialog_title">Name this location</string>
+  <string name="loi_name_dialog_body">A new data collection site will be created.</string>
```
anandwana001, line 149: If we are going full Jetpack Compose way, we can skip putting our string here, as it requires extra conversion due to the view system and XML.

We can go for `const val` and use the string in our Jetpack Compose UI.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
sufyanAbbasi: Good call, however, received a linter error: `Prefer using 'if' for binary conditions instead of 'when'. [UseIfInsteadOfWhen]`.

ground/src/main/res/values/strings.xml
sufyanAbbasi: Thank you! For now I will keep the string convention because it's fully supported in Compose and it enables us to better handle translations (if we get around to it).

https://developer.android.com/jetpack/compose/resources

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
sufyanAbbasi: Ah, the only reason I was sending in the fragment was for its context... which I now see is not necessary with Composable's `colorResource()`. I was able to find a solution to replace the color attribute as well.
https://stackoverflow.com/questions/62971113/how-to-reference-theme-attributes-in-jetpack-compose

Removed reference to `fragment`.

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
sufyanAbbasi: Good call, done.

ground/src/main/java/com/google/android/ground/ui/datacollection/components/LoiNameDialog.kt
sufyanAbbasi: You're right, removed.

gino-m (approved):
👏 !

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
sufyanAbbasi: ^I will make the Pair fields a little more clear, let me push that right now.

PR #2328, opened 2024-03-11
Update draw area info dialog text and use `onResume()` event to show the dialog.

### LINKED ISSUE #2311: [Draw an area] Dialog shown only after adding point, text incorrect
Screenshot pending

Expected:

image

sufyanAbbasi:
Fixes #2311

Updates text for draw area info. Uses `onResume()` event to trigger the info dialog, since the `setMenuVisibility()` is deprecated in newer versions of Android and was not reliable for determining the task's visibility state overall.

- [x] Updates text to match the mocks.
- [x] Use `onResume()` event + `isVisible` to definitively determine the trigger to show the info dialog.
- [x] [Verified the performance of `SharedPreference`](https://stackoverflow.com/questions/19148282/read-speed-of-sharedpreferences) to be acceptable for multiple rapid-fire calls to `onResume()`.
- [x] Standardized the dialog color.
- [x] Removed unnecessary `StylizedText`, since fontSize already existed.

- [x] Existing unit tests pass.
- [x] Locally by removing the setter to always trigger the dialog and determinine that the 

Screenshot 2024-03-09 at 5 09 14 PM

@gino-m @shobhitagarwal1612 ?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
@@ -92,11 +92,9 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     }
   }
 
-  override fun setMenuVisibility(menuVisible: Boolean) {
-    super.setMenuVisibility(menuVisible)
-    if (menuVisible) {
-      onTaskVisibleToUser()
-    }
+  override fun onResume() {
+    super.onResume()
+    onTaskResume()
   }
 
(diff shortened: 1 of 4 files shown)


gino-m (approved):
Lgtm! Is this working for Drop a pin and Capture location as well?

sufyanAbbasi:
> Lgtm! Is this working for Drop a pin and Capture location as well?

Testing now.

sufyanAbbasi:
- `Capture location` doesn't have a dialog.
- `Drop a pin` seems to be bound to the hint icon. I think we'd need to refactor it to match the mocks.
 
Screenshot 2024-03-11 at 3 58 38 PM

Screenshot 2024-03-11 at 3 53 01 PM

I'm happy to perform the refactor!

Issue #794, opened 2021-05-18
[Data collection] Add support for "Other" option in multiple choice questions


gino-m:
If "other" is a valid multiple choice option allow data collectors to add a free text field to specify details

![Image](https://github.com/google/ground-android/assets/228050/dfe0c209-681e-4e0a-8be8-bc4c7c354298)

scolsen:
@JSunde to update the status on this.

gino-m:
This is a new feature specific to the new Ground.

JSunde:
What is the expectation for this new functionality?
1) Will the Survey organizer need to enable the "Other" option for Multiple Choice tasks when they are defining the job? Or will it always be available?
2) Will there be a text field beside the option for the user to specify what the other category is?
3) Will we allow multiple "Other" options for the Checkbox use case?

JSunde:
@gino-m

JSunde:
I took a look and I still think that most of these questions are not directly answered by the mocks, should I follow up here or on Figma?

gino-m:
> I took a look and I still think that most of these questions are not directly answered by the mocks, should I follow up here or on Figma?

Thanks for following up on the Figma. Reiterating my reply there, for the initial version we should aim for min viable functionality and simplest possible experience for the organizer. To that end, I'd suggest having a special multiple choice value with fixed "Other" label, and an optional free text for the data collector to qualify the "other" value. Does that make sense?

shobhitagarwal1612:
Also fix the UX for radio button

gino-m:
It's not clear whether this will be needed for field testing. Let's reassign once we have clarity.

gino-m:
To coordinate with @os-micmec and @gino-m

gino-m:
See https://github.com/google/ground-android/issues/2202 for schema

gino-m:
We can handle https://github.com/google/ground-android/issues/2219 separately.

jcqli:
@justinkimtang Is this closed now that https://github.com/google/ground-android/pull/2340 is merged?

PR #2330, opened 2024-03-11
Updating MultipleChoiceSchema to include other option

justinkimtang:
Towards #794 

Updating MultipleChoiceEntity to include newly added "hasOtherOption" field read from  the TaskNestedObject.

@gino-m  PTAL?


gino-m (approved):
LGTM!

Issue #2314, opened 2024-03-07
[Sync status] App crashes when sync status screen open and transitioning from "MEDIA_UPLOADED" to "COMPLETE"


gino-m:
After disabling Airplane mode:

From Crashlytics:

```
Fatal Exception: com.google.android.ground.persistence.local.LocalDataConsistencyException: Unknown jobId in submission mutation 1
       at com.google.android.ground.persistence.local.room.converter.ConverterExtKt.toModelObject(ConverterExt.kt:341)
       at com.google.android.ground.persistence.local.room.stores.RoomSubmissionStore$getAllSurveyMutationsFlow$$inlined$map$1$2.emit(Emitters.kt:224)
       at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
       at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:15)
       at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:87)
       at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
       at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:37)
       at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
       at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(:14)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at android.os.Handler.handleCallback(Handler.java:959)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loopOnce(Looper.java:232)
       at android.os.Looper.loop(Looper.java:317)
       at android.app.ActivityThread.main(ActivityThread.java:8501)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:878)
```

[...]

PR #2333, opened 2024-03-12
SubmissionStore: Handle exceptions in the allSurveyMutationsFlow

scolsen:
Adds a `catch` call to handle exceptions thrown during db -> model object conversion for submissions, preventing these from reaching downstream collectors. As a result, these values will be ignored in any downstream flow or collector of the flow.

Towards #2314


scolsen:
@gino-m this should at least prevent the crash in #2314, but if it's true that the cause was a state change on an existing mutation, it's not obvious to me where the bug lies. We perform copies in every case, so there must be a subtle error somewhere either in the workers or when we write to the db....yet another reason it would be good for us to revisit sync design and have clearer separation of responsibilities.

I won't have time to look into the underlying issue this week, unfortunately.

gino-m (approved):
Thank you for the fix!

PR #2334, opened 2024-03-12
Make location permissions more robust to offline errors, remove disabling the location lock button.

### LINKED ISSUE #2310: [Draw an area] App crashes on secondary (non-LOI) geometry tasks when offline
On Pixel 7 physical device in Airplane Mode, as a normal data collection task while adding an LOI. Does not occur when connected:

```
Fatal Exception: com.google.android.gms.common.api.ApiException: 8502: SETTINGS_CHANGE_UNAVAILABLE
       at com.google.android.gms.common.internal.ApiExceptionUtil.fromStatus(com.google.android.gms:play-services-base@@18.1.0:3)
       at com.google.android.gms.common.api.internal.TaskUtil.setResultOrApiException(com.google.android.gms:play-services-base@@18.1.0:4)
       at com.google.android.gms.internal.location.zzcq.zzb(com.google.android.gms:play-services-location@@21.0.1:2)
       at com.google.android.gms.internal.location.zzr.zza(com.google.android.gms:play-services-location@@21.0.1:3)
       at com.google.android.gms.internal.location.zzb.onTransact(com.google.android.gms:play-services-location@@21.0.1:3)
       at android.os.Binder.execTransactInternal(Binder.java:1501)
       at android.os.Binder.execTransact(Binder.java:1440)
```

sufyanAbbasi:
Fixes #2310

Makes location permissions more robust to offline mode, where the act of checking a permission requires a network roundtrip to the Play Store. By making the location requests more robust to exceptions, we can detect when the user is offline and don't check for location permissions explicitly, then fail with an error message when the user doesn't have the permission we need.

- [x] Adds try/catch blocks to detect for offline `SETTINGS_CHANGE_UNAVAILABLE` error when checking permissions against Play Store, then greedily attempts to get the location in this case.
- [x] Prevent the disabling of the location button, in order to allow the user to retry the location fetching (and retrigger the permission error).
- [x] Prevents crashes due to user denying permission while using the app, as well as turning it off manually.
- [x] Note that the capture location task when permission is denied is unusable.

- [x] Created a survey with capture location LOI, drop pin, draw area, and another capture location task.
- [x] Verified that when online, the lock location button works correctly with permission.
- [x] Verified that when offline, as long as the user has permission, the tasks can be completed.
- [x] Verified that when online and the user has no location permission, the app doesn't crash, prompts user to give permission.
- [x] Verified the same but when offline. 
  

Screenshot 2024-03-12 at 3 29 11 PM

PTAL @gino-m @shobhitagarwal1612


ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -145,13 +145,27 @@ constructor(
...
+      locationManager.requestLocationUpdates()
+    } catch (throwable: ApiException) {
+      val statusCode = throwable.statusCode
+      if (statusCode == SETTINGS_CHANGE_UNAVAILABLE) {
+        Timber.e(
+          throwable,
+          "User is offline, so fallback to user's current permission, which may also fail."
+        )
+        enableLocationLock()
+        locationManager.requestLocationUpdates()
```
gino-m, line 161: Can we just move these two lines after the catch block and remove the duplicate lines from the try block?

gino-m (changes requested):
One nit, otherwise LGTM!

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
```diff
@@ -145,13 +145,27 @@ constructor(
...
+      } else {
+        throw throwable
```
shobhitagarwal1612, line 163: Instead of rethrowing, should we directly call `handleRequestLocationUpdateFailed`? Then we can get rid of the try-catch block in `CaptureLocationTaskMapFragment.kt` Line 57-61

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
sufyanAbbasi: Good call! Done, and coordinated with the change that @shobhitagarwal1612 suggested below.

ground/src/main/java/com/google/android/ground/ui/common/BaseMapViewModel.kt
sufyanAbbasi: Unified all the try/catch into the `enableLocationLockAndGetUpdates()` so that we can call it without fear of throwing errors that will crash the app. The error message will now always funnel through the location lock icon failure handler, which is pleasant.

gino-m (approved):
LGTM!

PR #2337, opened 2024-03-13
Move back to map after offline area download.

### LINKED ISSUE #2301: [Offline area download] Return to main map after download completes
...instead of going back to the offline area list.

@sufyanAbbasi PTAL?

sufyanAbbasi:
Fixes #2301

Moves back to the map after completing a download.

- [x] Replaces unused `nav_graph` direction, `backToOfflineAreas`, with `offline_area_back_to_homescreen`
- [x] Instead of navigating up the back stack, navigate to the home screen instead. 

- [x] Downloaded an offline area and saw that the map view was returned to, conveniently exactly where the download occurred!

Screenshot 2024-03-12 at 6 18 23 PM

Screenshot 2024-03-12 at 6 26 24 PM

@gino-m


gino-m (approved):
🎸 !

Issue #620, opened 2020-11-09
[UI] Add support for Dark Theme


gino-m:

gino-m:
@sufyanAbbasi Can we force light mode on for now, perhaps like [this](https://stackoverflow.com/a/64999338)?

sufyanAbbasi:
Unfortunately, the solution `<item name="android:forceDarkAllowed">false</item>` only works on Android >=29, but we are currently targeting 24, most likely for backwards compatibility reasons.

It is possible to make a `v29/styles.xml` which will presumably apply to those versions or higher, but it looks like the style rules can end up skewing if we're not careful.

gino-m:
> Unfortunately, the solution `<item name="android:forceDarkAllowed">false</item>` only works on Android >=29, but we are currently targeting 24, most likely for backwards compatibility reasons.
> 
> It is possible to make a `v29/styles.xml` which will presumably apply to those versions or higher, but it looks like the style rules can end up skewing if we're not careful.

Can [`setDefaultNightMode()`](https://developer.android.com/reference/androidx/appcompat/app/AppCompatDelegate#setDefaultNightMode(int)) be used here

sufyanAbbasi:
I'll give it a shot!

sufyanAbbasi:
It actually worked!

anandwana001:
@sufyanAbbasi What's the update on this Bug? Are we having dark mode or keeping in light theme only?
Also, if we are having Dark Theme, can you share where I can get the colors and themes?

gino-m:
I'd say this is a P2 for GA, relative to other tasks in the queue.

@rawbzz Would it be possible to provide a version of the designs and/or color scheme for "dark mode"?

gino-m:
@vittorino When you get a sec can you please paste in the light+dark color schemes to this bug?

PR #2344, opened 2024-03-13
Force light theme (sorry!)

sufyanAbbasi:
Addresses #620

Forces light theme until we have a better dark theme story.

- [x] Enable dark theme and verified that the UI and tasks are in light theme. Note that the start up splash (the one with the ground logo) is set to dark background which looks pretty cool.

Screenshot 2024-03-13 at 12 18 44 PM
Screenshot 2024-03-13 at 12 15 20 PM
Screenshot 2024-03-13 at 12 21 20 PM

PTAL @gino-m @scolsen !


gino-m (approved):
Thanks for the fix, and sorry!

PR #2349, opened 2024-03-14
Refactor a common InstructionsDialog class and ensure Drop Pin tasks …

### LINKED ISSUE #2315: [Drop a pin] Instruction popup does not appear
image

sufyanAbbasi:
Fixes #2315 

Factors out the `InstructionsDialog` from `CaptureLocationTaskFragment` and implements it on the `DropPinTaskFragment`.

- [x] Creates a shared component, `InstructionsDialog`, which displays a standardized `AlertDialog`.
- [x] Implements it for both the capture location and drop pin tasks.
- [x] Removes the info icon from the drop pin task.

- [x] Deleted app data to retrigger the dialogs, saw that both instructions get triggered, but don't persistent after being seen. 

Screenshot 2024-03-13 at 3 16 18 PM
Screenshot 2024-03-13 at 3 16 08 PM

PTAL @shobhitagarwal1612 @gino-m !


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -44,11 +42,6 @@ class DropPinTaskMapFragment(private val viewModel: DropPinTaskViewModel) :
     savedInstanceState: Bundle?,
   ): View {
     val root = super.onCreateView(inflater, container, savedInstanceState)
```
shobhitagarwal1612, line 44: We can cleanup this overridden method as well.

shobhitagarwal1612 (approved):
1 minor cleanup suggestion.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
sufyanAbbasi: Thank you! Cleaned up nicely!

PR #2369, opened 2024-03-15
Select other option when user types something in.

### LINKED ISSUE #2362: [Multiple select] Typing in "other" field doesn't select the "other" option

sufyanAbbasi:
Fixes #2362

Ensures that when the user enters some text into the option text area, the option is checked. Will also remove the checkmark if the user backspaces.

- [x] Checks option when the user types something in.
- [x] Unchecks when text is fully deleted.
- [x] Still allows empty other and unchecked other with text.
- [x] Tried using `onClick` and `onTextChange` on the fragment xmls, but they didn't work as expected (`onClick` required two clicks, `onTextChange` ate the first user input).

- [x] Verified correct behavior with checkboxes and radio inputs.
- [x] Verified server has the other option. 

Default state:
Screenshot 2024-03-14 at 8 47 19 PM
Entered text:
Screenshot 2024-03-14 at 8 47 26 PM
Deleted text:
Screenshot 2024-03-14 at 8 47 35 PM
Allows checking empty:
Screenshot 2024-03-14 at 8 48 40 PM
Allows unchecking with text:
Screenshot 2024-03-14 at 8 48 51 PM

PTAL @gino-m @justinkimtang !

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskViewModel.kt
@@ -22,6 +22,7 @@ import com.google.android.ground.model.job.Job
 import com.google.android.ground.model.submission.MultipleChoiceResponse
 import com.google.android.ground.model.submission.MultipleChoiceResponse.Companion.fromList
 import com.google.android.ground.model.submission.Value
+import com.google.android.ground.model.task.MultipleChoice.Cardinality.SELECT_MULTIPLE
 import com.google.android.ground.model.task.Option
 import com.google.android.ground.model.task.Task
 import com.google.android.ground.ui.datacollection.tasks.AbstractTaskViewModel
@@ -30,7 +31,6 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.filterNotNull
-import kotlinx.coroutines.flow.firstOrNull
 
 class MultipleChoiceTaskViewModel @Inject constructor(resources: Resources) :
   AbstractTaskViewModel(resources) {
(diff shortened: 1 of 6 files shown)


shobhitagarwal1612 (commented):
Can we add some test coverage as well?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskViewModel.kt
```diff
@@ -40,10 +41,14 @@ class MultipleChoiceTaskViewModel @Inject constructor(resources: Resources) :
...
 
-  val textWatcher =
+  val otherTextWatcher =
     object : TextWatcher {
       override fun onTextChanged(s: CharSequence, start: Int, before: Int, count: Int) {
         otherText = s.toString()
+        // Set the other option.
+        _items.value
+          .firstOrNull { it.isOtherOption }
+          ?.let { setItem(it, otherText != "", it.cardinality == SELECT_MULTIPLE) }
```
justinkimtang, line 51: Do we also want to select it when its a radio button and the user start input?

If so we can remove the cardinality check here and in the object itself and just always add the otherId to the selected list on text change.

gino-m: Should instead we make the "other" text required? If we follow the model that Forms uses, it's required if the field is required, optional if not.  In the latter case, we empty other text is a valid state, so we wouldn't want to deselect.

sufyanAbbasi:
Working on testing this out.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskViewModel.kt
sufyanAbbasi: > Do we also want to select it when its a radio button and the user start input?

Ah, yes! The implementation works for both types. In this case, we're setting the `canSelectMultiple` parameter based on the cardinality of the input.

Screenshot 2024-03-15 at 2 44 02 PM

> If the text is empty here i.e. the user deleted everything do we also want to deselect?
> In the latter case, we empty other text is a valid state, so we wouldn't want to deselect.
Thank you for raising this, I will definitely refactor to consider the `isRequired` state for the task.

sufyanAbbasi:
Unit tests are in now! PTAL @shobhitagarwal1612!

shobhitagarwal1612 (approved):

PR #2370, opened 2024-03-15
ADD Troubleshooting SHA1 matching issues

yongsuk44:
It seems like the issue of SHA1 mismatch during Google login could be encountered by everyone who runs the project for the first time. It would be a good idea to add this to a separate troubleshooting tab.

@gino-m PTAL


gino-m:
Thank you, @yongsuk44, this is very helpful!

gino-m (approved):

gino-m:
anandwana001@, does this resolve your issue in #2324?

Issue #2136, opened 2023-12-20
Show current survey title and description somewhere!


gino-m:

rawbzz:
testing

rawbzz:
Link for new designs [here](https://www.figma.com/file/r2uGtyOiWoDa29za7DNuUc/Ground-Acacia-UX?type=design&node-id=5076%3A15265&mode=design&t=1BrOFjFheJMmzsbU-1)

gino-m:
Thanks for the create design @rawbzz ! Reassigning for triage to eng.

jcqli:
@scolsen @gino-m This looks to be done, closing

PR #2386, opened 2024-03-17
Hide textview placeholder if survey doesn't contain a description

shobhitagarwal1612:
Towards #2136

Hide description placeholder if survey doesn't have one. Otherwise we see blank space between title and change survey button

@gino-m  PTAL?


gino-m (approved):
Good catch, thanks for the fix!

PR #2391, opened 2024-03-18
Refactor NavigationRequest `object` to `data object`

yongsuk44:
'data object' instead of a plain 'object' means that it automatically has a 'toString()' function without the need to override it manually

[Link](https://kotlinlang.org/docs/whatsnew19.html#stable-data-objects-for-symmetry-with-data-classes)


shobhitagarwal1612:
@yongsuk44 Can you please fix the failing build?

yongsuk44:
@shobhitagarwal1612 
sry, object's 'import' and Lint issues and fixed them.

ground/src/main/java/com/google/android/ground/ui/common/NavigationRequest.kt
```diff
@@ -17,10 +17,10 @@ package com.google.android.ground.ui.common
 
 import androidx.navigation.NavDirections
 
-sealed interface NavigationRequest
+sealed interface NavigationRequest {
+  data object NavigateUp : NavigationRequest
 
-data class NavigateTo(val directions: NavDirections) : NavigationRequest
+  data object FinishApp : NavigationRequest
```
anandwana001, line 23: What is the rational behind changing it from `object` to `data object`? other than `data object` gives you the power to have a toString() function as written in the PR description, which just prints the name of the object

looks like an extra work we are trying to give to compiler to generate some functions for us, which we are currently not using.

WDYT? 
Let me know if I am missing anything here.

gino-m: This also adds more boilerplate it seems in referencing code.

yongsuk44: Thanks for the comment. 😀

@anandwana001 
The functions generated by the `data object`, such as `toString()`, may not be used in production code, but they can be useful for debugging during development. This can be easily understood through a [link](https://kotlinlang.org/docs/whatsnew1720.html#improved-string-representations-for-singletons-and-sealed-class-hierarchies-with-data-objects).
Also, Requesting additional work from the compiler is also justified because the Kotlin compiler is optimized to handle methods like `toString`, `equals`, and `hashCode` (e.g., for data classes).

@gino-m 
I understand that this leads to some additional boilerplate, but I think it has improved readability.

anandwana001: Got the idea now, cool, let's have it.

gino-m:
Running integration tests, they should complete shortly:

/gcbrun

shobhitagarwal1612:
`data` to `data object` is already addressed in https://github.com/google/ground-android/pull/2483

PR #2395, opened 2024-03-18
[Auth] Wait for user to be initialized before checking if null

### LINKED ISSUE #2390: [BLOCKING] User logged out when app killed
This occurred while offline, so the user could not sign back into the app to continue collecting data.

I haven't been able to reproduce this, but here are the steps I took:

1. Sign into the Ground app
2. Collect data for a new site
3. Open sync status screen
4. Turn on Airplane mode
5. Kill the app (swipe app or force quit)
6. Turn off Airplane mode, wait ~1 minute
7. Turn Airplane mode on again
8. Reopen the app

### Expected result
The user is still signed in and can continue collecting data.

### Actual result
The user was signed out.

No relevant exceptions were found in Crashlytics.

gino-m:
`FirebaseAuth` doesn't block the main thread, so checking its state directly at startup may be producing a race condition on init of the current user. Waiting for auth events before checking should solve this issue.

Fixes #2390


shobhitagarwal1612 (approved):
🏆

PR #2400, opened 2024-03-21
Only trigger permission dialog for capture location when user arrives on the task.

### LINKED ISSUE #2388: [Capture location] Location permissions requested on wrong task
Location permission dialog was shown on preceding task rather than on the capture location task.

@sufyanAbbasi FYI

sufyanAbbasi:
Fixes #2388

Prevents the location permission dialog from being shown too early (see issue for explanation), by only triggering early when the user has already given permission, and otherwise waiting to trigger the dialog until the user has arrived at the capture location task.

- [x] Adds a new flow property, `enableLocationLockFlow`, that keeps track of three states: 
1.   whether the location lock has already been enabled (`null`)
2.  whether the location lock needs to be enabled (`false`)
3.  whether to actually enable the location lock  (`true`)
- [x] If permission was already given, the existing behavior of starting the location lock early is preserved.
- [x] If permission was not given, then the request to turn on location lock is deferred to when the task is resumed or the user hits "Capture" if they declined the permission.

- [x] Created survey with a text task before a capture location task. Verified that turning off permission does not trigger the permission early, and triggers when the user arrives on the task.
- [x] Verified that already having the permission replicates the existing behavior.

PTAL @gino-m @shobhitagarwal1612 !

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
@@ -18,6 +18,7 @@ package com.google.android.ground.ui.datacollection.tasks.location
 import android.view.LayoutInflater
 import android.view.View
 import android.widget.LinearLayout
+import androidx.lifecycle.lifecycleScope
 import com.google.android.ground.R
 import com.google.android.ground.model.submission.isNotNullOrEmpty
 import com.google.android.ground.model.submission.isNullOrEmpty
@@ -28,6 +29,7 @@ import com.google.android.ground.ui.datacollection.tasks.AbstractTaskFragment
 import com.google.android.ground.ui.map.MapFragment
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
+import kotlinx.coroutines.launch
 
 @AndroidEntryPoint
 class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskViewModel>() {
@@ -48,6 +50,13 @@ class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskView
       )
       .commit()
     return rowLayout
+  }
+
+  override fun onTaskResume() {
+    // Ensure that the location lock is enabled, if it hasn't been..
+    if (viewModel.enableLocationLockFlow.value == false) {
+      viewLifecycleOwner.lifecycleScope.launch { viewModel.enableLocationLockFlow.emit(true) }
+    }
   }
 
   override fun onCreateActionButtons() {
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
@@ -17,21 +17,34 @@ package com.google.android.ground.ui.datacollection.tasks.location
 
 import android.content.res.Resources
 import android.location.Location
+import androidx.lifecycle.viewModelScope
 import com.google.android.ground.ui.datacollection.tasks.AbstractTaskViewModel
 import com.google.android.ground.ui.datacollection.tasks.location.CaptureLocationTaskResult.Companion.toCaptureLocationResult
 import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.launch
 
 class CaptureLocationTaskViewModel @Inject constructor(resources: Resources) :
   AbstractTaskViewModel(resources) {
 
   private val lastLocation = MutableStateFlow<CaptureLocationTaskResult?>(null)
+  /**
+   * Allows control for triggering the location lock programmatically, and captures a ternary state:
+   * 1. null: The location lock was already enabled.
+   * 2. false: The location lock needs to be enabled.
+   * 3. true: The location lock will be enabled when emitted.
+   */
+  val enableLocationLockFlow = MutableStateFlow<Boolean?>(null)
 
   suspend fun updateLocation(location: Location) {
     lastLocation.emit(location.toCaptureLocationResult())
   }
 
   fun updateResponse() {
-    setValue(lastLocation.value)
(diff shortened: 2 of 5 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
```diff
@@ -50,6 +52,15 @@ class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskView
     return rowLayout
   }
 
+  override fun onTaskResume() {
+    // Ensure that the location lock is enabled, if it hasn't been.
+    if (viewModel.enableLocationLockFlow.value == LocationLockState.NEEDS_ENABLE) {
+      viewLifecycleOwner.lifecycleScope.launch {
+        viewModel.enableLocationLockFlow.emit(LocationLockState.ENABLE)
+      }
+    }
```
shobhitagarwal1612, line 61: We can simplify this by creating a `enableLocationLock()` method in viewModel and moving all these details into that. Then we won't have to expose any flows or other states to this fragment. Thoughts?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
```diff
@@ -17,21 +17,46 @@ package com.google.android.ground.ui.datacollection.tasks.location
...
     lastLocation.emit(location.toCaptureLocationResult())
   }
 
   fun updateResponse() {
-    setValue(lastLocation.value)
+    if (lastLocation.value == null) {
+      viewModelScope.launch { enableLocationLockFlow.emit(LocationLockState.ENABLE) }
+    } else {
+      setValue(lastLocation.value)
+    }
```
shobhitagarwal1612, line 60: Is this scenario even possible? Ideally we should disable the `Capture location` button if there is no location to be set.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
```diff
@@ -17,21 +17,46 @@ package com.google.android.ground.ui.datacollection.tasks.location
...
+
+  /** The location lock was already enabled, or an attempt was made. */
+  ALREADY_ENABLED,
+
+  /** The location lock was not already enabled. */
+  NEEDS_ENABLE,
+
+  /** Trigger to enable the location lock. */
+  ENABLE,
+}
```
shobhitagarwal1612, line 40: Can we move this to a shared class as location lock is present in other fragments as well?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
sufyanAbbasi: This is a great suggestion, I was even able to refactor the login in `onMapReady()` and set `enableLocationLockFlow` to private, which is nice separation of concerns. Done.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
sufyanAbbasi: I realized that because I was able to completely localize usage of this enum to one file, I marked it private instead. I'd prefer it be used for this specialized purpose rather than generally across the code base, if that's ok with you!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
shobhitagarwal1612: That makes sense! Thank you!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
```diff
@@ -17,21 +17,71 @@ package com.google.android.ground.ui.datacollection.tasks.location
...
+    if (enableLocationLockFlow.value == LocationLockEnabledState.NEEDS_ENABLE) {
+      viewModelScope.launch { enableLocationLockFlow.emit(LocationLockEnabledState.ENABLE) }
+    }
+  }
+
+  suspend fun onMapReady(mapViewModel: BaseMapViewModel) {
+    val locationLockEnabledState =
+      if (mapViewModel.hasLocationPermission()) {
+        // User has permission to enable location updates, enable it now.
+        mapViewModel.enableLocationLockAndGetUpdates()
```
shobhitagarwal1612, line 72: Won't this cause the dialog to appear in the previous task too?

shobhitagarwal1612: I thought we were calling `enableLocationLockAndGetUpdates` only via `collect` below

sufyanAbbasi: > Won't this cause the dialog to appear in the previous task too?
I tested to make sure that this is true, but the permission dialog does not show if the user already has permission, so we can preemptively start collecting location to preserve the existing behavior of front-loading the location tracking before the user arrives on the task (otherwise, there's a delay before the first data arrives).

> I thought we were calling enableLocationLockAndGetUpdates only via collect below
Meanwhile, the collect below is a fallback to turn on location tracking later manually when we need to. We could throw it in the `else` clause so that we only collect if we fall into the "no permissions" case, but I think having a noop consumer that's always there is somewhat useful. I'm happy to make the change though!

shobhitagarwal1612: Thanks! Marking this as resolved. Given the complexity of the method, maybe it can benefit with some refactoring (maybe in a follow-up)

shobhitagarwal1612 (approved):

PR #2401, opened 2024-03-22
[UI] Convert sign out confirmation dialog to Material 3

### LINKED ISSUE #2387: [UI Bug] Sign out confirmation dialog not Material 3 dialog

shobhitagarwal1612:
Fixes #2387

|  Before  |  After  |
|---|---|
|   r |    |

@gino-m @scolsen @sufyanAbbasi  PTAL?


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
```diff
@@ -147,9 +159,62 @@ class HomeScreenFragment :
       R.id.sync_status -> homeScreenViewModel.showSyncStatus()
       R.id.nav_offline_areas -> homeScreenViewModel.showOfflineAreas()
       R.id.nav_settings -> homeScreenViewModel.showSettings()
-      R.id.nav_sign_out -> homeScreenViewModel.showSignOutConfirmation()
+      R.id.nav_sign_out -> showSignOutConfirmationDialog()
     }
     closeDrawer()
     return true
   }
+
```
gino-m, line 167: Is there a way to move this and other side panel logic into their own files? Ok to do in a follow up.

gino-m (approved):

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
shobhitagarwal1612: Moved the compose code to a separate file.

PR #2409, opened 2024-03-24
Fix crash in data collection fragment

### LINKED ISSUE #2408: [Bug] App crash in data collection screen
Issue reported via crashlytics

Exception message:  `Fatal Exception: java.util.NoSuchElementException:  List is empty`

Stacktrace:

```
com.google.android.ground.ui.datacollection.DataCollectionViewModel.<init> (DataCollectionViewModel.kt:136)
com.google.android.ground.DaggerGroundApplication_HiltComponents_SingletonC$ViewModelCImpl$SwitchingProvider.get (DaggerGroundApplication_HiltComponents_SingletonC.java:1480)
dagger.hilt.android.internal.lifecycle.HiltViewModelFactory$2.createViewModel (HiltViewModelFactory.java:132)
dagger.hilt.android.internal.lifecycle.HiltViewModelFactory$2.create (HiltViewModelFactory.java:103)
dagger.hilt.android.internal.lifecycle.HiltViewModelFactory.create (HiltViewModelFactory.java:170)
androidx.lifecycle.ViewModelProvider.get (ViewModelProvider.kt:184)
```

The above stacktrace points to code below:

```
var currentTaskId: StateFlow<String> =
    savedStateHandle.getStateFlow(TASK_POSITION_ID, tasks.first().id)
```

`first()` expects the list to be non-empty and throws an exception. Handling that should atleast prevent the crash from happening.

shobhitagarwal1612:
Fixes #2408

Instead of crashing the app, reset the current task id to empty. 

@gino-m @sufyanAbbasi  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -133,7 +133,7 @@ internal constructor(
 
   // Tracks the current task ID to compute the position in the list of tasks for the current job.
   var currentTaskId: StateFlow<String> =
-    savedStateHandle.getStateFlow(TASK_POSITION_ID, tasks.first().id)
+    savedStateHandle.getStateFlow(TASK_POSITION_ID, tasks.firstOrNull()?.id ?: "")
 
   lateinit var submissionId: String
 


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -133,7 +133,7 @@ internal constructor(
 
   // Tracks the current task ID to compute the position in the list of tasks for the current job.
   var currentTaskId: StateFlow<String> =
-    savedStateHandle.getStateFlow(TASK_POSITION_ID, tasks.first().id)
+    savedStateHandle.getStateFlow(TASK_POSITION_ID, tasks.firstOrNull()?.id ?: "")
```
gino-m, line 136: Do we need to update anything else to check if taskId is empty?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: That is already handled. We use it to check current position and if it's empty, we fallback to first position

gino-m (approved):

PR #2414, opened 2024-03-29
Use jetpack compose for buttons in task fragments

### LINKED ISSUE #1795: Migrate to Jetpack Compose
Some early experimentation in #1629

shobhitagarwal1612:
Towards #1795

This PR replaces xml layouts for action buttons with jetpack compose.

[compose-task-button.webm](https://github.com/google/ground-android/assets/8918023/ac841e8a-f91c-4ab9-84cb-e62fbb9ed5c2)

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -192,21 +198,30 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
...
+  /** Adds the action buttons to the UI. */
+  private fun renderButtons() {
+    taskView.actionButtonsContainer.composeView.apply {
+      setContent {
+        AppTheme {
+          Row(
+            horizontalArrangement = Arrangement.SpaceBetween,
+            modifier = Modifier.fillMaxWidth().padding(8.dp),
+          ) {
+            // TODO: Previous button should always be positioned to the left of the screen. Rest
```
gino-m, line 216: Please add tracking bug.

gino-m (changes requested):
Overall looks good! @sufyanAbbasi Do you think this is safe to merge, or do we want to implement e2e testing first?

shobhitagarwal1612:
The tests are passing locally. Not sure what the problem is with remote builds.

gino-m:
> The tests are passing locally. Not sure what the problem is with remote builds.

The "Update build status" step failed, not the actual tests. Flake?

shobhitagarwal1612:
We upload the results in this step and then fail the build. So this is WAI

gino-m:
> We upload the results in this step and then fail the build. So this is WAI

I see. So the actual error is in the unit tests:

```
com.google.android.ground.ui.signin.SignInFragmentTest > Back press should finish activity FAILED
    kotlinx.coroutines.test.UncaughtExceptionsBeforeTest at TestScope.kt:242
```

It seems the activity isn't closing on "Back" on GCB for some reason?

shobhitagarwal1612:
I can see this test failing occasionally on local builds as well. Should we retry?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
shobhitagarwal1612: Done

gino-m (approved):
⏺️

shobhitagarwal1612:
Even before this change, the buttons were generated dynamically without a resource ID. So we can use text or content description (for image buttons)

PR #2418, opened 2024-04-01
Apply app theme to jetpack compose ui

### LINKED ISSUE #2419: [Code health] Apply app theme to jetpack compose UI
The app uses jetpack compose for rendering alert dialogs and custom dialog in various parts of the app. These components aren't using any themes, so each element has to be customized to apply consistent theming. 

This can be avoid by directly applying the `AppTheme` and rendering the compose ui inside it.

shobhitagarwal1612:
Fixes #2419

Replaces custom styling of compose with consistent app theme

Currently we have a few jetpack compose UI elements (e.g. alert dialog, custom dialog, etc) which aren't using the xml styled themes. Because of this, we have to manually apply theme colors to these texts and buttons. To avoid this, we can generate the theme and apply it consistently for all jetpack compose generated ui.

The theme was generated using material theme builder by selecting primary, secondary, tertiary and neutral colors from current xml theme. Rest of the colors were generated as per material guidelines by the theme builder.
https://m3.material.io/theme-builder

Note: There are small differences between the xml theme and the auto-generated theme. We should eventually move off of xml theme once all components are migrated to compose.

@gino-m  PTAL?


gino-m:
Thanks for doing this! If we plan to migrate to Compose we should document
and share our motivations and strategy. Would you be about to file a Code
Health issue for this and describe those there?

shobhitagarwal1612:
Filed a bug and added details.

shobhitagarwal1612:
@gino-m Gentle ping!

gino-m (approved):

PR #2423, opened 2024-04-01
Add custom keys to firebase crashlytics

### LINKED ISSUE #2421: [Code health] Survey ID and other debug details not showing in Crashlytics
This will help debug #2416 and other Crashlytics exceptions.

shobhitagarwal1612:
Fixes #2421

Currently, we were only logging survey id when the error is logged. But if the app crashes, then the app closes directly and the custom key isn't loaded. 

To fix that, update the selected survey id and screen name to firebase crashlytics proactively.

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
+++ b/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.google.android.ground
 
 import com.google.firebase.crashlytics.FirebaseCrashlytics
--- a/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
+++ b/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
@@ -32,7 +32,8 @@ class FirebaseCrashLogging @Inject constructor() {
   }
 
   private fun setCustomKeys(init: KeyValueBuilder.() -> Unit) {
-    val builder = KeyValueBuilder(FirebaseCrashlytics.getInstance())
-    builder.init()
+    if (BuildConfig.BUILD_TYPE.contentEquals("release")) {
+      KeyValueBuilder(FirebaseCrashlytics.getInstance()).init()
+    }
   }
 }
--- a/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
+++ b/ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
@@ -15,6 +15,7 @@
  */
 package com.google.android.ground
 
+import com.google.android.ground.Config.isReleaseBuild
 import com.google.firebase.crashlytics.FirebaseCrashlytics
 import com.google.firebase.crashlytics.KeyValueBuilder
 import javax.inject.Inject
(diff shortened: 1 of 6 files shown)


ground/src/main/java/com/google/android/ground/util/Debug.kt
```diff
@@ -15,13 +15,19 @@
...
 
 object Debug {
   fun logLifecycleEvent(instance: Any) {
     val stackTrace = Thread.currentThread().stackTrace
     val callingMethod = stackTrace[3].methodName + "()"
-    Timber.tag(instance.javaClass.simpleName).v("Lifecycle event: $callingMethod")
+    val className = instance.javaClass.simpleName
+    Timber.tag(className).v("Lifecycle event: $callingMethod")
+
+    if (callingMethod == "onResume()") {
```
gino-m, line 28: Nice trick!

ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
```diff
@@ -0,0 +1,39 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class FirebaseCrashLogging @Inject constructor() {
+
+  fun setSelectedSurveyId(surveyId: String?) {
+    setCustomKeys { key("selectedSurveyId", surveyId ?: "") }
+  }
+
+  fun setScreenName(name: String) {
```
gino-m, line 30: Can we call this something more specific so that it's more obvious where this comes from? I'm thinking "fragmentClass" or similar.

gino-m (changes requested):
Nice work, that was fast, too!

ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
```diff
@@ -0,0 +1,39 @@
...
+    setCustomKeys { key("screenName", name) }
+  }
+
+  private fun setCustomKeys(init: KeyValueBuilder.() -> Unit) {
+    if (BuildConfig.BUILD_TYPE.contentEquals("release")) {
```
gino-m, line 35: Can we use `isReleaseBuild()` here as well?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
gino-m: Then maybe `viewClass`?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/util/Debug.kt
```diff
@@ -15,13 +15,19 @@
...
 
+import com.google.android.ground.FirebaseCrashLogging
 import timber.log.Timber
 
 object Debug {
   fun logLifecycleEvent(instance: Any) {
     val stackTrace = Thread.currentThread().stackTrace
     val callingMethod = stackTrace[3].methodName + "()"
-    Timber.tag(instance.javaClass.simpleName).v("Lifecycle event: $callingMethod")
+    val className = instance.javaClass.simpleName
```
gino-m, line 25: This method relies on us remembering to call `logLifecycleEvent()` directly or via `super()`. Instead, can we use [registerActivityLifecycleCallbacks](https://developer.android.com/reference/android/app/Application.html#registerActivityLifecycleCallbacks) to do the same? We should probably move lifecycle event logging there as well.

ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/FirebaseCrashLogging.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/util/Debug.kt
shobhitagarwal1612: Thanks Gino! It looks like this callback only gets notified for activity lifecycle changes, and not fragment.

ground/src/main/java/com/google/android/ground/util/Debug.kt
```diff
@@ -15,13 +15,19 @@
...
+    Timber.tag(className).v("Lifecycle event: $callingMethod")
+
+    if (callingMethod == "onResume()") {
+      FirebaseCrashLogger().setScreenName(className)
```
gino-m, line 29: What do you think about remaining to something like fragmentClass so that what this string refers to is more self documenting?

Also, are we sure the fragment's is always called last? Are there any other lifecycle objects which might be called after? Since there's no compile check for this, we may want to check the type of the instance to make sure we got the right one.

gino-m (changes requested):
One last bit, LG!

gino-m (approved):

Issue #2420, opened 2024-04-01
[Code health] Set up end to end testing (local web, mobile, and database emulation)


sufyanAbbasi:
Formulating an overall strategy to run end to end testing on the mobile and web app. Will need to configure something like Docker or Github Hosted Runners which will spin up a local web server + Selenium WebDriver, local Android emulator + app, and a set up local Firebase database/emulator then run through a set of core tests.

sufyanAbbasi:
Mirror: https://github.com/google/ground-platform/issues/1699

sufyanAbbasi:
TODOs:

- [ ] Fix broken sign-in procedure on emulator (the `USE_EMULATOR` flag should prevent the sign-in page from showing up)
- [ ] Add additional test coverage for other CUJs like offline imagery
- [ ] Add test coverage for Android lifecycle issues like screen rotation.

PR #2425, opened 2024-04-06
Add e2eTest module that submits data to a test survey

sufyanAbbasi:
Addresses #2420
See web equivalent: https://github.com/google/ground-platform/pull/1700

Adds a new test module, `e2eTest`, that uses [UI Automator](https://developer.android.com/training/testing/other-components/ui-automator) to drive a test survey created by the web e2e test.

Note: It must be run using the `localDebug` variant, since it relies on the local Firebase emulator and test survey created by the web e2e test. See instructions in https://github.com/google/ground-platform/pull/1700 to run the web test and emulator. Also note that the test will fail if started before the survey creation has finished.

To run the test, first run the web e2e test until the survey has been created, then right click on the `e2eTest` module and press `Run All`, or find the `SurveyRunnerTest` file and run it from there.
Screenshot 2024-04-05 at 6 00 39 PM

- [x] Signs in to local emulator.
- [x] Select the test survey.
- [x] Use location lock to zoom in.
- [x] Clicks on the ad-hoc LOI card.
- [x] Fill out survey data.
- [x] Submit.
- [x] Fill out the survey again at the new LOI location.
- [x] Runs using `.gradlew` via `./gradlew :e2eTest:connectedLocalDebugAndroidTest`, or with the emulator (assuming `ground-platform` is a sibling directory): `firebase emulators:exec './gradlew :e2eTest:connectedLocalDebugAndroidTest' --project ground-local-dev --ui --config ../ground-platform/firebase.local.json --import ../ground-platform/data/local --export-on-exit ../ground-platform/data/local`

PTAL @gino-m @shobhitagarwal1612 !


e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -0,0 +1,235 @@
...
+    zoomIntoLocation()
+    startAdHocLoiTask()
+    fillOutTaskData(isAdHoc = true, TEST_SURVEY_TASKS_ADHOC)
+    clickSubmissionConfirmationDone()
+    startPredefinedLoiTask()
+    fillOutTaskData(isAdHoc = false, TEST_SURVEY_TASKS_ADHOC.drop(1))
+    clickSubmissionConfirmationDone()
+  }
+
+  private fun launchGround() {
```
gino-m, line 68: I wonder if we can store these UI actions in their own files so they can be reused and to keep the test free from boiler plate?

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -0,0 +1,235 @@
+/*
```
gino-m, line 1: Can we reuse the `androidTest` package instead of creating a new one? Or are these tests somehow fundamentally different from normal instrumented tests?

gino-m (changes requested):
This is super cool, @sufyanAbbasi ! A few initial comments.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
sufyanAbbasi: I put them in their own module following recommendation from the [Android dev guide](https://developer.android.com/studio/test/advanced-test-setup#use-separate-test-modules-for-instrumented-tests):

> If you want to have a dedicated module for instrumented tests, to isolate the rest of your code from your tests, create a separate test module and configure its build similar to that of a library module.

The nice thing about this setup is that they won't accidentally be included in the main module's unit tests, since the tests don't work without the web emulator also running in the background, plus we don't have to keep track of filtering them out of the main module's test.

I don't know why, but they recommend putting the tests in `main` instead of `connectedAndroidTest` which would be customary for instrumented tests.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
sufyanAbbasi: I considered making an `AutomatorHelper` class, but it was annoying to hop between files since there was only one test. We should consider refactoring more functionality into the `AutomatorRunner` interface to share functionality between future classes, but for now, it's nicer to have everything in one file in top to bottom order of more to less abstract, e.g.:

```
    launchGround()
    signIn()
    selectTestSurvey()
    zoomIntoLocation()
    startAdHocLoiTask()
    fillOutTaskData(isAdHoc = true, TEST_SURVEY_TASKS_ADHOC)
    clickSubmissionConfirmationDone()
    startPredefinedLoiTask()
    fillOutTaskData(isAdHoc = false, TEST_SURVEY_TASKS_ADHOC.drop(1))
    clickSubmissionConfirmationDone()
```

gino-m (changes requested):
This is brilliant! 🌟 

Just a few nits, overall LGTM.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -0,0 +1,235 @@
...
+  private val close = stringResource(R.string.close)
+  private val dropPin = stringResource(R.string.drop_pin)
+  private val addPoint = stringResource(R.string.add_point)
+  private val complete = stringResource(R.string.complete_polygon)
+  private val capture = stringResource(R.string.capture)
+  private val camera = stringResource(R.string.camera)
+  private val ok = stringResource(R.string.ok)
+  private val next = stringResource(R.string.next)
+  private val done = stringResource(R.string.done)
+  private val save = stringResource(R.string.save)
```
gino-m, line 52: Idea: Since these are mostly (all?) used in calls to `By.text()`, can we add a helper, e.g. `byText()` which take the id and resolve the string resource directly? This way we don't need to maintain this block of assignments.

settings.gradle
```diff
@@ -1,6 +1,23 @@
...
+ * limitations under the License.
+ */
+
 plugins {
     id 'org.gradle.toolchains.foojay-resolver-convention' version '0.4.0'
 }
 
 include ':ground', ':sharedTest'
 
+include ':e2eTest'
```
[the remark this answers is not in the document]
sufyanAbbasi: Right! Done.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -0,0 +1,235 @@
...
+    if (!waitClickGone(By.clazz("android.widget.Button"), LONG_TIMEOUT)) {
+      fail("Failed to sign in.")
+    }
+  }
+
+  private fun selectTestSurvey() {
+    device.wait(Until.hasObject(By.text(selectSurveyTitle)), LONG_TIMEOUT)
+    val testSurveySelector =
+      By.clazz("androidx.cardview.widget.CardView")
+        .hasDescendant(By.clazz("android.widget.TextView").textContains(TEST_SURVEY_IDENTIFIER))
```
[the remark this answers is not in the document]
sufyanAbbasi: Amazing! I had to tweak to: `fun <T : Any> byClass(kclass: KClass<T>): BySelector = By.clazz(kclass.java.name)` since the direct to class version wasn't working.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
sufyanAbbasi: Good call! The `R.string.<id>` is already doing the heavy lifting of standardization. Done.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -0,0 +1,235 @@
...
+    val loiCollectDataButtonSelector =
+      By.text(collectData)
+        .hasAncestor(dataCollectionCardSelector.hasDescendant(By.text(newDataCollectionSite)))
+    if (!waitClickGone(loiCollectDataButtonSelector)) {
+      fail("Failed to start ad-hoc loi data collection.")
+    }
+  }
+
+  private fun startPredefinedLoiTask() {
+    val dataCollectionCardSelector = By.clazz("androidx.cardview.widget.CardView")
```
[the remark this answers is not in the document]
sufyanAbbasi: Good call, I kept the "selector" part since I wanted to be clear about the variable type. Done.

e2eTest/src/main/java/com/google/android/ground/e2etest/AutomatorRunner.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.google.android.ground.e2etest.TestConfig.LONG_TIMEOUT
+import com.google.android.ground.e2etest.TestConfig.SHORT_TIMEOUT
+
+interface AutomatorRunner {
+  var device: UiDevice
+
+  fun stringResource(
+    @StringRes resId: Int,
+    context: Context = InstrumentationRegistry.getInstrumentation().targetContext
+  ): String {
```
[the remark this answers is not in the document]
sufyanAbbasi: Ah, I should have noticed this one. Done.

gino-m (approved):
🎸 !!!

PR #2426, opened 2024-04-07
[Bug] Ensure firebase auth is always initialized before accessing it

### LINKED ISSUE #2335: Manually disabling location permissions then reopening app puts app in non-loaded state
**Describe the bug**
By manually removing app permissions and revisiting the app, the app doesn't load any data for some reason, even when the network is on.

**To Reproduce**
1. Ensure app has location permissions, click on a survey.
2. While in the survey map view, before clicking on "Collect", go into app settings and remove location permissions.
3. Then re-open the app using the "Open" button on the settings. 
4. See that app is in a weird state where nothing is loaded.

**Expected behavior**
App is loaded but with no location set.

**Actual behavior**
- No surveys or data are loaded.
- No user data is loaded in the sidebar menu.

**Screenshots**

Screenshot 2024-03-12 at 3 58 43 PM
Screenshot 2024-03-12 at 3 55 53 PM
Screenshot 2024-03-12 at 3 55 58 PM

shobhitagarwal1612:
Fixes #2335

When opening the app by following the repro steps as per #2335, the startup fragment is not invoked. Due to this, the user isn't logged in and the app appears to be un-initialized.

This PR fixes that by ensuring that sdk is always initialized irrespective of how the app is launched.

Verified by following the below steps:

2. Ensure app has location permissions, click on a survey.
3. While in the survey map view, before clicking on "Collect", go into app settings and remove location permissions.
4. Then re-open the app using the "Open" button on the settings.
5. See that app is properly loaded including user details and survey list.

@gino-m @sufyanAbbasi @scolsen PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/repository/UserRepository.kt
+++ b/ground/src/main/java/com/google/android/ground/repository/UserRepository.kt
@@ -40,11 +40,21 @@ constructor(
   private val networkManager: NetworkManager,
   private val surveyRepository: SurveyRepository,
   private val remoteDataStore: RemoteDataStore,
-) : AuthenticationManager by authenticationManager {
+) {
+
+  fun getSignInState() = authenticationManager.signInState
+
+  fun init() = authenticationManager.init()
+
+  fun signIn() = authenticationManager.signIn()
+
+  fun signOut() = authenticationManager.signOut()
+
+  suspend fun getAuthenticatedUser() = authenticationManager.getAuthenticatedUser()
 
   /** Stores the current user's profile details into the local and remote dbs. */
   suspend fun saveUserDetails() {
-    localUserStore.insertOrUpdateUser(authenticationManager.getAuthenticatedUser())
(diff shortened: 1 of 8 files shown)


gino-m (approved):
Nice fix, @shobhitagarwal1612 🎉 

I guess we still call this from the startup viewmodel to force the sign in screen if the user isn't already authenticated?

Also, I noticed `UserRepository` both extends and contains `AuthenticationManager`. Is this intentional? Can we simplify by just injecting w/o extending?

shobhitagarwal1612:
> Also, I noticed UserRepository both extends and contains AuthenticationManager. Is this intentional? Can we simplify by just injecting w/o extending?

Yes, we used delegation to expose authentication manager's methods via user repository. https://kotlinlang.org/docs/delegation.html

Will simplify it by injecting and creating one method for each.

gino-m:
Thanks!

PR #2427, opened 2024-04-07
Disable fallbackToDestructiveMigration()

### LINKED ISSUE #932: [Local Storage] Local db wiped when app is updated to a newer version
Will add further details here once we determine if this is needed.

shobhitagarwal1612:
Fixes #932

We can't reset DB version to 1 since it is already launched to play store. With this change, the app will always require migration instead of clearing the db when upgrading the db version

@gino-m  PTAL?


gino-m (approved):
Sgtm. Thanks for the fix, @shobhitagarwal1612

PR #2428, opened 2024-04-07
Skip offline area list screen when there's no offline imagery

### LINKED ISSUE #2342: [Offline map imagery] Skip directly to area selector screen when there's no offline imagery
If there are no offline areas we should skip this screen:

image

And go directly to:

image

shobhitagarwal1612:
Fixes #2342

@gino-m @scolsen  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
@@ -90,12 +90,13 @@ internal constructor(
     )
   }
 
+  private suspend fun getOfflineAreas() = offlineAreaRepository.offlineAreas().first()
+
   fun showOfflineAreas() {
     viewModelScope.launch {
-      val count = offlineAreaRepository.offlineAreas().first().count()
       navigator.navigate(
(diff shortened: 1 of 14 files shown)


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
```diff
@@ -88,7 +91,13 @@ internal constructor(
   }
 
   fun showOfflineAreas() {
-    navigator.navigate(HomeScreenFragmentDirections.showOfflineAreas())
+    viewModelScope.launch {
+      val count = offlineAreaRepository.offlineAreas().first().count()
```
gino-m, line 95: Would it be more idiomatic to use `isEmpty()` and invert the if statement below instead of `first().count() > 0`?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenViewModel.kt
shobhitagarwal1612: Done

gino-m (approved):
LGTM!

PR #2433, opened 2024-04-09
Refactor: Rename Response/Result -> TaskData

### LINKED ISSUE #2430: [Code health] Disambiguate `PhotoResponse` and `PhotoResult`
Currently we have `PhotoResponse` in the model and `PhotoResult` in the view.

IMO, `PhotoResponse` in the model should be renamed to `PhotoResult` for consistency with `GeometryTaskResult`, and `PhotoResult` in the view should be renamed just `Photo` or `PhotoData`.

@scolsen WDYT?

scolsen:
Clarifies how these types are used in capturing data from user completed tasks. In addition:

- Deletes `PhotoResult` we only use `PhotoTaskData` directly instead.
- Simplifies the logic in the PhotoTaskViewModel.
- Moves LatLngConverter into the geometry model.
- Moves all TaskData types into the model package.

Fixes #2430

@gino-m


gino-m (approved):
This is _much_ clearer, thank you @scolsen !

gino-m (approved):

PR #2436, opened 2024-04-10
[DRAFT] MediaMutations: Implement initial model and persistent types.

scolsen:
Adds basic types for representing media and media mutations separately from other mutations. Media mutations are dependent on submission mutations, but using a more fine-grained representation allows us to both operate on these mutations in a more granular manner and allows us to give more specific information to users.

Note that we do not use these types yet, they are only introduced here.

@gino-m, @sufyanAbbasi - FYI.

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
+++ b/ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.mutation
+
+import java.util.Date
+
+/**
+ * Represents a local change to media associated with a submission.
+ *
+ * These mutations are directly dependent upon another mutation type: SubmissionMutation. At
+ * present, any MediaMutation will only have relevance in the context of a submission. To this end,
+ * several properties are derived from the mutation's associated submission mutation.
+ */
+data class MediaMutation(
+  override val type: Type,
+  val mediaId: String,
+  val submissionMutation: SubmissionMutation,
+  override val id: Long? = null,
+  override val syncStatus: SyncStatus = SyncStatus.PENDING,
+  override val retryCount: Long = 0,
+  override val lastError: String = "",
+) : Mutation() {
+  override val surveyId: String = submissionMutation.surveyId
+  override val locationOfInterestId: String = submissionMutation.locationOfInterestId
+  override val clientTimestamp: Date = submissionMutation.clientTimestamp
+  override val userId: String = submissionMutation.userId
+}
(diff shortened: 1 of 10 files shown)


ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Represents a local change to media associated with a submission.
+ *
+ * These mutations are directly dependent upon another mutation type: SubmissionMutation. At
+ * present, any MediaMutation will only have relevance in the context of a submission. To this end,
+ * several properties are derived from the mutation's associated submission mutation.
+ */
+data class MediaMutation(
+  override val type: Type,
+  val mediaId: String,
+  val submissionMutation: SubmissionMutation,
```
gino-m, line 30: This would require `SubmissionMutations` to always be loaded and attached when deserializing `MediaMutations`. Is that the intent? If not we may just want to pass the required fields here.

ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
```diff
@@ -0,0 +1,40 @@
...
+  override val id: Long? = null,
+  override val syncStatus: SyncStatus = SyncStatus.PENDING,
+  override val retryCount: Long = 0,
+  override val lastError: String = "",
+) : Mutation() {
```
gino-m, line 35: Not sure where this should be represented so dropping here for now: at present there's only one type of MediaMutation: "upload" mutation (create?), but no delete update.

gino-m (commented):
LGTM! Two questions/suggestions, looks ready for review.

ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
scolsen: Yes, my thinking here was that a number of the mutation fields on this type `loi ID`, `survey` etc should be derived from the corresponding submission. If we want these fields available on this type, I think this is how we should do it since it makes the derivation possible and the dependence clear.

Otherwise, we could simply omit these fields from this type, but it would require either refactoring `Mutation` a bit or moving this type out of the Mutation type tree.
 
I did have a  tentative change a while back that reduced `Mutation` to its truly common parts `id, type, syncstate, retrycount`.... we'd then define loiId etc. as fields on the subtypes. This sort of feels correct to me, so I'd be cool with doing it that way too. wdyt?

ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
scolsen: Good point! Maybe we want a separate enum here. This might be more evidence that this type is not *really* a complete mutation type, and maybe we should just move it out of the subtype tree. though certain properties (rerties, sync status, id) are certainly shared.

gino-m: It may or may not be more elegant to model all remote changes as mutations - in this case the mutation is the CREATE a new media entity in data storage. If we were to all "edit" we could also have an "UPDATE" operation which would just be implemented as a DELETE and CREATE. I'd say let's go with the enum we have for now and keep this in the back of our minds in the next PRS?

scolsen: sgtm

ground/src/main/java/com/google/android/ground/model/mutation/MediaMutation.kt
gino-m: `SubmissionMutation` can also be associated with an `LoiMutation`, but we don't actually represent that in the model. If you think it wouldn't be too much of a burden to deserialize the dependent mutation before deserializing  `MediaMutations` then this ok, but I think it will add some overhead in terms of impl.

Speaking of which, perhaps we should link `LoiMutation` to the `Submission` one somehow so they can be grouped/shown together in the sync status screen?

gino-m:
Closing this for now, we can come back to it under the community project.

PR #2437, opened 2024-04-11
Show location lock button in offline area selector

### LINKED ISSUE #739: [Offline area download] Show GPS lock toggle button when selecting offline areas
Forked from #734 filed by @karistenneson.  

 @scolsen I think showing the same "location lock" toggle shown on the main map view would work here. Is that what you had in mind as well?

shobhitagarwal1612:
Fixes #739

Adds "location lock" button to offline area selector and some minor code cleanup.

Screenshot_20240411_104558

@gino-m  PTAL?


gino-m (changes requested):
Thank you @shobhitagarwal1612!

One suggestion: Can we make the fab in this UI smaller and move it to the inside of the inner rectangle so that the bottom text can remain centered?

@rawbzz in case you have better or more concrete advise on placement

shobhitagarwal1612:
Updated screenshot:
Screenshot_20240411_214248

gino-m (approved):
🦾 🚀 !

PR #2448, opened 2024-04-20
Improve offline download performance

### LINKED ISSUE #2355: [Download offline imagery] Download of 12 MB imagery painfully slow on 5G

shobhitagarwal1612:
Fixes #2355

This PR aims to reduce total time taken for downloading offline tiles.

Context:
1. For a given `bounds` (SW + NE), we generate a list of `MogTileRequest` 
2. Then for each `MogTileRequest` (synchronously), we try to fetch all `MogTile`
  a. Creating an input stream from the given `MogTileRequest`
  b. Then using offsets of for each tile from metadata, we read contents of each tile (synchronously) and emit the raw data
3. Finally, we save the data to local storage

We cannot speed up step 2.b without duplicating step 2.a as all read tile steps reuse the input steam.
Therefore, this PR attempts to make step 2 asynchronous using corountines.

Approaches that I tried:
 * Using IO dispatcher 
 * Using unconfined dispatcher
 * Using IO dispatcher with extra parallelism 
 * Moving tile write within tile read method
 * Making tile writes async

For each approach, I download offline tiles of various sizes and plotted the points in a sheet to understand which approach improves the speed for various scenarios.

Please see: https://docs.google.com/spreadsheets/d/1W5TlbYVFGOIqrkkcstVY5W3OPrQ1nLNjt8DbnjFBSH4

I found IO + 200 parallelism to improve the download speed the most.

@gino-m @sufyanAbbasi  PTAL?


shobhitagarwal1612:
Hey @gino-m! Based on our previous discussion, I fixed the bounds to a constant value (approx 5 MB total size) and experimented with different thread values with IO dispatcher.

### Time taken (at HEAD): 48551 ms

### Time taken with this change

| **Threads** | **Time taken (ms)** |
| ----|----|
| default |  52901 <-- slowest |
| 1 |  48613 |
| 5 |  44859 |
| 10 |  44305 |
| 20 |  50499 |
| 50 |  45997 |
| 100 |  48365 |
| 150 |  43139 |
| 200 |  41987 <-- fastest |

### Outcome
This PR improved the overall time taken for downloading the 5 mb sample offline tile when using 200 threads by **~6.5s** (13% faster)

ground/src/main/java/com/google/android/ground/ui/map/gms/mog/MogTileReader.kt
```diff
@@ -27,6 +27,7 @@ class MogTileReader(private val inputStream: InputStream, initialOffset: Long) {
   private var offset: Long = initialOffset
 
   fun readTiles(tiles: List<MogTileMetadata>): Flow<MogTile> = flow {
+    // Tiles must be read sequentially since input stream is shared among all.
```
gino-m, line 30: "input stream is shared among all" what? Not sure what this is referring to.

shobhitagarwal1612: Updated the comment

gino-m (approved):
Great work, @shobhitagarwal1612, esp. on tuning the threading params. I'm surprised we can make 200 parallel requests without running into issues, but if it works it works. Thanks for sorting this out!

Issue #1754, opened 2023-07-20
[Offline map imagery] Add unit tests for MogClient


gino-m:

gino-m:
Required for #2134

sufyanAbbasi:
See #1730 and #2115. We will want to configure the MogClient to pull directly from Firebase Cloud Storage instead of arbitrary URLs.

PR #2449, opened 2024-04-20
Inject MogClient instead of building a new instance each time

shobhitagarwal1612:
Towards #1754

This also speeds up "updateDownloadSize" task as the client is not generated on each map camera move

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/map/gms/mog/MogProviderModule.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/map/gms/mog/MogProviderModule.kt
@@ -25,13 +25,11 @@ import dagger.hilt.components.SingletonComponent
 @InstallIn(SingletonComponent::class)
 @Module
 class MogProviderModule {
-  @Provides
-  fun provideMogClient(remoteStorageManager: RemoteStorageManager): MogClient {
-    return MogClient(DEFAULT_MOG_COLLECTION, remoteStorageManager)
-  }
 
-  companion object {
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/ui/map/gms/mog/MogProviderModule.kt
```diff
@@ -0,0 +1,37 @@
...
+class MogProviderModule {
+  @Provides
+  fun provideMogClient(remoteStorageManager: RemoteStorageManager): MogClient {
+    return MogClient(DEFAULT_MOG_COLLECTION, remoteStorageManager)
+  }
+
+  companion object {
+    private val DEFAULT_MOG_SOURCES = Config.getMogSources(Config.DEFAULT_MOG_TILE_LOCATION)
+    private val DEFAULT_MOG_COLLECTION = MogCollection(DEFAULT_MOG_SOURCES)
+  }
```
anandwana001, line 36: if both the variables are private, then we can keep them outside the companion object, as these variables are not planned to be accessible from outside. 

WDYT?

shobhitagarwal1612: Makes sense! Done

shobhitagarwal1612:
@gino-m Gentle ping!

gino-m (approved):
👍 LGTM!

PR #2453, opened 2024-04-22
Initial impl of converter from Firestore document to protobuf

gino-m:
Towards #1748


gino-m:
@shobhitagarwal1612 @sufyanAbbasi PTAL?

ground/src/test/java/com/google/android/ground/persistence/remote/firebase/protobuf/FirestoreToProtobufExtTest.kt
shobhitagarwal1612, line 1: Minor suggestion: We can convert this into a parameterized unit test. See `TiffTagDataTypeTest.kt`

ground/src/main/proto/com/google/ground/job.proto
shobhitagarwal1612, line 1: How do you feel about putting all protos in a single file? Or will these eventually live outside of this project?

shobhitagarwal1612 (commented):
Looks great!

ground/src/test/java/com/google/android/ground/persistence/remote/firebase/protobuf/FirestoreToProtobufExtTest.kt
gino-m: I thought of that, but if we need to test for exceptions I thought it would be less readable. I can give it a try.

gino-m: The companion object with the data would then look something like this:

```kt

  companion object {
    @JvmStatic
    @Parameterized.Parameters(name = "{0}")
    fun data() =
      listOf(
        arrayOf("converts document id", document(id = "12345"), survey { id = "12345" }),
        arrayOf(
          "converts string fields",
          document(data = mapOf("title" to "something")),
          survey { title = "something" },
        ),
        arrayOf("ignores unknown fields", document(data = mapOf("foo" to "bar")), survey {}),
        arrayOf(
          "ignores bad type in string field",
          document(data = mapOf("title" to 123)),
          survey {},
        ),
        arrayOf(
          "converts map<string, Message> fields",
          document(data = mapOf("jobs" to mapOf("job123" to mapOf("name" to "A job")))),
          survey { jobs["job123"] = job { name = "A job" } },
        ),
        arrayOf("ignores bad type in map", document(data = mapOf("jobs" to 123)), survey {}),
        arrayOf(
          "converts nested objects",
          document(
            data = mapOf("jobs" to mapOf("job123" to mapOf("style" to mapOf("color" to "#112233"))))
          ),
          survey { jobs["job123"] = job { style { color = "#112233" } } },
        ),
      )
  }
}
```

Even without testing for errors, I'm not convinced grokking this array is more readable than having the test data be inlined. If we were able to use named args this would be a bit better, but JUnit's parameterized testing is not yet kotlinic.

shobhitagarwal1612: Thanks for trying it out. Sounds good to leave the tests as is.

gino-m: Update: I think by creating a helper function we can actually make this more readable:

```kt
    @JvmStatic
    @Parameterized.Parameters(name = "{0}")
    fun data() =
      listOf(
        testCase(desc = "converts document id", id = "12345", expected = survey { id = "12345" }),
        testCase(
          desc = "converts string fields",
          data = mapOf("title" to "something"),
          expected = survey { title = "something" },
        ),
        testCase(
          desc = "ignores unknown fields",
          data = mapOf("foo" to "bar"),
          expected = survey {},
        ),
        testCase(
          desc = "ignores bad type in string field",
          data = mapOf("title" to 123),
          expected = survey {},
        ),
        testCase(
          desc = "converts map<string, Message> fields",
          data = mapOf("jobs" to mapOf("job123" to mapOf("name" to "A job"))),
          expected = survey { jobs["job123"] = job { name = "A job" } },
        ),
        testCase(
          desc = "ignores bad type in map",
          data = mapOf("jobs" to 123),
          expected = survey {},
        ),
        testCase(
          desc = "converts nested objects",
          data = mapOf("jobs" to mapOf("job123" to mapOf("style" to mapOf("color" to "#112233")))),
          expected = survey { jobs["job123"] = job { style { color = "#112233" } } },
        ),
      )
```

Applying these changes now.

ground/src/main/proto/com/google/ground/job.proto
shobhitagarwal1612: That's just a personal preference to keep all related protos in a single file. I don't know if there are any guidelines around it.

gino-m: Not sure, Gemini (public version) makes a few good points:

Here's a breakdown of the pros and cons of using a single file vs multiple files for your protocol buffers, considering you have 3-4 root messages with nested messages:

**Single File**

**Pros:**

* **Simpler project structure:**  Keeping everything in one file can make your project easier to navigate, especially for smaller projects.
* **Easier to find message definitions:** All your message definitions are in one place, which can be helpful for quick lookups.

**Cons:**

* **Larger file size can lead to slower compilation times:** As the file grows larger, compilation times can increase.
* **Changes to a single message can cause entire file to be recompiled:** Even a small change to a message  can trigger a recompile of the entire file.

**Many Files (one per message)**

**Pros:**

* **Faster compilation times:** Only the files you modify are recompiled, leading to faster build times.
* **Easier to manage dependencies between messages:** It's easier to see which messages depend on each other when they are in separate files.

**Cons:**

* **More complex project structure:**  Managing multiple files can add complexity, especially for larger projects.
* **Harder to find all message definitions:**  You'll need to search across multiple files to find specific message definitions.


**Recommendation for your case:**

With 3-4 root messages and nested messages, using multiple files is likely beneficial. Here's why:

* The compilation time improvements from only recompiling modified files can be significant, especially as your project grows.
* Managing dependencies between messages is easier with separate files. With nested messages, this becomes even more relevant.

**Additional factors to consider:**

* **Project size:** If your project is very small, the simplicity of a single file might outweigh the benefits of separate files.
* **Team size:** With a larger team, maintaining a clear project structure with separate files becomes more important.


I hope this helps you decide which approach is best for your project!

shobhitagarwal1612 (approved):

ground/src/main/proto/com/google/ground/job.proto
shobhitagarwal1612: Another thing is that if we choose to keep separate files, then would we be doing it for a single field proto message as well (e.g. color / style)? In our case, I think we will not be having too many protos in the beginning. If it ever reaches to a state where it's growing too large, then we can consider splitting it into smaller files. Thoughts?

PR #2471, opened 2024-05-14
Define survey schema and related (de)serializers

gino-m:
Towards #2049.


gino-m:
Almost there. Just need to add more proto doc. Will follow up with LOI, Submission, and Geometry models in separate PRs.

ground/src/main/proto/com/google/ground/job.proto
```diff
@@ -19,11 +19,189 @@ syntax = "proto3";
...
-    string name = 2;
-    Style default_style = 3;
-    // TODO(#1748): Add `tasks` and `strategy` fields.
+    // Index of the job when shown in lists.
+    int32 index = 2;
+    // User-facing name of the job.
+    string name = 3;
+    // The style used when rendering geometries and other UI elements associated
+    // with this job.
+    Style default_style = 4;
```
sufyanAbbasi, line 33: nit: Should this just be `style` if there is no way to set a style to override it?

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/MessageLiteReflectionExt.kt
```diff
@@ -148,3 +149,12 @@ private fun KProperty<*>.isMessageFieldProperty(): Boolean =
...
 
 fun String.toSnakeCase() = replace(Regex("[A-Z]")) { "_${it.value}" }.lowercase()
+
+private fun getEnumValues(enumClass: KClass<out Enum<*>>): List<Enum<*>> =
+  enumClass.java.enumConstants.toList()
+
+@Suppress("UNCHECKED_CAST")
+fun <T : EnumLite> KClass<T>.findByNumber(number: Int): T? {
+  require(isSubclassOf(Enum::class))
+  return getEnumValues(this as KClass<Enum<*>>).find { (it as EnumLite).number == number } as? T
```
sufyanAbbasi, line 159: Not sure if this is the same number, but `getNumber()` seems to be implemented on the `EnumLite` interface, which should be more persistent between library upgrades.

https://protobuf.dev/reference/java/api-docs/com/google/protobuf/Internal.EnumLite.html#getNumber()

ground/src/main/proto/com/google/ground/job.proto
```diff
@@ -19,11 +19,189 @@ syntax = "proto3";
 option java_package = "com.google.android.ground.proto";
 option java_multiple_files = true;
 
-import "com/google/ground/style.proto";
-
+// A single job, defined by one or more tasks, and zero or more locations of
```
sufyanAbbasi, line 22: > zero or more locations of interest

This is a tad confusing since LOIs are not nested in `Job`, unlike `tasks`. Can you clarify the relationship between LOIs and Jobs?

ground/src/main/proto/com/google/ground/job.proto
```diff
@@ -19,11 +19,189 @@ syntax = "proto3";
...
+    }
+    // Defines to which entity data collected in this task is associated with.
+    // Defaults to LOCATION_OF_INTEREST_DATA.
+    DataCollectionLevel level = 5;
+
+
+    // Conditions which control whether this task is shown during data
+    // collection. All conditions must be evaluate to `true` for the task to
+    // be triggered.
+    repeated Condition conditions = 6;
```
sufyanAbbasi, line 82: Just for my understanding, it seems that we're simplifying the [currently implemented `Condition` logic](https://github.com/google/ground-android/blob/462d28b787ad9b49e3a970b0e0d933e9c0a89af2/ground/src/main/java/com/google/android/ground/model/task/Condition.kt#L31), which is fine. Currently we have a single Condition with a match type of three types: match one, any, or all, and then for each sub expression, an any, all or one option selected.

In this new model, we are hardcoding "all" match conditions then "any" sub choice. If I did my math right, I think we should be able to cover the "match_all" case by creating multiple conditions with one option in each condition, and the "match_any" case by creating one condition with multiple options in a single condition, and we'll forgo a "match_one" type.

ground/src/main/proto/com/google/ground/job.proto
```diff
@@ -19,11 +19,189 @@ syntax = "proto3";
...
+        // A condition is defined by one of the following constraints:
+        oneof condition_type {
+            // A condition which passes if any of the specified values are
+            //selected.
+            MultipleChoiceSelection multiple_choice = 1;
+        }
+    }
+
+    // A condition which passed if any of the specified values are selected.
+    message MultipleChoiceSelection {
```
sufyanAbbasi, line 201: This feels like it should be pluralized, since the condition accounts for multiple options, but the style guide probably would discourage non-repeated fields to be plural. Instead, maybe `MultipleChoiceCondition` to make clear the association between this message type and the Condition one?

Or consider nesting this proto.

https://cloud.google.com/apis/design/naming_convention#repeated_field_names

ground/src/main/proto/com/google/ground/job.proto
```diff
@@ -19,11 +19,189 @@ syntax = "proto3";
...
+        // Possible values for the `allowed_methods` field.
+        enum Method {
+            UNSPECIFIED_DRAW_GEOMETRY_METHOD = 0;
+            DROP_PIN = 1;
+            DRAW_AREA = 2;
+        }
+        // One or more methods which the user may use to complete the geometry
+        // task. If multiple methods are specified, data collectors are prompted
+        // which method should be used.
+        repeated Method allowed_methods = 1;
```
sufyanAbbasi, line 176: I like this idea, we should implement it!

ground/src/test/proto/test.proto
```diff
@@ -16,9 +16,19 @@
...
+    string id = 1;
+    map<string, NestedTestObject> obj_map = 2;
+}
+
+message NestedTestObject {
+    string name = 1;
+    DeepNestedTestObject other_thing = 2;
+}
+
+message DeepNestedTestObject {
```
sufyanAbbasi, line 32: nit: `DeeplyNestedTestObject`

sufyanAbbasi (approved):
This looks fantastic! One thing I would recommend for another PR is to set up a proto linter if you haven't done so already, but otherwise, I just have mostly FMI (for my information) questions and small nits.

ground/src/main/proto/com/google/ground/job.proto
gino-m: If we apply YAGNI aggressively then yes! I assumed we'd need dynamic styling at some point, but since we don't have it already I can rename, eso, he new Firestore representation lets us change names easily!

ground/src/main/proto/com/google/ground/job.proto
gino-m: There will be many LOIs in the db, so keeping them in one document wasn't really scalable, whereas there are a small number tasks so nesting under job is ok. Any ideas on how I can represent that here without it seeming too out of context?

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/MessageLiteReflectionExt.kt
gino-m: Yes, `.number` is the Kotlinic way of calling `.getNumber()`.

ground/src/main/proto/com/google/ground/job.proto
gino-m: Yes, I didn't include those since they're not used and there aren't any FRs requesting them. I also think, as you said, that the various options can be represented sufficiently in this model. That said, additive changes to this model with reasonable defaults shouldn't break old clients, and I think the current model allows for adding those later without any structural changes (moving protos around in hierarchy etc).

ground/src/main/proto/com/google/ground/job.proto
gino-m: Thanks! This is for #1744, which was filed to address the EUDR reporting of smaller plots, which the data collector can opt to report as a point rather than a polygon.

gino-m:
Thanks! Love the idea of a proto linter! I couldn't find one using Google style, perhaps `bufLint` would work? https://github.com/bufbuild/buf-gradle-plugin

ground/src/main/proto/com/google/ground/job.proto
gino-m: Since this message type in nested, the selection options are already plural and are namespaced as`Task.Condition.MultipleChoiceSelection.option_ids`. In that case do you think it's clear?

ground/src/test/proto/test.proto
gino-m: Okaayyy :)

ground/src/main/proto/com/google/ground/job.proto
sufyanAbbasi: Oh gosh! I'm sorry, I couldn't read the proto nesting... Looks good!

ground/src/main/proto/com/google/ground/job.proto
sufyanAbbasi: Maybe something like, "Locations of interest (LOIs) will be associated with the job by the job ID in a separate document" ?

gino-m: Sorry just saw this comment now, I'll address these in the new PR!

PR #2474, opened 2024-05-15
Implement remaining protobuf definitions

gino-m:
Towards #2049.

@sufyanAbbasi PTAL? Feel free to approve and merge this if you review it before #2471 gets merged.


ground/src/main/proto/com/google/ground/geometry.proto
```diff
@@ -0,0 +1,70 @@
...
+
+// A single point on the map.
+message Point {
+    // Required. The point's coordinates..
+    Coordinates coordinates = 1;
+}
+
+// A linear ring geometry.
+message LinearRing {
+    // Required. The coordinates of the ring's vertices.
```
sufyanAbbasi, line 54: Maybe worth mentioning that the first and last coordinate should be identical. Also below.

ground/src/main/proto/com/google/ground/loi.proto
```diff
@@ -0,0 +1,57 @@
...
+    string owner_uid = 5;
+
+    // Required. Info about who created this LOI and when.
+    AuditInfo created = 6;
+
+    // Info about who last modified this LOI and when. If unset, it is assumed
+    // the LOI was never modified after it was created.
+    AuditInfo last_modified = 7;
+
+    // A user-defined ID for this location of interest.
```
sufyanAbbasi, line 50: Maybe "name" or "description" instead, to connote human readability? I wouldn't want it being used accidentally as a key. Ok to keep as is if that's what we have right now.

ground/src/main/proto/com/google/ground/submission.proto
```diff
@@ -0,0 +1,130 @@
...
+
+// Data provided by data collectors for a specific task and LOI.
+message TaskData {
+    // Required. A system-defined ID.
+    string id = 1;
+
+    // Required. ID of the task to which this task data refers.
+    string task_id = 2;
+
+    // Indicates this task was skipped pr
```
sufyanAbbasi, line 59: Typo?

sufyanAbbasi (approved):
Thank you! Just small edits to comments and such, so approving now!

ground/src/main/proto/com/google/ground/geometry.proto
```diff
@@ -0,0 +1,70 @@
...
+    repeated Coordinates coordinates = 1;
+}
+
+// A polygon geometry.
+message Polygon {
```
sufyanAbbasi, line 59: Technically there should be a boolean or something for "even-odd" or "left-inside" but that's probably not necessary if going by GeoJSON spec.
https://en.wikipedia.org/wiki/Point_in_polygon

ground/src/main/proto/com/google/ground/audit_info.proto
```diff
@@ -0,0 +1,41 @@
...
+
+    // Required. Time and date reported by the server.
+    google.protobuf.Timestamp server_timestamp = 3;
+
+    // The user-facing name of the user.
+    string display_name = 4;
+
+    // URL of the user's profile picture.
+    string photo_url = 5;
+
```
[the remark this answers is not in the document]
gino-m: Done.

ground/src/main/proto/com/google/ground/audit_info.proto
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+syntax = "proto3";
+
+option java_package = "com.google.android.ground.proto";
+option java_multiple_files = true;
+
+import "google/protobuf/timestamp.proto";
+
+// Audit info about "who" performed a particular action and "when".
```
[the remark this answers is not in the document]
gino-m: Good idea, done.

ground/src/main/proto/com/google/ground/audit_info.proto
```diff
@@ -0,0 +1,41 @@
...
+message AuditInfo {
+    // Required. The uid of the user performing the action.
+    string uid = 1;
+
```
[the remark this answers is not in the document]
gino-m: Good point! I would rather define these protos as Firestore agnostic if possible, so replacing with `user_id` instead.

ground/src/main/proto/com/google/ground/geometry.proto
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+syntax = "proto3";
+
+option java_package = "com.google.android.ground.proto";
+option java_multiple_files = true;
+
+// An arbitrary geometry, represented using WGS 84 Web Mercator coordinate
+// system. Designed for compatibility with Google Maps Platform and GeoJSON
+// format.
```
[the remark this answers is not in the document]
gino-m: Done.

ground/src/main/proto/com/google/ground/loi.proto
gino-m: This is used to store the `fid` / `feature_id` / `system:index` exported by different systems, as opposed to `name` which is a GeoJSON property. Which reminds me, I need to add those as well! I could also call it "feature_id", but we don't have the concept of "feature" anywhere anymore, so that might be confusing. Open to suggestions.

gino-m: How about something like `custom_tag`? or `tag`?

ground/src/main/proto/com/google/ground/submission.proto
gino-m: Oops.. the `skipped` field got deleted for some reason, adding it back!

ground/src/main/proto/com/google/ground/geometry.proto
gino-m: Another good point! Added some clarification to `LinearRing` itself. Re winding, the spec says:

```
Note: the [[GJ2008](https://datatracker.ietf.org/doc/html/rfc7946#ref-GJ2008)] specification did not discuss linear ring winding
   order.  For backwards compatibility, parsers SHOULD NOT reject
   Polygons that do not follow the right-hand rule.
```

We don't check/correct the orientation of user-drawn or imported polygons, so perhaps given the spec we leave this as underspecified for now? Note that shell and holes are represented separately, so there should be no ambiguity about which is which.

gino-m:
@sufyanAbbasi I'll merge for now, I can apply any late comments in the next PR. Thanks!

PR #2475, opened 2024-05-18
Allow storing altitude and accuracy in Firestore for CaptureLocationTaskData

### LINKED ISSUE #2473: [Capture location] Altitude and accuracy probably not stored in Firebase
This code is not reachable: 

https://github.com/google/ground-android/blob/462d28b787ad9b49e3a970b0e0d933e9c0a89af2/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/SubmissionMutationConverter.kt#L89-L91

...since `CaptureLocationTaskData` extends `GeometryTaskData`, which is handled in the previous `is` statement.

`CaptureLocationTaskData` doesn't need to extend `GeometryTaskData`, it can be an independent type with similar fields.

@shobhitagarwal1612 Can you take a look?

shobhitagarwal1612:
Fixes #2473

Moves the CaptureLocationTaskData case before GeometryTaskData to correctly convert the object to firestore map. Also adds missing test coverage for capture location task, time task and data task.

@gino-m @sufyanAbbasi  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/SubmissionMutationConverter.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/SubmissionMutationConverter.kt
@@ -83,11 +83,11 @@ internal object SubmissionMutationConverter {
       is DateTaskData -> {
         taskData.date
       }
-      is GeometryTaskData -> {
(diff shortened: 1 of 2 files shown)


gino-m:
Sorry for the delay! I'll be under water till Tuesday - @sufyanAbbasi Could you PTAL?

sufyanAbbasi (approved):
Looks good to me!

PR #2476, opened 2024-05-18
Merge multiple ui state flows into a single flow in SurveySelectorFragment

### LINKED ISSUE #2081: [Code health] Refactor SurveySelectorFragment
Merge the two flows, survey list & survey list state into one.

shobhitagarwal1612:
Fixes #2081

This PR aims at simplifying the way we are updating the UI in survey selector fragment. Currently, we have 3 separate flows for toggling progress dialog, survey list and errors. In this PR, we have merged all three flows into a single flow `UiState`  

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
@@ -23,7 +23,6 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.PopupMenu
 import android.widget.Toast
-import androidx.lifecycle.lifecycleScope
 import com.google.android.ground.R
 import com.google.android.ground.databinding.SurveySelectorFragBinding
 import com.google.android.ground.model.SurveyListItem
@@ -32,7 +31,6 @@ import com.google.android.ground.ui.common.BackPressListener
 import com.google.android.ground.util.visibleIf
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.flow.filterNotNull
-import kotlinx.coroutines.launch
 
 /** User interface implementation of survey selector screen. */
 @AndroidEntryPoint
@@ -70,8 +68,10 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
   }
 
   private fun handleSurveyListUpdated(surveys: List<SurveyListItem>) {
-    binding.container.visibleIf(surveys.isNotEmpty())
-    binding.emptyContainer.visibleIf(surveys.isEmpty())
+    with(binding) {
+      container.visibleIf(surveys.isNotEmpty())
+      emptyContainer.visibleIf(surveys.isEmpty())
+    }
     adapter.updateData(surveys)
   }
 
--- a/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
@@ -30,7 +30,6 @@ import com.google.android.ground.ui.common.AbstractFragment
 import com.google.android.ground.ui.common.BackPressListener
 import com.google.android.ground.util.visibleIf
 import dagger.hilt.android.AndroidEntryPoint
-import kotlinx.coroutines.flow.filterNotNull
 
 /** User interface implementation of survey selector screen. */
 @AndroidEntryPoint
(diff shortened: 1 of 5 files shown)


ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -55,38 +53,41 @@ internal constructor(
   private val userRepository: UserRepository,
 ) : AbstractViewModel() {
 
-  // TODO(#2081): Expose non-mutable state flow.
-  val surveyListState: MutableStateFlow<State?> = MutableStateFlow(null)
-  private val _errorFlow: MutableSharedFlow<Int?> = MutableSharedFlow()
-  val errorFlow: Flow<Int> = _errorFlow.asSharedFlow().filterNotNull()
+  private val _uiState: MutableStateFlow<UiState?> = MutableStateFlow(null)
+  val uiState: StateFlow<UiState?> = _uiState.asStateFlow()
```
anandwana001, line 57: rather than null as the initial value, we can keep some default `UiState`? WDYT?

ground/src/main/java/com/google/android/ground/ui/surveyselector/UiState.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  data object FetchingSurveys : UiState()
+
+  data class SurveyListAvailable(val surveys: List<SurveyListItem>) : UiState()
+
+  data object ActivatingSurvey : UiState()
+
+  data object SurveyActivated : UiState()
+
+  data class Error(val errorResId: Int) : UiState()
```
anandwana001, line 30: how about naming these objet and class in such a way where it can represent a state as it is part of UiState? 

example - `FetchingSurveys` => `FetchSurveys`, sounds more actionable, WDYT?

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -43,18 +46,35 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
...
     adapter = SurveyListAdapter(viewModel, this)
-    // TODO(#2081): Merge the survey list flow and survey list state flow into a single stream.
-    lifecycleScope.launch { viewModel.getSurveyList().collect { adapter.updateData(it) } }
-    lifecycleScope.launch {
-      viewModel.surveyListState.collect { state -> state?.let { handleSurveyListState(it) } }
-    }
-    lifecycleScope.launch {
-      viewModel.errorFlow.collect {
-        Toast.makeText(requireContext(), it, Toast.LENGTH_SHORT).show()
+    lifecycleScope.launch { viewModel.uiState.filterNotNull().collect { updateUI(it) } }
```
gino-m, line 49: Nit: Not sure if Kotlin follows the Java naming guidelines 100%, but if that's the case `UI` should written as `Ui` instead.

ground/src/main/java/com/google/android/ground/ui/surveyselector/UiState.kt
gino-m: I think gerund is ok here, if the state is emphasizing something that's ongoing.

Since the state is specified to the survey list, I think we can omit "Survey" simply and say `LoadingList`.

Also, `SurveyActivated` sounds more like an event than a UI state. What does it represent? 

On that note, please add ktdoc to describe what each of these states represents.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -43,18 +46,35 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
...
+  }
```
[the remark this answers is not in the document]
shobhitagarwal1612: Nice catch! Fixed

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -43,18 +46,35 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
...
+      is UiState.Error -> {
+        dismissProgressDialog()
+        Toast.makeText(requireContext(), uiState.errorResId, Toast.LENGTH_SHORT).show()
       }
     }
   }
 
+  private fun handleSurveyListUpdated(surveys: List<SurveyListItem>) {
+    binding.container.visibleIf(surveys.isNotEmpty())
+    binding.emptyContainer.visibleIf(surveys.isEmpty())
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done.

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
shobhitagarwal1612: done

gino-m (approved):

PR #2477, opened 2024-05-19
Use UiState in DataCollectionFragment

shobhitagarwal1612:
This is similar to https://github.com/google/ground-android/pull/2476. This allows hiding internals of the view model to fragment and exposing it via UiState. This makes the class more robust and manageable.

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
@@ -34,16 +34,18 @@ import com.google.android.ground.databinding.DataCollectionFragBinding
 import com.google.android.ground.model.task.Task
 import com.google.android.ground.ui.common.AbstractFragment
 import com.google.android.ground.ui.common.BackPressListener
+import com.google.android.ground.ui.common.EphemeralPopups
 import com.google.android.ground.ui.common.Navigator
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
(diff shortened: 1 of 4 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -102,6 +101,16 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
...
     )
+
+    lifecycleScope.launch { viewModel.uiState.filterNotNull().collect { updateUI(it) } }
+  }
+
+  private fun updateUI(uiState: UiState) {
+    when (uiState) {
+      is UiState.TaskListAvailable -> loadTasks(uiState.tasks, uiState.taskPosition)
+      is UiState.TaskUpdated -> onTaskChanged(uiState.taskPosition)
+      is UiState.Error -> popups.ErrorPopup().show(uiState.message)
```
gino-m, line 112: Showing an error popup is an event, not a state, since it's fired once then forgotten. To avoid the error reappearing in each config change or other lifecycle event, I would model this as a hot non-replaying stream instead (Coroutine).

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
shobhitagarwal1612: Removed error from UiState

gino-m (approved):

PR #2487, opened 2024-06-05
Add Github Actions to run end-to-end tests, and also hermetically

sufyanAbbasi:
Addresses #2420

Sets up Github Actions to be able to run end to end test against the `ground-platform` actions, and to define the `submit-test` action which will run an Firebase and Android emulator test against a local copy of the Firebase data hermetically.

- [x] Allows for trigger of end to end tests by commenting `/e2eTests` in a PR
- [x] Allows for trigger of just the submit survey response test by commenting `/submitTest`
- [x] Prints logcat output and also saves screenshot as an artifact when encountering an issue 
- [ ] TODO: The submit test is currently broken due to emulator/build configuration issues, will need to spend more time to fix it.

See test PR for run invocations: https://github.com/sufyanAbbasi/ground-android/pull/1

@gino-m  PTAL?


.github/actions/submit-test/action.yml
```diff
@@ -0,0 +1,179 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+name: Submit to survey
+
+inputs:
+  android-repository:
+    description: 'ground-android repository under test'
+    default: sufyanAbbasi/ground-android
```
gino-m, line 19: Did you mean to update this to google/?

Does this also work in forks?

.github/actions/submit-test/action.yml
```diff
@@ -0,0 +1,179 @@
...
+
+  platform-repository:
+    description: 'ground-platform repository under test (if applicable)'
+    default: sufyanAbbasi/ground-platform
```
gino-m, line 23: Same here

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,66 @@
+# Copyright 2024 The Ground Authors.
```
gino-m, line 1: In the Android repo we only have one author at the moment, so this would just be "Google LLC" for now. Same throughout.

e2eTest/build.gradle
```diff
@@ -43,15 +46,19 @@ android {
     productFlavors {
         local {
             dimension "backend"
+            buildConfigField "boolean", "USE_EMULATORS", "true"
```
gino-m, line 49: Ahh.. this is what was missing!

e2eTest/build.gradle
```diff
@@ -43,15 +46,19 @@ android {
...
         }
         dev {
             dimension "backend"
+            buildConfigField "boolean", "USE_EMULATORS", "false"
```
gino-m, line 53: Can we remove other flavors here, since they're not intended to be used?

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -65,25 +83,39 @@ class SurveyRunnerTest : AutomatorRunner {
 
     // Wait for the app to appear.
     if (!device.wait(Until.hasObject(By.pkg(GROUND_PACKAGE).depth(0)), LONG_TIMEOUT)) {
+      takeScreenshot("${javaClass.simpleName}_${nameRule.methodName}")
```
gino-m, line 86: `${javaClass.simpleName}_${nameRule.methodName}` is used a few times.. you might be able to extract it into to an `inline` helper, eg `this.getScreenshotName()`.

gino-m (changes requested):
🏅 !!! Very cool the 📸 work as well!

.github/actions/submit-test/action.yml
sufyanAbbasi: > Did you mean to update this to google/?

Whoops, this is for testing purposes since the main repo doesn't have the actions defined yet. Also updated in the other PR.

> Does this also work in forks?

If you want to end to end with two forks, then no, but most likely you'll want to test the currently forked repo against the prod other repo, aka the google one.

.github/workflows/test-e2e.yml
sufyanAbbasi: Ah, gotcha. I must have copied it over from the ground-platform. Done.

.github/actions/submit-test/action.yml
sufyanAbbasi: Done.

e2eTest/build.gradle
sufyanAbbasi: Yes! It was quite the find.

e2eTest/build.gradle
sufyanAbbasi: I tried this out when I was constructing this file... the weird thing is that the build fails when these other flavors are not defined. Same as in `/sharedTest`

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
sufyanAbbasi: Done.

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
```diff
@@ -260,12 +259,14 @@ class SurveyRunnerTest : AutomatorRunner {
...
     if (device.wait(Until.hasObject(byText(R.string.save)), SHORT_TIMEOUT) == null) {
-      takeScreenshot("${javaClass.simpleName}_${nameRule.methodName}")
+      takeScreenshot(getScreenshotName())
       fail("Failed to find loi name popup")
     }
     enterText("An loi name")
     waitClickGone(byText(R.string.save))
   }
+
+  private fun getScreenshotName() = "${javaClass.simpleName}_${nameRule.methodName}"
```
gino-m, line 271: I was going to suggest this but I didn't realize this actually gets the test method name. Cool!

gino-m: Stupid question - if takeScreenshot() always has the same screenshot name logic, we can calculate the name there. The noargs takeScreenshot() would collide with the imported one; prob a good idea to rename in any case to avoid confusion between the helper and the actual imported API method.

gino-m (changes requested):
One more nit, sorry!

e2eTest/src/main/java/com/google/android/ground/e2etest/SurveyRunnerTest.kt
sufyanAbbasi: That's true... Done.

gino-m (approved):
👏 💯 !

PR #2488, opened 2024-06-06
Remove upload sync max retry counts

### LINKED ISSUE #154: [Offline support] App doesn't sync data after max retries of sync errors

sufyanAbbasi:
Fixes #154 

- Allows sync operation to retry forever and relies on exponential backoff to prevent bogging down the network worker thread.

- Verified by throwing an exception in the `processMutations` routine and verified that the retries happen via exponential backoff.

@gino-m @scolsen PTAL!


gino-m (approved):
One nit, otherwise LGTM!

ground/src/main/java/com/google/android/ground/Config.kt
```diff
@@ -44,6 +44,9 @@ object Config {
    */
   const val CLUSTERING_ZOOM_THRESHOLD = 14f
 
+  /** Maximum number of attempts for retrying unsuccessful syncs. */
```
gino-m, line 47: It looks like this PR is adding max retry, rather than removing it.. or did I miss something?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/Config.kt
sufyanAbbasi: Ah, sorry about the confusion. Please check to see if the diff view is against HEAD since I accidentally also removed the survey sync retry one in the first pass, then put it back.

ground/src/test/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorkerTest.kt
```diff
@@ -138,37 +138,36 @@ class LocalMutationSyncWorkerTest : BaseHiltTest() {
...
-        val result = createAndDoWork(context, TEST_LOI_ID)
-
-        assertThat(result).isEqualTo(retry())
-        assertMutationsState(
-          failed = 2,
-          retryCount = listOf(i, i),
-          lastErrors = listOf(ERROR_MESSAGE, ERROR_MESSAGE),
-        )
-      }
+  fun `Worker retries indefinitely`() = runWithTestDispatcher {
```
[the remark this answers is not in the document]
sufyanAbbasi: Oh right! That was late night brain. I fixed the test to make a little more sense.

gino-m (approved):
`while (true) { If at first you don't succeed... }` 🥳

PR #2489, opened 2024-06-06
Filter saved data to those present in the task sequence

### LINKED ISSUE #2351: [Conditional Tasks] Edgecase: data will be saved for hidden tasks by hacking the flow.
**Describe the bug**
See https://github.com/google/ground-android/issues/1955#issuecomment-1966902122
When implementing conditional tasks in https://github.com/google/ground-android/pull/2255 I realized that I never implemented that I never implemented the spec that:

> only "valid" data should be captured.

**To Reproduce**
1. Set up a conditional task (see https://github.com/google/ground-android/pull/2255)
2. Select options in the flow to make a later task show.
3. Get to the later task and enter a value.
4. Go back to the previous task and make the latter task hide.
5. Go through the flow and submit.
6. See that a value was saved for the hidden task.

**Expected behavior**
The hidden tasks should not have their values saved.

**Actual behavior**
The hidden tasks have their values saved if you are able to get to them during the flow.

sufyanAbbasi:
Fixes #2351

Filters saved deltas to only tasks returned by the task sequence.

- [x] Ensures that only data from tasks which are validated by conditions are saved to the deltas list. This prevents hidden task data from being saved, for example, if a condition is unfulfilled by hitting previous.

- [x] Tested app locally by running a conditional survey and hacking the survey locally with breakpoints.

@shobhitagarwal1612 PTAL!


gino-m (approved):
LGTM! Thanks for the fix! 🔨

sufyanAbbasi:
> Can we add some unit test coverage for this?

I'm actually having a really rough time with this, since there is no `DataCollectionViewModelTest`, I wasn't able to figure out how to get the `SavedStateHandle` dependency injected. There is a `DataCollectionFragmentTest` that I could modify, but it's a tad difficult to figure out how to get the MultipleChoice selection to work since we're just testing the basic text inputs there. I will schedule some time with you, perhaps, to work through them!

shobhitagarwal1612:
Sounds good to discuss it over a call. FYI, we already have test coverage for data collection fragment. So we don't need to mock or manually inject the view model's dependencies.

sufyanAbbasi:
Nevermind, I got the test to work!

shobhitagarwal1612 (approved):

PR #2495, opened 2024-06-15
Cleanup SubmissionRepository

shobhitagarwal1612:
Removes unnecessary creation of `submission` object when attempting to save submission.

@sufyanAbbasi  @gino-m  PTAL?


gino-m (changes requested):
Hi Shobhit, thanks for starting work on this. I think replacing our model objects with Protos might be a huge effort, and in some cases where we need additional state, may require some wrapper classes or similar. On our call yesterday I was suggesting just converting between Protos and model objects on load/save from remote repo. Does that make sense? Any ideas or concerns with that approach?

shobhitagarwal1612:
I thought we had to replace all data classes. Thanks for clarifying it. However, this change still holds true as it simplifies the submission repository little bit.

shobhitagarwal1612:
updated PR title and description to reflect that this is related to cleanup and not protobuf migration

gino-m (approved):
Thanks for the cleanup!

PR #2496, opened 2024-06-15
Handle exception when refreshing user's profile

shobhitagarwal1612:
This prevents app from crashing if the remote call to update user's profile fails.

@gino-m @sufyanAbbasi  PTAL?


gino-m (approved):
👍🏼👍🏼!

PR #2508, opened 2024-06-21
Moves creation of LOI to LoiRepository

shobhitagarwal1612:
Moves logic for creation of LOI to LoiRepository. Also, removes the creation of intermediate state `LocationOfInterest` and directly generates the mutation.

This is similar to https://github.com/google/ground-android/pull/2495 where the goal is to simplify the save LOI flow by reducing the number of steps involved.

@gino-m @sufyanAbbasi  PTAL?


gino-m:
Thank you @shobhitagarwal1612 for the update!

gino-m (approved):

PR #2518, opened 2024-06-26
Added LineString Unit Test

### LINKED ISSUE #2517: [Test Coverage] Add Unit Test of LineString
Add Unit Test of data class LineString

anandwana001:
Fixes #2517 

Added two remaining test cases for `LineString` data class

Screenshot 2024-06-26 at 10 41 24

@shobhitagarwal1612  PTAL?

### DIFF
--- a/ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
+++ b/ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
@@ -74,7 +74,7 @@ class GeometryTest {
   @Test
   fun testPointArea() {
     val point = point(x, y)
-    assertThat(POINT_AREA).isEqualTo(point.area)
+    assertThat(ZERO_AREA).isEqualTo(point.area)
   }
 
   @Test
@@ -123,6 +123,17 @@ class GeometryTest {
     assertThat(LineString(CLOSED_LOOP).isClosed()).isTrue()
   }
 
+  @Test
+  fun testCenterOfLineString() {
+    assertThat(LineString(CLOSED_LOOP).center()).isEqualTo(COORDINATE_2)
+    assertThat(LineString(OPEN_LOOP).center()).isEqualTo(COORDINATE_2)
+  }
+
+  @Test
+  fun testAreaOfLineString() {
+    assertThat(LineString(CLOSED_LOOP).area).isEqualTo(ZERO_AREA)
+  }
+
   private fun point(x: Double, y: Double) = Point(Coordinates(x, y))
 
   private fun linearRing(path: Path) = LinearRing(toCoordinateList(path))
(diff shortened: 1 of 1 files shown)


ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
```diff
@@ -74,7 +74,7 @@ class GeometryTest {
   @Test
   fun testPointArea() {
     val point = point(x, y)
-    assertThat(POINT_AREA).isEqualTo(point.area)
+    assertThat(ZERO_AREA).isEqualTo(point.area)
```
shobhitagarwal1612, line 77: Can we change the order of actual result and expected result? Also, let's hardcode the value 0 instead of creating a variable for simplicity. Something like;

```
assertThat(point.area).isEqualTo(0.0)
```

ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
```diff
@@ -123,6 +123,17 @@ class GeometryTest {
...
 
+  @Test
+  fun testCenterOfLineString() {
+    assertThat(LineString(CLOSED_LOOP).center()).isEqualTo(COORDINATE_2)
+    assertThat(LineString(OPEN_LOOP).center()).isEqualTo(COORDINATE_2)
+  }
+
+  @Test
+  fun testAreaOfLineString() {
+    assertThat(LineString(CLOSED_LOOP).area).isEqualTo(ZERO_AREA)
```
shobhitagarwal1612, line 134: Same as above

shobhitagarwal1612 (changes requested):

ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
anandwana001: Done

ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
anandwana001: done

ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
```diff
@@ -68,13 +68,13 @@ class GeometryTest {
   @Test
   fun testPointIsEmpty() {
     val point = point(x, y)
-    assertThat(POINT_IS_EMPTY).isEqualTo(point.isEmpty())
+    assertThat(false).isEqualTo(point.isEmpty())
```
shobhitagarwal1612, line 71: Can we reverse the order here as well?

anandwana001: done

shobhitagarwal1612 (approved):

PR #2525, opened 2024-06-28
Added unit test for Polygon and Multipolygon

### LINKED ISSUE #2524: [Code health] Add Unit Test for Polygon and Multipolygon
Add Unit Test for Polygon and Multipolygon

anandwana001:
Fixes #2524 

Added unit test for Polygon and Multipolygon

@sufyanAbbasi @shobhitagarwal1612 PTAL?

Screenshot 2024-06-28 at 13 56 57
Screenshot 2024-06-28 at 13 56 46


ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
```diff
@@ -85,13 +85,49 @@ class GeometryTest {
...
+    assertThat(multiPolygon.center())
+      .isEqualTo(Coordinates(lat=-89.62333049749999, lng=41.89413522))
+  }
+
+  @Test
+  fun testMultiPolygonIsEmpty() {
+    val multiPolygon = multiPolygon(polygon(path1, path2), polygon(path3, path4))
+
+    assertThat(multiPolygon.isEmpty()).isEqualTo(false)
+  }
```
shobhitagarwal1612, line 129: Can we add a test case when isEmpty returns true?

shobhitagarwal1612 (changes requested):

ground/src/test/java/com/google/android/ground/model/geometry/GeometryTest.kt
anandwana001: yes, we can, but what I checked is it requires a complex params to pass to first make it all polygons non empty.
Will be taking this in next pr.

shobhitagarwal1612 (approved):

PR #2527, opened 2024-06-28
Added Unit Test for MapStateRepository

### LINKED ISSUE #2526: [Code health] Add unit test for MapStateRepository
Add unit test for MapStateRepository

anandwana001:
Fixes #2526 

40% -> 58% Code Coverage

@sufyanAbbasi @shobhitagarwal1612  PTAL?
Should we write test for the delegate property as well? I think the variables setter and getter are already tested and the remaining can be skip, but would like to know your thoughts on this.

Screenshot 2024-06-28 at 15 21 01


shobhitagarwal1612 (approved):

ground/src/test/java/com/google/android/ground/repository/MapStateRepositoryTest.kt
```diff
@@ -30,6 +33,7 @@ import org.robolectric.RobolectricTestRunner
 class MapStateRepositoryTest : BaseHiltTest() {
 
   @Inject lateinit var mapStateRepository: MapStateRepository
+  @Inject lateinit var localValueStore: LocalValueStore
 
   @Test
   fun getMapType_whenNotSet_returnsDefault() {
```
sufyanAbbasi, line 39: Can convert these to string function names and below too! Thank you!

sufyanAbbasi (approved):

sufyanAbbasi:
> Should we write test for the delegate property as well?  I think the variables setter and getter are already tested and the remaining can be skip

I think your analysis is correct here, we don't need to unit test those since they are Kotlin features we can assume to just work.

Thank you so much!

ground/src/test/java/com/google/android/ground/repository/MapStateRepositoryTest.kt
anandwana001: done

shobhitagarwal1612:
@anandwana001 Can you please ensure that the checks are passing locally as well?

```
* What went wrong:
Execution failed for task ':workspace:ground:ktfmtCheckTest'.
> [ktfmt] Found 1 files that are not properly formatted:
  src/test/java/com/google/android/ground/repository/MapStateRepositoryTest.kt
```

anandwana001:
Passing locally, running again now

PR #2533, opened 2024-07-01
Added unit test for TaskNotFoundException

### LINKED ISSUE #2532: [Code health] Add Unit Test for TaskNotFoundException
Add Unit Test for TaskNotFoundException

anandwana001:
Fixes #2532

Screenshot 2024-07-02 at 09 34 13

@shobhitagarwal1612 @sufyanAbbasi  PTAL?


ground/src/test/java/com/google/android/ground/model/job/TaskNotFoundExceptionTest.kt
```diff
@@ -0,0 +1,34 @@
...
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+class TaskNotFoundExceptionTest : BaseHiltTest() {
+
+  @Test
+  fun testCustomLocalizedMessage() {
+    val obj = Job.TaskNotFoundException(taskId = "1")
+    assertThat(obj.localizedMessage).isEqualTo("unknown task 1")
+  }
+}
```
shobhitagarwal1612, line 34: I don't feel we need a specific test class for the exception. Instead, we should look into how to add coverage for this class using user flows.

anandwana001: ok, this class was mentioned in the given test coverage list.


> how to add coverage for this class using user flows.

I think an independent is useful to confirm whether the custom message is correct or not. No matter which user flow is uses.

shobhitagarwal1612: The custom message is not user facing. @sufyanAbbasi Wdyt?

sufyanAbbasi: My apologies on this, I did include this on the list because it showed up in the test coverage class list but it does look like something we can probably skip, since the code under test is a one-liner:

```
class TaskNotFoundException(taskId: String) : Throwable(message = "unknown task $taskId")
```

I will look into making sure that it can be removed from the test coverage class list. If I can do that, then we can go ahead and remove this one from the list. If that's not possible, we can keep the test in to satisfy the test coverage runner.

anandwana001: ok, IMO it's ok to have a test, no matter class is a one liner, just to save time in future and if the custom message is changed accidently, the test file will be our reminder. 

Will close the PR as we are not planning to have it.

shobhitagarwal1612 (approved):

PR #2540, opened 2024-07-09
Cleanup unused hilt dependencies

shobhitagarwal1612:
Used android studio's inspection tool to identify all unused hilt injections.

@gino-m @sufyanAbbasi  PTAL?


sufyanAbbasi (approved):
Wow! Thank you for finding and fixing all these! I assume you were able to run through these on the emulator, so LGTM!

PR #2559, opened 2024-07-20
Fix flickering of offline icon in survey selector screen

### LINKED ISSUE #2558: Offline and overflow icon in Survey Selector screen flickers on load
Steps to reproduce:

1. Open survey selector screen
2. Observe that `offline` and `overflow` icons are visible for all cards and then disappear

Expected behavior: 

1. No flickering should happen

shobhitagarwal1612:
Fixes #2558

Due to data binding, there was a delay in offline icon's visibility change. Doing it manually resolves the issue.

[Before](https://github.com/user-attachments/assets/065375ae-c91f-460c-bb86-1a69146dca4a)

[After](https://github.com/user-attachments/assets/392acf70-e278-4385-90ed-f8d3576c8e4d)

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveyListAdapter.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveyListAdapter.kt
@@ -17,6 +17,7 @@
 package com.google.android.ground.ui.surveyselector
 
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import androidx.recyclerview.widget.RecyclerView
 import com.google.android.ground.databinding.SurveyCardItemBinding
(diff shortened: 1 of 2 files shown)


ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveyListAdapter.kt
```diff
@@ -42,10 +43,17 @@ class SurveyListAdapter(
...
+    with(holder.binding) {
+      val me = this@SurveyListAdapter
+      item = surveyListItem
+      viewModel = me.viewModel
+      fragment = me.fragment
+
+      val offlineVisibility = if (surveyListItem.availableOffline) View.VISIBLE else View.GONE
+      offlineIcon.visibility = offlineVisibility
+      overflowMenu.visibility = offlineVisibility
+    }
```
anandwana001, line 56: just a quick guess, but how about if we take the BindingAdapter annotation approach over here?

WDYT?

IMO, onBindViewHolder should be a mediator type, it should not consist the UI logics.

shobhitagarwal1612: We already had the bind adaptor approach which was leading to this problem. Given that we are moving towards jetpack compose, this recycler view / view holder should get replaced soon in the future with the migration.

sufyanAbbasi (approved):
Thank you!

PR #2568, opened 2024-07-23
Added unit test for MutationDetail

### LINKED ISSUE #2567: [Code health] Add Unit Test for MutationDetail
Add Unit Test for MutationDetail

anandwana001:
Fixes #2567 

@sufyanAbbasi  PTAL?

Screenshot 2024-07-23 at 1 17 40 PM

### DIFF
--- a/ground/src/test/java/com/google/android/ground/ui/syncstatus/MutationDetailTest.kt
+++ b/ground/src/test/java/com/google/android/ground/ui/syncstatus/MutationDetailTest.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
Why do we need test coverage for a data class?

anandwana001:
> Why do we need test coverage for a data class?

Yes, writing a separate test file might not require or can be covered with another dependent file, but
1. This file is mentioned in my task list as it's a 0% code coverage
2. IMO, writing test cases might look extra but it's good to have as we can say we have both dependent and independent test cases written which is helping us to achieve 100% code coverage 

WDYT?

shobhitagarwal1612:
Adding coverage via this way doesn't achieve anything. I'd recommend not adding it. Instead, we should see which code flow should have covered these files and add tests for that. 

For e.g. MutationDetail object is created when displaying the SyncStatusFragment. So, adding unit tests for this fragment should automatically cover these files as well.

anandwana001:
Got it, let me close this pr and we will cover it with the `SyncStatusFragment `. Thanks

PR #2579, opened 2024-07-25
Fix millisecond to second conversion

### LINKED ISSUE #2578: [Proto] Timestamps for submissions and LOIs serialized incorrectly
The timestamp for submission was serialized to `1721893085608000` which according to https://www.epochconverter.com/ is 1000 * milliseconds! Looks like we multiplied when we should have divided, since Kotlin/Java Dates are in milliseconds!

sufyanAbbasi:
Fixes #2578 

Fixes millisecond to second conversion

Unit tests

@shobhitagarwal1612 PTAL!


shobhitagarwal1612 (approved):
Thank you!

PR #2588, opened 2024-07-30
Prevent view ID conflict/crash by multiplying each generated view ID by a prime number

### LINKED ISSUE #2493: Force quit on consecutive geometry tasks (com.google.maps.api.android.lib6.impl.as cannot be cast to android.view.ViewGroup)
**Describe the bug**

Received the following error when two subsequent location tasks are being rendered: the map doesn't render and we see this in the stack trace:
```
java.lang.ClassCastException: com.google.maps.api.android.lib6.impl.as cannot be cast to android.view.ViewGroup
  at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:537)
  at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:272)
  at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1943)
  at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1839)
  at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1782)
  at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:565)
  at android.os.Handler.handleCallback(Handler.java:958)
  at android.os.Handler.dispatchMessage(Handler.java:99)
  at android.os.Looper.loopOnce(Looper.java:205)
  at android.os.Looper.loop(Looper.java:294)
  at android.app.ActivityThread.main(ActivityThread.java:8177)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
```

I placed breakpoints at the exact line where the crash occurs and inspected the thing that was failing to cast as a ViewGroup. It turns out that the crash happens when it tries to render the compass of all things! 

It happens specifically when I try to render the `DrawAreaTaskFragment` before or after a drop pin LOI task.

When I just have a Draw Area LOI task and no other, then it seems to work

**To Reproduce**
1. Publish a survey with the default capture location task and then an ad-hoc LOI task (e.g. capture location again).
2. Zoom in and start a job, then see that the first task, the default capture task, is broken with a white screen.

**Expected behavior**
1. The map is rendered.

sufyanAbbasi:
Fixes #2493

- [x] Increase the chances of uniqueness for generated View IDs by multiplying by a randomly chosen, 5-digit prime number in order to prevent crashes due to `generateViewId()` producing an ID number that is the same as an ID number assigned to a View in the GoogleMapsFragment.
  -  I think this occurs because `generateViewId()` is not thread safe when `GoogleMapsFragment` views are also being produced at the same time.

- [x] Created a survey via `npm run test:e2e:create` in `ground-platform` to create a survey to repro breakage.
- [x] Run local server via `npm run start:local` by modifying package.json to use `test/data-create`
- [x] Opened survey, verified that crash was avoided. 

Screenshot 2024-07-29 at 5 13 12 PM

@gino-m, @shobhitagarwal1612  PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
@@ -40,7 +40,7 @@ class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskView
   override fun onCreateTaskBody(inflater: LayoutInflater): View {
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical uniqueness.
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
-    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11149}
+    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11149 }
     parentFragmentManager
       .beginTransaction()
       .add(
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
@@ -38,7 +38,9 @@ class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskView
     TaskViewFactory.createWithCombinedHeader(inflater, R.drawable.outline_pin_drop)
 
   override fun onCreateTaskBody(inflater: LayoutInflater): View {
-    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() }
+    // NOTE(#2493): Multiplying by a random prime to allow for some mathematical uniqueness.
+    // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
+    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11149}
     parentFragmentManager
       .beginTransaction()
       .add(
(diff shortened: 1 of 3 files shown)


gino-m:
Nice sleuthing, @sufyanAbbasi ! Looks like some checks are failing?

sufyanAbbasi:
@scolsen: 
> Do we need to do it this way at all? The other task fragments don't call add and instead use bindings (see e.g. MultipleChoiceTaskFragment).

> If we need to use add here, can we either drop the LinearLayout container or define it directly in the layout for the fragment? add claims that the ID is optional and using an argument of 0 will just add it without using a parent—can we use the existing build-time ID of the general task fragment parent instead, to avoid this problem entirely?

Thank you so much for the alternatives! I have a strong feeling these would actually work, now understanding better what the root cause of it is. Unfortunately, I'm under time constraints and will take me a little bit to investigate and verify that the fix actually works. Can I schedule some time with you to peer program on this?

@gino-m:
> Looks like some checks are failing?

```
* What went wrong:
Execution failed for task ':ground:kaptGenerateStubsDevStagingUnitTestKotlin'.
> Could not resolve all files for configuration ':ground:detachedConfiguration83'.
   > A build operation failed.
         Could not read workspace metadata from /workspace/.gradle/caches/transforms-4/041b4d9711d44f92419df7bc9f3bdaac/metadata.bin
      > Could not read workspace metadata from /workspace/.gradle/caches/transforms-4/041b4d9711d44f92419df7bc9f3bdaac/metadata.bin
```

I can't tell if these are just transient, I'll try rerunning.

sufyanAbbasi:
@scolsen and I had a chance to meet, we determined that due to the nature of needing to render a view to return from `createTaskBody()` to be added to `addTaskView()`, we need to create the dummy `LinearLayout` to achieve that. Static IDs doesn't work either since we run into the same issue of ID conflicts. Finally, we could refactor the entire codebase to return a Fragment instead of a View there, then handle the Fragment add upstream, but that's far more work than it is worth.

Hence, we have decided to just go for this solution for now with comment.

scolsen (approved):

PR #2598, opened 2024-07-31
Hide "Collect data" button when the job has no more tasks to complete

### LINKED ISSUE #2539: Crash when clicking "Collect data" again in recently added geometry
**Describe the bug**

After having recorded a new site in my survey that does not require anything else but recording a geometry (no further questions, no photo demanded), there is a white pane at the bottom of my screen with the record's name and a green button "Collect data". If I click that button, the app crashes. This does not happen in another survey, where I can answer additional survey questions about a recorded item that I have previously skipped when I click "Collect data".
Happened on Xiaomi Poco X3 with Andoird 12 SQK1.211019.001.

**To Reproduce**
1. Create a new survey with nothing but one job: To walk or draw the perimeter.
2. Draw a perimeter, give it a name and save.
3. In the white pane with the item's name, click the green "Collect data" button again (don't swipe to the right to collect data for a new selection site, that works alright.)
4. Crash.

**Expected behavior**
I don't really know. Probably not crash. Rather have the "Collect data"-button greyed out.

**Actual behavior**
Button is there and clickable and triggers a crash of the application.

**Screenshots**
Collect_data_bug

sufyanAbbasi:
Fixes #2539

Don't start data collection and instead throw an error if there are not more valid tasks for the job.

- [x] Detects when there are no tasks or there is only an LOI task for pre-defined submission and displays an error.
- [x] Defensively put the user back to the beginning of the task list if for any reason the `startId` is not found.
- [ ] Couldn't find any good place to test the code, sadly... And I'm unfortunately in a time crunch right now to fix that.

- [x] Created a job that only has the LOI task using the local web server.
- [x] Started the job in the app and created an LOI (only task).
- [x]  Pressed "Collect data" on the task, shows an error instead.

Screenshot 2024-08-12 at 6 59 39 PM

Screenshot 2024-08-12 at 7 03 26 PM

This state should be impossible but is nice to have to prevent crashes:
Screenshot 2024-07-31 at 2 13 15 PM

@shobhitagarwal1612 PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -30,6 +30,7 @@ import com.google.android.ground.model.submission.isNullOrEmpty
 import com.google.android.ground.model.task.Condition
 import com.google.android.ground.model.task.Task
 import com.google.android.ground.persistence.local.room.converter.SubmissionDeltasConverter
+import com.google.android.ground.persistence.uuid.OfflineUuidGenerator
 import com.google.android.ground.repository.LocationOfInterestRepository
 import com.google.android.ground.repository.SubmissionRepository
 import com.google.android.ground.repository.SurveyRepository
@@ -82,6 +83,7 @@ internal constructor(
   @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
   private val savedStateHandle: SavedStateHandle,
   private val submissionRepository: SubmissionRepository,
+  private val offlineUuidGenerator: OfflineUuidGenerator,
   locationOfInterestRepository: LocationOfInterestRepository,
   surveyRepository: SurveyRepository,
 ) : AbstractViewModel() {
@@ -238,7 +240,7 @@ internal constructor(
       return
     }
 
-    data[taskViewModel.task] = taskViewModel.taskTaskData.firstOrNull()
+    data[taskViewModel.task] = taskViewModel.taskTaskData.value
 
     if (!isLastPosition()) {
       step(1)
@@ -263,8 +265,9 @@ internal constructor(
 
   /** Persists the changes locally and enqueues a worker to sync with remote datastore. */
   private fun saveChanges(deltas: List<ValueDelta>) {
+    val collectionId = offlineUuidGenerator.generateUuid()
     externalScope.launch(ioDispatcher) {
-      submitDataUseCase.invoke(loiId, job, surveyId, deltas, customLoiName)
+      submitDataUseCase.invoke(loiId, job, surveyId, deltas, customLoiName, collectionId)
     }
   }
 
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -321,9 +321,10 @@ internal constructor(
     if (tasks.isEmpty()) {
       error("Can't generate sequence for empty task list")
     }
+    val firstTaskId = tasks[0].id
     val startIndex =
       tasks
-        .indexOfFirst { it.id == (startId ?: tasks[0].id) }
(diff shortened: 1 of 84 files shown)


gino-m:
Thank you, @sufyanAbbasi ! Are missing tests easy to add, or should we disable these warnings for now?

sufyanAbbasi:
Unfortunately, the tests are really hard to add since we don't have any good Fragment or ViewModel tests on these components, just a small one that checks for the map type dialog opens. There's a ton we'd have to do get that fixed.... I have tested this locally on the emulator, if you'll at least believe that the crash won't happen.

gino-m (approved):
LGTM!

gino-m:
> Unfortunately, the tests are really hard to add since we don't have any good Fragment or ViewModel tests on these components, just a small one that checks for the map type dialog opens. There's a ton we'd have to do get that fixed.... I have tested this locally on the emulator, if you'll at least believe that the crash won't happen.

Ack. Please stamp and merge https://github.com/google/ground-android/pull/2610 and update this branch to rerun checks!

gino-m (approved):

gino-m:
What does it mean for a job to have no more tasks to complete? A job with no tasks is an invalid job and shouldn't be shown, right?

gino-m (changes requested):
What does it mean for a job to have no more tasks to complete? A job with no tasks is an invalid job and shouldn't be shown, right?

gino-m (changes requested):
@sufyanAbbasi On second thought - the correct thing to do would be to hide the "Add data" button altogether for jobs which only have a single addLoi task. Can you update this PR accordingly?

sufyanAbbasi:
> What does it mean for a job to have no more tasks to complete? A job with no tasks is an invalid job and shouldn't be shown, right?

Good question, definitely we can/should workshop the UX a little bit here, as I was mostly going for the simplest solution possible. For all intents and purposes, there are tasks on the job, it's just that it's only the one LOI task, so if you are trying to submit to a pre-defined LOI task, then there are technically no MORE tasks on the job, hence the message. I think a survey organizer would be able to understand why their user saw that; they configured the survey to only have the LOI task so if they are trying to submit to a predefined task, there's nothing the user can do with that.

I did see that it's possible to put the card in a read-only state, but it doesn't provide any feedback to the user about why they can't do anything with it. From a UX perspective, we prefer not to disable submit buttons, instead, let users click them and get feedback about why it doesn't work, and they can escalate to the survey organizers with a clear enough message.

If you have any suggestions for a better error message, please let me know!

gino-m:
It's quite clear imo - if your only task is to add an LOI (e.g, a survey where the users just need to drop pins on the map), "add data to an existing LOI" has no meaning, and as such it would be a bug to show the CTA in the UI. "No more tasks" would be confusing without context - "How do I get more tasks? Do I need to do something else? Or is this a bug"? Would it be hard to just remove it completely in these cases?

sufyanAbbasi:
Implemented the fix to remove the "Collect Data" button. I did keep the "No task" message even though now it's impossible to get to since the button doesn't render, but since we do the same thing for user permission, I figured it would be worth keeping around to prevent crashes.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -321,17 +321,18 @@ internal constructor(
...
-    if (task.isEmpty()) {
-      error(
-        "Unable to find a task with id startId=$startId, firstTaskId=${tasks[0].id}, allTasks=${tasks.map { it.id }}"
-      )
-    }
-
-    val startIndex = tasks.indexOf(task.first())
+    val startIndex =
+      tasks
+        .indexOfFirst { it.id == (startId ?: tasks[0].id) }
```
shobhitagarwal1612, line 326: Can we move `tasks[0].id` outside of the filtering logic?

sufyanAbbasi: Done.

shobhitagarwal1612 (approved):

gino-m (approved):
LGTM! Please just update the PR desc before merging!

PR #2600, opened 2024-08-01
Added test cases for HomeScreenViewModel

### LINKED ISSUE #2599: [Code health] Add test cases for HomeScreenViewModel
Add test cases for HomeScreenViewModel

anandwana001:
Fixes #2599 

@sufyanAbbasi @shobhitagarwal1612  PTAL?

Looks like the codecov is visible after the PR get's merge for outside contributors

Screenshot 2024-08-01 at 12 29 47 PM

### DIFF
--- a/ground/src/test/java/com/google/android/ground/ui/home/HomeScreenViewModelTest.kt
+++ b/ground/src/test/java/com/google/android/ground/ui/home/HomeScreenViewModelTest.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.home
+
+import com.google.android.ground.BaseHiltTest
+import com.google.android.ground.persistence.local.stores.LocalOfflineAreaStore
+import com.google.android.ground.testNavigateTo
+import com.google.android.ground.ui.common.Navigator
+import com.sharedtest.FakeData.OFFLINE_AREA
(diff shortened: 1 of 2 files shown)


ground/src/test/java/com/google/android/ground/ui/home/HomeScreenViewModelTest.kt
```diff
@@ -0,0 +1,106 @@
...
+
+  @Inject lateinit var navigator: Navigator
+  @Inject lateinit var homeScreenViewModel: HomeScreenViewModel
+  @Inject lateinit var localOfflineAreaStore: LocalOfflineAreaStore
+
+  @Test
+  fun `showAbout is navigating to correct screen`() = runWithTestDispatcher {
+    testNavigateTo(navigator.getNavigateRequests(), HomeScreenFragmentDirections.showAbout()) {
+      homeScreenViewModel.showAbout()
+    }
```
shobhitagarwal1612, line 42: Can we perform UI click instead of calling the nav directions?

shobhitagarwal1612 (changes requested):
Left a comment regarding the testing strategy

ground/src/test/java/com/google/android/ground/ui/home/HomeScreenViewModelTest.kt
anandwana001: ViewModel Testing vs Fragment Testing

shobhitagarwal1612: @sufyanAbbasi What is your opinion on testing fragments vs view models? I feel testing via fragment gives a better coverage including the UI components. If we add a test for view model right now, that part would still remain missing and having both fragment and view model unit tests would result in duplicate tests.

anandwana001: Putting my thoughts here:

1. In terms of coverage, yes the above lines are correct
2. Regarding testing strategy, we have two things, for example, A. writing test cases to check Is my view visible on screen or not, B. Are my ViewModel functions working as expected or not, are my ViewModel variables are having correct values or not, no matter how they are consumed by UI or not, but values in variables are correct or not. 
That's why, it's important to have Both the test cases to make sure all pieces of code are working correctly independently.

shobhitagarwal1612: discussed offline, we'll be going ahead with adding both coverages if possible.

shobhitagarwal1612 (approved):

PR #2602, opened 2024-08-02
Load drawn area from saved state

### LINKED ISSUE #2373: [Draw area task] Add support for partial loads of drawn areas (works for completed ones)
Split from https://github.com/google/ground-android/issues/708

sufyanAbbasi:
Towards #2373 

Ensures that draw area tasks can be loaded from saved state. Will enter "complete" mode by default and the user can  undo if they want to go back, similar to the drop pin save state logic.

- [x] Loads drawn areas from `taskData` by setting the vertices and completing the polygon.
- [x] Handles errors if for whatever reason the saved polygon is incomplete.

- [x] Created a job with draw area LOI task and quit after the draw area task was completed by force quitting the app.
- [x] Reloaded the app and ensured that the task started with the drawn area already loaded (see screenshot)

Screenshot 2024-08-01 at 6 04 32 PM

@shobhitagarwal1612 PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
@@ -74,9 +74,7 @@ class DrawAreaTaskFragment : AbstractTaskFragment<DrawAreaTaskViewModel>() {
     addPointButton =
       addButton(ButtonAction.ADD_POINT).setOnClickListener { viewModel.addLastVertex() }
     completeButton =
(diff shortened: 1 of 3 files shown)


shobhitagarwal1612:
This won't work if the drawn area isn't complete, right? My understanding was that we have to persist partial polygon as well, but that was causing some troubles as we create the task data only when the polygon is marked as complete. The current approach is a big improvement over not loading anything at all. 

@gino-m Can we confirm that we don't need to re-load partial polygons?

shobhitagarwal1612 (approved):

sufyanAbbasi:
Good point, won't mark as fixed for now and we can revisit this as a feature request if we want to do partial loads. We can probably also lower the priority

PR #2603, opened 2024-08-03
[Crash] Re-enable destructive migration

### LINKED ISSUE #2551: App doesn't start on some devices
The app crashes immediately after opening with no error message.

The likely culprit is that we've recently disabled destructive migrations, but we don't have any migrations implemented, so the app crashes when trying to upgrade the local db. 

https://console.firebase.google.com/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/5b84c7cbace4a03aa2c306d463cd4c8b?time=last-seven-days&sessionEventKey=66989F78037200016ECEE70338003D9A_1971279737149363807

Workaround: Uninstall and reinstall the app

To fix: Re-enable destructive migrations

shobhitagarwal1612:
Fixes #2551

@sufyanAbbasi @gino-m  PTAL?


gino-m (approved):
LGTM!

PR #2609, opened 2024-08-07
Sync: Fix double mutation detail rendering in sync status.

### LINKED ISSUE #2389: [Offline sync status] Duplicate entries shown in status screen
**Describe the bug**
When adding a new LOI, two entries appear, one of which never completes.

**To Reproduce**
1. Add a new LOI and collect data
2. Open sync status screen

**Expected behavior**
There is only one entry shown for the LOI, submission, and photo.

**Actual behavior**
Two entries are shown, one of which never finishes.

**Screenshots**

@scolsen PTAL? @jcqli for tracking.

scolsen:
Fixes #2389.

When a user collects data, they may need to add a new LOI depending on the job configuration. We treat the addition of the LOI and the eventual submission of data against it as two distinct data mutations. This resulted in duplicate mutation details in sync status for what, to the user, is perceived as a single instance of data collection/submission, which can prove confusing.

To resolve this, we now associate mutations together using a "collection ID" which indicates whether or not two distinct mutations originated from the same data collection flow. When we read mutations out of the local store, we then combine them using this identifier, taking only the more recent mutation as indicative of the overall status for the submission.

### DIFF
--- a/ground/src/main/java/com/google/android/ground/Config.kt
+++ b/ground/src/main/java/com/google/android/ground/Config.kt
@@ -25,7 +25,7 @@ object Config {
   const val SHARED_PREFS_MODE = Context.MODE_PRIVATE
 
   // Local db settings.
-  const val DB_VERSION = 118
+  const val DB_VERSION = 119
   const val DB_NAME = "ground.db"
 
   // Firebase Cloud Firestore settings.
--- a/ground/src/main/java/com/google/android/ground/domain/usecases/submission/SubmitDataUseCase.kt
+++ b/ground/src/main/java/com/google/android/ground/domain/usecases/submission/SubmitDataUseCase.kt
@@ -44,12 +44,13 @@ constructor(
     surveyId: String,
     deltas: List<ValueDelta>,
     loiName: String?,
+    collectionId: String,
   ) {
     Timber.v("Submitting data for LOI: $selectedLoiId")
     val deltasToSubmit = deltas.toMutableList()
     val submissionLoiId =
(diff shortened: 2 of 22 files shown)


scolsen:
Many of these changes a simple downstream consequences of adding a new field to the mutations. The core logical change is in ground/src/main/java/com/google/android/ground/repository/MutationRepository.kt, which we use only to read off mutations for the status screen. Note, importantly, that the actual workers do not use this logic. We generate the value for this field at submission time in the data collection view model.

gino-m (approved):
If we add `edit` functionality we may want to expand the meaning and scope of `collection` to be something like `operation`, but for now this LGTM!

gino-m:
@scolsen 

```
> [ktfmt] Found 2 files that are not properly formatted:
  src/test/java/com/google/android/ground/ui/datacollection/DataCollectionFragmentTest.kt
  src/test/java/com/google/android/ground/persistence/remote/firebase/protobuf/LoiMutationConverterTest.kt
```

scolsen:
formatting is fixed but it looks like there's some other issue... I can't see the logs unfortunately

scolsen:
...looks like it may have been checkcode stuff

sufyanAbbasi (approved):
Thank you for the fix!!!

gino-m:
> ...looks like it may have been checkcode stuff

Actually it's no longer finding the protoc java plugin for some reason. I was able to repro this locally in `master` as well.. 🤷‍♂️ Not sure why it's using `java` and not `java-lite`, and what would have changed to cause this, but specifying the dep manually seems to fix this. Fix in https://github.com/google/ground-android/pull/2620.

PR #2620, opened 2024-08-08
Specify protoc java plugin dep explicitly

gino-m:
Fixes failure running tests in CI in #2609. @scolsen PTAL?

Tested by syncing Gradle and running `gradle generateDevDebugUnitTestProto` locally.


scolsen (approved):

gino-m:
Looks like we get a Gradle `CheckDuplicatesRunnable` error now.. looking into this..

gino-m:
This doesn't seem like the right approach, I wonder if #2609 failed due to flakes and I just got unlucky locally with the same flake?.. Let's keep an eye out!

Issue #2618, opened 2024-08-08
[Multiple choice question task] Brackets still being added to "Other" text in db new format


gino-m:
Observed directly in the db, the value is being stored as "[ user text ]" rather than "user text".

The "other" text is also getting inserted into `selected_option_ids`, leading to https://github.com/google/ground-platform/issues/1954 on export:

![image](https://github.com/user-attachments/assets/ec03afb6-9ce8-4cb1-9c05-ede0b0e4e201)

@shobhitagarwal1612 PTAL? @jcqli FYI

PR #2621, opened 2024-08-09
Fix value for multiple choice tasks with "other" option selected

shobhitagarwal1612:
Fixes https://github.com/google/ground-android/issues/2618

Exclude "other" value from selected ids list and set it to correct field

Verified changes via unit tests and testing locally.

@gino-m  PTAL?


gino-m (approved):
Thank you, @shobhitagarwal1612 ! Can we just get rid of the `[ ]` workaround altogether? Let's merge this now to fix the issue and revisit when you have time?

shobhitagarwal1612:
The tests are working fine locally. Will merge this as is.

gino-m:
Flaky tests? Rerunning now...

PR #2629, opened 2024-08-10
Prevent overlapping of info card and location FAB

### LINKED ISSUE #2442: [Draw area task] Location card obscures location lock button
Screenshot 2024-04-12 at 8 24 28 AM

Screenshot 2024-04-12 at 8 25 41 AM

@rawbzz Is looking at this UI element now, may want to wait for final designs before fixing.

shobhitagarwal1612:
Fixes #2442

Inline the location fab and other elements in task fragment instead of embedded layout to allow constraints to be set correctly.

| Small screen | Large Screen |
| ----- | ----- |
| update-info-card | update-info-card-large |

@gino-m  PTAL?


gino-m (approved):
Thanks for the fix!

PR #2630, opened 2024-08-10
Don't zoom map when changing LOIs remotely

### LINKED ISSUE #2378: [Sync] Map zooms when LOIs are changed remotely
**Repro**

While online:

1. Open a survey with LOIs
2. Modify an LOI manually in Firebase (e.g., update the position of a pin)

**Expected behavior**

Map reflects changes in remote db

**Actual behavior**

Map updates then zooms out 1-2 zoom levels

@shobhitagarwal1612 @jcqli @sufyanAbbasi FYI

shobhitagarwal1612:
Fixes #2378

This PR resolves the following problems:
 * Remove implicit padding of `100` pixels when applying camera position via `bounds` and make it explicit.
 * Make the logic related to moving camera position more explicit than based on precedence. This also fixes the problem of duplicate position logic (bounds vs coordinates + zoom level)
 * Code cleanup to allow `CameraPosition` to have non-null co-ordinates which was making code awkward at certain places where it was needed always.

Changes verified by ensuring that the map doesn't zoom if the LOIs are changed remotely

@gino-m  PTAL?


shobhitagarwal1612:
Found a better solution to the problem. Closing this.

shobhitagarwal1612:
@gino-m This is now ready for your review.

gino-m (approved):
Thank you for the fix, @shobhitagarwal1612 !

Issue #2626, opened 2024-08-09
[Add LOI] LOI names added to wrong field in db


gino-m:
They're being added to `custom_tag`, which is reserved for imported ID. 

The desired proto path in LOI is `properties['name'].string_value`.

@shobhitagarwal1612 PTAL?

gino-m:
From the looks of the LOI card, it appears it's being set in both customTag and name fields:

![image](https://github.com/user-attachments/assets/245924bf-b340-470f-9520-dda88fdccb61)

sufyanAbbasi:
Sorry! This was my fault!

I filed: https://github.com/google/ground-android/issues/2576
and "fixed it": https://github.com/google/ground-android/pull/2577

I thought the customId was important but it wasn't as important as I thought...

gino-m:
> Sorry! This was my fault!
> 
> I filed: #2576 and "fixed it": #2577
> 
> I thought the customId was important but it wasn't as important as I thought...

There are two things:

* `customTag`: User-defined (imported) ID for the LOI
* `properties.name` User-defined label for the LOI

These get displayed to the user as defined in https://github.com/google/ground-android/issues/1926.

Iiuc, the changes you implemented treat those as interchangeable in the db, whereas they should actually be kept separately; they are only combined using the rules in #1926 at render time.

sufyanAbbasi:
Fixed by https://github.com/google/ground-android/pull/2645

PR #2645, opened 2024-08-13
Don't set customTag with property name value.

sufyanAbbasi:
Fixes #2577

Undo https://github.com/google/ground-android/pull/2577

@gino-m PTAL?


gino-m (approved):
Thanks! This addresses my comment in https://github.com/google/ground-android/issues/2626#issuecomment-2286824196 as well.

PR #2646, opened 2024-08-13
Update gradle dependencies

shobhitagarwal1612:
Routine gradle dependency updates.

Verified that there are no regressions via manual testing.

Note: I noticed that the latest update to map dependency updates the look and feel of base-map as well.

@gino-m  PTAL?


gino-m (approved):
🆙

anandwana001:
Suggestion

We should move to Version catlog approach or something similar, so that we don't have to update versions at multiple places and we can have a single source to manage our dependencies. 
Though, right now it's not that much, but it can be hard to manage once it grow.

Issue #1591, opened 2023-02-23
Wrap first time survey sync in a transaction so survey and LOI download is atomic


gino-m:
If `activateSurvey` in `SurveyRepository` fails halfway through, local db may be left in an invalid state.

gino-m:
@gino-m to verify

PR #2653, opened 2024-08-15
Prevent crash when displaying status of missing LOI

### LINKED ISSUE #2652: [Data collection / sync status] App crashes after collecting data offline
Crash:

1. Turn on airplane mode
2. Add new LOI with data
3. Turn off airplane mode
4. Tap "Sync status"

Stack trace from Crashlytics:

```
          Fatal Exception: java.lang.IllegalStateException: LOI not found: 6fwrqWRbPUBx6uOdlfWu
       at com.google.android.ground.repository.LocationOfInterestRepository.getOfflineLoi(LocationOfInterestRepository.kt:80)
       at com.google.android.ground.repository.LocationOfInterestRepository$getOfflineLoi$1.invokeSuspend(:15)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
       at android.os.Handler.handleCallback(Handler.java:959)
       at android.os.Looper.loopOnce(Looper.java:232)
       at android.os.Looper.loop(Looper.java:317)
       at android.app.ActivityThread.main(ActivityThread.java:8592)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:878)
```

scolsen:
Instead, we (for now) just print "Unkown Location". We should get a better fix in place hereafter.

Towards #2652

@gino-m

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/syncstatus/SyncStatusViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/syncstatus/SyncStatusViewModel.kt
@@ -64,7 +64,7 @@ internal constructor(
   ): List<MutationDetail> = mutations.map { toMutationDetail(it) }
 
   private suspend fun toMutationDetail(mutation: Mutation): MutationDetail {
-    // TODO: Find a better solution here; perhaps return a null detail and do not render.
+    // TODO(#2652): Find a better solution here; perhaps return a null detail and do not render.
     val user = userRepository.getAuthenticatedUser()
     return try {
       val loi =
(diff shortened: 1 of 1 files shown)


gino-m:
This solves the crash but not the data loss problem, which is arguable more problematic. Updated "Fixes" to "Towards".

gino-m (approved):
Approved, with minor edits

gino-m (changes requested):
@scolsen If we don't crash when the db is in an invalid state, we won't know when it's happening in the wild since Crashlytics won't be triggered, making it harder to identify. Are you able send a message to Crashlytics when this happens so we have some record of this?

gino-m:
#1591 should have resolved this issue at the source. In the future, we may want to introduce a true "failed" state to surface unrecoverable errors like this one to the user.

PR #2661, opened 2024-08-16
[Terms of service] Sign user out and show toast when ToS cannot be loaded

### LINKED ISSUE #2657: [Bug] Terms of service screen is skipped if network is not available when the app is opened for the first time.
**Describe the bug**
Terms of service screen is skipped if network is not available. Instead it skips to the survey selection screen and the terms of service screen is showed on the next app reopen.

gino-m:
To simplify error handling, this PR also stops loading ToS from remote before directing to the fragment; instead, we always direct to the fragment. An empty ToS will be shown if not found.

Fixes #2657 

@shobhitagarwal1612 @sufyanAbbasi @scolsen or @anandwana001 PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/MainActivity.kt
+++ b/ground/src/main/java/com/google/android/ground/MainActivity.kt
@@ -32,9 +32,7 @@ import com.google.android.ground.databinding.MainActBinding
 import com.google.android.ground.repository.UserRepository
 import com.google.android.ground.system.ActivityCallback
 import com.google.android.ground.system.ActivityStreams
-import com.google.android.ground.system.auth.AuthenticationManager
 import com.google.android.ground.ui.common.BackPressListener
-import com.google.android.ground.ui.common.EphemeralPopups
 import com.google.android.ground.ui.common.FinishApp
 import com.google.android.ground.ui.common.NavigateTo
 import com.google.android.ground.ui.common.NavigateUp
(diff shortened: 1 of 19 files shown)


gino-m:
Looks like `MainViewModelTest` and `TermsOfServiceRepositoryTest` need to be updated.

ground/src/main/java/com/google/android/ground/MainActivity.kt
```diff
@@ -58,7 +60,9 @@ class MainActivity : AbstractActivity() {
   @Inject lateinit var activityStreams: ActivityStreams
   @Inject lateinit var viewModelFactory: ViewModelFactory
   @Inject lateinit var navigator: Navigator
+  @Inject lateinit var authManager: AuthenticationManager
   @Inject lateinit var userRepository: UserRepository
+  @Inject lateinit var popups: EphemeralPopups
```
shobhitagarwal1612, line 65: Can we remove these unused dependencies?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/MainActivity.kt
gino-m: Done

shobhitagarwal1612 (approved):

Issue #2670, opened 2024-08-19
[Code health] Migrate Android CI to GitHub Actions


gino-m:
- [x] Unit tests
- [x] Code quality checks
- [x] Test coverage
- [x] Master branch build status
- [x] Caching
- [x] Ability to view logs / reports for debugging
- [x] Deprecate GCB

PR #2671, opened 2024-08-19
[Github Actions] Configure a workflow for running unit tests and static code analysis

shobhitagarwal1612:
Towards #2670

Things done:
 * Triggers a github action on every pull request on master or push to master
 * Runs unit tests and static code analysis checks
 * Also enables gradle caching

@gino-m  PTAL?


.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          cache: 'gradle'
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      - name: Run unit tests
+        run: ./gradlew testDevDebugUnitTest --no-daemon
```
gino-m, line 41: We already have a `localDebug` variant with a similar config. Perhaps we should use that one instead, and remove the duplicate local config in `google-services.json`?

gino-m (changes requested):
That was quick! One comment.

.github/workflows/unit-test.yml
shobhitagarwal1612: Done

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,41 @@
...
+name: Unit Tests
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  test:
```
gino-m, line 23: Can we split this into two jobs, `build` and `test`? `test` would just need `needs: build`.

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2024 The Ground Authors.
```
gino-m, line 1: I assume this workflow will be used for all CI jobs? (build, test, lint, etc?) If so, should we rename it `ci.yml` or similar?

gino-m (changes requested):
Naming nits incoming!

.github/workflows/unit-test.yml
shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m Can you add the codecov token to project secrets? I don't have access to add it.

gino-m (approved):
🏅 !!

Issue #2269, opened 2024-02-27
Investigate into gradle daemon failure for remote builds


shobhitagarwal1612:
Daemon is failing here too..could it be related to [this](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:disabling_the_daemon) recent change?

_Originally posted by @gino-m in https://github.com/google/ground-android/issues/2264#issuecomment-1965250692_

PR #2673, opened 2024-08-20
Re-enable Gradle daemon on CI

gino-m:
Builds frequently fail with:

> Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

For newer version of Gradle [docs](https://docs.gradle.org/8.3/userguide/gradle_daemon.html#continuous_integration) recommend the following:

> We recommend using the Daemon for both developer machines and Continuous Integration servers.

Related to #2269 

Also, double max Gradle heap size for good measure.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -85,7 +85,7 @@ steps:
     args:
       - '-c'
       - |
-        ./gradlew -PdisablePreDex assembleDevStaging assembleDevStagingUnitTest -PtestBuildType=staging --no-daemon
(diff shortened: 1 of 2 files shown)


gino-m:
@sufyanAbbasi PTAL?

sufyanAbbasi (approved):
Thank you!!

shobhitagarwal1612:
We are close to migrating the CI pipeline from GCB to Github Actions. So, this PR would be obsolete once that is done.

gino-m:
@shobhitagarwal1612 And it doesn't appear to solve the problem anyway:

```
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

Closing.

PR #2700, opened 2024-08-26
Update ktfmt to 0.20.0

shobhitagarwal1612:
https://github.com/cortinico/ktfmt-gradle/releases

@gino-m  PTAL?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = false
 indent_style = space
-indent_size = 2
+indent_size = 4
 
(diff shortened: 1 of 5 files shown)


.editorconfig
```diff
@@ -3,7 +3,7 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = false
 indent_style = space
-indent_size = 2
+indent_size = 4
```
gino-m, line 6: Wouldn't we need to reformat all code to conform?

shobhitagarwal1612: I had done this in https://github.com/google/ground-android/pull/2671/files which is causing all files to change indentation. So, this is a rollback for this file only

gino-m (approved):

PR #2709, opened 2024-08-30
Render Markdown in Terms of Service

### LINKED ISSUE #2707: [ToS screen] Render Markdown in Terms of Service screen
Can we use the same component used for data sharing terms to render this?

scolsen:
Fixes #2707.

I added at least a minimal unit test, but it'd be good if we could find a better solution here.


gino-m (approved):
Nice work!

PR #2713, opened 2024-09-05
Remove dependency on context from view models

### LINKED ISSUE #2520: [Code health] Android Resources in ViewModel is not good practice
Original discussion https://github.com/google/ground-android/issues/2519

As we know ViewModel is treated as a layer between UI and other layers. ViewModel consists of all the business logic that is needed for the UI to function as per the requirement. 

Now, if we see the class `AbstractTaskViewModel`, which is a common ViewModel under the tasks feature. One of the features here is showing of error message which is a MutableLiveData and XML is accessing it directly based on it's value. 

So far, everything is correct, but now, to get the error string message, we are using the `android.content.res.resources` which is acting as a ViewModel dependency here, and it is not a good practice to use Android specific objects in ViewModel layer, as it might create a memory leak issue. 

Screenshot 2024-06-26 at 15 08 07

## Solution
We can shift the String assignment work to the respective Fragment or Activity. Or maybe we can use BindningAdapter.

### Memory Leak
**Lifecycle Mismatch**: ViewModels are designed to outlive Activity and Fragment lifecycles, meaning they can be retained across configuration changes (e.g., screen rotations). If a Context or Resources object tied to an Activity or Fragment is held by the ViewModel, it can prevent the Activity or Fragment from being garbage collected, leading to memory leaks.

### Increased Coupling
**Tight Coupling**: Passing a Context or Resources directly into a ViewModel increases the coupling between the ViewModel and the Android framework. This makes the ViewModel less reusable and harder to test, as it now depends on Android-specific classes.

### Testing Complexity
**Difficult to Mock**: Unit testing ViewModels becomes more complex if they depend on Android framework classes. This is because you need to mock or provide fake implementations of Context or Resources, which can be cumbersome.
**Reduced Testability**: Ideally, ViewModels should be tested in isolation without relying on Android framework classes. Dependency injection and repository patterns help achieve this by abstracting dependencies.

@sufyanAbbasi @shobhitagarwal1612 WDYT?

shobhitagarwal1612:
Fixes #2520

@anandwana001 @gino-m   PTAL?

### DIFF
--- a/ground/src/main/res/layout/date_task_frag.xml
+++ b/ground/src/main/res/layout/date_task_frag.xml
@@ -17,7 +17,6 @@
   -->
 
 <layout xmlns:android="http://schemas.android.com/apk/res/android"
-  xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools">
 
   <data>
@@ -46,7 +45,6 @@
       android:maxLines="1"
       android:onClick="@{__ -> fragment.showDateDialog()}"
       android:text="@{viewModel.responseText}"
-      app:error="@{viewModel.error}"
       tools:text="@string/date" />
   </com.google.android.material.textfield.TextInputLayout>
 </layout>
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskViewModel.kt
@@ -15,9 +15,7 @@
  */
 package com.google.android.ground.ui.datacollection.tasks
 
-import android.content.res.Resources
 import androidx.lifecycle.LiveData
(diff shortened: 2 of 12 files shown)


ground/src/main/res/layout/date_task_frag.xml
```diff
@@ -46,7 +45,6 @@
       android:maxLines="1"
       android:onClick="@{__ -> fragment.showDateDialog()}"
       android:text="@{viewModel.responseText}"
-      app:error="@{viewModel.error}"
```
anandwana001, line 49: hope this is not breaking anything

anandwana001 (approved):

ground/src/main/res/layout/date_task_frag.xml
shobhitagarwal1612: This was only done for date/time tasks and nothing else. But the error popup is shown for all tasks in a similar way.

PR #2716, opened 2024-09-06
Update bug_report.md

jcqli:
Include app versions in bug report template. Towards https://github.com/google/ground-android/issues/2731

@gino-m  PTAL?


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -10,6 +10,11 @@ assignees: ''
 **Describe the bug**
 A clear and concise description of what the bug is.
 
+**App versions**
+Android:
+
+Web:
```
gino-m, line 16: Same comment as on the other related PR - there's no version in the web app atm. Even if there were, the user probably wouldn't know with which version of the web app was used to create the survey.

gino-m (changes requested):
Is it possible to add the issue to the "Ground" project via this template? It would be helpful, since atm we need to first save the bug for an issue to be auto-added.

jcqli:
> Is it possible to add the issue to the "Ground" project via this template? It would be helpful, since atm we need to first save the bug for an issue to be auto-added.

Github automation should already be doing that -- e.g. https://github.com/google/ground-android/issues/2735 
image

.github/ISSUE_TEMPLATE/bug_report.md
jcqli: Removed. I'll update https://github.com/google/ground-platform/issues/2015 as well

gino-m:
> > Is it possible to add the issue to the "Ground" project via this template? It would be helpful, since atm we need to first save the bug for an issue to be auto-added.
> 
> Github automation should already be doing that for every new issue -- e.g. #2735 image

@jcqli  It only does that after the bug is saved though; none of the project attributes are editable until then.

shobhitagarwal1612 (approved):

gino-m (approved):

Issue #2717, opened 2024-09-06
Address TODOs in proto converters


jcqli:
After the proto migration, there are several TODOs outstanding in the proto converters related to inconsistency in some proto fields. Need to go through them and address them (remove, fix, document).

- [x] Add "skipped" field in taskData when the value is missing or task is skipped
~~Add "timestamp" in CaptureLocationResult~~
- [ ] Replace "isPredefined" with "Source" in LOI model.
- [ ] Rename "customId" to "customTag" in LOI model.

For full list of tasks, please refer to this [sheet](https://docs.google.com/spreadsheets/d/1Ognh710debgAEI01Z2qvl-jAIfOoxnH6O8z4npiL9_g/r/0-ZZ16BvSYLPvARlghXBtX4A/edit?gid=0#gid=0)

jcqli:
FYI @shobhitagarwal1612 based on standup today

shobhitagarwal1612:
@gino-m I've audited all proto converters when submitting data and added them to the below sheet. Please let me know your thoughts on whether we need these or not?

https://docs.google.com/spreadsheets/d/1Ognh710debgAEI01Z2qvl-jAIfOoxnH6O8z4npiL9_g/r/0-ZZ16BvSYLPvARlghXBtX4A/edit?gid=0#gid=0

gino-m:
@shobhitagarwal1612 Answers in the sheet. HTH.

PR #2718, opened 2024-09-08
[Data submission] Set "skipped" to true when user skips a task 

shobhitagarwal1612:
Towards #2717

Set "skipped" field when user skips a task 

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/model/submission/SkippedTaskData.kt
+++ b/ground/src/main/java/com/google/android/ground/model/submission/SkippedTaskData.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.submission
+
+import kotlinx.serialization.Serializable
+
+/** User skipped the response for the given task. */
+@Serializable
+class SkippedTaskData : TaskData {
+
+  override fun getDetailsText(): String = ""
+
+  override fun isEmpty(): Boolean = false
+}
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/ModelToProtoExt.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/ModelToProtoExt.kt
@@ -38,7 +38,6 @@ import com.google.android.ground.model.submission.TaskData
 import com.google.android.ground.model.submission.TextTaskData
 import com.google.android.ground.model.submission.TimeTaskData
 import com.google.android.ground.model.submission.isNotNullOrEmpty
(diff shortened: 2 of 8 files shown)


ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/ModelToProtoExt.kt
```diff
@@ -128,11 +129,16 @@ fun LocationOfInterestMutation.createLoiMessage(user: User) = locationOfInterest
   }
 }
 
-private fun ValueDelta.toMessage() = taskData {
-  val me = this@toMessage
+private fun toSkippedTaskData(id: String) = taskData {
+  // TODO: What should be the ID?
+  taskId = id
+  skipped = true
```
gino-m, line 135: We can't know at this point if the task was skipped by the user, or if it's just missing data.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/protobuf/ModelToProtoExt.kt
shobhitagarwal1612: Ah, I misinterpreted it based on the proto comments. Thanks for pointing it out.

gino-m: 👍  Would you be able to update the comments to make it more explicit? We should make sure to say "explicitly skipped by the user".

shobhitagarwal1612: Done. I'll update the comment separately

shobhitagarwal1612:
@gino-m PTAL?

ground/src/main/java/com/google/android/ground/model/submission/SkippedTaskData.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.model.submission
+
+import kotlinx.serialization.Serializable
+
+/** User skipped the response for the given task. */
+@Serializable
+class SkippedTaskData : TaskData {
```
gino-m, line 22: This isn't a new task type, so why create a new `TaskData` type? Could we instead just add a `skipped` flag to `TaskData` base class?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/model/submission/SkippedTaskData.kt
shobhitagarwal1612: I had started with this approach but faced couple of challenges:

1. It was possible to have both skipped and value set at the same time. This led to adding more complexity in each task type fragment / view model to ensure skipped and value are never used simultaneously.

2. Serialization of task data became more complicated as now we have to store one extra bit in each task type.

Both of these problems were solved by creating a new task type which works like a empty value but with a special intent (user skipped).

Please let me know if you feel otherwise regarding the currently chosen approach.

gino-m: I don't really have a sense for why serialization would be more complete - you could always check the `skipped` field first and short circuit if set. As for (1), it almost sounds as if you're tricking the UI into resetting its state by setting a different task data.. this sounds somewhat magical and likely adds complexity.

That said, if you've tried this already and it's expensive to implement, I defer to your judgement.

shobhitagarwal1612: Can you please LG if the changes look ok to you?

gino-m (approved):

PR #2719, opened 2024-09-08
[Data Submission] Save timestamp when capturing current location

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/2717

@gino-m PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/common/AbstractMapFragmentWithControls.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/common/AbstractMapFragmentWithControls.kt
@@ -25,7 +25,7 @@ import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
 import com.google.android.ground.R
 import com.google.android.ground.databinding.MapTaskFragBinding
-import com.google.android.ground.model.submission.CaptureLocationTaskData.Companion.toCaptureLocationResult
+import com.google.android.ground.model.geometry.Coordinates
 import com.google.android.ground.ui.map.CameraPosition
 import com.google.android.ground.ui.map.MapFragment
 import com.google.android.ground.util.toDmsFormat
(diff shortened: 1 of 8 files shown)


ground/src/main/java/com/google/android/ground/ui/common/AbstractMapFragmentWithControls.kt
```diff
@@ -55,7 +56,7 @@ abstract class AbstractMapFragmentWithControls : AbstractMapContainerFragment()
     viewLifecycleOwner.lifecycleScope.launch {
       repeatOnLifecycle(Lifecycle.State.STARTED) {
         getMapViewModel().location.collect {
-          val taskData = it?.toCaptureLocationResult()
+          val taskData = it?.toCaptureLocationResult(timestampMillis = Date().time)
```
gino-m, line 59: Why do we save the current time in two places, here and in the `CaptureLocationTaskMapFragment`? 

@kenstershiro My initial intuition was that we should store timestamp when the user clicks "Capture Location" button, not when the OS gives a location update, which would usually be slightly sooner. The degenerate case would be if the OS stops giving location updates (no GPS signal, location services disabled) - in that case, would we want to capture the time "Capture location" was clicked, or the time the last location was known? Can you ask @lecrabe and @jo-spek?

shobhitagarwal1612: We wanted the formatted lat/lng for the info card. Removed the dependency of `CaptureLocationTaskData` from this class and directly created the formatted location text.

gino-m (approved):

ground/src/main/java/com/google/android/ground/ui/common/AbstractMapFragmentWithControls.kt
shobhitagarwal1612: Discussed with @jo-spek

His main concern is that no other task stores a timestamp. Putting this PR on hold until the discussion is finalized.

PR #2720, opened 2024-09-09
Added partial load support for all data types except Draw

### LINKED ISSUE #2373: [Draw area task] Add support for partial loads of drawn areas (works for completed ones)

anandwana001:
Fixes #2373 

@shobhitagarwal1612 PTAL?


anandwana001:
@shobhitagarwal1612 I think we need to do this PR in two parts, as the code and number of files is increasing.

1. Partial Support for all data types, kill support
2. Partial Support for Draw Type, kill support

This PR is handling all data types apart from Draw type.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -257,6 +257,16 @@ internal constructor(
...
   }
 
+  suspend fun saveCurrentState(taskViewModel: AbstractTaskViewModel) {
+    val validationError = taskViewModel.validate()
+    if (validationError != null) {
+      return
+    }
+
+    data[taskViewModel.task] = taskViewModel.taskTaskData.value
+    step(0)
```
shobhitagarwal1612, line 267: Why are stepping when saving the current state?

shobhitagarwal1612: We should call the relevant methods which save to draft for clarity.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -110,6 +110,11 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     onTaskResume()
   }
 
+  override fun onPause() {
+    super.onPause()
+    saveCurrentState()
+  }
```
shobhitagarwal1612, line 116: Would it cause the value to be saved twice when pressing "next" or "previous" button? Can we make it more efficient?

shobhitagarwal1612 (changes requested):
Left a few comments

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
anandwana001: hm..earlier I was planning to control from parent fragment to avoid this condition, as I was not thinking to handle using any extra param.
let me try some other approach.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
anandwana001: fixed

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
anandwana001: fixed

shobhitagarwal1612: If we are saving the state in `onPause`, then do we even need to save it when pressing "previous" or "next" button?

anandwana001: as per the current logic, our function will not work in onPause and onNext and onPrevious has there own way of saying and emiting the data, if we bring all functions under one roof, lot of checks will require, so I choose the each one has there own way of doing things and responsibility

gino-m:
> @shobhitagarwal1612 I think we need to do this PR in two parts, as the code and number of files is increasing.
> 
> 1. Partial Support for all data types, kill support
> 2. Partial Support for Draw Type, kill support
> 
> This PR is handling all data types apart from Draw type.

Wasn't case #1 already addressed by @shobhitagarwal1612? Draw Area was the only case left iiuc?

shobhitagarwal1612:
[test-changes-partial-save.webm](https://github.com/user-attachments/assets/4cb1de6d-0d48-4209-9d72-647528c10e65)

I patched your changes locally and there is some problem with it.

shobhitagarwal1612:
Can you please add some unit test coverage as well for your changes?

anandwana001:
> test-changes-partial-save.webm 
> I patched your changes locally and there is some problem with it.

But the current logic was, never load for the first one, right?
@shobhitagarwal1612

shobhitagarwal1612:
> But the current logic was, never load for the first one, right?

No, the problem was that we were only saving the draft when pressing "previous" or "next". We should save the draft even when the app is killed without pressing any button

anandwana001:
> > But the current logic was, never load for the first one, right?
> 
> No, the problem was that we were only saving the draft when pressing "previous" or "next". We should save the draft even when the app is killed without pressing any button

yes, that's what this PR is doing, as we discussed onPause and all the stuff.

As per the video, the first task, never get to draft due as per the currect product flow.

anandwana001:
https://github.com/user-attachments/assets/854b26fc-6e57-4d6c-8e85-c745bf6f13dd


@shobhitagarwal1612 it's working

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -248,6 +248,20 @@ internal constructor(
...
+  fun saveCurrentState(taskViewModel: AbstractTaskViewModel) {
+    if (!data.containsKey(taskViewModel.task)) {
+      val validationError = taskViewModel.validate()
+      if (validationError != null) {
+        return
+      }
+
+      data[taskViewModel.task] = taskViewModel.taskTaskData.value
+      savedStateHandle[TASK_POSITION_ID] = taskViewModel.task.id
+      clearDraft()
```
gino-m, line 260: Why do we need to call `clear` first?

anandwana001: missed this one, fixed now.

shobhitagarwal1612:
@anandwana001 I found a bug. It also saves empty value and resumes the draft with empty state. Can you please check again?

anandwana001:
> @anandwana001 I found a bug. It also saves empty value and resumes the draft with empty state. Can you please check again?

The empty check is already there, not happening for me. Can ou tell how did you tested this?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
```diff
@@ -40,6 +42,10 @@ class DateTaskFragment : AbstractTaskFragment<DateTaskViewModel>() {
     taskBinding.lifecycleOwner = this
     taskBinding.fragment = this
     taskBinding.viewModel = viewModel
+    if (viewModel.taskTaskData.value.isNotNullOrEmpty()) {
+      val timestamp = (viewModel.taskTaskData.value as DateTimeTaskData).timeInMillis
+      viewModel.updateResponse(getDateFormatter(), Date(timestamp))
```
shobhitagarwal1612, line 47: We already update the response with initial value when loading from draft. So, here we are doing this for updating the UI and not the response. The issue is that text displayed on the UI is tied to `updateResponse` of `DateTaskViewModel`. Can we pull the `dateText` state flow from view model to the fragment to fix this dependency?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/time/TimeTaskFragment.kt
```diff
@@ -40,6 +42,10 @@ class TimeTaskFragment : AbstractTaskFragment<TimeTaskViewModel>() {
     taskBinding.lifecycleOwner = this
     taskBinding.fragment = this
     taskBinding.viewModel = viewModel
+    if (viewModel.taskTaskData.value.isNotNullOrEmpty()) {
+      val timestamp = (viewModel.taskTaskData.value as DateTimeTaskData).timeInMillis
+      viewModel.updateResponse(getTimeFormatter(), Date(timestamp))
```
shobhitagarwal1612, line 47: Same as previous comment.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -110,6 +110,13 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     onTaskResume()
   }
 
+  override fun onPause() {
+    super.onPause()
+    if (!dataCollectionViewModel.isNavigatingUp) {
+      saveCurrentState()
+    }
+  }
```
shobhitagarwal1612, line 118: Can this be moved to `DataCollectionFragment` instead?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -179,6 +186,10 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     lifecycleScope.launch { dataCollectionViewModel.onNextClicked(viewModel) }
   }
 
+  private fun saveCurrentState() {
+    dataCollectionViewModel.saveCurrentState(viewModel)
+  }
+
```
shobhitagarwal1612, line 192: Same as previous comment

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
anandwana001: can we access `viewModel ` which is a AbstractTaskViewModel inside the DatCollectionfragment?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
anandwana001: Keeping it in ViewModel helps in managing the state of the dateText, even in config change, is shifting it to fragment will be a good idea?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
shobhitagarwal1612: DataCollectionViewModel can directly get the currently visible task's view model. We don't need to pass it to the `DataCollectionFragment`

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
shobhitagarwal1612: You can chain the `dateText` state flow with view model's task data. Then everything will work as expected even for config changes.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -140,6 +140,8 @@ internal constructor(
 
   lateinit var submissionId: String
 
+  var isNavigatingUp = false
```
shobhitagarwal1612, line 143: Can this be moved to the fragment as well and made private? I don't think it is needed outside of the fragment.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
```diff
@@ -20,26 +20,36 @@ import android.text.format.DateFormat
...
+import kotlinx.coroutines.flow.asStateFlow
 import org.jetbrains.annotations.TestOnly
 
 @AndroidEntryPoint
 class DateTaskFragment : AbstractTaskFragment<DateTaskViewModel>() {
 
   private var datePickerDialog: DatePickerDialog? = null
 
+  private var _dateText: MutableStateFlow<String> = MutableStateFlow("")
+  val dateText = _dateText.asStateFlow()
```
shobhitagarwal1612, line 40: Can we chain the task data to date conversion here directly instead of calling it manually from the below two places?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
```diff
@@ -28,16 +30,21 @@ import com.google.android.ground.ui.datacollection.tasks.AbstractTaskFragment
...
 import org.jetbrains.annotations.TestOnly
 
 @AndroidEntryPoint
 class DateTaskFragment : AbstractTaskFragment<DateTaskViewModel>() {
 
   private var datePickerDialog: DatePickerDialog? = null
 
-  private var _dateText: MutableStateFlow<String> = MutableStateFlow("")
-  val dateText = _dateText.asStateFlow()
+  private val _timestamp: MutableStateFlow<Date?> = MutableStateFlow(null)
```
shobhitagarwal1612, line 41: We can reuse the taskData from view model instead.

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/date/DateTaskFragment.kt
anandwana001: viewmodel might fail to access, let me check

shobhitagarwal1612: You can do something like this:

```
  lateinit var dateText: LiveData<String>

  override fun onTaskViewAttached() {
    super.onTaskViewAttached()
    dateText =
      viewModel.taskTaskData
        .filterIsInstance<DateTimeTaskData?>()
        .map { taskData ->
          if (taskData != null) {
            val calendar = Calendar.getInstance()
            calendar.timeInMillis = taskData.timeInMillis
            getDateFormatter().format(calendar.time) ?: ""
          } else {
            ""
          }
        }
        .asLiveData()
  }
  ```

Also, please test all that date/time task data also gets saved without clicking any buttons.

File: ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt [hunk 11]
```diff
@@ -248,6 +248,22 @@ internal constructor(
     }
   }
 
+  fun saveCurrentState() {
+    getTaskViewModel(currentTaskId.value)?.let {
+      if (!data.containsKey(it.task)) {
+        val validationError = it.validate()
+        if (validationError != null) {
+          clearDraft()
```
shobhitagarwal1612, line 256: Do we really want to clear the draft (i.e. all tasks) even if the current one is failing validation checks? IMO, we should just return and not clear everything. wdyt?

anandwana001: the reason to clear is, when we move back to home screen, there is a check which opens the draft collections, so need to clear

did you try removing this and see how app behave?

anandwana001: fixed

shobhitagarwal1612: Consider a scenario where you have 5 tasks and 3 of them are already filled. On 4th task, the user decided to kill the app abruptly. If we don't clear the draft, then atleast 3 tasks would be revived.

shobhitagarwal1612 (approved):

PR #2724, opened 2024-09-17
Make sure "Done" button is rendered when the last task is hidden due to unsatisfied condition.

### LINKED ISSUE #2394: [Data collection] Done button not showing when conditional task is final task
The button label should say "Done", instead it shows "Next".

Survey (with ACL removed)..

```
{
  "description": "",
  "jobs": {
    "wr7jv1ZXAAKJ50x2z21y": {
      "name": "Label a point",
      "id": "wr7jv1ZXAAKJ50x2z21y",
      "defaultStyle": {
        "color": "#F37C22"
      },
      "tasks": {
        "R3bgHzIrNZywrUrzkQJy": {
          "addLoiTask": false,
          "label": "Land cover",
          "options": {
            "iwR633IAkxzdILsByP3I": {
              "code": "",
              "label": "Forestland",
              "index": 0
            },
            "CWNRTdcdMJih8QxvjRSj": {
              "label": "Other",
              "code": "",
              "index": 5
            },
            "MSx7wqwadWxwUADyLyrm": {
              "code": "",
              "index": 3,
              "label": "Grassland"
            },
            "P1Q8Nk8mqpT0u3syCDGb": {
              "index": 4,
              "code": "",
              "label": "Wetland"
            },
            "jiEa0g8vziTuNy2BEy97": {
              "label": "Settlements",
              "code": "",
              "index": 2
            },
            "i71ttovXowWXwmyGxBCd": {
              "index": 1,
              "label": "Cropland",
              "code": ""
            }
          },
          "index": 1,
          "type": "multiple_choice",
          "cardinality": "select_one",
          "hasOtherOption": false,
          "required": false
        },
        "VVm0OjJKpNXwETo3rULk": {
          "required": true,
          "label": "Define plot boundaries",
          "type": "draw_area",
          "index": 0,
          "addLoiTask": true
        },
        "GAie6HrMvn3q2BTYu6uN": {
          "options": {
            "lvdc5KuRM2MnqgtnHoxC": {
              "index": 0,
              "label": "Palm",
              "code": ""
            },
            "wDmg0bu7aOWMtyUtrsse": {
              "index": 2,
              "code": "",
              "label": "Coffee"
            },
            "labNfMorqBAqtDSlirCS": {
              "code": "",
              "index": 1,
              "label": "Cocoa"
            },
            "1jICyjEFlLrGXBHtEemd": {
              "label": "Rubber",
              "code": "",
              "index": 5
            },
            "OZ7uGXZmi0kdPhBXX6GN": {
              "index": 4,
              "code": "",
              "label": "Beef"
            },
            "GRPbVwNbYoKAmgQzK6nI": {
              "label": "Soy",
              "code": "",
              "index": 3
            }
          },
"condition": {
    "matchType": "MATCH_ALL",
    "expressions": [
         {
            "expressionType": "ONE_OF_SELECTED",
            "taskId": "R3bgHzIrNZywrUrzkQJy",
            "optionIds": ["i71ttovXowWXwmyGxBCd"]
         }
      ]
    },
          "required": false,
          "label": "Subtypes cropland",
          "index": 2,
          "hasOtherOption": false,
          "addLoiTask": false,
          "cardinality": "select_one",
          "type": "multiple_choice"
        },
        "oHTFUT8m3Pa4wHJGb58t": {
          "label": "North facing",
          "type": "photo",
          "index": 3,
          "addLoiTask": false,
          "required": false
        }
      },
      "index": 1,
      "strategy": "MIXED"
    }
  },
  "title": "Land cover descriptors",
  "acl": {
   // ...
  }
}
```

@sufyanAbbasi FYI

sufyanAbbasi:
Fixes #2394 

Dynamically updates the "Next/Done" button when 

- [x] Converted the `buttonAction` field to a `MutableState` so it can rerender the button dynamically when changed.
- [x] Added a routine, `testLastPosition()` that takes the current task ID and the currently selected value to determine whether the current task WOULD be the last task in the sequence.
- [x] Dynamically updates the "Next" button based on the results of the test.
- [ ] Does not fix the progress bar (that's a very different beast!)

- [x] Tested by creating a survey with the last task as a conditional task based on whether "option 1" is checked.

Shows Done button disabled:
Screenshot 2024-09-16 at 10 17 49 PM
Option 1 triggers the next task, Next button is enabled
Screenshot 2024-09-16 at 10 17 57 PM
Option 2 checked, Next button still enabled
Screenshot 2024-09-16 at 10 18 03 PM
Option 1 unchecked, Next button becomes Done button, still enabled
Screenshot 2024-09-16 at 10 18 09 PM

@anandwana001, @shobhitagarwal1612  PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -325,7 +325,11 @@ internal constructor(
    * start ID will always generate a sequence with the start ID as the first element, and if
    * reversed is set, will generate the previous tasks from there.
    */
-  private fun getTaskSequence(startId: String? = null, reversed: Boolean = false): Sequence<Task> {
+  private fun getTaskSequence(
+    startId: String? = null,
+    reversed: Boolean = false,
+    taskValueOverride: Pair<String, TaskData?>? = null,
+  ): Sequence<Task> {
     if (tasks.isEmpty()) {
       error("Can't generate sequence for empty task list")
     }
@@ -345,7 +349,9 @@ internal constructor(
         tasks.subList(startIndex, tasks.size)
       }
       .let { tasks ->
-        tasks.asSequence().filter { it.condition == null || evaluateCondition(it.condition) }
+        tasks.asSequence().filter {
+          it.condition == null || evaluateCondition(it.condition, taskValueOverride)
+        }
       }
   }
 
@@ -381,10 +387,36 @@ internal constructor(
   fun isLastPosition(taskId: String? = null): Boolean =
     (taskId ?: currentTaskId.value) == getTaskSequence().last().id
 
+  /**
+   * Returns true if the given [taskId] and task data would be last, or the current active task.
+   * Useful for handling conditional tasks.
+   */
+  fun testLastPosition(taskId: String? = null, value: TaskData?): Boolean =
+    (taskId ?: currentTaskId.value) ==
+      getTaskSequence(taskValueOverride = (taskId ?: currentTaskId.value) to value).last().id
+
   /** Evaluates the task condition against the current inputs. */
-  private fun evaluateCondition(condition: Condition): Boolean =
+  private fun evaluateCondition(
+    condition: Condition,
+    taskValueOverride: Pair<String, TaskData?>? = null,
+  ): Boolean =
     condition.fulfilledBy(
-      data.mapNotNull { (task, value) -> value?.let { task.id to value } }.toMap()
+      data
+        .mapNotNull { (task, value) -> value?.let { task.id to it } }
+        .let { pairs ->
+          if (taskValueOverride != null) {
+            if (taskValueOverride.second == null) {
+              // Remove pairs with the testTaskId if testValue is null.
+              pairs.filterNot { it.first == taskValueOverride.first }
+            } else {
+              // Override any task IDs with the test values.
+              pairs + (taskValueOverride.first to taskValueOverride.second!!)
+            }
+          } else {
+            pairs
+          }
+        }
+        .toMap()
     )
 
   companion object {
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -383,17 +383,17 @@ internal constructor(
   /** Returns true if the given [taskId] is first in the sequence of displayed tasks. */
   fun isFirstPosition(taskId: String): Boolean = taskId == getTaskSequence().first().id
 
-  /** Returns true if the given [taskId] is last if set, or the current active task. */
-  fun isLastPosition(taskId: String? = null): Boolean =
-    (taskId ?: currentTaskId.value) == getTaskSequence().last().id
-
   /**
-   * Returns true if the given [taskId] and task data would be last, or the current active task.
-   * Useful for handling conditional tasks.
+   * Returns true if the given [taskId] with task data would be last in sequence. Defaults to the
+   * current active task if not set. Useful for handling conditional tasks, see #2394.
    */
   fun checkLastPositionWithTaskData(taskId: String? = null, value: TaskData?): Boolean =
     (taskId ?: currentTaskId.value) ==
       getTaskSequence(taskValueOverride = (taskId ?: currentTaskId.value) to value).last().id
+
+  /** Returns true if the given [taskId] is last if set, or the current active task. */
+  fun isLastPosition(taskId: String? = null): Boolean =
+    (taskId ?: currentTaskId.value) == getTaskSequence().last().id
 
   /** Evaluates the task condition against the current inputs. */
   private fun evaluateCondition(
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -391,7 +391,7 @@ internal constructor(
    * Returns true if the given [taskId] and task data would be last, or the current active task.
    * Useful for handling conditional tasks.
    */
-  fun testLastPosition(taskId: String? = null, value: TaskData?): Boolean =
+  fun checkLastPositionWithTaskData(taskId: String? = null, value: TaskData?): Boolean =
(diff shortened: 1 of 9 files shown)


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -381,10 +387,36 @@ internal constructor(
   fun isLastPosition(taskId: String? = null): Boolean =
     (taskId ?: currentTaskId.value) == getTaskSequence().last().id
 
+  /**
+   * Returns true if the given [taskId] and task data would be last, or the current active task.
+   * Useful for handling conditional tasks.
+   */
+  fun testLastPosition(taskId: String? = null, value: TaskData?): Boolean =
```
anandwana001, line 394: is it testing the last position, or taskLastPosition or isTaskOnLastPosition?

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
```diff
@@ -28,7 +28,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.res.vectorResource
 import com.google.android.ground.model.submission.TaskData
 
-class TaskButton(val action: ButtonAction) {
+class TaskButton(var action: MutableState<ButtonAction>) {
```
anandwana001, line 31: is it a good idea to pass a mutable property? 
Only one place should have a mutation responsibility!!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -227,9 +233,22 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
     }
   }
 
-  /** Returns true if the given [ButtonAction] should be replace with "Done" button. */
-  private fun ButtonAction.shouldReplaceWithDoneButton() =
-    this == ButtonAction.NEXT && dataCollectionViewModel.isLastPosition(taskId)
+  /** Returns true if the survey is in the last position. */
```
shobhitagarwal1612, line 236: rename: survey -> task

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -150,7 +150,10 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
   protected fun addNextButton() =
     addButton(ButtonAction.NEXT)
       .setOnClickListener { handleNext() }
-      .setOnValueChanged { button, value -> button.enableIfTrue(value.isNotNullOrEmpty()) }
+      .setOnValueChanged { button, value ->
+        button.enableIfTrue(value.isNotNullOrEmpty())
+        button.applyDoneState(testLastPosition(value))
```
shobhitagarwal1612, line 155: We add the next button in `CaptureLocationTask` and `DropPinTask` as well. This logic should be kept in some shared location.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612, line 394: Can we call this method `isLastPosition` can omit `test` from the method name?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
sufyanAbbasi: Renamed to be slightly more clear, `checkLastPositionWithTaskData()`!

ground/src/main/java/com/google/android/ground/ui/datacollection/components/TaskButton.kt
sufyanAbbasi: Good point, it's a private variable now with a getter and setters, specifically "done()" and "next()".

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
sufyanAbbasi: Done.

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
sufyanAbbasi: Yes, looks like this method's name needs to be clarified, as per @anandwana001's comment too!

So we essentially have two methods, `isLastPosition()`, which checks that the task is truly the last in the sequence, and I renamed this one to `checkLastPositionWithTaskData()`. It would also be trivial to combine these together into one, but I'd prefer this weird edge case thing to be in its own method so that someone can investigate its origin later. I left the bug number as a bread crumb.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/AbstractTaskFragment.kt
sufyanAbbasi: Ah, good catch! Done. I consolidated some of into `TaskButton` but in this case, since it's an overridden method with its own custom action button logic, the best I could do was just add the new callback there.

Technically, we only needed to do this for the MultipleChoice tasks since that's the only way conditional tasks can be triggered, but maybe in the far far future, we'll have conditions based on geographical location or something :)

I tested it with the CaptureLocation as the last task and everything seems to work fine there.

anandwana001 (approved):

sufyanAbbasi:
Sorry! Looks like I need to fix `DataCollectionFragmentTest`. I was certain I ran all the tests in Android Studio, but for some reason they got skipped over.

sufyanAbbasi:
Gosh darn, now `MultipleChoiceTaskFragmentTest` seems to be failing too... looking into it!

sufyanAbbasi:
Oh! I actually just had a silly bug here:
```
    if (action.value == ButtonAction.NEXT && done) {
      done()
    } else if (action.value == ButtonAction.DONE) {
                // ^^^ Always sets DONE to NEXT :P
      next()
    }
```

shobhitagarwal1612 (approved):

PR #2740, opened 2024-09-18
Ignore survey clicks while the survey is being activated.

### LINKED ISSUE #2729: Required survey answers logged as "Skipped"
Relates to issues raised by @jabramowitz5. (Bug report is a bit messy, since it's been drawn from our common Google Chat and with a lot of uncertainties. Please add further findings in comments.)

Collectors in Ghana on mission on Sep 16, 2024 reported frequent crashes. It seems that when they restarted the app and picked up surveys again, some questions had been skipped. This problem occurred with both regular questions set to 'required' as well as with conditional questions set to 'required'. Partially relates to https://github.com/google/ground-platform/issues/1720.

It is already being discussed, that the survey in question (https://ground.openforis.org/survey/SB0BQw8vcz5fUTHzK2aB) might have been too large with 38 tasks, and with a lot of conditionals set to 'required'. But keep in mind that **also regular questions set to 'required' were skipped**.

Crashes could possibly have been triggered by phototaking (no evidence yet).

Could have something to do with crash report https://github.com/google/ground-android/issues/2726.

<br><br><br>
Copy-pasted chat history for reference:

<i>I also noticed some of the entries do not have all of the questions answered, despite all tasks (except for one) being required. Some of the tasks say skipped because they are conditionals and the conditional did not lead to them, which makes sense, but there are several required and non-conditional tasks that show up as "skipped" in the web data viewer. I am not sure how data collectors could have skipped a required task and still have been able to submit the data collection?

For example, all the questions with answers "skipped" were required in the survey and the data collector somehow got past them and answered questions after them and submitted the data collection. I thought they would have had to have answered those questions to be able to move forward in the survey

Just spoke with the project lead in Ghana. There were 7 data collectors out of 30 that had the issues with no "add new data collection site" card popping up and were therefore unable to use Ground app in the field today. Also looking through the data some data entries have required tasks that are missing responses (as mentioned above). The project lead told me he was hearing from data collectors that the app was crashing very often in the field, so perhaps the issue was that the app crashed during the survey and then when they reopened app it has skipped questions in their current collection (just a thought, since I am not sure how they would have otherwise by-passed required questions and then answered later questions and submitted an entry)?

Could there be an issue with large conditionals? For example we have a “select one” question for regions with 6 options, each leading to its own “select one” question with 2-5 options for district, each of those leading to a “select one” question with many answers for community (I think 1 question had close to 80 options listed).

Could this be the source of crashing? Maybe you can see that in Crashlytics?

It looks like we had 124 data entries yesterday from the team. Only 59 however have all of the required questions answered so about half are somehow incomplete entries

I am thinking either this conditional question or the “take a photo” task is causing the crash? Some data collectors reported the “take a photo” task was not working for them. Perhaps certain devices cannot handle that task?</i>

sufyanAbbasi:
Fixes #2729

Prevent secondary loading of survey while the current survey is still being loaded. Unfortunately, there still exists UX problems of the user not getting any feedback about the survey still in the loading state even after the loading bar has gone away.

@gino-m PTAL!

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
@@ -72,6 +72,10 @@ internal constructor(
   fun activateSurvey(surveyId: String) =
     viewModelScope.launch {
       runCatching {
+          // Ignore extra clicks while survey is loading, see #2729.
+          if (_uiState.value is UiState.ActivatingSurvey) {
+            return@runCatching
+          }
           _uiState.emit(UiState.ActivatingSurvey)
           activateSurveyUseCase(surveyId)
         }
--- a/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
@@ -73,11 +73,10 @@ internal constructor(
     viewModelScope.launch {
       runCatching {
           // Ignore extra clicks while survey is loading, see #2729.
-          if (_uiState.value is UiState.ActivatingSurvey) {
-            return@runCatching
+          if (_uiState.value !is UiState.ActivatingSurvey) {
+            _uiState.emit(UiState.ActivatingSurvey)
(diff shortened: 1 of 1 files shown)


gino-m (approved):
LGTM!

PR #2742, opened 2024-09-19
Navigate To Home only after Survey Activated and Subscribed

anandwana001:
ActivateSurveyUseCase and MakeSurveyAvailableOfflineUseCase are somehow working in parallel internally

Earlier, `onSuccess` is not called only after complete all the steps of activating + making survey offline

and because of that, onSuccess invoked, emitting SurveyActivated, in a partial phase, user was able to navigateToHome and the ongoing job, which is making the survey offline gets cancelled 

https://github.com/user-attachments/assets/8ee6898a-d246-445b-aff2-8dc254f741eb


anandwana001:
Closing this PR as we have other workaround 
https://github.com/google/ground-android/pull/2745

need a better look later on

PR #2744, opened 2024-09-19
Replace "isPredefined" with "source" field in LOI and mutation

shobhitagarwal1612:
Towards #2717

Discussed in https://docs.google.com/spreadsheets/d/1Ognh710debgAEI01Z2qvl-jAIfOoxnH6O8z4npiL9_g/r/0-ZZ16BvSYLPvARlghXBtX4A/edit?gid=0#gid=0

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/Config.kt
```diff
@@ -25,7 +25,7 @@ object Config {
   const val SHARED_PREFS_MODE = Context.MODE_PRIVATE
 
   // Local db settings.
-  const val DB_VERSION = 120
+  const val DB_VERSION = 121
```
shobhitagarwal1612, line 28: @gino-m Is it ok for destructive migration or should we start adding migrations to prevent data loss?

gino-m: We should aim to preserve data collected so far in the field, in which case we would want to implement migrations. But do we know if the user stays signed into the app after upgrading the app in Play Store?

shobhitagarwal1612: Not today, if there are any breaking changes in the db due to destructive migration.

gino-m: I would advise punting on this issue, saving the migration for when we switch over to serialized protos locally.

gino-m: Apologies that my comment on the sheet didn't take into consideration that the local db would need to be updated.

PR #2745, opened 2024-09-19
[Survey list] Ignore clicks on survey while loading

gino-m:
Follow-up to #2740, which didn't address the case of multiple clicks before the survey activation process start. This should handle all cases.


anandwana001 (approved):
LGTM for a work around for now

PR #2748, opened 2024-09-19
ViewModel is now access based on taskId using Bundle

### LINKED ISSUE #2725: 0.1.9-Openforis DrawAreaTaskMapFragment App Crash
```
          Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.openforis.ground/com.google.android.ground.MainActivity}: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.google.android.ground.ui.datacollection.tasks.polygon.DrawAreaTaskMapFragment: could not find Fragment constructor
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4035)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4201)
       at android.app.ActivityThread.handleRelaunchActivityInner(ActivityThread.java:6347)
       at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:6212)
       at android.app.servertransaction.ActivityRelaunchItem.execute(ActivityRelaunchItem.java:71)
       at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2438)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8663)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:567)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
```

anandwana001:
Fixes #2725 

@shobhitagarwal1612 @gino-m @sufyanAbbasi  PTAL?


gino-m (approved):
@anandwana001  Could you please do the same for other TaskFragments, either in this PR or in a follow-up?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
```diff
@@ -46,18 +46,27 @@ class CaptureLocationTaskMapFragment(private val viewModel: CaptureLocationTaskV
...
     binding.locationLockBtn.isClickable = false
-    viewLifecycleOwner.lifecycleScope.launch { viewModel.onMapReady(mapViewModel) }
+    viewLifecycleOwner.lifecycleScope.launch {
+      if (this@CaptureLocationTaskMapFragment::viewModel.isInitialized) {
+        viewModel.onMapReady(mapViewModel)
+      }
+    }
+  }
+
+  fun setViewModel(viewModel: CaptureLocationTaskViewModel) {
```
trietbui85, line 65: Might I know why don't we use `private val viewModel: CaptureLocationTaskViewModel by viewModels()`, but use the setter instead?

anandwana001: because we don't want another instance of it, and need to access the one which was there in previous fragment, and  the other side, if we try accessing it using the parentFragmentManager, the parent is DataCollectionFragment not the CaptureLocationTaskFragment,

Any other approach if you could here?

anandwana001: Also, it will throw error

As an example
```
java.lang.RuntimeException: Cannot create an instance of class com.google.android.ground.ui.datacollection.tasks.polygon.DrawAreaTaskViewModel
```

anandwana001: The ViewModel instance is created by `AbstractTaskFragment`

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
```diff
@@ -41,13 +41,10 @@ class CaptureLocationTaskFragment : AbstractTaskFragment<CaptureLocationTaskView
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical uniqueness.
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
     val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11149 }
+    val fragment = CaptureLocationTaskMapFragment.newInstance(map, super.taskId)
```
shobhitagarwal1612, line 44: We can use `taskId` without `super` keyword

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
```diff
@@ -22,19 +22,33 @@ import android.view.ViewGroup
...
+      try {
+        val taskId = it.getString("taskId")
+        taskId?.let {
+          viewModel =
+            (requireParentFragment() as DataCollectionFragment).viewModel.getTaskViewModel(taskId)
+              as CaptureLocationTaskViewModel
+        }
+      } catch (e: Exception) {
+        Timber.e("CaptureLocationTaskMapFragment - $e")
+      }
```
shobhitagarwal1612, line 50: Please cleanup the debug logs.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskFragment.kt
```diff
@@ -47,9 +47,10 @@ class DropPinTaskFragment : AbstractTaskFragment<DropPinTaskViewModel>() {
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical "uniqueness".
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
     val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11617 }
+    val fragment = DropPinTaskMapFragment.newInstance(map, super.taskId)
```
shobhitagarwal1612, line 50: Same as above.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -18,34 +18,60 @@ package com.google.android.ground.ui.datacollection.tasks.point
...
+      try {
+        val taskId = it.getString("taskId")
+        taskId?.let {
+          viewModel =
+            (requireParentFragment() as DataCollectionFragment).viewModel.getTaskViewModel(taskId)
+              as DropPinTaskViewModel
+        }
+      } catch (e: Exception) {
+        Timber.e("DropPinTaskMapFragment - $e")
+      }
```
shobhitagarwal1612, line 46: Please cleanup debug logs

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -18,34 +18,60 @@ package com.google.android.ground.ui.datacollection.tasks.point
...
   override fun onMapCameraMoved(position: CameraPosition) {
     super.onMapCameraMoved(position)
-    viewModel.updateCameraPosition(position)
+    if (this@DropPinTaskMapFragment::viewModel.isInitialized) {
+      viewModel.updateCameraPosition(position)
+    }
+  }
+
+  fun setViewModel(viewModel: DropPinTaskViewModel) {
+    this.viewModel = viewModel
```
shobhitagarwal1612, line 66: Can we remove this setter now?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -59,7 +59,7 @@ class DrawAreaTaskFragment : AbstractTaskFragment<DrawAreaTaskViewModel>() {
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical "uniqueness".
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
     val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11411 }
-    drawAreaTaskMapFragment = DrawAreaTaskMapFragment.newInstance(viewModel, map)
+    drawAreaTaskMapFragment = DrawAreaTaskMapFragment.newInstance(map, super.taskId)
```
shobhitagarwal1612, line 62: Same as above.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -19,36 +19,52 @@ import android.os.Bundle
...
+      try {
+        val taskId = it.getString("taskId")
+        taskId?.let {
+          viewModel =
+            (requireParentFragment() as DataCollectionFragment).viewModel.getTaskViewModel(taskId)
+              as DrawAreaTaskViewModel
+        }
+      } catch (e: Exception) {
+        Timber.e("DrawAreaTaskMapFragment - $e")
+      }
```
shobhitagarwal1612, line 49: Please cleanup

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
anandwana001: We do need this, so that we can get in production, right?

sufyanAbbasi: Maybe also add the new method, `FirebaseCrashLogger().logException(t)` here if it's safe to do so, and in the other log areas?

And one more thing, by catching the error without rethrowing it, the app will be in a weird state where no `viewModel` is set and we skip everything that is in the `isInitialized` check. But these are critical for the functionality of the app, like updating map pins or capturing the location.

My question is, if the app finds itself in an impossible to recover state like this, should we log the exception and try to continue, or should we go back to the home screen and show and error to the user "An unexpected error has occurred. Please try again or contact your survey organizer"?  I'm not sure how to do that off the top of my head, so maybe just navigate to the home screen after logging. Also true for below.

sufyanAbbasi (commented):
Thank you so much! I'm not Android expert enough to be able to verify the correctness of this, at least not without pulling your PR and giving it a test, but I had a bigger question about error handling here, since swallowing the error means the app is in an even weirder state.

@gino-m, do you have any thoughts about how to handle these irrecoverable states without crashing?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
gino-m: It's better that the app crashes than remains in an invalid state. But if we know that that can happen, we should try to fix the root cause rather than eating the exception.

gino-m: Also, @anandwana001 messages logged with `Timber.e()` aren't captured anywhere in prod. If you need to communicate critical debug info via Crashlytics, the method @sufyanAbbasi mentioned should work.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -18,34 +18,56 @@ package com.google.android.ground.ui.datacollection.tasks.point
...
+    arguments?.let {
+      try {
+        val taskId = it.getString("taskId")
+        taskId?.let {
+          viewModel =
+            (requireParentFragment() as DataCollectionFragment).viewModel.getTaskViewModel(taskId)
+              as DropPinTaskViewModel
+        }
+      } catch (e: Exception) {
+        Timber.e("DropPinTaskMapFragment - $e")
```
gino-m, line 45: `Timber.e` already reports the name of the class. Also `$e` calls `e.toString()`. Instead, you can just let Timber format the message stack trace `Timber.e(t)`

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -18,34 +18,56 @@ package com.google.android.ground.ui.datacollection.tasks.point
...
-    viewModel.features.observe(this) { map.setFeatures(it) }
+    if (this@DropPinTaskMapFragment::viewModel.isInitialized) {
+      viewModel.features.observe(this) { map.setFeatures(it) }
+    }
   }
 
   override fun onMapCameraMoved(position: CameraPosition) {
     super.onMapCameraMoved(position)
-    viewModel.updateCameraPosition(position)
+    if (this@DropPinTaskMapFragment::viewModel.isInitialized) {
```
gino-m, line 60: Swallowing the invalid state here as well will make the issue impossible to identify in prod and even difficult to find locally.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
anandwana001: Isn't it a regular error handling where we pass arguments to a new fragment and use it?
Not in a weird app state because this is something required to do. 
yes, we can use `FirebaseCrashLogger().logException(t)`

anandwana001: > It's better that the app crashes than remains in an invalid state. But if we know that that can happen, we should try to fix the root cause rather than eating the exception.

This is more of a workaround we can do and user can just do a back and forth and keep continuing the data collection, no more crashes.

Later, as a core solution, we can work on the abstract layer and the ViewModel and support the orientation change to work things better
@gino-m

gino-m: > Isn't it a regular error handling where we pass arguments to a new fragment and use it?

Not sure what you mean about passing arguments to a new fragment is a regular error handling.. Can you rephrase?

> Not in a weird app state because this is something required to do.

It's generally considered bad practice to catch `Throwable` without managing potential consequences. The app will be in an invalid state if the view model couldn't be set since the view requires it to function, no?

anandwana001:
solved in PR https://github.com/google/ground-android/pull/2761#pullrequestreview-2323896876

Issue #2682, opened 2024-08-22
App Crash - Data Collection Screen - Orientation Change


anandwana001:
**Describe the bug**
While collecting data for `DropPinTaskMapFragment`, If we rotate the device, the app gets crashes with following logs

```
FATAL EXCEPTION: main (Ask Gemini)
Process: com.google.android.ground, PID: 4605
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.google.android.ground/com.google.android.ground.MainActivity}: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.google.android.ground.ui.datacollection.tasks.point.DropPinTaskMapFragment: could not find Fragment constructor
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4047)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4234)
at android.app.ActivityThread.handleRelaunchActivityInner(ActivityThread.java:6303)
at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:6194)
at android.app.servertransaction.ActivityRelaunchItem.execute(ActivityRelaunchItem.java:79)
at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:60)
at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:174)
at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:109)
at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:81)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2635)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loopOnce(Looper.java:232)
at android.os.Looper.loop(Looper.java:317)
at android.app.ActivityThread.main(ActivityThread.java:8699)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:886)
Caused by: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.google.android.ground.ui.datacollection.tasks.point.DropPinTaskMapFragment: could not find Fragment constructor
at androidx.fragment.app.Fragment.instantiate(Fragment.java:692)
[...]

Issue #2624, opened 2024-08-09
Auto-rotate (landscape mode) doesn't work on many screens


gino-m:
@anandwana001 Please add subtasks here and create related issues:

- [x] #2682
- [x] #2681
- [x] #2794
- [x] #2795
- [x] #2796 
- [x] #2798
- [x] Crash on rotate of ToS screen
- [x] Thoroughly test rotation on all screens (including all task types)
- [x] Ensure location lock is restored and works properly after rotate
- [x] #3053
- [x] Remove `android:screenOrientation="portrait"` from AndroidManifest.xml to enable screen rotation
- [x] #3029

kenstershiro:
@jcqli to move sub-tasks into next iteration

gino-m:
@anandwana001 Can you please test rotation on all app screens and add UI tests for all screens.

gino-m:
Rotation is still not enabled, reopening.

PR #2761, opened 2024-09-23
[Draw area task] Fix crashes on config change

### LINKED ISSUE #2725: 0.1.9-Openforis DrawAreaTaskMapFragment App Crash

gino-m:
* Instead of passing the view model in constructor, this PR gets the vm from the parent fragment.
* The task ID is passed as an arg to the child fragment so that it knows which view model to request from its parent.
* The camera move handler requires the map to be initialized, so the `collect` call is moved to after the map has loaded.
* Uses an injected `Provider` to create new instances so that `@Inject` works.
* Reinitialize `DataCollectionViewModel` on config changes so that view is restored.

Fixes #2725
Towards #2624,  #2682

[Screen_recording_20240918_224809.webm](https://github.com/user-attachments/assets/03b06b17-47f7-4351-9e47-cc8ea1593ad7)

The data collection flow still reverts to the previous step in the flow on config change, and exits if in the first position, but this is likely an unrelated issue.


ground/src/main/AndroidManifest.xml
```diff
@@ -60,7 +65,6 @@
     <activity
       android:name=".MainActivity"
       android:exported="true"
-      android:screenOrientation="portrait"
```
gino-m, line 63: Removed this for debugging purposes, will revert before sending for review.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -19,26 +19,33 @@ import android.os.Bundle
...
   private lateinit var mapViewModel: BaseMapViewModel
+  private lateinit var viewModel: DrawAreaTaskViewModel
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+
     mapViewModel = getViewModel(BaseMapViewModel::class.java)
+    val taskId = arguments?.getString("taskId")
+    val dcf = requireParentFragment() as DataCollectionFragment
+    viewModel = dcf.viewModel.getTaskViewModel(taskId!!) as DrawAreaTaskViewModel
```
anandwana001, line 42: `taskId!!` is something can crash

similar approach https://github.com/google/ground-android/pull/2748 but wrapped up in try catch or maybe some other error. handling approach we need

anandwana001: maybe we can have an appraoch, where we don't have to do `dcf.viewModel.getTaskViewModel` for every task

gino-m: How would you recover from taskId not being set in the fragment args? I don't think we can...

gino-m: > 
> maybe we can have an appraoch, where we don't have to do `dcf.viewModel.getTaskViewModel` for every task

What do you suggest?

gino-m:
Config changes are now supported on DrawAreaTask:

[Screen_recording_20240923_172230.webm](https://github.com/user-attachments/assets/635d4901-50db-4450-bdc6-f896b1b6aa49)

Unsaved draft Changes are still restored when the app is killed:

[Screen_recording_20240923_172342.webm](https://github.com/user-attachments/assets/fd083d2f-d5a3-4961-95c4-a088d52fa70c)

@anandwana001 would you like to apply a similar fix to remaining Task(Map)Fragments?

ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
```diff
@@ -95,6 +96,8 @@ abstract class BaseTaskFragmentTest<F : AbstractTaskFragment<VM>, VM : AbstractT
       preTransactionAction = {
         fragment = this as F
         fragment.taskId = task.id
+        // TODO: Attach `fragment` to mock DataCollectionFragment
```
gino-m, line 99: @shobhitagarwal1612 @anandwana001 `DrawAreaTaskFragmentTest` is failing with:

```
Fragment DrawAreaTaskMapFragment{44fdc5b3} (07ec492d-1efe-474f-94ab-b361fdc1063e id=0x2c93 tag=DrawAreaTaskMapFragment) is not a child Fragment, it is directly attached to dagger.hilt.android.internal.managers.ViewComponentManager$FragmentContextWrapper@33500213
java.lang.IllegalStateException: Fragment DrawAreaTaskMapFragment{44fdc5b3} (07ec492d-1efe-474f-94ab-b361fdc1063e id=0x2c93 tag=DrawAreaTaskMapFragment) is not a child Fragment, it is directly attached to dagger.hilt.android.internal.managers.ViewComponentManager$FragmentContextWrapper@33500213
	at androidx.fragment.app.Fragment.requireParentFragment(Fragment.java:1183)
...
```

I believe we have to update the way we mock out the task view model here, but not sure how. Any ideas?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -19,26 +19,33 @@ import android.os.Bundle
...
 
   private lateinit var mapViewModel: BaseMapViewModel
+  private lateinit var viewModel: DrawAreaTaskViewModel
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+
     mapViewModel = getViewModel(BaseMapViewModel::class.java)
+    val taskId = arguments?.getString(TASK_ID_ARG_KEY) ?: error("null taskId arg")
+    val dcf = requireParentFragment() as DataCollectionFragment
```
anandwana001, line 41: `requireParentFragment()` can throw an exception, can we handle it?

This is the edge case, if this fails, no ViewModel will be there

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -57,7 +64,6 @@ class DrawAreaTaskMapFragment(private val viewModel: DrawAreaTaskViewModel) :
   }
 
   companion object {
-    fun newInstance(viewModel: DrawAreaTaskViewModel, map: MapFragment) =
-      DrawAreaTaskMapFragment(viewModel).apply { this.map = map }
+    const val TASK_ID_ARG_KEY = "taskId"
```
anandwana001, line 67: AbstractTaskFragment already has this, how about using the same rather than creating this new?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
gino-m: Can that ever happen in practice, and is it something we can recover from?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
gino-m: AbstractTaskFragment has the key used in `Bundle`, not Fragment args. They happen to have the same value, but that's coincidental rather than necessary.

ground/src/test/java/com/google/android/ground/ui/datacollection/tasks/BaseTaskFragmentTest.kt
gino-m: Resharing @anandwana001 's suggestion via chat:

adding childfragment:

```kt
inline fun <reified T : Fragment> launchFragmentWithNavController(
  fragmentArgs: Bundle? = null,
  @StyleRes themeResId: Int = R.style.FragmentScenarioEmptyFragmentActivityTheme,
  destId: Int,
  fragment: Fragment?=null,
  crossinline preTransactionAction: Fragment.() -> Unit = {},
  crossinline postTransactionAction: Fragment.() -> Unit = {},
): ActivityScenario<HiltTestActivity> =
  hiltActivityScenario(themeResId).launchFragment<T>(
    fragmentArgs,
    {
      // Attach `fragment` as a child of the mock `DataCollectionFragment`
      parentFragment?.childFragmentManager?.beginTransaction()
        ?.add(R.id.data_collection_fragment, fragment!!, "childFragmentTag")
        ?.commitNow()

      this.preTransactionAction()
      viewLifecycleOwnerLiveData.observeForever { viewLifecycleOwner ->
        if (viewLifecycleOwner != null) {
          val navController = TestNavHostController(getApplicationContext())
          navController.setViewModelStore(ViewModelStore())
          // Required for graph scoped view models.
          navController.setGraph(R.navigation.nav_graph)
          navController.setCurrentDestination(destId, fragmentArgs ?: bundleOf())

          // Bind the controller after the view is created but before onViewCreated is called.
          Navigation.setViewNavController(requireView(), navController)
        }
      }
    },
  ) {
    this.postTransactionAction()
  }
```

and pass the parent fragment

```kt
launchFragmentWithNavController<T>(
  destId = R.id.data_collection_fragment,
  fragment = DataCollectionFragment(),
  preTransactionAction = {
    // Set up fragment-specific properties
    fragment = this as F
    fragment.taskId = [task.id](http://task.id/)
  }
)
```

gino-m: @shobhitagarwal1612 Any ideas on the best way to create and attach `DataCollectionFragment` as the parent to the task under test here? We need to get the view model used by the tasks.

gino-m: @shobhitagarwal1612 Gentle ping

anandwana001 (approved):

PR #2768, opened 2024-09-25
Keep Showing Progress Dialog on while Survey Loads

### LINKED ISSUE #2767: Show Progress Dialog until Survey Load
Currently, the progress dialog gets dismissed and user can click on survey multiple times but nothing happens. We need to show the progress dialog to let user know that the selected survey is in the loading state and will soon home screen opens up

anandwana001:
Fixes #2767 

@gino-m @shobhitagarwal1612  PTAL?

https://github.com/user-attachments/assets/2f37f938-fea0-4c3c-8e87-e11bf8dc52ad


ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -55,11 +55,15 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
         showProgressDialog()
       }
       UiState.SurveyActivated -> {
-        dismissProgressDialog()
+        if (!viewModel.surveyActivationInProgress.value) {
```
gino-m, line 58: Why would `SurveyActivated` state be set while the survey activation is still in progress? Is this check really just masking the root cause?

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -55,11 +55,15 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
...
+          dismissProgressDialog()
+        }
       }
       is UiState.SurveyListAvailable -> {
         handleSurveyListUpdated(uiState.surveys)
-        dismissProgressDialog()
+        if (!viewModel.surveyActivationInProgress.value) {
```
gino-m, line 64: Same question here.

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -53,7 +53,9 @@ internal constructor(
 
   private val _uiState: MutableStateFlow<UiState> = MutableStateFlow(UiState.FetchingSurveys)
   val uiState: StateFlow<UiState> = _uiState.asStateFlow()
-  private var surveyActivationInProgress = false
+
+  private val _surveyActivationInProgress = MutableStateFlow<Boolean>(false)
+  var surveyActivationInProgress = _surveyActivationInProgress.asStateFlow()
```
gino-m, line 58: This is never actually used as a flow in this PR, since it's always being accessed via `value`. In that case, why define it as such?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
anandwana001: Not masking the root cause

User selects a survey -> ActivatingSurvey State -> Show progress dialog -> **SurveyListAvailable State** (**this state is getting emitted again, due to re-arrangment of the survey activation list, emitting SurveyListAvailable state and the dialog is getting dismissed before onSuccess**) -> onSuccess -> SurveyActivated State -> navigate to home

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
anandwana001: changed to primitive

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
anandwana001: answered above
https://github.com/google/ground-android/pull/2768#discussion_r1776222807

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
anandwana001: The reason to use the flow is because of baking property, without it, the value can be changed from outside, which we don't want, and if going with a primitive type, it will not trigger as soon the value changes.

Updated to primitive tye with one variable, but not a healthy approach as the value can be changed from outside which we don't want/.

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
gino-m: Ok, so it sounds like there's a circular state dependency.. perhaps the list should simplfy stop updating once a survey is tapped?

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
gino-m: What do you mean "can be changed from outside"? The member was `private`, so the value could not be changed from outside this class.

anandwana001: yes, but this PR made the variable public, so that fragment can access it

gino-m: Sorry I missed that. Can you just make it a `val` instead of a `var`?

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorFragment.kt
gino-m: Ok, but perhaps the UI state should pertain to the initial load and not reactive updates? The root cause seems to be the circle dependency on the state flow and the user tap. Ideally we would break that loop.

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -53,7 +53,8 @@ internal constructor(
 
   private val _uiState: MutableStateFlow<UiState> = MutableStateFlow(UiState.FetchingSurveys)
   val uiState: StateFlow<UiState> = _uiState.asStateFlow()
-  private var surveyActivationInProgress = false
+
+  var surveyActivationInProgress = false
```
anandwana001, line 57: This is the latest one

ground/src/main/java/com/google/android/ground/ui/surveyselector/SurveySelectorViewModel.kt
anandwana001: jumping the conversation to different thread
https://github.com/google/ground-android/pull/2768#discussion_r1782042009

gino-m (approved):
Approving this as a workaround, but we should rethink how states and events are managed here. Filed https://github.com/google/ground-android/issues/2781 to track.

PR #2770, opened 2024-09-27
Add values-fr folder with French translation of app

jo-spek:

Issue #2676, opened 2024-08-21
French translation


jo-spek:
Would it take long to implement a language switch button in the settings?

I'm happy to take the string.xml and translate everything to French if that helps. Doesn't seem to be all that much.

gino-m:
The app currently uses the device language, but shouldn't be hard to allow users to select the language under settings. @shobhitagarwal1612 wdyt?

Translating sounds good, we should also tidy up the file a bit to make it easier to translate:

- [ ] Clean up unused strings in strings.xml
- [ ] Add comments describing where each string is used
- [ ] Do actual translations

shobhitagarwal1612:
Yes, we can add a setting to switch the language of the app. However, that won't change the language of the questions in the data collection screen since they are populated from firestore.

jo-spek:
> Yes, we can add a setting to switch the language of the app. However, that won't change the language of the questions in the data collection screen since they are populated from firestore.

Do you mean the questions that the survey organizer creates?

shobhitagarwal1612:
> Do you mean the questions that the survey organizer creates?

Yes

jo-spek:
That's alright. I'm just talking about translating the GUI here. We'll have a training in Togo soon and having the GUI in French would make things a lot easier. The individual survey questions would then also be made in French by the participants.
I'm halfway through translating the strings.xml. Can I just send it to you later or what's the best way to go about this?

shobhitagarwal1612:
Please feel free to raise a PR.

jo-spek:
> Please feel free to raise a PR.

I created a new folder "values-fr" under "ground/src/main/res/" containing the translated strings.xml in my fork. The CI workflow run outputs that 'All jobs have failed'.

The errors all look like this:
` /home/runner/work/ground-android/ground-android/ground/src/main/res/values-fr/strings.xml:63:2: string/admin_request_access does not contain a valid string resource.
/home/runner/work/ground-android/ground-android/ground/src/main/res/values-fr/strings.xml:62:2: Failed to flatten XML for resource 'agree_checkbox' with error: Invalid unicode escape sequence in string
"{str}"`

When uploading the original strings.xml in the same directory it passes.

Should I do a Pull Request nonetheless?

Here's the translated strings.xml: [strings.zip](https://github.com/user-attachments/files/16760274/strings.zip)

jo-spek:
[...]

PR #2771, opened 2024-09-28
Add french translations

shobhitagarwal1612:
Towards #2676

Forks https://github.com/google/ground-android/pull/2770 

Changes made:
 * Removes obsolete strings which have been removed from the app
 * Escapes special characters

@gino-m @jo-spek PTAL?


gino-m:
Let's let @jo-spek finish #2770 instead so he can retain authorship. Feel free to commit the tweaks you did here to that PR!

PR #2793, opened 2024-10-13
Adds offline warning to signout dialog and minor refactors

### LINKED ISSUE #2643: Warn user that they won't be able to sign back in if they sign out while offline
Currently we only warn that unsynced data will be lost. We should also inform the user that:

* Offline surveys will no longer be available until their sign back in
* They won't be able to sign in until they're back online

image

@amegantz Suggestions here?

shobhitagarwal1612:
Fixes #2643

This PR updates the message in signout dialog to warn users regarding not being able to sign in again when in offline mode.
Also adds some minor readability refactors in the related dialog composable code.

[signout_dialog_message.webm](https://github.com/user-attachments/assets/d5bd3484-6b3f-4fa9-a5ed-5a7c4d11ed45)

@anandwana001 @gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
@@ -185,9 +185,13 @@ class HomeScreenFragment :
             UserDetailsDialog(showUserDetailsDialog, showSignOutDialog, user)
           }
           if (showSignOutDialog.value) {
-            SignOutConfirmationDialog(showUserDetailsDialog, showSignOutDialog) {
-              homeScreenViewModel.signOut()
-            }
+            SignOutConfirmationDialog(
+              { homeScreenViewModel.signOut() },
+              {
+                showUserDetailsDialog.value = false
+                showSignOutDialog.value = false
+              },
+            )
           }
         }
       }
--- a/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
@@ -173,35 +173,33 @@ class HomeScreenFragment :
     // compose.
     binding.composeView.apply {
       setContent {
-        val showUserDetailsDialog = remember { mutableStateOf(true) }
+        val showUserDetailsDialog = remember { mutableStateOf(false) }
         val showSignOutDialog = remember { mutableStateOf(false) }
 
-        // reset the state for recomposition
-        showUserDetailsDialog.value = true
-        showSignOutDialog.value = false
+        fun showUserDetailsDialog() {
+          showUserDetailsDialog.value = true
+          showSignOutDialog.value = false
+        }
+
+        fun showSignOutDialog() {
+          showUserDetailsDialog.value = false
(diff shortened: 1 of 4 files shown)


ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
```diff
@@ -173,21 +173,33 @@ class HomeScreenFragment :
...
+          showSignOutDialog.value = true
+        }
+
+        fun hideAllDialogs() {
+          showUserDetailsDialog.value = false
+          showSignOutDialog.value = false
+        }
+
+        // Init state for composition
+        showUserDetailsDialog()
```
anandwana001, line 195: We can do this with LaunchedEffect, right?

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
```diff
@@ -173,21 +173,33 @@ class HomeScreenFragment :
...
+
+        fun showSignOutDialog() {
+          showUserDetailsDialog.value = false
+          showSignOutDialog.value = true
+        }
+
+        fun hideAllDialogs() {
+          showUserDetailsDialog.value = false
+          showSignOutDialog.value = false
+        }
```
anandwana001, line 192: can we manage this using state rather then nested function?
WDYT?

shobhitagarwal1612: I haven't changed the business logic. Just created new helper methods for improved readability. What exactly are you proposing? Can you please elaborate?

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
shobhitagarwal1612: We don't want it to be triggered again and again on state change.

ground/src/main/java/com/google/android/ground/ui/home/HomeScreenFragment.kt
anandwana001: Rather than having these three nested functions which maintains the `value - true/false`, can we use, some state for managing it?

```kt
showUserDetails -> { }
showSignOut -> { }
hideAll -> { }
```

shobhitagarwal1612: We already have 2 states: `showSignOutDialog` and `showUserDetailsDialog` which are bound to the dialogs. Are you proposing to replace those as well?

anandwana001: no, I mean removing these functions

```kt
fun showUserDetailsDialog() {}
fun showSignOutDialog() {}
fun hideAllDialogs() {}
```

and updating the states based on some single source state or any stateflow which can emit some data telling what to make false, what to make true
```
 showUserDetailsDialog.value = true
          showSignOutDialog.value = false

   showUserDetailsDialog.value = false
          showSignOutDialog.value = true

showUserDetailsDialog.value = false
          showSignOutDialog.value = false
```

anandwana001 (approved):

PR #2814, opened 2024-11-08
Added more debug info for missing jobs

### LINKED ISSUE #2813: [Code health] Add Missing Job Debug Info
Add Missing Job Debug Info

anandwana001:
Fixes #2813 

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
@@ -42,7 +42,11 @@ object LoiConverter {
     val loiProto = LocationOfInterestProto::class.parseFrom(doc, 1)
     val geometry = loiProto.geometry.toGeometry()
     val jobId = loiProto.jobId
-    val job = DataStoreException.checkNotNull(survey.getJob(jobId), "job $jobId")
+    val job =
+      DataStoreException.checkNotNull(
+        survey.getJob(jobId),
+        "job $jobId \n loiId = $loiId \n loiProto = $loiProto",
+      )
     // Degrade gracefully when audit info missing in remote db.
     val created = AuditInfoConverter.toAuditInfo(loiProto.created)
     val lastModified =


shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
```diff
@@ -42,7 +42,11 @@ object LoiConverter {
     val loiProto = LocationOfInterestProto::class.parseFrom(doc, 1)
     val geometry = loiProto.geometry.toGeometry()
     val jobId = loiProto.jobId
-    val job = DataStoreException.checkNotNull(survey.getJob(jobId), "job $jobId")
+    val job =
+      DataStoreException.checkNotNull(
+        survey.getJob(jobId),
+        "job $jobId \n loiId = $loiId \n loiProto = $loiProto",
+      )
```
gino-m, line 49: This still doesn't make it to Crashlytics, it's getting swallowed and written to local logs here:

https://github.com/google/ground-android/blob/d704635e2b4885615472a2423dc26f8b6e6eedf1/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiCollectionReference.kt#L63-L70

Instead of just logging a warning we should send the error to Crashlytics.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
gino-m: Example:

https://github.com/google/ground-android/blob/d704635e2b4885615472a2423dc26f8b6e6eedf1/ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt#L58

anandwana001: updated

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
```diff
@@ -43,10 +44,17 @@ object LoiConverter {
     val geometry = loiProto.geometry.toGeometry()
     val jobId = loiProto.jobId
     val job =
-      DataStoreException.checkNotNull(
-        survey.getJob(jobId),
-        "job $jobId \n loiId = $loiId \n loiProto = $loiProto",
-      )
+      try {
```
gino-m, line 47: Why use try / catch instead of just adding `if (job == null)` block here?

But more generally, can't we add this logging higher up in the call stack, in the code I referenced in my original comment? That is where all errors in the conversion process are handled.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/remote/firebase/schema/LoiConverter.kt
anandwana001: > Why use try / catch instead of just adding if (job == null) block here?

Because `checkNotNull` internally throws the exception rather than just returning null object. So, this way the job object will never become null


> But more generally, can't we add this logging higher up in the call stack, in the code I referenced in my original comment? That is where all errors in the conversion process are handled.

Updated to the higher entry point

gino-m (approved):

PR #2818, opened 2024-11-11
[Capture location] Prompt for location permissions when capture location task is opened

### LINKED ISSUE #2658: [Capture location] Prompt for location permissions when capture location task is opened

anandwana001:
Fixes #2658 

@shobhitagarwal1612  PTAL?

## Case 1
Happy Case - Location Permission Given on Home Page, before Data Collection
https://github.com/user-attachments/assets/31e28419-f758-4c7a-8c02-6b9e2822ca56

## Case 2
If the user does not allow the location requested inside the CaptureLocation Data Collection
https://github.com/user-attachments/assets/e9b86938-a9ab-4b33-b804-142c6695ae1b

## Case 3
If the user allow the location by using the dialog allow feature and turning on off device location
https://github.com/user-attachments/assets/03b1a197-07a1-4896-95f0-56b09270c7f6

## Other cases happening
- User is able to click on location lock button now
- If User Cancel, then the capture button will be disabled
- As per the Android SDK, is user has selected, Don't Allow, the Android permission request Dialog will not be shown to user again and again
- User can either skip, or if task is required, user can cancel by using the cross button


ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
```diff
@@ -57,9 +61,42 @@ class CaptureLocationTaskMapFragment @Inject constructor() : AbstractMapFragment
...
+          val openAlertDialog = remember { mutableStateOf(true) }
+          when {
+            openAlertDialog.value -> {
+              AppTheme {
+                LocationPermissionDialog(
+                  onCancel = {
+                    binding.locationLockBtn.isClickable = false
+                    openAlertDialog.value = false
+                  },
+                  onDismiss = { openAlertDialog.value = false },
```
gino-m, line 93: Do we also need to remove the dialog from the view when closed to avoid a leak due to dialogs being added multiple times to the view?

ground/src/main/res/values/strings.xml
```diff
@@ -188,4 +188,8 @@
   <string name="data_submitted_image">Data submitted image</string>
   <string name="data_collection_complete">Data collection complete!</string>
   <string name="data_collection_complete_details">Your data has been saved and will be automatically synced when you’re online.</string>
+
+  <string name="allow_location_confirmation">Allow Location</string>
```
gino-m, line 192: Please update based on design in https://github.com/google/ground-android/issues/2658#issuecomment-2402992661. Be sure to use "Sentence case" for headings instead of "Title Case".

gino-m (changes requested):
LG, just a few nits.

ground/src/main/res/values/strings.xml
anandwana001: done

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskMapFragment.kt
anandwana001: done, removing on dispose off

gino-m (approved):

PR #2842, opened 2024-11-17
Remove logic specific to handling multi-user data sync

### LINKED ISSUE #2837: [Code health] Delete unused multi-user code in LocalMutationSyncWorker
There will no longer be mutations in the db for multiple user, only for the currently signed in user. Related code can be deleted.

shobhitagarwal1612:
Fixes #2837

Removes logic specific to handling data sync for multiple users as it is not supported at the moment. 

@gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
@@ -110,11 +110,9 @@ constructor(
     if (userIds.size != 1) {
       Timber.e("Expected exactly 1 user, but found ${userIds.size}")
     }
-    val filteredMutations = mutations.filter { it.userId == user.id }
-    if (mutations.isNotEmpty() && filteredMutations.isEmpty()) {
-      Timber.e("Could not find mutations for the logged in user")
-    }
-    return filteredMutations
+    val (validMutations, invalidMutations) = mutations.partition { it.userId == user.id }
+    invalidMutations.forEach { Timber.e("Invalid mutation: $it") }
+    return validMutations
   }
 
   companion object {


ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -82,63 +83,55 @@ constructor(
...
-  private suspend fun getUser(userId: String): User? {
-    val user = localUserStore.getUserOrNull(userId)
+  /**
+   * Returns the user associated with the mutations. If missing, removes the mutations from local
+   * storage as well.
+   */
+  private suspend fun getUserFromMutations(mutations: List<Mutation>): User? {
+    val userIds = mutations.map { it.userId }.toSet()
+
+    check(userIds.size == 1) { "Found mutations for multiple users" }
```
gino-m, line 113: Do we really want to crash the app in this case? Wouldn't we instead want to log something to Crashlytics and ignore?

gino-m: Actually, this really can never occur in practice, since multiple users can't create data with the same LOI id. 

Instead of this, we should be checking that all user IDs are equal to the id of the current user, ignoring ones that are not and reporting to Crashlytics without crashing the app.

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -82,63 +83,55 @@ constructor(
...
 
+  /**
+   * Marks all mutations as failed since the remote datastore only commits when all mutations have
+   * succeeded.
+   */
+  private suspend fun handleMutationSyncFailed(mutations: List<Mutation>, throwable: Throwable) {
+    mutationRepository.markAsFailed(mutations, throwable)
+    val crashlytics = FirebaseCrashLogger()
+    crashlytics.setSelectedSurveyId(mutations.first().surveyId)
+    crashlytics.logException(throwable)
```
gino-m, line 132: I noticed this earlier - imo we should update `logException` to require a custom message to help us identify the source of these errors more easily in Crashlytics.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
shobhitagarwal1612: > Do we really want to crash the app in this case? Wouldn't we instead want to log something to Crashlytics and ignore?

The reason for crashing the app is because the state is never expected to occur. Ignoring the error might lead to other side-effects which may go unnoticed.

> Instead of this, we should be checking that all user IDs are equal to the id of the current user, ignoring ones that are not and reporting to Crashlytics without crashing the app.

I've left a TODO for adding the check. I'm not sure what would happen if the app is opened in the background and the worker tries to fetch the logged in user which gets initialized only when opening the app.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
shobhitagarwal1612: Why aren't we using `Timber.e` instead? That would tag the exception and log to crashlytics.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
gino-m: > The reason for crashing the app is because the state is never expected to occur. Ignoring the error might lead to other side-effects which may go unnoticed.

I agree we would want to know about this, but I think ignoring the unexpected mutations and processing the expected ones wouldn't make the situation worse, but it might allow users to continue working, and minimize the amount of data stuck on the device.

> I've left a TODO for adding the check. I'm not sure what would happen if the app is opened in the background and the worker tries to fetch the logged in user which gets initialized only when opening the app.

The worker is only cued due to user action, which means the app had to be opened at some point in the past. Those credentials are serialized used by the worker to access Firebase.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
gino-m: Does it? How does that work?

shobhitagarwal1612: Timber.e() reaches this method.

https://github.com/google/ground-android/blob/17422bdd7f9b3cd1b8bdec94f0a6f1f4a10a5245/ground/src/main/java/com/google/android/ground/GroundApplication.kt#L64

which then logs the exception to crashlytics if the log level is Error

https://github.com/google/ground-android/blob/17422bdd7f9b3cd1b8bdec94f0a6f1f4a10a5245/ground/src/main/java/com/google/android/ground/FirebaseCrashLogger.kt#L32

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
```diff
@@ -53,9 +54,8 @@ constructor(
       Timber.d("Syncing survey $surveyId")
       syncSurvey(surveyId)
     } catch (t: Throwable) {
-      val logger = FirebaseCrashLogger()
-      logger.setSelectedSurveyId(surveyId)
-      logger.logException(t)
+      firebaseCrashLogger.setSelectedSurveyId(surveyId)
+      Timber.e(t)
```
gino-m, line 58: Let's also log text explaining what may have caused this.

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
shobhitagarwal1612: Added check for currently logged in user

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -80,63 +84,55 @@ constructor(
...
-    val user = localUserStore.getUserOrNull(userId)
+  /**
+   * Returns the user associated with the mutations. If missing, removes the mutations from local
+   * storage as well.
+   */
+  private suspend fun getUserFromMutations(mutations: List<Mutation>): User? {
+    val userIds = mutations.map { it.userId }.toSet()
+
+    check(userIds.size == 1) { "Found mutations for multiple users" }
+    check(userRepository.getAuthenticatedUser().id == userIds.first()) {
```
gino-m, line 115: Let's `Timber.e()` and skip these instead to make sure data collectors don't end up with data stuck on their devices?

ground/src/main/java/com/google/android/ground/persistence/sync/SurveySyncWorker.kt
gino-m: You can disregard this, I added these changes to my PR.

shobhitagarwal1612: Ack

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/persistence/sync/LocalMutationSyncWorker.kt
```diff
@@ -80,63 +84,55 @@ constructor(
...
+      "User ${userIds.first()} doesn't match logged in user ${userRepository.getAuthenticatedUser().id} "
+    }
+
+    val user = localUserStore.getUserOrNull(userIds.first())
     if (user == null) {
-      Timber.e("User account removed before mutation processed")
+      Timber.e("User removed before mutation processed. Removing %d mutations", mutations.size)
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m PTAL

Issue #2854, opened 2024-11-20
Non-fatal Exception: kotlinx.coroutines.JobCancellationException


gino-m:
This has been logged in prod 319 times in the past 7 days. There are some potentially useful log lines under "Logs & Breadcrumbs", primarily related to:

* Photo upload failed. local path: .. <-- likely due to no connection, should be fixed in #2852
* Value not found for task .. <-- @shobhitagarwal1612 Could this be related to #2847 ?

shobhitagarwal1612:
This exception occurs when the coroutine is cancelled via `job.cancel()` or due to a timeout. It is automatically propagated to all it's child coroutines to cancel everything as expected. Therefore, we shouldn't be logging it as an exception.

shobhitagarwal1612:
One way to prevent this exception from getting thrown is by checking whether the couroutine is active or not before calling `await()`

shobhitagarwal1612:
On further debugging, I found that it happens when calling `await()` if the job is already cancelled.

shobhitagarwal1612:
Wait for #2869 to be released in next candidate and then check again if the issue is resolved or not via crash logs.

kenstershiro:
To verify crashlytics stats once next release cut by @rfontanarosa

kenstershiro:
@gino-m to take a look at the stats to see if this is fixed

PR #2869, opened 2024-11-23
Adds exception handling for coroutine cancellation when using await()

shobhitagarwal1612:
Towards #2854

As per documentation, calling `await()` when the coroutine is cancelled throws `JobCancellationException`. Since it is a normal part of the coroutine lifecycle, these can be safely ignored and the error can be handled accordingly. 

The classes are identified based on logs/stacktraces from the crashlytics dashboard. There could be more of them, so we should keep the issue open until no more of these errors are observed.

@gino-m  PTAL?

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -273,6 +273,7 @@ exceptions:
     SwallowedException:
         active: true
         ignoredExceptionTypes:
+            - 'CancellationException'
             - 'InterruptedException'
             - 'MalformedURLException'
             - 'NumberFormatException'
--- a/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirebaseStorageManager.kt
+++ b/ground/src/main/java/com/google/android/ground/persistence/remote/firebase/FirebaseStorageManager.kt
@@ -22,7 +22,9 @@ import java.io.File
 import java.util.StringJoiner
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.CancellationException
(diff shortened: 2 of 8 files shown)


shobhitagarwal1612:
Catching `CancellationException` at the top level disrupts coroutine cancellation, risking resource leaks, broken structured concurrency, and inconsistent application states. We must always ensure that `CancellationException` is properly propagated. `await()` is one of the use-case where it is justified to handle it explicitly.

I'm open to refactoring the common code by using an extension function. Is that what you were suggesting?



Always propagate it unless explicitly needed for specific handling.

gino-m:
> Catching `CancellationException` at the top level disrupts coroutine cancellation, risking resource leaks, broken structured concurrency, and inconsistent application states. We must always ensure that `CancellationException` is properly propagated. `await()` is one of the use-case where it is justified to handle it explicitly.
> 
> I'm open to refactoring the common code by using an extension function. Is that what you were suggesting?
> 
> Always propagate it unless explicitly needed for specific handling.

Thanks for clarifying. I agree keeping the handling closer to the invocation makes things more clear and predictable. LGTM.

gino-m (approved):

shobhitagarwal1612:
> Does that mean they're ignored in the suspend functions exposed by platform classes like Worker, but not when invoked from suspend functions under a launch* invocation?

All exceptions which are not handled inside a coroutine are finally caught by a default exception handler once the coroutine is finished. This handler only logs the exception and doesn't re-throw them. This default implementation ignores `CancellationException` and doesn't log to console.

PR #2876, opened 2024-11-27
Enable exporting table schema to code repository

shobhitagarwal1612:
Allow exporting table schema to code repository. 

This will enable compile time failures if room cannot auto-migrate schema and a manual migration is missing.

https://developer.android.com/training/data-storage/room/migrating-db-versions

Needed for https://github.com/google/ground-android/pull/2871

@gino-m  PTAL?


gino-m (approved):
👍 🚢

PR #2877, opened 2024-11-27
Define DrawAreaTaskFragment in XML instead of creating programmatically

### LINKED ISSUE #2749: [TRENDING CRASH] IllegalArgumentException: No view found for id 0x42dc8 (unknown) for fragment
Occurred 17 times over last 7 days in Open Foris Ground 0.1.9 prod on Android 12, 13, and 14 on multiple devices. 

```
Fatal Exception: java.lang.IllegalArgumentException: No view found for id 0x42dc8 (unknown) for fragment DrawAreaTaskMapFragment{3f3486a} (a77cc8b6-7841-4ca9-a1f3-1a66d701904b id=0x42dc8 tag=DrawAreaTaskMapFragment)
       at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:567)
       at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:286)
       at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2164)
       at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:2059)
       at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:2002)
       at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:702)
       at android.os.Handler.handleCallback(Handler.java:958)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:230)
       at android.os.Looper.loop(Looper.java:319)
       at android.app.ActivityThread.main(ActivityThread.java:8913)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:608)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1103)
  ```

### PR DESCRIPTION
**Pull request by anandwana001**
Fixes #2749

We are seeing such error in DrawAreaTaskFragment particular fragment only from past 90 days for removing the dynamic view and replacing it with XML file. 

@shobhitagarwal1612 @gino-m  PTAL?

### CODE REVIEW

**Comment by anandwana001** [comment]
This PR is a suggested approach, as the bug is no repro.

@sufyanAbbasi any suggestions?

File: ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt [hunk 1]
```diff
@@ -58,16 +57,21 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
   override fun onCreateTaskBody(inflater: LayoutInflater): View {
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical "uniqueness".
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
-    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11411 }
+    val rootView = inflater.inflate(R.layout.fragment_draw_area_task, null)
```
gino-m, line 60: If the second arg is required, please use named args to clarify what it represents.

ground/src/main/res/layout/fragment_draw_area_task.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+  ~ Copyright 2022 Google LLC
```
gino-m, line 4: Please use 2024 for new files

gino-m (changes requested):
Why do we need an XML layout for an empty Fragment, especially if the rest of the Fragment is built dynamically?

Also, please include more content in the PR title, eg "Define DrawAreaTaskFragment in  XML instead of creating programmatically"

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
anandwana001: fixed

anandwana001:
> Why do we need an XML layout for an empty Fragment, especially if the rest of the Fragment is built dynamically?

As of the error is coming to this particular fragment where we are dynamically generating the view using prime number, so this is one approach where rather than dependent on newly built view, we have our own XML view to use every time which will prevent this error to occur.

gino-m (changes requested):
Sorry I might be a bit out of the loop. It looks like `drawAreaTaskMapFragment` is still being passed in as an args, and that this replaces the `LinearLayout` with a new fragment defined in XML. This doesn't look like it affects the ID of the DrawAreaTaskMapFragment at all? @shobhitagarwal1612 @sufyanAbbasi Do you have any ideas here?

shobhitagarwal1612:
This was introduced by @sufyanAbbasi in https://github.com/google/ground-android/pull/2588 for all map based fragments. So whatever we end up using should be same across all map based fragments.

> Sorry I might be a bit out of the loop. It looks like drawAreaTaskMapFragment is still being passed in as an args, and that this replaces the LinearLayout with a new fragment defined in XML. This doesn't look like it affects the ID of the DrawAreaTaskMapFragment at all? @shobhitagarwal1612 @sufyanAbbasi Do you have any ideas here?

Yes, it does replace dynamically generating the view ids with default implementation.

@anandwana001 How have you verified that this doesn't have any side effects? Did you try creating a job with multiple map based tasks and trying to pause/resume app (intentionally ending up in edge case scenarios)?

anandwana001:
> @anandwana001 How have you verified that this doesn't have any side effects? Did you try creating a job with multiple map based tasks and trying to pause/resume app (intentionally ending up in edge case scenarios)?


As this PR is based for DrawAreaTask which is generally one map task only in one job. Anything I am missing here?

gino-m:
I think two things would help to unblock this:

1) Describe in unambitious terms what is causing the original IAE exception 
2) Write a UI test which fails before this fix, and passes after

If 2 isn't feasible for some reason, can you describe how to test this fix manually?

anandwana001:
println("=========  CaptureLocationTaskFragment  ${rowLayout.id} ---- ${View.generateViewId()} ")
println("=========  ${rowLayout.id} ---- ${View.generateViewId()} ")

```
2024-12-02 15:36:28.611  7160-7160  System.out              com.google.android.ground            I  =========  114110 ---- 11 
2024-12-02 15:36:28.711  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  144937 ---- 14 
2024-12-02 15:36:46.044  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  267576 ---- 25 
2024-12-02 15:36:46.371  7160-7160  System.out              com.google.android.ground            I  =========  319508 ---- 29 
2024-12-02 15:36:59.458  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  356768 ---- 33 
2024-12-02 15:36:59.624  7160-7160  System.out              com.google.android.ground            I  =========  410796 ---- 37 
2024-12-02 15:37:19.558  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  445960 ---- 41 
2024-12-02 15:37:19.740  7160-7160  System.out              com.google.android.ground            I  =========  502084 ---- 45 
2024-12-02 15:37:34.924  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  535152 ---- 49 
2024-12-02 15:37:35.106  7160-7160  System.out              com.google.android.ground            I  =========  593372 ---- 53 
2024-12-02 15:37:49.272  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  624344 ---- 57 
2024-12-02 15:37:49.457  7160-7160  System.out              com.google.android.ground            I  =========  684660 ---- 61 
2024-12-02 15:38:03.274  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  713536 ---- 65 
2024-12-02 15:38:03.456  7160-7160  System.out              com.google.android.ground            I  =========  775948 ---- 69 
2024-12-02 15:39:19.506  7160-7160  System.out              com.google.android.ground            I  =========  CaptureLocationTaskFragment  802728 ---- 73 
2024-12-02 15:39:19.739  7160-7160  System.out              com.google.android.ground            I  =========  867236 ---- 77 

```



When doing a lot of back and forth with the mapFragments, the viewId generation thing keep happening all the time due to which the rowLayout.id is not in sync with the view generated id and causing this error



```
2024-12-02 15:46:40.680  8851-8851  AndroidRuntime          com.google.android.ground            D  Shutting down VM
2024-12-02 15:46:40.682  8851-8851  AndroidRuntime          com.google.android.ground            E  FATAL EXCEPTION: main (Ask Gemini)
                                                                                                    Process: com.google.android.ground, PID: 8851
                                                                                                    java.lang.IllegalArgumentException: No view found for id 0x1bdbe (unknown) for fragment DrawAreaTaskMapFragment{e08cac8} (9abbd048-780f-41c1-a975-63d8103742c0 id=0x1bdbe tag=DrawAreaTaskMapFragment)
                                                                                                    	at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:567)
                                                                                                    	at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:286)
                                                                                                    	at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2164)
                                                                                                    	at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:2059)
                                                                                                    	at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:2002)
                                                                                                    	at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:702)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:959)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:232)
                                                                                                    	at android.os.Looper.loop(Looper.java:317)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8674)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:886)
2024-12-02 15:46:40.714  8851-8862  .android.ground         com.google.android.ground            W  Cleared Reference was only reachable from finalizer (only reported once)
2024-12-02 15:46:43.685  8851-8851  FirebaseCrashlytics     com.google.android.ground            E  Cannot send reports. Timed out while fetching settings.
2024-12-02 15:46:43.712  8851-8851  Process                 com.google.android.ground            I  Sending signal. PID: 8851 SIG: 9

```


The viewId generation value is always bigger than the previous one, and as per the Android system if value goes beyond 0x00FFFFFF, system internally truncates it which might becomes duplicate of other id in the process and cause this issue.


I tried attaching the duplicate view id and it gives this error because that particular id is not able to find in the fragment DrawAreaTaskFragment



### Fix

Now the fix I'm proposing here is to use the XML file which always gives the same view id so no more issues exist regarding the view id.

anandwana001:
> @anandwana001 Thank you for the additional context. Can you please refresh my memory as to why we can't we just use generateViewId() directly, without the prime number multiplier? I don't believe we would pass 0x00FFFFFF (16,777,215) as an identifier - they would only reach that ID if the user reopened the app every hour for over 1800 years! Something else must be happening here.
> 
> Also @sufyanAbbasi


As mentioned in the comment in codebase, 

// NOTE(#2493): Multiplying by a random prime to allow for some mathematical "uniqueness".
    // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
    
    
 "might conflict" - because of this, I proposed why even to have dynamic view and why not to go with XML layout?

gino-m:
@anandwana001 I see, thanks for reminding me of the context of the original bug. Am I right in understanding that we only need the ID so the view is restored on configuration change, and that this is only required because the view is being added dynamically, and that if/when we move to Compose we wouldn't need this workaround at all?

anandwana001:
> @anandwana001 I see, thanks for reminding me of the context of the original bug. Am I right in understanding that we only need the ID so the view is restored on configuration change, and that this is only required because the view is being added dynamically, and that if/when we move to Compose we wouldn't need this workaround at all?

Yes

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -58,16 +58,21 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
   override fun onCreateTaskBody(inflater: LayoutInflater): View {
     // NOTE(#2493): Multiplying by a random prime to allow for some mathematical "uniqueness".
     // Otherwise, the sequentially generated ID might conflict with an ID produced by Google Maps.
-    val rowLayout = LinearLayout(requireContext()).apply { id = View.generateViewId() * 11411 }
+    val rootView = FragmentDrawAreaTaskBinding.inflate(inflater)
```
gino-m, line 61: Please replace the above comment/note with:
```kt
// XML layout is used to provide a static view ID which does not collide with Google Maps view ID (https://github.com/google/ground-android/issues/2493). 
// The ID is needed when restoring the view on config change since the view is dynamically created.
// TODO(https://github.com/google/ground-android/issues/1795):
//   Remove this workaround once this UI is migrated to Compose.
```

gino-m (changes requested):
LGTM, please just apply requested edit and reformat. Thanks!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
anandwana001: done

gino-m (approved):

PR #2885, opened 2024-11-29
[Capture location task] Prompted to Correct enable location task

### LINKED ISSUE #2855: [Capture location task] Prompted to enable location on wrong task
The user is getting prompted to enable location permissions on the previous task, which in this case was a "take a photo" tasks. See start of screen recording for repro.

https://github.com/user-attachments/assets/e875aef3-f687-4cf8-beea-9ba0b4d2752a

@anandwana001 PTAL?

anandwana001:
Fixes #2855 

@shobhitagarwal1612  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
@@ -24,9 +24,7 @@ import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.ComposeView
-import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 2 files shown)


shobhitagarwal1612 (changes requested):
Instead of moving the code to CaptureLocationTaskFragment, can you do something like:

```

  override fun onResume() {
    super.onResume()
    if (isVisible) {
       // add existing code for showing permission dialog.
    }
  }
```

anandwana001:
> Instead of moving the code to CaptureLocationTaskFragment, can you do something like:
> 
> ```
> 
>   override fun onResume() {
>     super.onResume()
>     if (isVisible) {
>        // add existing code for showing permission dialog.
>     }
>   }
> ```



it will not work, because as this fragment is a part of pagination where the code is getting executed when the index is +1.

shobhitagarwal1612:
We care about the fragment being visible to the user. Can you try out the above code once?

anandwana001:
> We care about the fragment being visible to the user. Can you try out the above code once?

Tried, not working

shobhitagarwal1612 (approved):

PR #2905, opened 2024-12-09
[Data collection flow] Disable animation on previous/next

### LINKED ISSUE #2829: [Data collection] Jank when cycling between "Previous" and "Next"
Other tasks "flash" on the screen when navigating to/from a conditional task.

anandwana001:
Fixes #2829 

@shobhitagarwal1612 @gino-m  PTAL?

https://github.com/user-attachments/assets/3047ddc2-7d69-48fc-b304-945a63a6354c

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
@@ -155,6 +155,7 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
   }
 
   private fun onTaskChanged(taskPosition: TaskPosition) {
+    // Pass false to parameter smoothScroll to avoid smooth scrolling animation.
     viewPager.setCurrentItem(taskPosition.absoluteIndex, false)
     updateProgressBar(taskPosition, true)
   }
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
This also removes the animation. Can we keep that?

anandwana001:
> This also removes the animation. Can we keep that?

@shobhitagarwal1612 
for this we can create custom pageTranform, but we also have another ticket where we want to have dynamic page creation, as currently the pages are getting created at dataCollection starts.

So, should we wait for that and remove this completely or for now should we go with this fix with no animation and start work on that?

shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -155,7 +155,7 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
   }
 
   private fun onTaskChanged(taskPosition: TaskPosition) {
-    viewPager.currentItem = taskPosition.absoluteIndex
+    viewPager.setCurrentItem(taskPosition.absoluteIndex, false)
```
gino-m, line 158: Please add named arg to clarify what `false` refers to

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
anandwana001: added comment.
As this is a java function, it doesn't support named argument as of kotlin.

gino-m (approved):

Issue #2919, opened 2024-12-11
[Startup] "Initializing" spinner gets stuck on startup


gino-m:
This occurs when the app is opened in the background (e.g., via Android Studio) while the phone is locked, or after completing the Google Play Services install flow and returning to the app. 

The root cause is that the sign in event occurs while the `MainActivity` isn't running. `MainViewModel`  emits the `navigationRequests`, but `MainActivity` is not there to receive it. To resolve this:

* The `signInState` in `AuthenticationManager` should be a `StateFlow`. We have one now, but we convert to a `Flow` because we call `filterNotNull()`.
*  Either Navigation events should only be triggered from within the Fragment lifecycle.

gino-m:
Detailed solution propose here:
https://github.com/google/ground-android/pull/2924#issuecomment-2573357452

kenstershiro:
@shobhitagarwal1612 is working on https://github.com/google/ground-android/issues/1795 which will also address this (needs to move all screens to compose first)

kenstershiro:
Moving back to backlog for now, impact is low on users

Issue #2783, opened 2024-10-08
GoogleApiManager$GooglePlayServicesMissingException - Google play services not available


anandwana001:
GoogleApiManager.installGooglePlayServices
com.google.android.ground.system.GoogleApiManager$GooglePlayServicesMissingException - Google play services not available

```console
 Fatal Exception: com.google.android.ground.system.GoogleApiManager$GooglePlayServicesMissingException: Google play services not available
       at com.google.android.ground.system.GoogleApiManager.installGooglePlayServices(GoogleApiManager.kt:45)
       at com.google.android.ground.ui.startup.StartupViewModel.initializeLogin(StartupViewModel.kt:32)
       at com.google.android.ground.ui.startup.StartupFragment$onResume$1.invokeSuspend(StartupFragment.kt:55)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.internal.DispatchedContinuationKt.resumeCancellableWith(DispatchedContinuation.kt:367)
       at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:30)
       at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable$default(Cancellable.kt:25)
       at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:110)
       at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:126)
       at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch(Builders.common.kt:56)
       at kotlinx.coroutines.BuildersKt.launch(:1)
       at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch$default(Builders.common.kt:47)
       at kotlinx.coroutines.BuildersKt.launch$default(:1)
       at com.google.android.ground.ui.startup.StartupFragment.onResume(StartupFragment.kt:53)
       at androidx.fragment.app.Fragment.performResume(Fragment.java:3210)
       at androidx.fragment.app.FragmentStateManager.resume(FragmentStateManager.java:666)
       at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:310)
       at androidx.fragment.app.FragmentStore.moveToExpectedState(FragmentStore.java:114)
       at androidx.fragment.app.FragmentManager.moveToState(FragmentManager.java:1675)
[...]

PR #2920, opened 2024-12-11
Handle Play services failures more gracefully

gino-m:
Some GMS strangeness:

* When Play services are present, but disabled,  after "ENABLE" is clicked in the resolution dialog, `onActivityResult()` is called with a failure result even though the flow is not complete.
* When Play services are not supported on a device, an error dialog is shown, but `onActivityResult()` is not called when dismissed.

To work around this, if the error is identified as "recoverable, this PR waits indefinitely for Play services to be installed.

Towards #2783 - we may want to follow-up with a PR to add a distinct message "Waiting for Play services..." while waiting.

Filed #2919 as follow-up for edge case.

### DIFF
--- a/ground/src/main/java/com/google/android/ground/system/GoogleApiManager.kt
+++ b/ground/src/main/java/com/google/android/ground/system/GoogleApiManager.kt
@@ -16,14 +16,19 @@
 package com.google.android.ground.system
 
 import android.content.Context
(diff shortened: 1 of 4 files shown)


gino-m:
@anandwana001 @shobhitagarwal1612 PTAL?

anandwana001 (approved):

PR #2925, opened 2024-12-15
Update auto-assign config to only add 1 reviewer

shobhitagarwal1612:
Disable auto-assigner and reduce reviewer to 1

@gino-m  PTAL?


gino-m:
Why disable auto assign?

shobhitagarwal1612:
Assignees and reviewers generated by the workflow are not always the same person. So, this is to reduce confusion

anandwana001 (approved):

PR #2937, opened 2024-12-17
Added values in dimens.xml

### LINKED ISSUE #2913: [Code health] Ensure all UI sizes and dimensions are defined in dimens.xml

anandwana001:
Fixes #2913 

@gino-m  PTAL?

Can we generalize the values or should we go feature by feature?


ground/src/main/res/values/dimens.xml
```diff
@@ -39,4 +39,16 @@
...
+  <dimen name="spacing_small">4dp</dimen>
+  <dimen name="spacing_medium">8dp</dimen>
+  <dimen name="spacing_large">16dp</dimen>
+  <dimen name="spacing_xlarge">24dp</dimen>
+
+  <!-- Text Sizes -->
+  <dimen name="text_small">12sp</dimen>
+  <dimen name="text_medium">14sp</dimen>
+  <dimen name="text_large">16sp</dimen>
+  <dimen name="text_xlarge">20sp</dimen>
```
shobhitagarwal1612, line 53: I don't think the intention is to add arbitrary dimens to this file. But rather to move the hardcoded values to this file. However, I feel we should not spend too much time on this since Compose UI doesn't use this file. Do you know what is the recommended approach for compose UI?

anandwana001: As per documentation 
`dimensionResource(R.dimen.padding_small)`

Hence using dimens.xml

Same as what we are doing for strings

shobhitagarwal1612 (approved):

ground/src/main/res/values/dimens.xml
anandwana001: wdyt about naming?

gino-m: LGTM

ground/src/main/res/values/dimens.xml
```diff
@@ -39,4 +39,16 @@
   <dimen name="job_card_corner_radius">16dp</dimen>
   <dimen name="job_card_start_end_margin">8dp</dimen>
   <dimen name="zoom_on_cluster_click_padding">50dp</dimen>
+
+  <!-- Standard Spacing -->
+  <dimen name="spacing_small">4dp</dimen>
```
gino-m, line 44: CI is failing because these are unused. Maybe add the first reference to them to this PR and use elsewhere as needed in follow-ups?

gino-m (changes requested):

ground/src/main/java/com/google/android/ground/ui/offlineareas/OfflineAreaListItem.kt
```diff
@@ -99,6 +105,16 @@ fun OfflineAreaListItem(
   }
 }
 
+@Composable
+private fun OfflinePinImage() {
+  Icon(
+    imageVector = ImageVector.vectorResource(id = R.drawable.ic_offline_pin),
+    contentDescription = stringResource(id = R.string.offline_area_list_item_icon),
+    tint = MaterialTheme.colorScheme.primary,
+    modifier = Modifier.size(dimensionResource(R.dimen.spacing_xlarge)),
```
shobhitagarwal1612, line 114: Using spacing dimen for size seems incorrect. Can you create new dimens for size?

anandwana001: the plan was to use the generic values wherever required like 8dp 4dp etc

are you suggesting duplicating the values with _size suffix?

anandwana001:
@shobhitagarwal1612 keeping the name as `dimen_Xdp` to be used as the generic one wherever required.

shobhitagarwal1612:
@anandwana001 Can we close this if it is no longer relevant?

anandwana001:
> > @shobhitagarwal1612 keeping the name as dimen_Xdp to be used as the generic one wherever required.
> 
> But how is this going to be provide any value? This is as good as hardcoding the value in the code. Before we make any more changes, can we discuss this further in the bug regarding which approach should be undertaken?
> 
> Possible options:
> 
> 1. Current approach i.e. use generic `dimen_Xdp`
> 2. Creating a new dimen based on the type of usage e.g. `spacing_{small|medium|large}`, `image_size_{small|medium|large}`, etc.
> 
> We should also clearly mention in the bug why we are doing this refactor and the potential benefits of it.

@gino-m

gino-m:
@anandwana001 Asymmetrical spacing/padding would be represented by the container hierarchy (e.g., the outer container has a 4dp padding, the inner one is right-aligned also with 4dp padding, etc.). Since that's not the case today, we should keep the hardcoded dimens. 

tl;dr; In an ideal world, we'd just have `container_padding = 4dp`, and apply additional padding using nested containers. Since that's not the case today, we can just keep the hard-coded values.

PR #2943, opened 2024-12-17
Update todo-to-list github action to commit to PR instead of master branch

shobhitagarwal1612:
Instead of merging the commit to master branch, it commits to the PR's source branch when the PR is approved.

@gino-m  PTAL?

### DIFF
--- a/.github/workflows/todo-to-issue.yml
+++ b/.github/workflows/todo-to-issue.yml
@@ -16,9 +16,8 @@
 
 name: "Run TODO to Issue"
 on:
-  push:
-    branches:
-      - replace-with-master-when-fixed
+  pull_request:
+    branches: [ "master" ]
 
 jobs:
   build:
@@ -26,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          ref: ${{ github.head_ref }}
       - name: "TODO to Issue"
         uses: alstr/todo-to-issue-action@v5.1.7
         with:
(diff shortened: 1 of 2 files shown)


gino-m:
@shobhitagarwal1612 Should we move this to draft for now?

gino-m (approved):

shobhitagarwal1612:
@gino-m The workflow isn't able to find the email if the workflow is run after PR is approved automation.

gino-m:
> @gino-m The workflow isn't able to find the email if the workflow is run after PR is approved automation.

Which email are you referring to?

shobhitagarwal1612:
In order to bypass cla/google check, I added a step to get the email from `github.head_ref` and use that when committing the new changes from workflow. This works fine when the workflow is executed on commit push. When the workflow is triggered on PR approval, then the step to get email doesn't work.

gino-m:
> In order to bypass cla/google check, I added a step to get the email from `github.head_ref` and use that when committing the new changes from workflow. This works fine when the workflow is executed on commit push. When the workflow is triggered on PR approval, then the step to get email doesn't work.

I see. Only maintainers can push to this repo, so perhaps allowing creation of TODOs on commit is acceptable?

shobhitagarwal1612:
> I see. Only maintainers can push to this repo, so perhaps allowing creation of TODOs on commit is acceptable?

Agreed. If you also agree, I can update this PR to run the workflow on each commit push.

gino-m:
> > I see. Only maintainers can push to this repo, so perhaps allowing creation of TODOs on commit is acceptable?
> 
> Agreed. If you also agree, I can update this PR to run the workflow on each commit push.

Sure, let's do it!

PR #2946, opened 2024-12-17
SyncStatus: Use new mutations repository upload queue flow; renaming.

### LINKED ISSUE #2894: [Code health] Use UploadQueueEntry from MutationRepository.getUploadQueueFlow() to drive "Upload status" UI
Once this is done please remember to delete unused `MutationRepository` methods formerly used to load the queue.

@scolsen Adding to Iteration 27. Thanks for taking this on!

scolsen:
This commit uses the newly introduced getUploadQueueFlow() in the MutationRepository to fetch mutations for the sync status UI. In additiion, I have renamed some classes and added some documentation for clarity. In addition, we now handle SubmissionMutations separately from LOI Mutations, allowing us to better inform users. We also no longer check for the authenticated user in the sync view model, since this is done in the mutation repository before the uplaod queue is returned.

Fixes #2894

@gino-m

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/syncstatus/SyncStatusDetail.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/syncstatus/SyncStatusDetail.kt
@@ -17,10 +17,16 @@ package com.google.android.ground.ui.syncstatus
 
 import com.google.android.ground.model.mutation.Mutation
 
-/** A tiny helper class for bundling mutation history display data. */
-data class MutationDetail(
+/**
+ * Defines the set of data needed to display the human-readable status of a queued local [Mutation].
+ */
+data class SyncStatusDetail(
+  /** The username of the user who made this change. */
   val user: String,
+  /** The underlying [Mutation]. */
   val mutation: Mutation,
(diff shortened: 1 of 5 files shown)


shobhitagarwal1612 (approved):

ground/src/main/java/com/google/android/ground/ui/syncstatus/SyncStatusDetail.kt
```diff
@@ -17,10 +17,16 @@ package com.google.android.ground.ui.syncstatus
...
 
-/** A tiny helper class for bundling mutation history display data. */
-data class MutationDetail(
+/**
+ * Defines the set of data needed to display the human-readable status of a queued local [Mutation].
+ */
+data class SyncStatusDetail(
+  /** The username of the user who made this change. */
   val user: String,
+  /** The underlying [Mutation]. */
```
gino-m, line 26: Will we replace this with `UploadStatus` in a follow-up?

scolsen: sgtm, will do!

PR #2949, opened 2024-12-19
Fix/2843

### LINKED ISSUE #2843: [Code health] Rename packages from `com.google.android.ground` to `org.groundplatform.android`

anileshwar12:
Fixes #2843


sufyanAbbasi:
Sorry about not seeing this for a while! It looks like there are some "replace all" where some instances of "google" should be preserved and not all instances "com.google" have been converted to "org.groundplatform." It might be worth doing another pass, perhaps loading it up into Android Studio and building it. 

Also, we've recently migrated the top level folder from "ground" to "app" so starting over from scratch might be a good call here!

gino-m:
Closing due to inactivity

PR #2950, opened 2024-12-20
Cleanup remaining usages of getDetailsText() from TaskData

### LINKED ISSUE #1733: [Code health] Move creation of string labels for responses from model to view
As an early hack, we returned the string representation of responses by extending `getDetailsText()` in the base `TaskData`. To allow for i18n and proper SoC, these UI representations should be produce be ViewModels or related helpers, not in the model itself.

shobhitagarwal1612:
Fixes #1733 

getDetailsText() was previously used for displaying value of response for the current task fragment. This method is extended by all response model classes (TaskData). All text should be created on view layer and not is model due to i18n.

In this PR, I've migrated all remaining usages of getDetailsText() with appropriate values and cleaned up all unused methods.

This is a pure refactor and doesn't change any existing user behavior. 
Verified that this doesn't introduce any regressions via unit tests and manually running the app in an emulator.

@gino-m  PTAL?


gino-m (changes requested):
Ah, finally!!! I've dreamt of this moment for 2+ years 😆 

Would it be possible to add tests to the ViewModelTests to check this presentation logic while you're at it?

shobhitagarwal1612:
> Would it be possible to add tests to the ViewModelTests to check this presentation logic while you're at it?

Done. PTAL?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskFragment.kt
```diff
@@ -52,7 +52,7 @@ class DropPinTaskFragment @Inject constructor() : AbstractTaskFragment<DropPinTa
     val args = Bundle()
     args.putString(TASK_ID_FRAGMENT_ARG_KEY, taskId)
     fragment.arguments = args
-    parentFragmentManager
+    childFragmentManager
```
gino-m, line 55: What does this do? Should this be in a separate PR in case it breaks something in prod?

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -68,7 +68,7 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
     val args = Bundle()
     args.putString(TASK_ID_FRAGMENT_ARG_KEY, taskId)
     drawAreaTaskMapFragment.arguments = args
-    parentFragmentManager
+    childFragmentManager
```
gino-m, line 71: Same.

ground/src/main/java/com/google/android/ground/ui/map/gms/ConverterExt.kt
```diff
@@ -38,3 +38,7 @@ fun Coordinates.toLatLng(): LatLng = LatLng(lat, lng)
 fun List<Coordinates>.toLatLngList(): List<LatLng> = map { it.toLatLng() }
 
 fun Location.toCoordinates(): Coordinates = Coordinates(latitude, longitude)
+
+fun Location.toAltitude(): Double? = if (hasAltitude()) altitude else null
```
gino-m, line 42: These new "to" methods seem like misnomers. Maybe `getAltitudeOrNull()? and `getAccuracyOrNull()` would more correct and kotlinic?

gino-m (changes requested):
LGTM, just a few more nits. Thanks!

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/point/DropPinTaskFragment.kt
shobhitagarwal1612: The relevance of parentFragmentManager vs childFragmentManager is dependent on what the child fragment needs access to. If the child fragment only needs to interact with the parent fragment, then `childFragmentManager` is needed. If the child fragment needs to interact with the activity or the parent fragment's siblings, then only `parentFragmentManager` is needed.

In our case, `childFragmentManager` is enough. This fixes the scope of the access. In addition to that, this also resolves the issue of unit tests. Earlier, the tests couldn't be added for the mapTaskFragments as the parent fragment's viewmodel couldn't be injected. This was because nav controller set for the parent fragment wasn't applied to the child fragment. However, this got auto fixed when using `childFragmentManager`. This was needed for adding tests for CaptureLocationTaskFragment. Since all map based tasks are same, I added the same change to drop pin and draw area tasks as well. Verified that the parent VM injection works fine even after this change.

ground/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
shobhitagarwal1612: Responded in previous comment.

gino-m (approved):
🥳

PR #2953, opened 2024-12-24
Do not auto-close the issues when TODO is cleaned up

shobhitagarwal1612:
Do not auto-close the issues linked to TODOs with the todo-to-issue workflow.

This is to prevent the issues from getting closed even before the PR is merged.

Example: https://github.com/google/ground-android/issues/1733#issuecomment-2560054435

@gino-m  PTAL?

### DIFF
--- a/.github/workflows/todo-to-issue.yml
+++ b/.github/workflows/todo-to-issue.yml
@@ -31,6 +31,7 @@ jobs:
         with:
           AUTO_ASSIGN: "true"
           INSERT_ISSUE_URLS: "true"
+          CLOSE_ISSUES: "false"
       - name: Set Git user
         run: |
           git config --global user.name ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


.github/workflows/todo-to-issue.yml
```diff
@@ -31,7 +31,6 @@ jobs:
         with:
           AUTO_ASSIGN: "true"
           INSERT_ISSUE_URLS: "true"
-          CLOSE_ISSUES: "true"
```
gino-m, line 34: Nit: Use `CLOSE_ISSUES: "false"` instead. Even though that's the default, it would make this behavior more discoverable to readers.

shobhitagarwal1612: Done

gino-m (approved):

PR #2955, opened 2024-12-27
[Code health] Replace .emit() with .update() for state flows wherever applicable

### LINKED ISSUE #2954: [Code health] Replace .emit with .update for State Flows wherever applicable
Description:

We need to replace `.emit` with `.update` for state flows in the application only where the value update does not need to suspend a coroutine. This change improves thread safety and aligns with best practices for state flow management, while ensuring that coroutine suspension scenarios remain correctly handled with `.emit`.

* `.emit` is not thread-safe and can lead to race conditions when multiple threads modify the state concurrently.
* `.update` ensures atomic updates, preventing these risks in non-suspending scenarios by synchronizing modifications.

Steps:
1. Identify all instances of .emit in state flows across the codebase.
2. Replace .emit with .update for state updates that do not require coroutine suspension.
3. Verify that all updated state flows function correctly, especially in concurrent update scenarios.

This improves the reliability and safety of state management in non-suspending scenarios, while preserving the functionality of suspending state updates. It ensures the app adheres to best practices for state flow management.

shobhitagarwal1612:
Fixes #2954

Verified for regressions by running the app locally

@gino-m  PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
```diff
@@ -207,7 +205,7 @@ class DataCollectionFragment : AbstractFragment(), BackPressListener {
       false
     } else {
       // Otherwise, select the previous step.
-      lifecycleScope.launch { viewModel.step(-1) }
```
gino-m, line 210: Perhaps for a follow up - why do we expose "step" with an arbitrary number of pages, rather than "next" and "previous"?

gino-m (changes requested):
LG! Can you please explain in the PR description why this is needed and why this change should be a no-op?

shobhitagarwal1612:
> LG! Can you please explain in the PR description why this is needed and why this change should be a no-op?

Please see for detailed explanation: https://github.com/google/ground-android/issues/2954#issue-2760712879

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionFragment.kt
shobhitagarwal1612: Added a todo

gino-m (approved):
LGTM!

Issue #2958, opened 2024-12-30
[Code health] Move task sequence calculation / handling into a separate handler


shobhitagarwal1612:
Currently, `DataCollectionViewModel` does a lot of things incl. creation of task sequence and performing various operations using the generated task sequence. This bloats up the class and reduces the code readability / maintainability. 

This could be handled in a better way by abstracting out all task sequence related methods into a single wrapper class. This would greatly improve the overall structure of the class.

shobhitagarwal1612:
Proposed structure for the sequence handler:

```
interface TaskSequenceHandler {

  /**
   * Retrieves the task sequence based on the provided inputs and conditions.
   *
   * This function determines the order of tasks to be presented, taking into account any overrides
   * specified by [taskValueOverride].
   *
   * @param taskValueOverride An optional pair where the first element is the task ID and the second
   *   element is the [TaskData] to override the default task data. If null, no override is applied.
   * @return A [Sequence] of [Task] objects representing the ordered tasks.
   */
  fun getTaskSequence(taskValueOverride: Pair<String, TaskData?>? = null): Sequence<Task>

  /**
   * Checks if the specified task is the first task in the displayed sequence.
   *
   * @param taskId The ID of the task to check.
   * @return `true` if the task is the first in the sequence, `false` otherwise.
   */
  fun isFirstPosition(taskId: String): Boolean

  /**
   * Checks if the specified task is the last task in the displayed sequence.
   *
   * @param taskId The ID of the task to check.
   * @return `true` if the task is the last in the sequence, `false` otherwise.
   */
  fun isLastPosition(taskId: String): Boolean

  /**
   * Checks if the specified task with the given data is the last task in the sequence.
   *
   * This method allows checking if a specific task with a particular [TaskData] is the last one.
   *
   * @param taskId The ID of the task to check.
   * @param value The [TaskData] associated with the task.
   * @return `true` if the task with the given data is the last in the sequence, `false` otherwise.
   */
  fun isLastPosition(taskId: String, value: TaskData?): Boolean

  /**
   * Retrieves the ID of the task that precedes the specified task in the sequence.
   *
   * @param taskId The ID of the task for which to find the previous task.
   * @return The ID of the previous task in the sequence.
   */
  fun getPreviousTask(taskId: String): String

  /**
   * Retrieves the ID of the task that follows the specified task in the sequence.
   *
   * @param taskId The ID of the task for which to find the next task.
   * @return The ID of the next task in the sequence.
   */
[...]

PR #2959, opened 2024-12-31
Introduce TaskSequenceHandler for data collection

shobhitagarwal1612:
Towards #2958 

Introduce a new handler `TaskSequenceHandler` which exposes all the functionality required for computing the sequence and related helper methods.

This will be later used in `DataCollectionViewModel` to replace all task sequence related code with this handler. 

@sufyanAbbasi  PTAL?


ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandler.kt
```diff
@@ -0,0 +1,105 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.datacollection
+
+import com.google.android.ground.model.submission.TaskData
+import com.google.android.ground.model.task.Task
+
+interface TaskSequenceHandler {
```
scolsen, line 21: Do we really expect to have to override this? If not, consider making this concrete and removing the need for a separate Impl class. Alternatively, we could make this an `open` class to bundle the interface with the default implementation (if we do expect to need to override these with alternatives).

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.datacollection
+
+import com.google.android.ground.model.submission.TaskData
+import com.google.android.ground.model.task.Task
+
+class TaskSequenceHandlerImpl(
```
scolsen, line 21: If we do keep the interface, the distinguishing feature here, and the required data is the list of tasks, which must be non empty. You can probably better distinguish this class by naming it after that requirement, e.g.:

`NonEmptyTaskListSequenceHandler` or similar.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+  private val tasks: List<Task>,
+  private val shouldIncludeTask:
+    (task: Task, taskValueOverride: Pair<String, TaskData?>?) -> Boolean,
+) : TaskSequenceHandler {
+
+  init {
+    if (tasks.isEmpty()) {
```
scolsen, line 28: Nit: Consider using `require`, throwing an `IllegalArgument` exception instead.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.datacollection
+
+import com.google.android.ground.model.submission.TaskData
+import com.google.android.ground.model.task.Task
+
+class TaskSequenceHandlerImpl(
+  private val tasks: List<Task>,
+  private val shouldIncludeTask:
```
scolsen, line 23: I think it might get a little tricky to use this correctly, and it seems overly general for now. IIUC, the intention here is to use `TaskData` instead of the given `Task` whenever the task in question matches the id.

If that's the case, I would inline the ID check directly instead of letting callers provide their own Boolean function. 

I think we should start with relatively limited functionality here and generalize and expand only if it eventually becomes necessary (unless we already need to do this based on the existing patterns in the codebase, in which case, ignore this).

scolsen: Another thing to consider is that this is basically doing two things. 1. Giving users a way to simply filter tasks (don't include x if it matches id y) and 2. giving users a way to replace tasks (replace x with y if id of x == some string). Separating these two functions might make it easier to use this and clearer, even though they are both instances of the same general idea.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  override fun isLastPosition(taskId: String): Boolean {
+    checkInvalidTaskId(taskId)
+    return taskId == getTaskSequence().last().id
+  }
+
+  override fun isLastPosition(taskId: String, value: TaskData?): Boolean {
+    checkInvalidTaskId(taskId)
+    return taskId == getTaskSequence(taskValueOverride = taskId to value).last().id
```
scolsen, line 61: I found this confusing to understand at first. Wouldn't the basic boolean check:

```
val last = getTaskSequence.last()
taskId == last.id && value == last.getValue()
```

be simpler?

scolsen: Note also that the current implementation will throw if the sequence happens to be empty after filtering.

If you separate out the call to last instead you can check for this and simply return false.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+
+  override fun getAbsolutePosition(taskId: String): Int {
+    val index = tasks.indexOfFirst { it.id == taskId }
+    checkInvalidIndex(taskId, index)
+    return index
+  }
+
+  override fun getRelativePosition(taskId: String): Int {
+    val index = getTaskSequence().indexOfFirst { it.id == taskId }
+    checkInvalidIndex(taskId, index)
```
scolsen, line 82: Consider returning a sentinel `-1` value instead of throwing. This is in keeping with the semantics of `indexOfFirst` and would allow you to make this just `return getTaskSequence().indexOfFirst { it.id == taskId }`.

I would generally avoid throwing unless we need to really flag checking for a task no in the collection as an exceptional case (which might be the correct thing here, I trust your call on this)

scolsen (changes requested):
Nice! A couple of suggestions. Nothing crazy, and you can reject them if needed/

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandler.kt
shobhitagarwal1612: That's a great point! We won't have another fake implementation of this interface as the normal one can be directly used in unit tests. Will drop the interface and merge with concrete class

shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: Already replaced the interface with concrete class. Resolving this.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: The `Task` entity includes a `condition` field, which contains a list of `expressions` used to determine whether a multiple-choice task has a specific set of selected values. During sequence computation, we evaluate this condition (if present) to decide if the task should be included or filtered out. To perform this evaluation, access to the `taskDataMap` for all tasks is necessary. Instead of passing the `taskDataMap` to each method, I opted for a callback in the implementation to avoid direct propagation of the map. However, I acknowledge this approach may appear complex and unintuitive. Do you have suggestions for a cleaner way to retrieve task values in this handler or delegate condition evaluation to a dedicated class?

FYI, this is how it was planned to be used in the current structure.
https://github.com/google/ground-android/pull/2986/files#diff-0f12b194c69d844eb8ddb61645fd78f49d11f76701c884cef7464c66fb978764R141-R142

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: Up until this point, we were assuming -1 means first task. IMO, that added fragility to the code as instead of failing the application, we were sometimes jumping tasks if something couldn't be found. So, to identify all those default cases, I added the task id empty check and index out of bounds check. 

Let me re-evaluate if we can live without throwing the errors. Thanks again!

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: Added a TODO to address this issue while integrating with the view model.

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandler.kt
```diff
@@ -0,0 +1,154 @@
...
+
+  private fun checkInvalidIndex(taskId: String, index: Int) {
+    checkInvalidTaskId(taskId)
+    require(index >= 0) { "Task $taskId not found in task list" }
+  }
+
+  /**
+   * Retrieves the task sequence based on the provided inputs and conditions.
+   *
+   * This function determines the order of tasks to be presented, taking into account any overrides
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandlerImpl.kt
shobhitagarwal1612: Done

shobhitagarwal1612:
Thanks @scolsen for the thorough review! I'll try to resolve some of the comments in the subsequent PR where this handler will be used in the VM.

Issue #2338, opened 2024-03-13
[Code health] Cleanup `gradle:checkCode` lint errors and deprecations.


sufyanAbbasi:
Seems like there's a growing list of deprecations produced by the `gradle:checkCode` workflow. We should consider going through them and fixing them all!

PR #2968, opened 2025-01-02
Fix activeNetworkInfo Deprecation

### LINKED ISSUE #2967: activeNetworkInfo is Deprecated
```kt
   val networkInfo = cm.activeNetworkInfo
    return networkInfo?.isConnected ?: false
```

The activeNetworkInfo method in ConnectivityManager has been deprecated since API level 29 (Android 10).

anandwana001:
Fixes #2967 

Parent Ticket #2338 

This PR fixes the `activeNetworkInfo` deprecation by using `activeNetwork`

@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
Can you please add details on how you verified that no regressions are introduced with this change?

anandwana001:
> Can you please add details on how you verified that no regressions are introduced with this change?


Manul testing, should we create a separate ticket for unit test network manager class?

shobhitagarwal1612:
Some tests are failing, can you please look into it?

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
```diff
@@ -61,11 +61,13 @@ class NetworkManager @Inject constructor(@ApplicationContext private val context
...
-  /** Returns true iff the device has internet connectivity, false otherwise. */
+  /** Returns true if the device has internet connectivity, false otherwise. */
   @RequiresPermission("android.permission.ACCESS_NETWORK_STATE")
   fun isNetworkConnected(): Boolean {
-    val cm = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
-    val networkInfo = cm.activeNetworkInfo
-    return networkInfo?.isConnected ?: false
+    val connectivityManager =
+      context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+    val activeNetwork = connectivityManager.activeNetwork
```
shobhitagarwal1612, line 69: Can `activeNetwork` be null? If yes, can we add null check for safety?

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
```diff
@@ -61,11 +61,13 @@ class NetworkManager @Inject constructor(@ApplicationContext private val context
...
+    val networkCapabilities = connectivityManager.getNetworkCapabilities(activeNetwork)
+    return networkCapabilities?.hasCapability(NET_CAPABILITY_INTERNET) == true
```
shobhitagarwal1612, line 71: Let's add a log with the result boolean value.

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
anandwana001: as the next LoC accepts null and eventually it will return false, so I think we don't need extra null check.

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
shobhitagarwal1612: Yes. It would be helpful when debugging the app locally to understand how frequently the function is invoked or if it fails what was the last checked value.

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
shobhitagarwal1612: SG

ground/src/main/java/com/google/android/ground/system/NetworkManager.kt
shobhitagarwal1612, line 71: `== true` can be replaced with `?: false`

shobhitagarwal1612 (approved):

PR #2974, opened 2025-01-02
ViewModelProviders is updated with ViewModelProvider

### LINKED ISSUE #2973: ViewModelProviders is deprecated

anandwana001:
Fixes #2973 

Parent Ticket #2338

`ViewModelProviders` is deprecated and updated with ViewModelProvider.

@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
Have we tested for regressions? Can you please specify all the edge cases tested?

anandwana001:
> Have we tested for regressions? Can you please specify all the edge cases tested?

Did a manual testing for screens which are using this factory class.

shobhitagarwal1612:
Can you please resolve the build failures before I take a pass?

ground/src/main/java/com/google/android/ground/ui/common/ViewModelFactory.kt
```diff
@@ -37,8 +36,22 @@ constructor(
...
+        @Suppress("UNCHECKED_CAST")
+        return viewModel as T
+      } else {
+        error(
+          "Expected instance of ${modelClass.simpleName}, but got ${viewModel::class.simpleName}"
+        )
+      }
+    } catch (e: Exception) {
+      throw ViewModelCreationException("Failed to create ViewModel for class $modelClass", e)
+    }
```
shobhitagarwal1612, line 54: Why are we catching and re-throwing the errors?

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/common/ViewModelFactory.kt
anandwana001: The initial idea of having 3 throws:
1. if the modelClass is unknown
2. if the modelClass is not a viewModel class type then throw in else part
3. If we get any error while `val viewModel = creator.get()` getting then our custom throw `ViewModelCreationException`.

shobhitagarwal1612: I would argue that it is better to throw directly and now catch and rethrow as it doesn't provide any extra benefits. We won't be using `ViewModelCreationException` for any special handling.

anandwana001: fixed by removing the extra throw

PR #2984, opened 2025-01-06
Use extension function for creating bitmap from vector drawables

shobhitagarwal1612:
This is a no-op change which replaces custom implementation for creating bitmaps with extension function `toBitmap()`.

Verified that this doesn't introduce any regressions by opening the draw area fragment which uses this util function. Also adds ktdoc to all public methods.

@anandwana001 @gino-m  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/util/BitmapUtil.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/util/BitmapUtil.kt
@@ -18,43 +18,44 @@ package com.google.android.ground.ui.util
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
-import android.graphics.Canvas
 import android.graphics.ImageDecoder
 import android.net.Uri
 import android.os.Build
+import androidx.annotation.DrawableRes
 import androidx.core.content.ContextCompat
+import androidx.core.graphics.drawable.toBitmap
 import dagger.hilt.android.qualifiers.ApplicationContext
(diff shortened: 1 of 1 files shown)


ground/src/main/java/com/google/android/ground/ui/util/BitmapUtil.kt
```diff
@@ -18,43 +18,44 @@ package com.google.android.ground.ui.util
...
 
-  /** Retrieves an image for the given url as a [Bitmap]. */
-  fun fromUri(url: Uri?): Bitmap {
-    require(url != null) { "Uri cannot be null" }
-
-    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+  /**
+   * Retrieves an image for the given URI as a [Bitmap].
+   *
+   * For Android P (API 28) and above, it uses [ImageDecoder] for better performance and support for
```
scolsen, line 35: Nit: This isn't hardcoded in the impl of this function--so it might change if the constants we reference change. If it's totally unlikely that that will happen I say leave this otherwise consider referring to the constants directly.

scolsen (approved):
Nice!

ground/src/main/java/com/google/android/ground/ui/util/BitmapUtil.kt
shobhitagarwal1612: Removed the impl details from ktdoc.

PR #2990, opened 2025-01-08
Show Toast when offline in Download

### LINKED ISSUE #662: [Offline basemap download] Do not let user open "Select and download" when offline
Instead, show a modal dialog "Offline basemaps may only be downloaded when online" or similar.

anandwana001:
Towards #662 

This is a part 1 PR where we are showing a popup if device is not connected while trying to download the offline imagery. 

- [ ] When offline, show message "You are offline. Connect to download offline map imagery".

Part 2 will be
- [ ] If download fails, show error ("Connection lost") or similar.
The part 2 will close other tickets as well where we need to build download cancel feature,

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (commented):
Are you planning to add some unit tests for this?

anandwana001:
> Are you planning to add some unit tests for this?

in second pr, for complete testing.

ground/src/main/java/com/google/android/ground/ui/offlineareas/selector/OfflineAreaSelectorViewModel.kt
```diff
@@ -79,7 +81,16 @@ internal constructor(
   private val _navigate = MutableSharedFlow<UiState>(replay = 0)
   val navigate = _navigate.asSharedFlow()
 
+  private val _showPopupEvent = MutableSharedFlow<Boolean>()
+  val showPopupEvent = _showPopupEvent.asSharedFlow()
```
shobhitagarwal1612, line 85: Can we rename this to `networkUnavailableEvent`? Also, we aren't using the value of the emitted value. So the type can be changed to `Nil`.

ground/src/main/java/com/google/android/ground/ui/offlineareas/selector/OfflineAreaSelectorViewModel.kt
```diff
@@ -79,7 +81,16 @@ internal constructor(
...
+
   fun onDownloadClick() {
+    val isConnected = networkManager.isNetworkConnected()
+    if (!isConnected) {
```
shobhitagarwal1612, line 89: Let's inline this

shobhitagarwal1612 (changes requested):

ground/src/main/java/com/google/android/ground/ui/offlineareas/selector/OfflineAreaSelectorFragment.kt
```diff
@@ -81,6 +86,12 @@ class OfflineAreaSelectorFragment : AbstractMapContainerFragment() {
         }
       }
     }
+
+    lifecycleScope.launch {
+      viewModel.networkUnavailableEvent.collect { show ->
```
shobhitagarwal1612, line 91: Can we remove `show ->` as well?

shobhitagarwal1612 (changes requested):

shobhitagarwal1612 (approved):

ground/src/main/res/values/strings.xml
```diff
@@ -56,6 +56,7 @@
   <!-- Label of button used enter the offline area selector. -->
   <string name="offline_area_selector_select">Select area</string>
   <string name="no_basemaps_downloaded">No map imagery downloaded for offline use</string>
+  <string name="connect_to_download_message">You are offline. Connect to download offline map imager</string>
```
gino-m, line 59: @amegantz For edits here as well when you have cycles

PR #2986, opened 2025-01-07
[WIP] Integrate TaskSequenceHandler with DataCollectionViewModel

shobhitagarwal1612:
Fixes #2958

@... PTAL?

PR #2996, opened 2025-01-10
Refactor DataCollectionViewModel

shobhitagarwal1612:
This PR is a pure refactor which is split from https://github.com/google/ground-android/pull/2986 so that main logic change can be separated from refactors.

Things done:
 * Instead of using `currentTaskId.value` as default task id in all method, pass the id from the callers 
 * Prevent re-creating task sequence for each value when storing values to local db.
 * Cleanup unused code.

@anandwana001  PTAL?

### DIFF
--- a/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -209,14 +209,14 @@ internal constructor(
     val task = taskViewModel.task
     val taskValue = taskViewModel.taskTaskData.value
 
-    // Skip validation if the task is empty
-    if (taskValue.isNotNullOrEmpty()) {
-      val validationError = taskViewModel.validate()
-      if (validationError != null) {
-        popups.get().ErrorPopup().show(validationError)
-        return
+    taskValue
+      ?.takeIf { it.isNotNullOrEmpty() } // Skip validation if the task is empty
+      ?.let {
+        taskViewModel.validate()?.let {
+          popups.get().ErrorPopup().show(it)
+          return
+        }
       }
-    }
 
     data[task] = taskValue
     moveToPreviousTask()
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
@anandwana001 PTAL?

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -210,16 +210,12 @@ internal constructor(
...
-
-    val validationError = taskViewModel.validate()
-    if (validationError != null) {
-      popups.get().ErrorPopup().show(validationError)
-      return
+    if (taskValue.isNotNullOrEmpty()) {
+      val validationError = taskViewModel.validate()
+      if (validationError != null) {
+        popups.get().ErrorPopup().show(validationError)
+        return
```
anandwana001, line 217:
```suggestion
taskValue?.takeIf { it.isNotEmpty() }?.let {
    taskViewModel.validate()?.let { validationError ->
        popups.get().ErrorPopup().show(validationError)
        return
    }
}
```

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -277,23 +277,23 @@ internal constructor(
     }
   }
 
-  private fun getAbsolutePosition(): Int {
-    if (currentTaskId.value == "") {
+  private fun getAbsolutePosition(taskId: String): Int {
+    if (taskId == "") {
       return 0
     }
```
anandwana001, line 283:
```suggestion
if (taskId.isEmpty()) return 0
```

anandwana001 (approved):

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: Done

ground/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: Done

PR #2995, opened 2025-01-09
Introduce TaskDataHandler

shobhitagarwal1612:
Towards #2993 

Introduces a wrapper class for handling task data values. Provides a way to subscribe for changes to the task value. 

This is a no-op change. This will be integrated along with TaskSequenceHandler in DataCollectionViewModel in future CLs.

See https://github.com/google/ground-android/issues/2993#issue-2776885263 for detailed description.

@scolsen @anandwana001 PTAL?

Issue #2993, opened 2025-01-09
Add a method to update the cached sequence whenever necessary.


shobhitagarwal1612:
https://github.com/google/ground-android/blob/6972d362376b7f342d1ff1264546b9bb448bf607/ground/src/main/java/com/google/android/ground/ui/datacollection/TaskSequenceHandler.kt#L76-L77

```text

  }

  /**
   * Generates the task sequence based on whether a task should be included or not.
   *
   * This determines the order of tasks to be presented to the user, taking into account any
   * overrides specified by [taskValueOverride].
   *
   * @param taskValueOverride An optional pair where the first element is the task ID and the second
   *   element is the [TaskData] to override the default task data. If null, no override is applied.
   * @return A [Sequence] of [Task] objects representing the ordered tasks.
   */
  fun generateTaskSequence(taskValueOverride: Pair<String, TaskData?>? = null): Sequence<Task> =
    tasks.filter { task -> shouldIncludeTask(task, taskValueOverride) }.asSequence()

  fun getTaskSequence(): Sequence<Task> {
    if (!isSequenceInitialized) {
      taskSequence = generateTaskSequence()
      isSequenceInitialized = true
    }
    return taskSequence
  }

  // TODO: Add a method to update the cached sequence whenever necessary.
  // Issue URL: https://github.com/google/ground-android/issues/2993

  /**
   * Checks if the specified task is the first task in the displayed sequence.
   *

```

shobhitagarwal1612:
TaskSequence is generated from the list of all tasks. Tasks whose `condition` isn't satisfied are filtered out. For evaluating this `condition`, we use the `data` object in `DataCollectionViewModel` which stores `taskdata` corresponding to each task fragment.

So, we only need to update the sequence if `data` is mutated. Until then, the cached value can be used.

Proposal:
1. Rename current `getTaskSequence()` in `TaskSequenceHandler` to `createTaskSequence()`
2. Introduce a new method `refreshSequence()` in the same class which calls `createTaskSequence()` and saves the value in memory.
3. Create a new wrapper class for `data` called `TaskDataHandler` which exposes `setData` and `getData` for the `DataCollectionViewModel` for saving/fetching task data.
4. Only call `refreshSequence()` when `setData` is called to ensure that the sequence is updated only when the data is mutated.
5. Refactor the existing code base to use `getTaskSequence()` instead of `createTaskSequence()` (This would be done when integrating these classes with `DataCollectionViewModel`

[...]

PR #3005, opened 2025-01-15
Use TaskDataHandler in DataCollectionViewModel

shobhitagarwal1612:
Towards #2993

In #2995, we introduced `TaskDataHandler` for managing data state. This PR aims at using it in `DataCollectionViewModel`.

This shouldn't change the behavior and is no-op.

Split from https://github.com/google/ground-android/pull/2986

@anandwana001 @scolsen  PTAL?

### DIFF
--- a/app/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
+++ b/app/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
@@ -51,8 +51,6 @@ import com.google.android.ground.ui.datacollection.tasks.time.TimeTaskViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 import javax.inject.Provider
-import kotlin.collections.component1
-import kotlin.collections.component2
 import kotlin.collections.set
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -141,7 +139,7 @@ internal constructor(
   private val taskViewModels: MutableStateFlow<MutableMap<String, AbstractTaskViewModel>> =
     MutableStateFlow(mutableMapOf())
 
-  private val data: MutableMap<Task, TaskData?> = LinkedHashMap()
+  private val taskDataHandler = TaskDataHandler()
```
anandwana001, line 142: is it a good place for object creation?

anandwana001 (approved):

app/src/main/java/com/google/android/ground/ui/datacollection/DataCollectionViewModel.kt
shobhitagarwal1612: We later need this as a dependency for TaskSequenceHandler which will also be initialized at class level.

PR #3013, opened 2025-01-18
[Bug] Prevent saving to draft after task is submitted

shobhitagarwal1612:
Fixes an issue in data collection flow where the draft was getting saved after task submission causing the data collection fragment to be reopened just after collecting data.

Changes verified by running the app locally with and without this fix.

@anandwana001  PTAL?


anandwana001:
@shobhitagarwal1612 issue link?

anandwana001 (approved):

PR #3024, opened 2025-01-30
Show a warning dialog when attempting to close data collection screen

### LINKED ISSUE #2635: [Data collection] Clicking "x" in top left corner discards data without prompting user
User should be prompted whether they want to lose/discard the data they've collected so far:

Screenshot_20240812_153956

@amegantz @rawbzz for input.

shobhitagarwal1612:
Fixes #2635 

Adds a new warning dialog before discarding the in-progress data collection screen.

https://github.com/user-attachments/assets/a33557e6-c1cf-448e-a39d-d29e4768ba29

@anandwana001  PTAL?

### DIFF
--- a/app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
+++ b/app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
@@ -32,8 +32,8 @@ import com.google.android.ground.ui.theme.AppTheme
 fun ConfirmationDialog(
   @StringRes title: Int,
   @StringRes description: Int,
-  @StringRes dismissButton: Int = R.string.cancel,
-  @StringRes confirmButton: Int,
+  @StringRes dismissButtonText: Int = R.string.cancel,
+  @StringRes confirmButtonText: Int,
   onConfirmClicked: () -> Unit,
 ) {
   val showRemoveWarningDialog = remember { mutableStateOf(true) }
@@ -53,10 +53,10 @@ fun ConfirmationDialog(
       title = { Text(text = stringResource(title)) },
       text = { Text(text = stringResource(description)) },
       dismissButton = {
-        TextButton(onClick = { onDismiss() }) { Text(text = stringResource(dismissButton)) }
+        TextButton(onClick = { onDismiss() }) { Text(text = stringResource(dismissButtonText)) }
       },
       confirmButton = {
-        TextButton(onClick = { onConfirm() }) { Text(text = stringResource(confirmButton)) }
+        TextButton(onClick = { onConfirm() }) { Text(text = stringResource(confirmButtonText)) }
       },
     )
   }
@@ -70,7 +70,7 @@ fun PreviewConfirmationDialog() {
     ConfirmationDialog(
       title = R.string.data_collection_cancellation_title,
       description = R.string.data_collection_cancellation_description,
-      confirmButton = R.string.data_collection_cancellation_confirm_button,
+      confirmButtonText = R.string.data_collection_cancellation_confirm_button,
       onConfirmClicked = {},
     )
   }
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
```diff
@@ -20,27 +20,58 @@ import androidx.compose.material3.AlertDialog
...
 @Composable
 fun ConfirmationDialog(
   @StringRes title: Int,
   @StringRes description: Int,
-  @StringRes dismissButtonText: Int = R.string.cancel,
-  @StringRes confirmButtonText: Int,
-  signOutCallback: () -> Unit,
-  dismissCallback: () -> Unit,
+  @StringRes dismissButton: Int = R.string.cancel,
+  @StringRes confirmButton: Int,
```
anandwana001, line 36: technically we are passing text only, so why `dismissButtonText` -> `dismissButton`?

app/src/main/res/values/strings.xml
```diff
@@ -196,4 +196,8 @@
   <string name="allow_location_confirmation">Allow location</string>
   <string name="allow_location_title">Allow location sharing</string>
   <string name="allow_location_description">If you don&#8217;t allow Ground to access this device&#8217;s location, you won&#8217;t be able to continue collecting data for this site.</string>
+
+  <string name="data_collection_cancellation_title">Stop collecting data?</string>
+  <string name="data_collection_cancellation_description">If you exit, data you haven&#8217;t submitted yet will be discarded.</string>
+  <string name="data_collection_cancellation_confirm_button">Confirm</string>
```
anandwana001, line 202: we should add other translated strings

app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
```diff
@@ -20,27 +20,58 @@ import androidx.compose.material3.AlertDialog
...
+        TextButton(onClick = { onConfirm() }) { Text(text = stringResource(confirmButton)) }
+      },
+    )
+  }
+}
+
+@Composable
+@Preview
+@ExcludeFromJacocoGeneratedReport
+fun PreviewConfirmationDialog() {
```
anandwana001, line 68: just a suggestion, how about creating our own custom annotation which internally includes all these three annotations so that in future we have to add one annotation only and can prevent anyone forgetting exclude thing?

WDYT?

```kt
@Target(AnnotationTarget.FUNCTION)
@Retention(AnnotationRetention.RUNTIME)
@Composable
@Preview
@ExcludeFromJacocoGeneratedReport
annotation class ComposablePreview
```


```kt
@ComposablePreview
fun PreviewConfirmationDialog() {
   ...
}
```

app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
shobhitagarwal1612: Fixed

app/src/main/java/com/google/android/ground/ui/compose/ConfirmationDialog.kt
shobhitagarwal1612: Thanks for the suggestion! I would to keep `Preview` out of the proposed annotation as it gives us the flexibility to add other preview options which may or may not be applicable to all use cases (e.g. I don't like to add background to dialog, for single items i don't prefer showSystemUi so that multiple previews can be fit into one). Can we move this discussion to a new bug?

anandwana001: Sure, SG

app/src/main/res/values/strings.xml
shobhitagarwal1612: Done

shobhitagarwal1612:
@gino-m Who should we tag for validating the non-english translation strings?

gino-m:
@jo-spek

jo-spek (changes requested):
Minor change to French

shobhitagarwal1612:
Thank you @jo-spek

shobhitagarwal1612:
@jo-spek Can you please re-stamp the PR?

jo-spek (approved):

PR #3039, opened 2025-02-10
Move logic for showing data sharing terms to domain layers

### LINKED ISSUE #3031: [Code health] Cleanup logic for showing terms of service dialog
Currently, `HomescreenMapContainerFragment` contains a lot of key logic for showing data sharing terms dialog or error message. A lot of this can be moved to view models and the fragment can just handle the UI rendering logic. In addition to that, some of the coroutine / flow can be simplified by just using a getter from `SurveyRepository`

Data sharing terms was introduced in https://github.com/google/ground-android/pull/2637/

shobhitagarwal1612:
Fixes #3031

Simplifies the logic for showing data sharing terms by replacing flows with getter and moving core logic to domain layer with unit tests.

@anandwana001 @gino-m @sufyanAbbasi  PTAL?

### DIFF
--- a/app/src/main/java/com/google/android/ground/domain/usecases/datasharingterms/GetDataSharingTermsUseCase.kt
+++ b/app/src/main/java/com/google/android/ground/domain/usecases/datasharingterms/GetDataSharingTermsUseCase.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.domain.usecases.datasharingterms
+
+import com.google.android.ground.persistence.local.LocalValueStore
+import com.google.android.ground.proto.Survey.DataSharingTerms
+import com.google.android.ground.repository.SurveyRepository
+import javax.inject.Inject
+
+class GetDataSharingTermsUseCase
+@Inject
+constructor(
+  private val localValueStore: LocalValueStore,
+  private val surveyRepository: SurveyRepository,
+) {
+
+  /** Returns the data sharing terms for the currently active survey, if not already accepted. */
+  operator fun invoke(): DataSharingTerms? {
+    val survey = surveyRepository.activeSurvey ?: return null
+
+    if (localValueStore.getDataSharingConsent(survey.id)) {
+      // User previously agreed to the terms.
+      return null
+    }
+
+    return survey.dataSharingTerms
+  }
+}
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/google/android/ground/domain/usecases/datasharingterms/GetDataSharingTermsUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+class GetDataSharingTermsUseCase
+@Inject
+constructor(
+  private val localValueStore: LocalValueStore,
+  private val surveyRepository: SurveyRepository,
+) {
+
+  /** Returns the data sharing terms for the currently active survey, if not already accepted. */
+  operator fun invoke(): Result<DataSharingTerms?> = runCatching {
+    val survey = surveyRepository.activeSurvey ?: error("No active survey")
```
anandwana001, line 32: is it good idea to crash the app here? `error("No active survey")`

app/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -116,49 +100,55 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
...
-      ) {
-        ephemeralPopups.ErrorPopup().show(getString(R.string.invalid_data_sharing_terms))
-        return
+
+    mapContainerViewModel
+      .getDataSharingTerms()
+      .onSuccess { terms ->
+        if (terms == null) {
+          // Data sharing terms already accepted or missing.
+          navigateToDataCollectionFragment(cardUiData)
```
anandwana001, line 135:
```suggestion
          // Data sharing terms already accepted.
          navigateToDataCollectionFragment(cardUiData)
```

right?

app/src/main/res/values/strings.xml
```diff
@@ -200,4 +200,5 @@
   <string name="data_collection_cancellation_title">Stop collecting data?</string>
   <string name="data_collection_cancellation_description">If you exit, data you haven&#8217;t submitted yet will be discarded.</string>
   <string name="data_collection_cancellation_confirm_button">Confirm</string>
+  <string name="something_went_wrong">Something went wrong</string>
```
anandwana001, line 203: can we add translations

app/src/main/java/com/google/android/ground/domain/usecases/datasharingterms/GetDataSharingTermsUseCase.kt
shobhitagarwal1612: We don't crash the app. The method is wrapped with `runCatching`. So, the value is wrapped in a `Result`. In the fragment, we show a toast message for failures.

shobhitagarwal1612: Also added unit tests to confirm that the exception isn't thrown, instead failing result is returned.

app/src/main/res/values/strings.xml
shobhitagarwal1612: Done

app/src/main/java/com/google/android/ground/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
shobhitagarwal1612: We don't show the dialog if the survey hasn't configured a data sharing terms. This PR doesn't change the existing behavior.

anandwana001 (approved):

PR #3042, opened 2025-02-10
Separate UI and logic by moving composable components out of JobMapComposables class

shobhitagarwal1612:
Reduces complexity of JobMapComposables class by moving composable components to separate files. Also adds preview to each and simplifies the composable logic wherever possible.

Verified for regressions using previews.

@anandwana001 @gino-m @sufyanAbbasi  PTAL?

### DIFF
--- a/app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
+++ b/app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.home.mapcontainer.jobs
+
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Add
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.Icon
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.google.android.ground.ExcludeFromJacocoGeneratedReport
+import com.google.android.ground.R
+import com.google.android.ground.ui.theme.AppTheme
+
+@Composable
+fun ActionButton(
```
anandwana001, line 38: what if there is another variant of action button with no icon?

app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
```diff
@@ -0,0 +1,70 @@
...
+    modifier = modifier.size(width = 100.dp, height = 100.dp),
+    colors = getActionButtonColors(),
+    shape = RoundedCornerShape(25),
+  ) {
+    Icon(imageVector = icon, contentDescription = contentDescription, Modifier.size(65.dp))
+  }
+}
+
+@Composable
+private fun getActionButtonColors() =
```
anandwana001, line 55: why separate function?

anandwana001 (commented):
Can we confirm the default theme behaviour of these buttons on both light and dark system on level 34?

app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
shobhitagarwal1612: This composable is local to map container UI. If we ever need to add one with icon, we'll refactor accordingly and pull to common code.

app/src/main/java/com/google/android/ground/ui/home/mapcontainer/jobs/ActionButton.kt
shobhitagarwal1612: this is for readability purposes since the nesting was making harder to read the code. What would you suggest?

anandwana001: I was thinking as it was just a single line and not a big one, so we can keep inside function only. 
But for now, let's keep this way, later we can introduce a design system and all such stuff will come from there.

anandwana001 (approved):

PR #3057, opened 2025-02-20
[Draw area task] Show length of last vertex in tooltip

### LINKED ISSUE #2141: ["Draw area" and "drop pin" tasks] Show current and selected location and distance

anandwana001:
Fixes #2141 

@shobhitagarwal1612  PTAL?

https://github.com/user-attachments/assets/b51d80bd-0f20-4564-b771-26760c87f0e0


shobhitagarwal1612:
Looks like the PR is still in progress. The calculated value isn't being displayed anywhere. Am I missing something?

anandwana001:
> Looks like the PR is still in progress. The calculated value isn't being displayed anywhere. Am I missing something?

Need some suggestion on the UI part, as currently the line we are drawing is a custom one, how should we proceed with this new UI of chip?

shobhitagarwal1612:
As per the mocks, the distance chip should be shown over the line in the center position.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -43,6 +56,7 @@ constructor(private val resources: Resources, private val bitmapUtil: BitmapUtil
     val bitmap = bitmapUtil.fromVector(R.drawable.ic_circle_marker)
     CustomCap(BitmapDescriptorFactory.fromBitmap(bitmap))
   }
+  private var distanceMarker: Marker? = null
```
gino-m, line 59: Same here - "Distance marker" is specific to one particular use of LineStrings. If we were thinking in terms of generic linestrings, we could cll this the "tooltipMarker".

gino-m (changes requested):
@anandwana001 Can we please use `m` as the default unit for now? Eventually we should us `LocaleData.getMeasurementSystem(ULocale.getDefault())` to get the system and show `m` or `ft` accordingly.

Also - ideally the edge and distance would update while the drag is in progress, not only when it stops, but that's out of scope of this PR.

app/src/main/java/org/groundplatform/android/model/geometry/Geometry.kt
```diff
@@ -92,13 +93,27 @@ data class LineString(val coordinates: List<Coordinates>) : Geometry {
   override val area: Double
     get() = 0.0
 
+  val tooltipDistance: Double?
```
gino-m, line 96: Does "tooltip distance" make sense as a property of all geometries? This doesn't seem to be something specific to geometries at all and should be calculated in domain layer.

app/src/main/java/org/groundplatform/android/model/geometry/Geometry.kt
```diff
@@ -92,13 +93,27 @@ data class LineString(val coordinates: List<Coordinates>) : Geometry {
...
+    }
+
   override fun center(): Coordinates = coordinates.centerOrError()
 
   override fun isEmpty() = coordinates.isEmpty()
 
   fun isClosed(): Boolean =
     coordinates.size >= 4 && coordinates.firstOrNull() == coordinates.lastOrNull()
 
+  private fun calculateDistanceInMeters(start: Coordinates, end: Coordinates): Double {
```
gino-m, line 111: This isn't specific to Geometry; perhaps move into an existing Ext.kt file as `Coordinates.getDistanceInMeters`?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -72,6 +85,90 @@ constructor(private val resources: Resources, private val bitmapUtil: BitmapUtil
       jointType = JointType.ROUND
       zIndex = POLYLINE_Z
     }
+
+    updateTooltipMarker(map, geometry)
     return polyline
   }
+
+  private fun updateTooltipMarker(map: GoogleMap, geometry: LineString) {
+    if (geometry.coordinates.size < 2) {
```
gino-m, line 94: If the marker is generic, why does number of coordinates matter? Shouldn't it be removed if the tooltip is null or empty?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -72,6 +85,90 @@ constructor(private val resources: Resources, private val bitmapUtil: BitmapUtil
...
+      removeTooltipMarker()
+      return
+    }
+
+    val lastVertex = geometry.coordinates.last()
+    val secondLastVertex = geometry.coordinates[geometry.coordinates.size - 2]
+
+    val isPolygonClosed = geometry.isClosed()
+
+    if (isPolygonClosed) {
```
gino-m, line 104: Same here - the renderer should only care whether there's a tooltip set on the LineString or not, not the number of vertices; it's up to creator of the LineString to set the correct tooltip (or not).

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
```diff
@@ -0,0 +1,62 @@
...
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      val measurementSystem = LocaleData.getMeasurementSystem(ULocale.getDefault())
+      measurementSystem == LocaleData.MeasurementSystem.US
+    } else {
+      isImperialSystemFallback()
+    }
+
+  val (convertedDistance, unitString, unitMeasure) =
+    if (isImperial) {
+      Triple(distanceInMeters * 3.28084, resources.getString(R.string.unit_feet), MeasureUnit.FOOT)
```
gino-m, line 39: Let's move the conversion of meters to feet into its own extension function.

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.icu.text.MeasureFormat
+import android.icu.util.LocaleData
+import android.icu.util.Measure
+import android.icu.util.MeasureUnit
+import android.icu.util.ULocale
+import android.os.Build
+import java.util.Locale
+import org.groundplatform.android.R
+
+fun formatDistance(resources: Resources, distanceInMeters: Double): String {
```
gino-m, line 28: The filename seems to imply this is a class which represents a conversion, rather than a series of functions. Can these be moved into a new or existing Ext.kt file?

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
```diff
@@ -0,0 +1,62 @@
...
+  val (convertedDistance, unitString, unitMeasure) =
+    if (isImperial) {
+      Triple(distanceInMeters * 3.28084, resources.getString(R.string.unit_feet), MeasureUnit.FOOT)
+    } else {
+      Triple(distanceInMeters, resources.getString(R.string.unit_meters), MeasureUnit.METER)
+    }
+
+  val roundedDistance =
+    if (convertedDistance < 10) {
+      "%.2f".format(Locale.US, convertedDistance)
```
gino-m, line 46: We should use the system locale when formatting, not `US`

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
gino-m, line 39: Can this be code simplified to make it more understandable?

gino-m, line 39: Please move conversion factor in a constant.

gino-m (changes requested):
A few comments to improve separation of concerns and overall readability.

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
anandwana001: The idea was to use this new file "FormatConversion" as a single place where all the functions which are responsible of format conversion will get placed here

app/src/main/java/org/groundplatform/android/ui/map/gms/features/TooltipMarkerRenderer.kt
```diff
@@ -0,0 +1,92 @@
...
+import android.graphics.Typeface
+import com.google.android.gms.maps.GoogleMap
+import com.google.android.gms.maps.model.BitmapDescriptor
+import com.google.android.gms.maps.model.BitmapDescriptorFactory
+import com.google.android.gms.maps.model.LatLng
+import com.google.android.gms.maps.model.Marker
+import com.google.android.gms.maps.model.MarkerOptions
+import javax.inject.Inject
+
+class TooltipMarkerRenderer @Inject constructor() {
```
gino-m, line 33: I see you created a new rendered for tooltips, good idea. Can you please at ktdoc to public classes and methods briefly defining its scope?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringOverlayRenderer.kt
```diff
@@ -0,0 +1,49 @@
...
+  private val tooltipMarkerRenderer: TooltipMarkerRenderer,
+) {
+
+  fun renderOverlayFor(map: GoogleMap, line: LineString) {
+    if (line.coordinates.size < 2) {
+      remove()
+      return
+    }
+    val coords = line.coordinates
+    val distance = coords.tooltipDistanceIfLineStringClosed() ?: return
```
gino-m, line 39: It seems you've created this class to define the distance behavior, but I think we agreed that would be up to the ViewModel or business layer to calculate and set; this package only cares about rendering whatever tooltip is set on the `LineString`; it shouldn't know that this particular tooltip is computing using distance. I don't think we need this class at all actually.

gino-m (changes requested):
Structural comments

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
```diff
@@ -0,0 +1,85 @@
...
+    "$rounded $unitLabel"
+  }
+}
+
+private fun formatWithMeasureFormat(value: String, unit: MeasureUnit, fallback: String): String {
+  return try {
+    val formatter = MeasureFormat.getInstance(ULocale.getDefault(), MeasureFormat.FormatWidth.SHORT)
+    val number = value.toDoubleOrNull() ?: return fallback
+    formatter.format(Measure(number, unit))
+  } catch (e: Exception) {
```
gino-m, line 62: When might could this exception be thrown, ie what case is "fallback" trying to handle?

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
```diff
@@ -0,0 +1,85 @@
...
+    formatWithMeasureFormat(rounded, measureUnit, "$rounded $unitLabel")
+  } else {
+    "$rounded $unitLabel"
+  }
+}
+
+private fun formatWithMeasureFormat(value: String, unit: MeasureUnit, fallback: String): String {
+  return try {
+    val formatter = MeasureFormat.getInstance(ULocale.getDefault(), MeasureFormat.FormatWidth.SHORT)
+    val number = value.toDoubleOrNull() ?: return fallback
```
gino-m, line 60: Why are we passing value as a String here?

app/src/main/java/org/groundplatform/android/ui/util/FormatConversion.kt
gino-m: Great. Want to turn this into an extension function, `Double.formatDistance(Resources): String`? I'd suggest keeping this file more narrowly scoped, eg `DistanceFormatExt.kt`

gino-m (changes requested):

gino-m:
If distance is 0 I think we shouldn't show the label, as it obscures the underlying UI.

app/src/main/java/org/groundplatform/android/ui/util/DistanceFormatExt.kt
```diff
@@ -0,0 +1,108 @@
...
+  if (distance < DISTANCE_THRESHOLD) {
+    (distance * DECIMAL_MULTIPLIER).roundToInt() / DECIMAL_MULTIPLIER
+  } else {
+    distance.roundToInt().toDouble()
+  }
+
+private fun formatDistanceFallback(number: Double, unitLabel: String, locale: Locale): String {
+  val formatted =
+    if (number < 10) {
+      String.format(locale, "%.2f", number)
```
gino-m, line 70: The only thing which changes here is the format String, so no need to repeat the whole `String.format` call twice.

app/src/main/java/org/groundplatform/android/ui/util/DistanceFormatExt.kt
```diff
@@ -0,0 +1,108 @@
...
+  fallback: String,
+  locale: Locale,
+): String =
+  runCatching {
+      val formatter =
+        MeasureFormat.getInstance(ULocale.forLocale(locale), MeasureFormat.FormatWidth.SHORT)
+      formatter.format(Measure(number, measureUnit))
+    }
+    .getOrElse { exception ->
+      Timber.w("FormatDistance", "MeasureFormat failed", exception)
```
gino-m, line 89: * `Throwable` should be the first arg. 
* `FormatDistance` looks like an other reference to the file name, which is not needed since the stacktrace will include everything.
* `MeasureFormat` is the name of the class, which will also be in the stack trace i ; ideally we'd provide info about the potential cause. Why would an exception be thrown here?

app/src/main/java/org/groundplatform/android/ui/util/DistanceFormatExt.kt
```diff
@@ -0,0 +1,108 @@
...
+    runCatching {
+        LocaleData.getMeasurementSystem(ULocale.forLocale(locale)) ==
+          LocaleData.MeasurementSystem.US
+      }
+      .getOrElse { exception ->
+        Timber.w("MeasurementSystem", "Failed to get measurement system from LocaleData", exception)
+        isImperialSystemFallback(locale)
+      }
+  } else {
+    isImperialSystemFallback(locale)
```
gino-m, line 104: Can we short circuit to reduce nesting here, checking and returning this value first on older Androids?

app/src/main/java/org/groundplatform/android/util/CoordinatesExt.kt
```diff
@@ -37,3 +38,23 @@ private fun decimalToDms(latOrLong: Double): String {
...
+  return LatLng((start.lat + end.lat) / 2, (start.lng + end.lng) / 2)
+}
+
+private fun Coordinates.distanceTo(other: Coordinates): Double {
+  val result = FloatArray(1)
+  Location.distanceBetween(this.lat, this.lng, other.lat, other.lng, result)
+  return result[0].toDouble()
+}
+
+private fun List<Coordinates>.isClosedLineString(): Boolean = size >= 4 && first() == last()
```
gino-m, line 60: This doesn't actually test whether these coords are part of a `LineString`, which is a semantic distinction it cannot know about.  I think `isClosed` would be more correct here, in which case `size` must be `>0`.

We shouldn't encode the UI-specific logic (only show when label size >=4) here, but rather in the business layer.

In Koltin you can remove the return type, since it's immediate clear this is a Boolean expression.

gino-m (changes requested):
Good progress, but needs more work and unit tests. PTAL!

app/src/main/java/org/groundplatform/android/ui/map/gms/features/TooltipMarkerRenderer.kt
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * If no tooltip marker exists, a new one is created at [midPoint] with [tooltipText]. If a
+   * tooltip marker already exists, its position is updated to [midPoint] and its icon is updated
+   * using [tooltipText].
+   *
+   * @param map The [GoogleMap] instance where the tooltip marker should be rendered.
+   * @param midPoint The [LatLng] position for the tooltip marker.
+   * @param tooltipText The text to be displayed on the tooltip marker.
+   */
+  fun update(map: GoogleMap, midPoint: LatLng?, tooltipText: String?) {
```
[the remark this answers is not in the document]
anandwana001: We did that earlier, but as it needs to call from LineStringRenderer and as per our discussion that it should not be LineStringRenderer's responsibility, hence removed. Now, this file will handle such cases itself.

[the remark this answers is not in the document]
gino-m: The there point was that use-case specific business logic shouldn't be in the data or rendering layer. Each layer should expose the smallest surface possible; having two nullables here introduced 4 possible combinations of state of nullability state. Instead, this method should be hermetic and should only be used to update, not to remove.

gino-m:
@anandwana001 I've provided my suggestions as commits. Could you please provide unit tests for the new functionality?

gino-m:
@anandwana001 Any idea why `assertButtonIsHidden("Next")` might be failing?

gino-m:
@shobhitagarwal1612 Can you PTAL the new tests here? I'll have limited internet access this week.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
shobhitagarwal1612, line 1: Why is this change needed? Does it mean that this PR introduces regressions?

shobhitagarwal1612 (changes requested):

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
anandwana001: not sure why this test is passing earlier, but 
if you click on Complete button, the next button is not hidden and complete button is no more there on screen.

shobhitagarwal1612: Is this issue reproducible on HEAD branch too?

anandwana001: No, but the flow in production app is this only what we see below, 

1. click on complete
2. next button is visible
3. complete button is no more visible


Just test on production app from play store @shobhitagarwal1612

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -210,18 +217,33 @@ internal constructor(
...
+      geometry = LineString(vertices),
+      style = Feature.Style(strokeColor, Feature.VertexStyle.CIRCLE),
+      clusterable = false,
+      selected = true,
+      tooltipText = getDistanceTooltipText(),
+    )
+
+  /**
+   * Returns the distance tooltip if there is more than one vertex, if the last two vertices do not
+   * overlap, and if the polygon has not yet been marked as complete by the user.
```
shobhitagarwal1612, line 238: It is not recommended to write the low-level implementation details in the method doc. Instead it should reflect what the function is doing. In this case, it can look something like:

```
Returns the distance in meters between the last two vertices for displaying in the tooltip.
```

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -72,6 +78,13 @@ constructor(private val resources: Resources, private val bitmapUtil: BitmapUtil
       jointType = JointType.ROUND
       zIndex = POLYLINE_Z
     }
+    updateTooltipMarker(geometry, map, tooltipText)
```
shobhitagarwal1612, line 81: Should we add a check for null text to prevent showing empty markers?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/TooltipMarkerRenderer.kt
shobhitagarwal1612, line 55: Let's rename `midPoint` to `position` and make it non-nullable. `tooltipText` should also be non-nullable.

app/src/main/java/org/groundplatform/android/ui/util/LocaleAwareMeasureFormatter.kt
```diff
@@ -0,0 +1,61 @@
...
+import kotlin.math.floor
+import kotlin.math.pow
+
+private const val METERS_TO_FEET = 3.28084
+/** Distances equal to or greater than this value are rounded down to the nearest integer. */
+private const val MIN_ROUNDED_DISTANCE = 10
+/** The number of decimal places shown for distances < MIN_ROUNDED_DISTANCE. */
+private const val SMALL_DISTANCE_DECIMAL_PLACES = 1
+
+class LocaleAwareMeasureFormatter @Inject constructor(private val locale: Locale) {
```
shobhitagarwal1612, line 33: Can we add unit tests for this class?

app/src/main/java/org/groundplatform/android/util/CoordinatesExt.kt
```diff
@@ -37,3 +38,22 @@ private fun decimalToDms(latOrLong: Double): String {
...
+  val result = FloatArray(1)
+  Location.distanceBetween(this.lat, this.lng, other.lat, other.lng, result)
+  return result[0].toDouble()
+}
+
+/**
+ * Returns true iff the provided coordinates for a closed shape. A shape is considered closed if it
+ * has at least three (3) sized and the first and last coordinates are equal.
+ */
+fun List<Coordinates>.isClosed() = size >= 3 && first() == last()
```
shobhitagarwal1612, line 59: Let's add unit tests for this method as it is a core logic for polygons

app/src/main/java/org/groundplatform/android/util/CoordinatesExt.kt
```diff
@@ -37,3 +38,22 @@ private fun decimalToDms(latOrLong: Double): String {
   val (degrees, minutes, seconds) = dmsFormat.split(":".toRegex())
   return "$degrees°$minutes'$seconds\""
 }
+
+fun Coordinates.midpoint(other: Coordinates) =
```
shobhitagarwal1612, line 42: Please add unit tests

shobhitagarwal1612 (changes requested):

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModelTest.kt
```diff
@@ -171,6 +171,26 @@ class DrawAreaTaskViewModelTest : BaseHiltTest() {
     }
   }
```
shobhitagarwal1612, line 173: Let's add more tests where the tooltip isn't rendered

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
anandwana001: handled internally to TooltipMarkerRenderer

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModelTest.kt
anandwana001: added

app/src/main/java/org/groundplatform/android/util/CoordinatesExt.kt
anandwana001: no more required, as we are not using this function anymore

app/src/main/java/org/groundplatform/android/util/CoordinatesExt.kt
anandwana001: done

shobhitagarwal1612 (approved):

app/src/main/java/org/groundplatform/android/ui/util/LocaleAwareMeasureFormatter.kt
anandwana001: added

app/src/main/java/org/groundplatform/android/ui/map/gms/features/TooltipMarkerRenderer.kt
anandwana001: taking in another pr

anandwana001:
Merging for the coming release @gino-m

PR #3058, opened 2025-02-20
Added a new Instruction task

### LINKED ISSUE #2690: [Data collection tasks] Add organizer-defined "instructions" task
It would be great to have a feature where a text block can be shown before, in between or after a task when collecting data with the Ground Android app.

This was identified as a useful addition for SERVIRs Ghana campaign to provide instructions to data collectors or to show/ read out statements to farmers about what is being done on their fields.

To jump to the next task, one has to press an 'OK' button.

This feature request relates to the Ground platform request https://github.com/google/ground-platform/issues/1991.

anandwana001:
Fixes #2690 

Screenshot 2025-03-19 at 7 52 20 PM

@... PTAL?


anandwana001:
Proto PR - https://github.com/google/ground-platform/pull/2146

anandwana001:
@rfontanarosa Added some comments on chat, I think something is missing in the proto or backend from where the instruction task type is coming as unknown. Let's connect

shobhitagarwal1612:
What would happen if someone opens the older version of the app which doesn't support the new task type? Would it crash or not?

anandwana001:
> What would happen if someone opens the older version of the app which doesn't support the new task type? Would it crash or not?

Yes, it will crash

app/src/main/java/org/groundplatform/android/model/submission/InstructionTaskData.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.model.submission
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class InstructionTaskData(val text: String) : TaskData {
+  override fun isEmpty(): Boolean = text.trim { it <= ' ' }.isEmpty()
```
shobhitagarwal1612, line 22: Let's rethink what it means for this task. We use this method to check if the task has any value or not. Given this is a read-only task, let's ensure we are doing the right thing here. Also, we should add a unit test for the same

app/src/main/java/org/groundplatform/android/model/submission/InstructionTaskData.kt
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.model.submission
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class InstructionTaskData(val text: String) : TaskData {
```
shobhitagarwal1612, line 21: Are we planning to upload this task's value to remote storage? IMO, this is a read-only task and it doesn't make much sense to include this task in the submission mutation. If yes, we need to ensure that the value is skipped when uploading the data to remote storage.

@gino-m Wdyt?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/instruction/InstructionTaskFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+  @Composable
+  private fun ShowTextField() {
+    Box(
+      modifier =
+        Modifier.fillMaxSize()
+          .background(color = Color.White)
+          .border(2.dp, MaterialTheme.colorScheme.primary, RoundedCornerShape(2.dp))
+          .padding(16.dp)
+    ) {
+      Text(text = viewModel.instructionsText, fontSize = 24.sp)
```
shobhitagarwal1612, line 64: Can we directly use `viewModel.task.label` to simplify the view model class?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/instruction/InstructionTaskFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+  override fun onCreateTaskBody(inflater: LayoutInflater): View = createComposeView {
+    ShowTextField()
+  }
+
+  override fun onCreateActionButtons() {
+    addButton(ButtonAction.INSTRUCTION_OK).setOnClickListener { moveToNext() }
+  }
+
+  @Composable
+  private fun ShowTextField() {
```
shobhitagarwal1612, line 56: Let's add a preview method as well for verifying the UI locally.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/instruction/InstructionTaskViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.groundplatform.android.ui.datacollection.tasks.AbstractTaskViewModel
+
+class InstructionTaskViewModel @Inject constructor() : AbstractTaskViewModel() {
+
+  lateinit var instructionsText: String
+
+  override fun initialize(job: Job, task: Task, taskData: TaskData?) {
+    super.initialize(job, task, taskData)
+    instructionsText = task.label
+  }
```
shobhitagarwal1612, line 31: See the previous comment to eliminate this change.

app/src/main/res/values/strings.xml
```diff
@@ -208,4 +208,6 @@
   <string name="clear">Clear</string>
 
   <string name="area_message">Area: %1$.2f m²</string>
+
+  <string name="instruction_ok_button_text">OK</string>
```
shobhitagarwal1612, line 212: As per mocks, the button text should be `Okay`

shobhitagarwal1612: https://github.com/google/ground-android/issues/2690

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/model/submission/InstructionTaskData.kt
shobhitagarwal1612: Can we add a unit test for this?

anandwana001: removed this file

shobhitagarwal1612: I meant test to confirm that we don't upload the task's data to remote storage. See `SubmissionMutationConverterTest` for more reference.

anandwana001: added test

shobhitagarwal1612 (approved):

PR #3059, opened 2025-02-22
Hide close button after data is collected

### LINKED ISSUE #3050: Hide close button when data is collected
The "Close" (X) button in the top-left corner of the data collection screen remains visible on the final step of the data collection process, *after* data has been collected. 

At this stage, the button's functionality is no longer relevant and could lead to user confusion.

The design in Figma should be changed as well. Wdyt @gino-m ?

Image

Image

shobhitagarwal1612:
Fixes #3050

Things done in this PR:
 1. Hide the close button when displaying the confirmation screen
 2. Handle edge case for back press on confirmation screen to directly navigate up
 3. Simplify compose UI creation of the dialog

Tested locally

@gino-m  PTAL?


gino-m:
Thanks, @shobhitagarwal1612 !

PR #3061, opened 2025-02-24
Added vibrate function when pressing "Add point" 

### LINKED ISSUE #2674: Haptic feedback / wait message when pressing buttons
**Problem:** Field work in Kenya showed that many people had rather old and slow Android smartphones that usually wouldn't react instantly when clicking a button. That became quite a problem when adding points, because people would push the "Add point" button again and again until finally a registered point was displayed on the map. By then however, they would have pushed so often that multiple points appeared, sometimes in far away locations.

**Possible solutions:** 
- Haptic feedback: Short vibration when pressing a button, especially the 'Add point' button. 
- Instant visual feedback: "Please wait" message or some loading icon displaying as the very first thing when pressing a button before the point registration classes are even called.
<br>
Both together would be great, too.

anandwana001:
Fixes #2674 

Added vibration on every point added to screen.

Need to add visual effect feedback.

@shobhitagarwal1612  @gino-m  PTAL?

### DIFF
--- a/app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
+++ b/app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
@@ -15,7 +15,12 @@
  */
 package com.google.android.ground.ui.datacollection.tasks.polygon
 
+import android.content.Context
+import android.os.Build
 import android.os.Bundle
+import android.os.VibrationEffect
+import android.os.Vibrator
+import android.os.VibratorManager
 import android.view.LayoutInflater
 import android.view.View
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -80,11 +85,31 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
...
-      addButton(ButtonAction.ADD_POINT).setOnClickListener { viewModel.addLastVertex() }
+      addButton(ButtonAction.ADD_POINT).setOnClickListener {
+        viewModel.addLastVertex()
+        vibrate()
+      }
     completeButton =
       addButton(ButtonAction.COMPLETE).setOnClickListener { viewModel.completePolygon() }
   }
 
+  private fun vibrate() {
```
shobhitagarwal1612, line 96: Let's move the logic for vibration to a new system manager class and put it under the appropriate package. Also, let's call the vibrate function from within the view model and not fragment

shobhitagarwal1612 (changes requested):

app/src/main/java/com/google/android/ground/ui/util/VibrationHelper.kt
```diff
@@ -0,0 +1,51 @@
...
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+        val vibratorManager =
+          context.getSystemService(Context.VIBRATOR_MANAGER_SERVICE) as VibratorManager
+        vibratorManager.defaultVibrator
+      } else {
+        context.getSystemService(Context.VIBRATOR_SERVICE) as Vibrator
+      }
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      vibrator.vibrate(VibrationEffect.createOneShot(100, VibrationEffect.DEFAULT_AMPLITUDE))
```
shobhitagarwal1612, line 46: Let's make `duration` a parameter of the function with default value `100`

shobhitagarwal1612 (changes requested):

gino-m (approved):
Thanks for the fix!

PR #3063, opened 2025-02-26
Show estimated area when drawing polygon

### LINKED ISSUE #2096: Show estimated area when drawing polygon
Requested by @lecrabe:

> [estimating area] will only be useful when we do ad hoc, anticipating… but that will be the primary use of GROUND for EUDR DPI so I want it to be kept on top of the priority list

anandwana001:
Fixes #2096 

Showing toast of the calculated area.
Once the design is finalized, in next phase of this PR we can put this area into a specific UI component as well. 

@shobhitagarwal1612  @gino-m  PTAL?

Screenshot 2025-02-26 at 1 00 10 PM


app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -104,6 +105,10 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
     if (isVisible && !viewModel.instructionsDialogShown) {
       showInstructionsDialog()
     }
+    viewModel.polygonArea.observe(
+      viewLifecycleOwner,
+      { area -> Toast.makeText(requireContext(), "Area: $area m²", Toast.LENGTH_SHORT).show() },
```
gino-m, line 110: Please move text template to strings.xml

app/src/main/java/com/google/android/ground/ui/util/PolygonUtil.kt
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+package com.google.android.ground.ui.util
+
+import com.google.android.ground.model.geometry.Coordinates
+import java.util.Locale
+import kotlin.math.abs
+import kotlin.math.cos
+
+// Calculates the area of a polygon using the Shoelace formula
```
gino-m, line 23: Convert to ktdoc, clarify return value units

app/src/main/java/com/google/android/ground/ui/util/PolygonUtil.kt
```diff
@@ -0,0 +1,54 @@
...
+  var area = 0.0
+  for (i in points.indices) {
+    val j = (i + 1) % points.size
+    val term = points[i].first * points[j].second - points[j].first * points[i].second
+    area += term
+  }
+
+  val finalArea = abs(area) / 2.0
+
+  return String.format(Locale.US, "%.2f", finalArea).toDouble()
```
gino-m, line 39: Why convert to a String then back to a double? Rounding etc should be done in the UI layer, not done during calculation.

gino-m (changes requested):

app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -107,6 +109,18 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
     if (isVisible && !viewModel.instructionsDialogShown) {
       showInstructionsDialog()
     }
+    viewModel.polygonArea.observe(
+      viewLifecycleOwner,
+      { area ->
+        val formattedArea = String.format(Locale.US, "%.2f", area)
```
gino-m, line 115: We should use `Locale.getDefault()` here instead.

app/src/main/java/com/google/android/ground/ui/util/PolygonUtil.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlin.math.abs
+import kotlin.math.cos
+
+/**
+ * Calculates the area of a polygon using the Shoelace formula.
+ *
+ * This function computes the area of a simple, non-self-intersecting polygon based on its vertex
+ * coordinates. The first coordinate is used as a reference to convert all other points to meters.
+ *
+ * **Note:** This function returns the raw area value without rounding. Formatting (e.g., rounding
```
gino-m, line 28: I think you can remove the mention about formatting, which doesn't make sense in the context of a Double and can be removed. Also no need to say `**Note:**`.

gino-m (changes requested):
LG, a few more tweaks

app/src/main/java/com/google/android/ground/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -112,7 +112,7 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
     viewModel.polygonArea.observe(
       viewLifecycleOwner,
       { area ->
-        val formattedArea = String.format(Locale.US, "%.2f", area)
+        val formattedArea = String.format(Locale.getDefault(), "%.2f", area)
```
gino-m, line 115: Can you just put the `%.2f` inside `R.string.area_message` instead of formatting it explicitly here?

gino-m (approved):

PR #3070, opened 2025-03-17
Rotate Image only on preview

### LINKED ISSUE #2663: [Photo task] Preserve compression and EXIF tags from device camera
We currently decompress and recompress the image before uploading, which could be lossy, wastes battery, and causes the EXIF info to be lost. @sufyanAbbasi FYI. Related to #2552.

anandwana001:
Fixes #2663 

@gino-m  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
@@ -25,7 +25,7 @@ import androidx.databinding.BindingAdapter
 import com.google.android.gms.common.SignInButton
 import com.squareup.picasso.Picasso
 import org.groundplatform.android.R
-import org.groundplatform.android.ui.util.RotateTransformation
+import org.groundplatform.android.ui.util.RotateUsingExif
 import timber.log.Timber
 
 /**
@@ -55,7 +55,7 @@ object BindingAdapters {
       Picasso.get()
         .load(uri)
         .placeholder(R.drawable.ic_photo_grey_600_24dp)
-        .transform(RotateTransformation(uri, view.context))
+        .transform(RotateUsingExif(uri, view.context))
         .into(view)
     } catch (exception: IllegalStateException) {
       // Needed for running unit tests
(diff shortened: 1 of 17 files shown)


gino-m (changes requested):
Nicely done! Can you please add a test that the image is rotated when expected?

app/src/main/java/org/groundplatform/android/ui/common/RotateTransformation.kt
shobhitagarwal1612, line 1: Can we move this class to `util` directory?

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
```diff
@@ -48,8 +48,14 @@ object BindingAdapters {
...
   @BindingAdapter("imageUri")
   fun bindUri(view: ImageView?, uri: Uri?) {
+    if (view == null || uri == null) return
+
     try {
-      Picasso.get().load(uri).placeholder(R.drawable.ic_photo_grey_600_24dp).into(view)
+      Picasso.get()
+        .load(uri)
+        .placeholder(R.drawable.ic_photo_grey_600_24dp)
+        .transform(RotateTransformation(uri, view.context)) // Apply transformation
```
shobhitagarwal1612, line 57: Can we drop the comment? It's implicit that the line is applying a transformation.

gino-m (approved):

app/src/main/java/org/groundplatform/android/ui/util/RotateTransformation.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.Matrix
+import android.net.Uri
+import androidx.exifinterface.media.ExifInterface
+import com.squareup.picasso.Transformation
+import java.io.IOException
+
+class RotateTransformation(private val uri: Uri, private val context: Context) : Transformation {
```
gino-m, line 26: This class automatically rotates the image specifically based on Exif tags, but that's not mentioned anywhere in the class name.   Naming this something more obvious, like `RotateUsingExif` would make the type of transformation clearer.

gino-m, line 26: Please add class ktdoc.

gino-m (changes requested):

PR #3093, opened 2025-04-04
PhotoTask: Don't show draft toast on photo capture

### LINKED ISSUE #3080: [Take a photo] "Unsaved data restored" shown after returning from the camera, picture not captured
The main map screen also flashes before restoring the draft data collection flow.

@shobhitagarwal1612 Are you able to take a look?

scolsen:
This PR adds a new flag to track whether or not we are restoring the application from an instantaneous photo capture. If so, we do not bother showing the draft resotred toast.

Since the HomeScreenFragment controls the toast display, we unfortunately need to track this state as shared state in the HomeScreenVM, giving access to this state in the PhotoTaskFragment. We might want to consider moving the toast display into the Task fragments themselves to better encapsulate this state in the future.

Fixes #3080

@gino-m

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/home/HomeScreenViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/home/HomeScreenViewModel.kt
@@ -37,7 +37,7 @@ import org.groundplatform.android.ui.common.AbstractViewModel
 import org.groundplatform.android.ui.common.SharedViewModel
 import timber.log.Timber
 
-private const val WAITING_PHOTO_CAPTURE_KEY = "awaiting_photo_capture"
+private const val AWAITING_PHOTO_CAPTURE_KEY = "awaiting_photo_capture"
 
 @SharedViewModel
 class HomeScreenViewModel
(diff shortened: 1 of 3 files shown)


anandwana001 (approved):

app/src/main/java/org/groundplatform/android/ui/home/HomeScreenViewModel.kt
```diff
@@ -36,6 +37,8 @@ import org.groundplatform.android.ui.common.AbstractViewModel
 import org.groundplatform.android.ui.common.SharedViewModel
 import timber.log.Timber
 
+private const val WAITING_PHOTO_CAPTURE_KEY = "awaiting_photo_capture"
```
gino-m, line 40: Nit: WAITING vs awaiting - let's use "AWAITING" for consistency?

gino-m (changes requested):
Great! Cool to see `SavedHandleState` in action. On nit, LGTM!

app/src/main/java/org/groundplatform/android/ui/home/HomeScreenViewModel.kt
scolsen: Done!

scolsen:
The TODO to issue check keeps failing -- is that expected?

gino-m:
> The TODO to issue check keeps failing -- is that expected?

I suspect this is because you're using a fork. We can force for now since you don't have any new TODOs - next time try creating a branch, `scolsen/<issue no>/<some-text>` or similar in the main repo instead.

Issue #2859, opened 2024-11-21
[Offline support] Some survey data not deleted on "remove offline access"


gino-m:
We should instead:

1. Show the same big scary warning show when signing out to warn users that unsynced data will be lost
2. Delete pending mutations, jobs, and any other survey [meta]data we may have missed from the local device

anandwana001:
https://github.com/user-attachments/assets/788abc18-e54b-4569-b8eb-0ef02eaa85a7


@gino-m @shobhitagarwal1612  Am I missing any step here?

kenstershiro:
From @gino-m , will need to manually delete the associated LOIs and submissions.

gino-m:
@anandwana001 I can take this one if you haven't started on it yet.

PR #3103, opened 2025-04-13
Fix updating and deletion of survey in local db 

gino-m:
To allow cascading delete to be used in a follow-up PR without risking deletion of job data:

- [x] Jobs are inserted or updated and deleted jobs are removed rather than all jobs being deleted and reinserted.

Also resolved several inconsistencies in local db in the process.

- [x] Deleted multiple choice options are removed.
- [x] Deleted tasks are removed.
- [x] When updating a task, conditional expressions, which have no natural identifier, are now deleted and reinserted.

Code health:

- [x] Extracts and cleans up `LocalSurveyDataStore` tests into their own class
- [x] Adds coverage for above behaviors.

Towards #2859
Towards #1491


app/src/main/java/org/groundplatform/android/persistence/local/room/stores/RoomSurveyStore.kt
```diff
@@ -69,9 +69,12 @@ class RoomSurveyStore @Inject internal constructor() : LocalSurveyStore {
    */
   override suspend fun insertOrUpdateSurvey(survey: Survey) =
     localDatabase.withTransaction {
+      // Update survey.
       surveyDao.insertOrUpdate(survey.toLocalDataStoreObject())
-      jobDao.deleteBySurveyId(survey.id)
+      // Add or update jobs and tasks.
       insertOrUpdateJobs(survey.id, survey.jobs)
+      // Delete removed jobs.
+      jobDao.deleteNotIn(survey.jobs.map { it.id })
```
shobhitagarwal1612, line 77: IIUC, this will remove jobs of other surveys as well. Is that intentional?

gino-m: Yikes, good catch!

shobhitagarwal1612 (approved):

PR #3116, opened 2025-04-15
Refactor: Replace check for invalid addLoiTask using task type instead of task value

### LINKED ISSUE #2981: Replace check for valid addLoiTask using task type instead of TaskValue's type.
https://github.com/google/ground-android/blob/861846839d3a220ff5dc95edd2b55e1cab7c7700/ground/src/main/java/com/google/android/ground/domain/usecases/submission/SubmitDataUseCase.kt#L71-L72

```text

    val addLoiTaskId = deltas.indexOfFirst { it.taskId == addLoiTask.id }
    if (addLoiTaskId < 0) error("Add LOI task response missing")
    val addLoiValue = deltas.removeAt(addLoiTaskId).newTaskData
    // TODO: Replace check for valid addLoiTask using task type instead of TaskValue's type.
    // Issue URL: https://github.com/google/ground-android/issues/2981
    if (addLoiValue !is GeometryTaskData) error("Invalid add LOI task response")
    return locationOfInterestRepository.saveLoi(
      addLoiValue.geometry,

```

shobhitagarwal1612:
Fixes #2981 

This caused an issue when AddLocationTask's value was also of type geometry, causing it to be considered as valid AddLoi task.

@anandwana001  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/model/job/Job.kt
+++ b/app/src/main/java/org/groundplatform/android/model/job/Job.kt
@@ -52,9 +52,6 @@ data class Job(
       .apply { check(size <= 1) { "Expected 0 or 1, found $size AddLoiTasks" } }
       .firstOrNull()
 
-  /** Returns true if the job has one or more tasks. */
-  fun hasTasks() = tasks.values.isNotEmpty()
-
   /** Returns whether the job has non-LOI tasks. */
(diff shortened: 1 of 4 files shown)


anandwana001 (approved):
do we need to update any test for this change?

shobhitagarwal1612:
Let me add more test coverage for the added lines. Thanks for flagging!

PR #3117, opened 2025-04-15
Added app language change option

### LINKED ISSUE #2826: Language switch button
Ground gets displayed in English / French / Spanish / Portuguese based on the users system locale. 

If it is not too much work to implement at the moment, it would be nice to have a language switch button.

(First suggested in https://github.com/google/ground-android/issues/2676. Other FR closed because it was also about French translations which exist now.)

anandwana001:
Fixes #2826

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/AbstractActivity.kt
+++ b/app/src/main/java/org/groundplatform/android/AbstractActivity.kt
@@ -15,6 +15,7 @@
  */
 package org.groundplatform.android
 
+import android.content.Context
 import android.os.Bundle
 import android.view.View
 import android.view.ViewGroup
@@ -23,11 +24,17 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
 import javax.annotation.OverridingMethodsMustInvokeSuper
+import org.groundplatform.android.ui.settings.onAttach
 import org.groundplatform.android.util.Debug
 import org.groundplatform.android.util.systemInsets
 
 /** Base activity class containing common helper methods. */
 abstract class AbstractActivity : AppCompatActivity() {
+
+  override fun attachBaseContext(newBase: Context) {
+    val context = onAttach(newBase)
+    super.attachBaseContext(context)
+  }
 
   override fun onCreate(savedInstanceState: Bundle?) {
     Debug.logLifecycleEvent(this)
--- a/app/src/main/java/org/groundplatform/android/GroundApplication.kt
+++ b/app/src/main/java/org/groundplatform/android/GroundApplication.kt
@@ -15,6 +15,7 @@
  */
 package org.groundplatform.android
 
+import android.content.Context
 import android.os.StrictMode
 import android.os.StrictMode.VmPolicy
 import android.util.Log
@@ -24,6 +25,7 @@ import androidx.work.Configuration
 import dagger.hilt.android.HiltAndroidApp
 import javax.inject.Inject
 import org.groundplatform.android.Config.isReleaseBuild
+import org.groundplatform.android.ui.settings.onAttach
 import timber.log.Timber
 
 @HiltAndroidApp
@@ -44,6 +46,10 @@ class GroundApplication : MultiDexApplication(), Configuration.Provider {
       // Log failures when trying to do work in the UI thread.
       setStrictMode()
     }
+  }
+
+  override fun attachBaseContext(base: Context) {
+    super.attachBaseContext(base?.let { onAttach(it) })
   }
 
   private fun setStrictMode() {
--- a/app/src/main/java/org/groundplatform/android/ui/settings/Keys.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/settings/Keys.kt
@@ -25,5 +25,7 @@ object Keys {
   const val VISIT_WEBSITE = "visit_website"
   const val FEEDBACK = "feedback"
 
-  val ALL_KEYS = arrayOf(UPLOAD_MEDIA, VISIT_WEBSITE, FEEDBACK)
+  const val SELECT_LANGUAGE = "select_language"
+
+  val ALL_KEYS = arrayOf(UPLOAD_MEDIA, VISIT_WEBSITE, FEEDBACK, SELECT_LANGUAGE)
 }
--- a/app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.ui.settings
+
+import android.content.Context
+import android.content.res.Configuration
+import androidx.preference.PreferenceManager
+import java.util.Locale
+
+fun onAttach(context: Context, defaultLanguage: String = "en"): Context {
+  val lang = getPersistedLanguage(context, defaultLanguage)
+  return setLocale(context, lang)
+}
+
+private fun getPersistedLanguage(context: Context, defaultLanguage: String): String {
+  val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+  return preferences.getString("select_language", defaultLanguage) ?: defaultLanguage
+}
+
+fun setLocale(context: Context, language: String): Context {
+  val locale = Locale(language)
+  Locale.setDefault(locale)
+
+  // Create a new configuration with the selected locale
+  val config = Configuration(context.resources.configuration)
+  config.setLocale(locale)
+
+  // Return a context with the updated configuration.
+  return context.createConfigurationContext(config)
+}
--- a/app/src/main/java/org/groundplatform/android/ui/settings/SettingsFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/settings/SettingsFragment.kt
@@ -19,8 +19,10 @@ import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
 import android.widget.Toast
+import androidx.preference.DropDownPreference
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
+import androidx.preference.PreferenceManager
 import androidx.preference.SwitchPreferenceCompat
 import org.groundplatform.android.Config
 import org.groundplatform.android.R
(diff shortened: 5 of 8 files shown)


app/src/main/java/org/groundplatform/android/AbstractActivity.kt
```diff
@@ -23,12 +24,18 @@ import androidx.appcompat.app.AppCompatActivity
...
 import org.groundplatform.android.util.Debug
 import org.groundplatform.android.util.systemInsets
 
 /** Base activity class containing common helper methods. */
 abstract class AbstractActivity : AppCompatActivity() {
 
+  override fun attachBaseContext(newBase: Context) {
+    val context = onAttach(newBase)
+    super.attachBaseContext(context)
+  }
```
shobhitagarwal1612, line 37: Why do we need this change both here and in the application class? Is only the application class change sufficient?

app/src/main/java/org/groundplatform/android/ui/settings/Keys.kt
```diff
@@ -25,5 +25,7 @@ object Keys {
   const val VISIT_WEBSITE = "visit_website"
   const val FEEDBACK = "feedback"
 
-  val ALL_KEYS = arrayOf(UPLOAD_MEDIA, VISIT_WEBSITE, FEEDBACK)
+  const val SELECT_LANGUAGE = "select_language"
```
shobhitagarwal1612, line 28: Should we call it just `LANGUAGE`?

app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.groundplatform.android.ui.settings
+
+import android.content.Context
+import android.content.res.Configuration
+import androidx.preference.PreferenceManager
+import java.util.Locale
+
+fun onAttach(context: Context, defaultLanguage: String = "en"): Context {
```
shobhitagarwal1612, line 23: Can we improve the method name to be more appropriate? This function doesn't need to know if it is called during a lifecycle event. Instead the name should reflect what the function is doing. Also, please add some ktdoc to it.

app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+  val lang = getPersistedLanguage(context, defaultLanguage)
+  return setLocale(context, lang)
+}
+
+private fun getPersistedLanguage(context: Context, defaultLanguage: String): String {
+  val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+  return preferences.getString("select_language", defaultLanguage) ?: defaultLanguage
```
shobhitagarwal1612, line 30: Default value shouldn't be a parameter and can be moved to `Config.kt`

app/src/main/java/org/groundplatform/android/ui/settings/SettingsFragment.kt
```diff
@@ -48,6 +51,28 @@ class SettingsFragment : PreferenceFragmentCompat(), Preference.OnPreferenceClic
 
     val switchPreference = findPreference<SwitchPreferenceCompat>(Keys.UPLOAD_MEDIA)
     switchPreference?.isChecked = loadSwitchPreferenceState()
+
+    val languagePreference = findPreference<DropDownPreference>(Keys.SELECT_LANGUAGE)
+    val selectedLanguage =
+      preferenceManager.sharedPreferences?.getString(Keys.SELECT_LANGUAGE, "en") ?: "en"
```
shobhitagarwal1612, line 57: Let's reuse the same default value here as well

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
```diff
@@ -0,0 +1,44 @@
...
+
+fun onAttach(context: Context, defaultLanguage: String = DEFAULT_LANGUAGE): Context {
+  val lang = getPersistedLanguage(context, defaultLanguage)
+  return setLocale(context, lang)
+}
+
+private fun getPersistedLanguage(context: Context, defaultLanguage: String): String {
+  val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+  return preferences.getString(Keys.LANGUAGE, defaultLanguage) ?: DEFAULT_LANGUAGE
+}
```
shobhitagarwal1612, line 32: Let's move this to `LocalValueStore` class where we keep get/set logic for all shared preferences.

app/src/main/java/org/groundplatform/android/ui/settings/LocaleHelper.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.groundplatform.android.Config.DEFAULT_LANGUAGE
+
+fun onAttach(context: Context, defaultLanguage: String = DEFAULT_LANGUAGE): Context {
+  val lang = getPersistedLanguage(context, defaultLanguage)
+  return setLocale(context, lang)
+}
+
+private fun getPersistedLanguage(context: Context, defaultLanguage: String): String {
+  val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+  return preferences.getString(Keys.LANGUAGE, defaultLanguage) ?: DEFAULT_LANGUAGE
```
gino-m, line 31: Let's also not hide the "default" behavior in the persistence layer. Note that the "default" should be the system language. `en` should only be the fallback when the system language is not supported.

anandwana001: Is this approach looks correct?


```kt
fun getAppLanguage(): String {
    preferences.getString(Keys.LANGUAGE, null)?.let { return it }
    return locale.language ?: DEFAULT_LANGUAGE
  }
```

app/src/main/java/org/groundplatform/android/AbstractActivity.kt
```diff
@@ -23,12 +24,18 @@ import androidx.appcompat.app.AppCompatActivity
...
 import androidx.core.view.WindowInsetsCompat
 import javax.annotation.OverridingMethodsMustInvokeSuper
+import org.groundplatform.android.ui.settings.onAttach
 import org.groundplatform.android.util.Debug
 import org.groundplatform.android.util.systemInsets
 
 /** Base activity class containing common helper methods. */
 abstract class AbstractActivity : AppCompatActivity() {
 
+  override fun attachBaseContext(newBase: Context) {
```
gino-m, line 34: Why is this needed at all?

app/src/main/java/org/groundplatform/android/Config.kt
```diff
@@ -44,6 +44,8 @@ object Config {
    */
   const val CLUSTERING_ZOOM_THRESHOLD = 14f
 
+  const val DEFAULT_LANGUAGE = "en"
```
gino-m, line 47: ktdoc please:

```
/**
 * ISO 639-1 code of language used when the system language is not supported.
 */
```

gino-m (changes requested):
Hi @anandwana001, have you tried following https://developer.android.com/guide/topics/resources/app-languages#impl-overview re adding language picker?

>Enable automatic per-app language support to generate a LocaleConfig file and add your app's languages to system settings.
>Optionally, if you want to add an in-app language picker: use the AndroidX library and opt in to our [API implementation](https://developer.android.com/guide/topics/resources/app-languages#androidx-impl) to support backward compatibility through autoStoreLocales.

shobhitagarwal1612:
I tested this branch locally and found a bug. Even after changing the language, the toast on the home screen still appears in english locale "zoom in to start collecting data". The string does contain translation in other languages.

anandwana001:
> I tested this branch locally and found a bug. Even after changing the language, the toast on the home screen still appears in english locale "zoom in to start collecting data". The string does contain translation in other languages.

Yes, we need to change approach as discussed earlier. Let me move the PR to draft for now.

anandwana001:
Screenshot 2025-05-29 at 11 22 27 AM


Tests are passing locally, but failing here.

app/src/main/java/org/groundplatform/android/GroundApplication.kt
```diff
@@ -44,6 +50,11 @@ class GroundApplication : MultiDexApplication(), Configuration.Provider {
       // Log failures when trying to do work in the UI thread.
       setStrictMode()
     }
+    val selectedLanguage =
+      PreferenceManager.getDefaultSharedPreferences(this)
+        ?.getString(Keys.LANGUAGE, DEFAULT_LANGUAGE) ?: Locale.getDefault().language
```
shobhitagarwal1612: We should be first using the system's language as default value

shobhitagarwal1612: Also, let's move this logic to a util class and use here and in setting fragment.

app/src/main/java/org/groundplatform/android/ui/settings/SettingsFragment.kt
```diff
@@ -48,6 +54,29 @@ class SettingsFragment : PreferenceFragmentCompat(), Preference.OnPreferenceClic
 
     val switchPreference = findPreference<SwitchPreferenceCompat>(Keys.UPLOAD_MEDIA)
     switchPreference?.isChecked = loadSwitchPreferenceState()
+
+    val languagePreference = findPreference<DropDownPreference>(Keys.LANGUAGE)
+    val selectedLanguage =
+      preferenceManager.sharedPreferences?.getString(Keys.LANGUAGE, DEFAULT_LANGUAGE)
+        ?: Locale.getDefault().language
```
shobhitagarwal1612: See the previous comment.

app/src/main/res/values/arrays.xml
```diff
@@ -0,0 +1,32 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
+  <string-array name="language_entries">
+    <item>English</item>
+    <item>French</item>
+    <item>Spanish</item>
+    <item>Portuguese</item>
```
shobhitagarwal1612: Shouldn't these also be using translated strings?

app/src/main/java/org/groundplatform/android/Config.kt
```diff
@@ -55,6 +55,12 @@ object Config {
   /** Limit on the permitted character length for free text question responses. */
   const val TEXT_DATA_CHAR_LIMIT = 255
 
+  /**
+   * ISO 639-1 code of language used when the system language is not supported, and the user has not
+   * yet manually selected a supported language.
+   */
+  const val DEFAULT_LANGUAGE = "en"
```
shobhitagarwal1612: Instead of hardcoding it to `en`, can we use `Locale.getDefault().language` here?

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/util/PrefUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * This function checks the user's language preference stored in shared preferences. If no
+ * preference is set, it falls back to the device's default language.
+ *
+ * @param context The context used to access shared preferences.
+ * @return The selected language code (e.g., "en", "fr").
+ */
+fun getSelectedLanguage(context: Context): String {
+  val prefs = PreferenceManager.getDefaultSharedPreferences(context)
+  val defaultLanguage = Locale.getDefault().language
```
shobhitagarwal1612: Default locale is already provided by `GroundApplicationModule`. Can we inject it into this class?

app/src/main/java/org/groundplatform/android/util/PrefUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+  return prefs.getString(Keys.LANGUAGE, defaultLanguage) ?: defaultLanguage
```
shobhitagarwal1612: Let's use `LocalValueStore` for fetching the pref value and use `''` as default.

shobhitagarwal1612 (changes requested):

shobhitagarwal1612 (approved):

PR #3119, opened 2025-04-16
Run code quality checks before unit tests in CI

shobhitagarwal1612:
Running unit tests takes ~5 mins but code quality checks take ~2 mins. Running code quality checks first would allow for faster failures of builds

@anandwana001  PTAL?


trietbui85:
@shobhitagarwal1612  Why don't we run both test & quality check parallel? It would be faster.

anandwana001 (approved):

PR #3124, opened 2025-04-24
Open surveys via QR Code or URL

### LINKED ISSUE #3046: Open surveys via QR Code or URL
- [ ] Opening a link from a QR code should open the app and activate the survey with the ID encoded in the related intent. Depends on https://github.com/google/ground-platform/issues/2121
- [ ] This may require the user to complete the sign in flow (sign in and ToS) if they are not already signed in - in other words we'll need to record the open survey intent and evaluate it after the sign in flow is completed.
- [ ] "Unlisted" surveys which are available offline should be included survey list
- [ ] An error should be shown if there is an issue loading the survey (permissions, doesn't exists, etc), and the user directed to the survey list.

@lecrabe

anandwana001:
Fixes #3046

@... PTAL?

https://github.com/user-attachments/assets/2030a90c-76c9-46fa-b433-4ebffe780825

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -79,7 +79,7 @@
         <category android:name="android.intent.category.BROWSABLE"/>
         <data
           android:scheme="https"
-          android:host="groundforis.com"
+          android:host="groundplatform.org"
           android:pathPrefix="/survey/" />
       </intent-filter>
     </activity>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -72,6 +72,16 @@
 
         <category android:name="android.intent.category.LAUNCHER" />
       </intent-filter>
+
+      <intent-filter android:autoVerify="true">
+        <action android:name="android.intent.action.VIEW"/>
+        <category android:name="android.intent.category.DEFAULT"/>
+        <category android:name="android.intent.category.BROWSABLE"/>
+        <data
+          android:scheme="https"
+          android:host="groundforis.com"
+          android:pathPrefix="/survey/" />
+      </intent-filter>
     </activity>
     <activity
       android:name="org.groundplatform.android.SettingsActivity"
--- a/app/src/main/java/org/groundplatform/android/Config.kt
+++ b/app/src/main/java/org/groundplatform/android/Config.kt
@@ -44,6 +44,8 @@ object Config {
    */
   const val CLUSTERING_ZOOM_THRESHOLD = 14f
 
+  const val SURVEY_PATH_SEGMENT = "survey"
+
   // TODO: Make sub-paths configurable and
   //  stop hardcoding here.
   // Issue URL: https://github.com/google/ground-android/issues/1730
--- a/app/src/main/java/org/groundplatform/android/Config.kt
+++ b/app/src/main/java/org/groundplatform/android/Config.kt
@@ -44,6 +44,12 @@ object Config {
    */
   const val CLUSTERING_ZOOM_THRESHOLD = 14f
 
+  /**
+   * The path segment used in deep‑link URIs to identify the survey screen.
+   *
+   * When handling incoming deep links, compare the first segment of the URI’s path to this constant
+   * to determine whether to navigate to the survey flow.
+   */
   const val SURVEY_PATH_SEGMENT = "survey"
 
   // TODO: Make sub-paths configurable and
--- a/app/src/main/java/org/groundplatform/android/Config.kt
+++ b/app/src/main/java/org/groundplatform/android/Config.kt
@@ -25,7 +25,7 @@ object Config {
   const val SHARED_PREFS_MODE = Context.MODE_PRIVATE
 
   // Local db settings.
-  const val DB_VERSION = 121
+  const val DB_VERSION = 122
   const val DB_NAME = "ground.db"
 
   // Firebase Cloud Firestore settings.
--- a/app/src/main/java/org/groundplatform/android/MainActivity.kt
+++ b/app/src/main/java/org/groundplatform/android/MainActivity.kt
@@ -62,6 +62,8 @@ class MainActivity : AbstractActivity() {
 
   private var signInProgressDialog: AlertDialog? = null
 
+  private var pendingDeepLink: Uri? = null
+
   override fun onCreate(savedInstanceState: Bundle?) {
     // Make sure this is before calling super.onCreate()
     setTheme(R.style.AppTheme)
@@ -83,13 +85,17 @@ class MainActivity : AbstractActivity() {
     navHostFragment =
       supportFragmentManager.findFragmentById(R.id.nav_host_fragment) as NavHostFragment
 
-    intent.data?.let {
-      navHostFragment.navController.handleDeepLinkIfNeeded(it)
-    }
-
     viewModel = viewModelFactory[this, MainViewModel::class.java]
 
     lifecycleScope.launch { viewModel.navigationRequests.filterNotNull().collect { updateUi(it) } }
+
+    intent.data?.let {
+      if (navHostFragment.navController.currentDestination?.id != R.id.sign_in_fragment) {
+        navHostFragment.navController.handleDeepLinkIfNeeded(it)
+      } else {
+        pendingDeepLink = it
+      }
+    }
 
     onBackPressedDispatcher.addCallback(
       this,
@@ -189,11 +195,6 @@ class MainActivity : AbstractActivity() {
 
   /** Override up button behavior to use Navigation Components back stack. */
   override fun onSupportNavigateUp(): Boolean = navigateUp()
-
-  override fun onNewIntent(intent: Intent) {
-    super.onNewIntent(intent)
-    navHostFragment.navController.handleDeepLink(intent)
-  }
 
   private fun navigateUp(): Boolean {
     val navController = navHostFragment.navController
--- a/app/src/main/java/org/groundplatform/android/MainActivity.kt
+++ b/app/src/main/java/org/groundplatform/android/MainActivity.kt
@@ -128,7 +128,7 @@ class MainActivity : AbstractActivity() {
         navigateTo(SignInFragmentDirections.showTermsOfService(false))
       }
       MainUiState.NoActiveSurvey -> {
-        navigateTo(SurveySelectorFragmentDirections.showSurveySelectorScreen(true, null))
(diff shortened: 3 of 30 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -72,6 +72,16 @@
...
         <category android:name="android.intent.category.LAUNCHER" />
       </intent-filter>
+
+      <intent-filter android:autoVerify="true">
+        <action android:name="android.intent.action.VIEW"/>
+        <category android:name="android.intent.category.DEFAULT"/>
+        <category android:name="android.intent.category.BROWSABLE"/>
+        <data
+          android:scheme="https"
+          android:host="groundforis.com"
```
gino-m, line 82: You can use `groundplatform.org` as the host for the main codebase. @rfontanarosa will override to `ground.openforis.org` in the Open Foris fork.

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -75,6 +76,8 @@ class MainActivity : AbstractActivity() {
       }
     }
 
+    handleDeepLinkIntent(intent)
```
gino-m, line 79: The way this is written implies it will always handle the deep link intent. To make clear what's happening, let's check here: `if (intent.data) { handleDeepLinkIntent(intent.data) }`

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/ui/tos/TermsOfServiceFragment.kt
```diff
@@ -128,7 +128,7 @@ class TermsOfServiceFragment : AbstractFragment() {
     lifecycleScope.launch {
       viewModel.navigateToSurveySelector.collect {
         findNavController()
-          .navigate(SurveySelectorFragmentDirections.showSurveySelectorScreen(true))
+          .navigate(SurveySelectorFragmentDirections.showSurveySelectorScreen(true, null))
```
gino-m, line 131: Does this work without `, null`?

app/src/main/java/org/groundplatform/android/ui/home/HomeScreenFragment.kt
```diff
@@ -93,7 +93,7 @@ class HomeScreenFragment :
     navHeader.findViewById<TextView>(R.id.switch_survey_button).setOnClickListener {
       findNavController()
         .navigate(
-          HomeScreenFragmentDirections.actionHomeScreenFragmentToSurveySelectorFragment(false)
+          HomeScreenFragmentDirections.actionHomeScreenFragmentToSurveySelectorFragment(false, null)
```
gino-m, line 96: Same Q here

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -238,13 +246,10 @@ class MainActivity : AbstractActivity() {
     }
   }
 
-  private fun handleDeepLinkIntent(intent: Intent) {
-    val viewModel = viewModelFactory[this, SurveySelectorViewModel::class.java]
-    intent.data?.let { uri ->
+  private fun NavController.handleDeepLinkIfNeeded(uri: Uri) {
+    if (uri.host == "groundplatform.org" && uri.pathSegments.firstOrNull() == "survey") {
```
gino-m, line 250: We can't hard-code the host here since it will be different for different deployments. Is checking the host necessary at all? Iiuc the app will only receive deep links for the host defined in `AndroidManifest.xml`, no?

Ok to check the path segment, in case we support other paths in the future, but please move to a string const `SURVEY_PATH_SEGMENT`.

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/MainActivity.kt
anandwana001: Fixed

shobhitagarwal1612:
Can we attach a screen recording of the feature?

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -83,7 +89,19 @@ class MainActivity : AbstractActivity() {
 
     viewModel = viewModelFactory[this, MainViewModel::class.java]
 
-    lifecycleScope.launch { viewModel.navigationRequests.filterNotNull().collect { updateUi(it) } }
+    lifecycleScope.launch {
+      viewModel.navigationRequests.filterNotNull().first()
```
shobhitagarwal1612, line 93: What's the purpose of this line?

app/src/main/java/org/groundplatform/android/Config.kt
```diff
@@ -44,6 +44,8 @@ object Config {
    */
   const val CLUSTERING_ZOOM_THRESHOLD = 14f
 
+  const val SURVEY_PATH_SEGMENT = "survey"
```
shobhitagarwal1612, line 47: Let's add some comments.

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -234,4 +252,11 @@ class MainActivity : AbstractActivity() {
       navHostFragment.navController.navigate(directions)
     }
   }
+
+  private fun NavController.handleDeepLinkIfNeeded(uri: Uri) {
+    if (uri.pathSegments.firstOrNull() == SURVEY_PATH_SEGMENT) {
+      val surveyId = uri.lastPathSegment
+      navigate(SurveySelectorFragmentDirections.showSurveySelectorScreen(false, surveyId))
```
shobhitagarwal1612, line 259: As discussed previously, why are we using `survey selector fragment` for this flow? Is this a UX requirement?

app/src/main/java/org/groundplatform/android/ui/tos/TermsOfServiceFragment.kt
anandwana001: removed

app/src/main/java/org/groundplatform/android/ui/home/HomeScreenFragment.kt
anandwana001: removed

app/src/main/java/org/groundplatform/android/MainActivity.kt
anandwana001: Discussed in webdnesday meeting, showed the flow, we can do this for now.

app/src/main/java/org/groundplatform/android/Config.kt
anandwana001: added

app/src/main/java/org/groundplatform/android/MainActivity.kt
anandwana001: to make sure we have the data before we act on the intent data, to understand the current navigation request ongoing.

shobhitagarwal1612: Won't this block the UI when opening the app without a deeplink?

anandwana001: Yes, we did test, with roberto also, this is not blocking the UI.

app/src/main/java/org/groundplatform/android/ui/tos/TermsOfServiceFragment.kt
```diff
@@ -139,9 +140,19 @@ class TermsOfServiceFragment : AbstractFragment() {
...
     lifecycleScope.launch {
       viewModel.navigateToSurveySelector.collect {
-        findNavController()
-          .navigate(SurveySelectorFragmentDirections.showSurveySelectorScreen(true))
+        activity?.intent?.data?.let { uri ->
+          if (uri.pathSegments.firstOrNull() == SURVEY_PATH_SEGMENT) {
+            val surveyId = uri.lastPathSegment
+            openSurveySelector(surveyId)
+          }
+        } ?: run { openSurveySelector(null) }
```
shobhitagarwal1612, line 148: We don't need to provide null here since it's the default value.

app/src/main/java/org/groundplatform/android/usecases/survey/ListAvailableSurveysUseCase.kt
```diff
@@ -57,8 +57,11 @@ constructor(
   private suspend fun getRemoteSurveyList(): Flow<List<SurveyListItem>> {
     val user = userRepository.getAuthenticatedUser()
     val remoteSurveyListFlow = remoteDataStore.getSurveyList(user)
-    return remoteSurveyListFlow.combine(getLocalSurveyList()) { remoteSurveys, localSurveys ->
-      remoteSurveys.map { remoteSurvey -> addOfflineStatus(remoteSurvey, localSurveys) }
+    val localFlow = getLocalSurveyList()
+    return combine(remoteSurveyListFlow, localFlow) { remoteSurveys, localSurveys ->
+      val enrichedRemote = remoteSurveys.map { remote -> addOfflineStatus(remote, localSurveys) }
```
shobhitagarwal1612, line 62: Can we improve the name of the variable? Maybe something more explicit like `remoteSurveysWithOfflineStatus`

app/src/main/java/org/groundplatform/android/ui/tos/TermsOfServiceFragment.kt
anandwana001: fixed

app/src/main/java/org/groundplatform/android/usecases/survey/ListAvailableSurveysUseCase.kt
anandwana001: fixed

app/src/main/java/org/groundplatform/android/usecases/survey/ListAvailableSurveysUseCase.kt
```diff
@@ -57,8 +57,11 @@ constructor(
...
+      val onlyLocal = localSurveys.filter { local -> remoteSurveys.none { it.id == local.id } }
```
shobhitagarwal1612, line 64: `onlyLocal` --> `localOnlySurveys`

anandwana001: fixed

shobhitagarwal1612 (approved):

gino-m (approved):

PR #3143, opened 2025-05-22
Show Sites based on visibility 

### LINKED ISSUE #3068: [Map] Show sites added by data collectors if allowed by organizer
Depends on https://github.com/google/ground-platform/issues/1418

anandwana001:
Fixes #3068 

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/persistence/remote/firebase/schema/LoiCollectionReference.kt
```diff
@@ -59,6 +59,13 @@ class LoiCollectionReference internal constructor(ref: CollectionReference) :
         .whereEqualTo(OWNER_FIELD, ownerUserId),
     )
 
+  /** Retrieves LOIs created by the anyone in the specified survey. Main-safe. */
```
gino-m, line 62: Why "created by.."? Also, what are "public" LOIs? we only have to types of LOIs, predefined and ad hoc. If a survey allows data collectors to view each others LOI's, we just need to query LOIs with the `predefined !== false`  filter.

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/persistence/remote/firebase/schema/LoiCollectionReference.kt
anandwana001: Fixed

app/src/main/java/org/groundplatform/android/persistence/remote/RemoteDataStore.kt
```diff
@@ -55,6 +55,12 @@ interface RemoteDataStore {
   /** Returns LOIs owned by the specified user in the specified survey. Main-safe. */
   suspend fun loadUserLois(survey: Survey, ownerUserId: String): List<LocationOfInterest>
 
+  /**
+   * Returns LOIs that have been marked as publicly visible for all participants of the specified
+   * survey.
+   */
```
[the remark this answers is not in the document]
anandwana001: Screenshot 2025-05-26 at 11 21 08 AM

using public based on the web flow

app/src/main/java/org/groundplatform/android/persistence/remote/firebase/schema/SurveyConverter.kt
```diff
@@ -68,5 +71,6 @@ internal object SurveyConverter {
       jobMap = jobMap.toPersistentMap(),
       acl = surveyProto.aclMap.entries.associate { it.key to it.value.toString() },
       dataSharingTerms = dataSharingTerms,
+      dataVisibility = dataVisibility,
```
shobhitagarwal1612, line 74: Can we mark the model field as non-null?

anandwana001: As we are not saving this field in local db, hence it is null.

shobhitagarwal1612: If we are not saving it, then how will we get the value in offline state?

anandwana001: what happening here is:

1. User is online - able to see all the sites on map
2. user go offline
3. user still able to see all the sites as all the lois and etc are getting saved, but saving the visibility tag is not required in our case, hence we are not saving

shobhitagarwal1612: Maybe for offline, the code is using the default value which uses `show` behavior?

anandwana001: I think not, it is based on the save saved loi we have in the db

shobhitagarwal1612: Thanks for the extra context. I'd like to spend some more time reviewing it to ensure there is no risk.

anandwana001: Sure

app/src/main/java/org/groundplatform/android/model/SurveyListItem.kt
```diff
@@ -16,12 +16,15 @@
...
 package org.groundplatform.android.model
 
+import org.groundplatform.android.proto.Survey.DataVisibility
+
 data class SurveyListItem(
   val id: String,
   val title: String,
   val description: String,
   val availableOffline: Boolean,
+  val dataVisibility: DataVisibility?,
```
shobhitagarwal1612: This variable is not being used anywhere. Let's clean it up.

app/src/main/java/org/groundplatform/android/model/Survey.kt
```diff
@@ -26,6 +26,7 @@ data class Survey(
   val jobMap: Map<String, Job>,
   val acl: Map<String, String> = mapOf(),
   val dataSharingTerms: Survey.DataSharingTerms? = null,
+  val dataVisibility: Survey.DataVisibility? = null,
```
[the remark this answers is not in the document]
shobhitagarwal1612: I gave it another thought and it seems that this variable is only set when fetching the survey object from remote. However, given that the `Survey` object is used in multiple places, it might be worth either saving it to local db for consistency or creating another object which holds remote-only object for 'Survey'. 

I'm leaning towards storing the value to local db. @gino-m thoughts?

gino-m:
> I gave it another thought and it seems that this variable is only set when fetching the survey object from remote. However, given that the Survey object is used in multiple places, it might be worth either saving it to local db for consistency or creating another object which holds remote-only object for 'Survey'.

Under the current design, the local db is the source of truth for the ViewModel layer, so all`Survey` metadata needs to be persisted there. We already have local- and remote- only representations of the survey - `SurveyEntity` for local db, and `SurveyDocument` for remote. `Survey` is domain object which is agnostic of how/where it's persisted.

app/src/main/java/org/groundplatform/android/model/Survey.kt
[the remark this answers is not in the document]
gino-m: I agree. The local Survey model would be incomplete without it.

app/src/main/java/org/groundplatform/android/persistence/remote/firebase/schema/SurveyConverter.kt
gino-m: Let's store this field in the local db.

app/src/main/java/org/groundplatform/android/model/SurveyListItem.kt
gino-m: +1. We can remove it here.

shobhitagarwal1612 (dismissed):

app/src/test/java/org/groundplatform/android/usecases/survey/SyncSurveyUseCaseTest.kt
```diff
@@ -65,4 +70,50 @@ class SyncSurveyUseCaseTest : BaseHiltTest() {
 
     assertThrows(IllegalStateException::class.java) { runBlocking { syncSurvey(SURVEY.id) } }
   }
+
```
shobhitagarwal1612, line 73: We should move these tests out of `SyncSurveyUseCaseTest`. Since `LocationOfInterestRepository` contains the core logic we need to test and is mocked here, these tests are misaligned. Let's relocate them to the correct class and verify that the proper LOIs are saved to the database.

anandwana001: done

PR #3150, opened 2025-05-29
Added version check through Remote Congi

### LINKED ISSUE #2769: Force version update when new features added on backend/web app
Suggested by @cdanielw - we may need to force data collectors to upgrade their app version when a survey is written with an incompatible newer version of the app, for ex when adding a new required task type as in #2469.

@kenstershiro @rawbzz for discussion.

anandwana001:
Fixes #2769 

@... PTAL?

Screenshot 2025-05-29 at 3 43 19 PM

### DIFF
--- a/app/src/main/java/org/groundplatform/android/GroundApplication.kt
+++ b/app/src/main/java/org/groundplatform/android/GroundApplication.kt
@@ -21,6 +21,7 @@ import android.util.Log
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.multidex.MultiDexApplication
 import androidx.work.Configuration
+import com.google.firebase.remoteconfig.FirebaseRemoteConfig
 import dagger.hilt.android.HiltAndroidApp
 import javax.inject.Inject
 import org.groundplatform.android.Config.isReleaseBuild
@@ -31,6 +32,7 @@ class GroundApplication : MultiDexApplication(), Configuration.Provider {
 
   @Inject lateinit var crashReportingTree: CrashReportingTree
   @Inject lateinit var workerFactory: HiltWorkerFactory
+  @Inject lateinit var remoteConfig: FirebaseRemoteConfig
 
   override val workManagerConfiguration: Configuration
     get() = Configuration.Builder().setWorkerFactory(workerFactory).build()
@@ -44,6 +46,8 @@ class GroundApplication : MultiDexApplication(), Configuration.Provider {
       // Log failures when trying to do work in the UI thread.
       setStrictMode()
     }
+    remoteConfig.setDefaultsAsync(R.xml.firebase_remote_config_defaults)
+    remoteConfig.fetchAndActivate()
   }
 
   private fun setStrictMode() {
--- a/app/src/main/java/org/groundplatform/android/MainActivity.kt
+++ b/app/src/main/java/org/groundplatform/android/MainActivity.kt
@@ -16,6 +16,7 @@
 package org.groundplatform.android
 
 import android.app.AlertDialog
+import android.content.ActivityNotFoundException
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
@@ -31,8 +32,10 @@ import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavController
 import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
+import com.google.firebase.remoteconfig.FirebaseRemoteConfig
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
+import kotlin.compareTo
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
@@ -58,6 +61,7 @@ class MainActivity : AbstractActivity() {
   @Inject lateinit var activityStreams: ActivityStreams
   @Inject lateinit var viewModelFactory: ViewModelFactory
   @Inject lateinit var userRepository: UserRepository
+  @Inject lateinit var remoteConfig: FirebaseRemoteConfig
 
   private lateinit var viewModel: MainViewModel
   private lateinit var navHostFragment: NavHostFragment
@@ -197,6 +201,7 @@ class MainActivity : AbstractActivity() {
   override fun onResume() {
     super.onResume()
     viewModel.checkAuthStatus()
+    checkForUpdate()
   }
 
   /** Override up button behavior to use Navigation Components back stack. */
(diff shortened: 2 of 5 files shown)


app/src/main/res/xml/firebase_remote_config_defaults.xml
```diff
@@ -0,0 +1,28 @@
...
+  -->
+
+<defaultsMap>
+    <entry>
+        <key>force_update</key>
+        <value>false</value>
+    </entry>
+    <entry>
+        <key>latest_version_code</key>
+        <value>14914</value>
```
shobhitagarwal1612, line 26: Is this the current version at head? Do we need defaults at all or can it be nullable?

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -58,6 +61,7 @@ class MainActivity : AbstractActivity() {
   @Inject lateinit var activityStreams: ActivityStreams
   @Inject lateinit var viewModelFactory: ViewModelFactory
   @Inject lateinit var userRepository: UserRepository
+  @Inject lateinit var remoteConfig: FirebaseRemoteConfig
```
shobhitagarwal1612, line 64: Would it create problems with unit testing since firebase can't be initialized for unit tests? Have you tried running the tests locally?

shobhitagarwal1612 (changes requested):

app/src/main/res/xml/firebase_remote_config_defaults.xml
anandwana001: We need default because if the value gets deleted from firebase or any accident happen, there should be a default value in codebase to handle the case.

shobhitagarwal1612: If the value gets deleted from firebase, shouldn't the default values be null?

anandwana001: 1. firebase remote confg doesn't support null values
2. We can keep the value 0, so that even if the value gets deleted, it will work well and no crashes.

app/src/main/java/org/groundplatform/android/MainActivity.kt
```diff
@@ -261,4 +263,38 @@ class MainActivity : AbstractActivity() {
       navigate(action)
     }
   }
+
+  private fun checkForUpdate() {
+    val forceUpdate = viewModel.remoteConfig.getBoolean("force_update")
+    val latestVersion = viewModel.remoteConfig.getLong("latest_version_code").toInt()
+    val currentVersion = BuildConfig.VERSION_CODE
+
+    if (forceUpdate && currentVersion < latestVersion) {
```
shobhitagarwal1612: Can we move this to the view model?

shobhitagarwal1612 (changes requested):

anandwana001:
one test is failing when running - `./gradlew testLocalDebugUnitTest`

```kt
signInProgressDialog_whenSignedOutAfterSignInState_isNotDisplayed
```

app/src/main/java/org/groundplatform/android/GroundApplication.kt
```diff
@@ -44,6 +46,8 @@ class GroundApplication : MultiDexApplication(), Configuration.Provider {
       // Log failures when trying to do work in the UI thread.
       setStrictMode()
     }
+    remoteConfig.setDefaultsAsync(R.xml.firebase_remote_config_defaults)
```
shobhitagarwal1612: Please add some comments before these or encapsulate in a method for clarity

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -55,6 +56,7 @@ constructor(
   private val reactivateLastSurvey: ReactivateLastSurveyUseCase,
   @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
   private val authenticationManager: AuthenticationManager,
+  val remoteConfig: FirebaseRemoteConfig,
```
shobhitagarwal1612: Can this be declared as private?

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -138,4 +140,10 @@ constructor(
 
   /** Returns true if the user has already accepted the Terms of Service. */
   private fun isTosAccepted(): Boolean = termsOfServiceRepository.isTermsOfServiceAccepted
+
+  fun shouldForceUpdate(currentVersion: Int = BuildConfig.VERSION_CODE): Boolean {
```
shobhitagarwal1612: Let's rename this to `isAppUpdateAvailable` for clarity

shobhitagarwal1612:
The UI for this dialog looks a bit different from the rest of the app. Can we make it consistent?

anandwana001:
> The UI for this dialog looks a bit different from the rest of the app. Can we make it consistent?

As discussed earlier in meeting, we will be doing the UI UX stuff as soon as we get the design. For this, the font, styles colors are matching

anandwana001:
yes, it is only failing here, not sure why, looking into this

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -55,6 +56,7 @@ constructor(
   private val termsOfServiceRepository: TermsOfServiceRepository,
   private val reactivateLastSurvey: ReactivateLastSurveyUseCase,
   @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+  private val remoteConfig: FirebaseRemoteConfig,
```
gino-m, line 59: How does reading the config from a file in the same build allow us to remotely force a version update? Wouldn't the value need to live in Firebase?

Also, I don't see any way to compare version numbers in the format `x.y.z`. Simple string comparison won't work.

gino-m:
@anandwana001  I would add the min version number to a new doc, `/config/app` with `{minAppVersion: '0.2.3'}`. Only force the version update if the doc is present and if the app version is less than the major.minor.revision number.

app/src/main/java/org/groundplatform/android/MainViewModel.kt
anandwana001: > How does reading the config from a file in the same build allow us to remotely force a version update? Wouldn't the value need to live in Firebase?



the config is there in firebase console as well as the config xml file in codebase for fallbacks

https://firebase.google.com/docs/remote-config

anandwana001: Also, I don't see any way to compare version numbers in the format x.y.z. Simple string comparison won't work.

We are comparing - VERSION_CODE - https://github.com/openforis/ground-android/blob/1daeb5d0ffd921c9c39c749319c2d1cd1da9230d/app/build.gradle#L79

x.y.z is the version name which is itself a string

gino-m: I see. Can we compare version name instead to make it easier to know what value to set in the db config?

gino-m: > > How does reading the config from a file in the same build allow us to remotely force a version update? Wouldn't the value need to live in Firebase?
> 
> the config is there in firebase console as well as the config xml file in codebase for fallbacks
> 
> https://firebase.google.com/docs/remote-config

That's pretty cool, I'll look into it a bit more, it could be useful for other purposes as well.

anandwana001: Done and fixed the test as well.

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -160,9 +160,28 @@ constructor(
   /** Returns true if the user has already accepted the Terms of Service. */
   private fun isTosAccepted(): Boolean = termsOfServiceRepository.isTermsOfServiceAccepted
 
-  fun isAppUpdateAvailable(currentVersion: Int = BuildConfig.VERSION_CODE): Boolean {
+  private fun isOlderVersion(current: String, minRequired: String): Boolean {
```
shobhitagarwal1612, line 163: Can we add some tests for this method?

shobhitagarwal1612: Maybe in a follow up PR?

app/src/main/res/xml/firebase_remote_config_defaults.xml
```diff
@@ -0,0 +1,28 @@
...
+  ~ limitations under the License.
+  -->
+
+<defaultsMap>
+    <entry>
+        <key>force_update</key>
+        <value>false</value>
+    </entry>
+    <entry>
+        <key>latest_version_code</key>
```
gino-m: This should be called `min_app_version`

app/src/main/java/org/groundplatform/android/MainViewModel.kt
anandwana001: yes, we can do in follow up PR

gino-m (approved):

PR #3154, opened 2025-06-05
Not including Incomplete draft surveys 

### LINKED ISSUE #2445: Incomplete draft surveys are visible in the app to the survey organizer

anandwana001:
Fixes #2445

Incomplete draft surveys are visible in the app to the survey organizer, we are preventing the survey which are in draft state to be in the app.

@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
Can you please add some description to the PR? It helps in understanding the problem we are trying to solve.

shobhitagarwal1612:
Is this change backward compatible? What happens with surveys where this bit is not available? Are we planning to backfill this value for all existing surveys?

anandwana001:
this is not the new parameter in proto, so it is backward compatible. 
all the surveys which are in draft will not be available in survey selector from now.

shobhitagarwal1612 (approved):

PR #3158, opened 2025-06-08
Auto-enable location lock when activating a survey without any LOIs or last saved camera position

### LINKED ISSUE #2130: When no previous viewport is saved and there are no LOIs

shobhitagarwal1612:
Fixes #2130

This PR auto-enables the location lock button iff:
 * There are no LOIs to move the camera position to
 * There is no last saved camera position

Verified locally using unit tests and testing on emulator

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
@@ -175,6 +175,7 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
     )
     binding.bottomContainer.bringToFront()
     showDataCollectionHint()
+    mapContainerViewModel.maybeEnableLocationLock()
   }
 
   /**
--- a/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
@@ -35,6 +35,7 @@ import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
 import org.groundplatform.android.Config.CLUSTERING_ZOOM_THRESHOLD
 import org.groundplatform.android.model.Survey
 import org.groundplatform.android.model.job.Job
@@ -57,8 +58,10 @@ import org.groundplatform.android.ui.home.mapcontainer.jobs.AdHocDataCollectionB
 import org.groundplatform.android.ui.home.mapcontainer.jobs.DataCollectionEntryPointData
 import org.groundplatform.android.ui.home.mapcontainer.jobs.SelectedLoiSheetData
 import org.groundplatform.android.ui.map.Feature
+import org.groundplatform.android.ui.map.FeatureType
 import org.groundplatform.android.ui.map.isLocationOfInterest
 import org.groundplatform.android.usecases.datasharingterms.GetDataSharingTermsUseCase
+import org.groundplatform.android.usecases.location.ShouldEnableLocationLockUseCase
 
 @OptIn(ExperimentalCoroutinesApi::class)
 @SharedViewModel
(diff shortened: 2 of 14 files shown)


shobhitagarwal1612:
@gino-m Gentle ping!

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -162,6 +166,15 @@ internal constructor(
       }
   }
 
+  fun maybeEnableLocationLock() {
```
gino-m: Please add ktdoc.

app/src/main/java/org/groundplatform/android/usecases/location/ShouldEnableLocationLockUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package org.groundplatform.android.usecases.location
+
+import javax.inject.Inject
+import org.groundplatform.android.repository.LocationOfInterestRepository
+import org.groundplatform.android.repository.MapStateRepository
+
+/** Use case to determine if location lock should be enabled for a survey. */
+class ShouldEnableLocationLockUseCase
```
gino-m: A "use case" is a user-visible piece of application behavior, not a function whose return value is used internally by the app. If we want to model this as a use case, it would likely be part of the `ActivateSurveyUseCase`. If that's not possible given the current architecture, then it should live in the ViewModel associated with the home screen setup on survey activation.

gino-m (changes requested):
Thanks for the fix!

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/usecases/location/ShouldEnableLocationLockUseCase.kt
shobhitagarwal1612: Thanks for pointing it out! Removed the use case class and it's test cases.

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -175,6 +175,8 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
     )
     binding.bottomContainer.bringToFront()
     showDataCollectionHint()
+
+    launchWhenStarted { mapContainerViewModel.maybeEnableLocationLock() }
```
gino-m, line 181: Does this always happen *after* LOIs have been downloaded? Iirc this screen is now only shown after they're synced? A comment could help here to clarify.

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
```diff
@@ -176,6 +176,8 @@ class HomeScreenMapContainerFragment : AbstractMapContainerFragment() {
     binding.bottomContainer.bringToFront()
     showDataCollectionHint()
 
+    // Always ensure that the LOIs are synced before enabling location lock. During first load of a
```
gino-m: Should the reader ensure they're synced, or is it saying the code ensures they're synced? I maybe would avoid that phrasing altogether, and just say that they're synced as a matter of fact.

gino-m (changes requested):
One nit, LGTM

app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerFragment.kt
shobhitagarwal1612: Done

gino-m (approved):

PR #3165, opened 2025-06-17
Add Point Button size fix

### LINKED ISSUE #3164: Add Point button displays incorrectly
Image

anandwana001:
Fixes #3164 

Screenshot 2025-06-17 at 8 05 31 PM

@rfontanarosa  @shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/components/TaskButton.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/components/TaskButton.kt
@@ -15,6 +15,7 @@
  */
 package org.groundplatform.android.ui.datacollection.components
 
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
Can we ask which device was used in https://github.com/google/ground-android/issues/3164? 

Let's use the same device for confirming that it works with the above changes. Also, please add a before/after image as well to help visualize any other side effects of this change. 

Thanks!

anandwana001:
in the above attached screenshot, i kept the smallest dp size through developer options so no matter which device, it will work fine now. @shobhitagarwal1612

shobhitagarwal1612 (approved):

PR #3167, opened 2025-06-18
[PhotoTaskFragment] Remove decodeBitmap approach

### LINKED ISSUE #3102: Photo-taking task crashes app
eMail from user:

_"Picture taking form field in campaign A makes the app crash every time we try to take a picture, so we are just skipping it for the time being. It works in survey B though, which makes me think that it is the picture taking form field, in combination with some other data field(s), that the app is not tolerating."_

I can confirm that collecting for another survey in my app (picture taking required & can't be skipped) does not work, instead a "Unsaved data restored" pop-up appears. Survey: https://ground.openforis.org/survey/N5hZgxQ8MnKDD8lhDwMs Build 0.1.17

Same does not work reproduced in a survey where the picture-taking can be skipped.
Skipping and getting to the last page ("Data collected successfully") the app fully crashes.

anandwana001:
Fixes #3102 

@shobhitagarwal1612 @rfontanarosa  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
+++ b/app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
@@ -17,7 +17,6 @@ package org.groundplatform.android.repository
 
 import android.content.Context
 import android.graphics.Bitmap
-import android.graphics.BitmapFactory
 import android.net.Uri
 import android.os.Environment
 import android.provider.MediaStore
@@ -81,42 +80,13 @@ constructor(
     withContext(Dispatchers.IO) {
       val file = createImageFile(fieldId)
 
-      val bitmap = decodeSampledBitmapFromUri(uri)
-      FileOutputStream(file).use { output ->
-        bitmap.compress(Bitmap.CompressFormat.JPEG, 85, output)
-      }
+      context.contentResolver.openInputStream(uri)?.use { input ->
+        FileOutputStream(file).use { output -> input.copyTo(output) }
+      } ?: throw IOException("Unable to open URI: $uri")
 
       Timber.d("Photo saved ${file.path} : ${file.exists()}")
       file
     }
-
-  private fun decodeSampledBitmapFromUri(uri: Uri): Bitmap =
-    context.contentResolver.openInputStream(uri)?.use { inputStream ->
-      val options = BitmapFactory.Options().apply { inJustDecodeBounds = true }
-      BitmapFactory.decodeStream(inputStream, null, options)
-
-      options.inSampleSize = calculateInSampleSize(options)
-
-      options.inJustDecodeBounds = false
-      context.contentResolver.openInputStream(uri)?.use { newInputStream ->
-        BitmapFactory.decodeStream(newInputStream, null, options)
-      }
-    } ?: throw IOException("Unable to decode bitmap from URI: $uri")
-
-  private fun calculateInSampleSize(options: BitmapFactory.Options): Int {
-    val (height: Int, width: Int) = options.run { outHeight to outWidth }
-    var inSampleSize = 1
-
-    if (height > 1024 || width > 1024) {
-      val halfHeight: Int = height / 2
-      val halfWidth: Int = width / 2
-
-      while (halfHeight / inSampleSize >= 1024 && halfWidth / inSampleSize >= 1024) {
-        inSampleSize *= 2
-      }
-    }
-    return inSampleSize
-  }
 
   @Throws(FileNotFoundException::class)
   fun addImageToGallery(filePath: String, title: String): String =
--- a/app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
+++ b/app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
@@ -17,6 +17,7 @@ package org.groundplatform.android.repository
 
 import android.content.Context
 import android.graphics.Bitmap
+import android.graphics.BitmapFactory
 import android.net.Uri
 import android.os.Environment
 import android.provider.MediaStore
@@ -27,6 +28,8 @@ import java.io.FileOutputStream
 import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import org.groundplatform.android.Config
 import org.groundplatform.android.persistence.remote.RemoteStorageManager
 import org.groundplatform.android.persistence.uuid.OfflineUuidGenerator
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -62,25 +62,30 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
...
   private var taskWaitingForPhoto: String? = null
   private var capturedPhotoPath: String? = null
   private lateinit var capturedPhotoUri: Uri
 
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    capturePhotoLauncher =
+      registerForActivityResult(ActivityResultContracts.TakePicture()) { result: Boolean ->
+        externalScope.launch(Dispatchers.IO) {
+          if (result) {
```
gino-m: `if (!result) return`

gino-m (approved):
Lgtm, one nit.

app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
```diff
@@ -74,16 +77,46 @@ constructor(
    * memory, preserving the original quality, metadata, and resolution. It avoids potential
    * OutOfMemoryError caused by loading large images into a Bitmap.
    */
-  @Throws(IOException::class)
-  suspend fun savePhotoFromUri(uri: Uri, fieldId: String): File {
-    val file = createImageFile(fieldId)
+  suspend fun savePhotoFromUri(uri: Uri, fieldId: String): File =
+    withContext(Dispatchers.IO) {
+      val file = createImageFile(fieldId)
```
gino-m, line 81: `UserMediaRepository` shouldn't know about the `fieldId` - in the future we should abstract this away from this class

app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
```diff
@@ -74,16 +77,46 @@ constructor(
...
 
-    context.contentResolver.openInputStream(uri)?.use { input ->
-      FileOutputStream(file).use { output -> input.copyTo(output) }
-    } ?: throw IOException("Unable to open URI: $uri")
+      val bitmap = decodeSampledBitmapFromUri(uri)
+      FileOutputStream(file).use { output ->
+        bitmap.compress(Bitmap.CompressFormat.JPEG, 85, output)
```
gino-m: Why do we need to recompress the image? Isn't it written as JPEG to the location specified in `MediaStore.EXTRA_OUTPUT` when the photo Intent is triggered?

app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
```diff
@@ -74,16 +77,46 @@ constructor(
...
+      }
 
-    Timber.d("Photo saved %s : %b", file.path, file.exists())
-    return file
+      Timber.d("Photo saved %s : %b", file.path, file.exists())
```
gino-m: You can use template strings in Kotlin for better readability.

app/src/main/java/org/groundplatform/android/repository/UserMediaRepository.kt
anandwana001: we are not using `MediaStore.EXTRA_OUTPUT` anywhere, unable to find it.
@shobhitagarwal1612 anything on this do you know?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -135,6 +132,23 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
     outState.putString(CAPTURED_PHOTO_PATH, capturedPhotoPath)
   }
 
+  private fun handleCaptureResult(result: Boolean) {
+    if (!result || viewModel.capturedUri == null) {
```
gino-m, line 137: Move null check into launch block

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -58,18 +65,21 @@ constructor(
...
-      val bitmap = bitmapUtil.fromUri(uri)
-      val file = userMediaRepository.savePhoto(bitmap, currentTask)
-      userMediaRepository.addImageToGallery(file.absolutePath, file.name)
-      val remoteFilename = FirebaseStorageManager.getRemoteMediaPath(surveyId, file.name)
-      setValue(PhotoTaskData(remoteFilename))
-    } catch (e: IOException) {
-      Timber.e(e, "Error getting photo selected from storage")
+        withContext(Dispatchers.Main) { setValue(PhotoTaskData(remoteFilename)) }
+      } catch (e: IOException) {
+        Timber.e(e, "Error getting photo selected from storage")
```
gino-m: We need to let user know if saving photo fails, perhaps with toast

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -58,18 +65,21 @@ constructor(
...
+      } catch (e: Exception) {
```
gino-m: Don't need to cancel all Exception imo

gino-m (changes requested):
Please fix PR title

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -135,6 +132,23 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
...
+      Timber.e("Photo capture failed or URI not set")
+      return
+    }
+
+    externalScope.launch(Dispatchers.IO) {
```
gino-m: Does this while thing need to happen on IO thread, or just `savePhotoTaskData`?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -58,18 +65,21 @@ constructor(
    * Saves photo data stored on an on-device URI in Ground-associated storage and prepares it for
    * inclusion in a data collection submission.
    */
-  suspend fun savePhotoTaskData(uri: Uri) {
+  fun savePhotoTaskData(uri: Uri) {
     val currentTask = taskWaitingForPhoto
     requireNotNull(currentTask) { "Photo captured but no task waiting for the result" }
+    externalScope.launch(Dispatchers.IO) {
```
gino-m: We spoke, @shobhitagarwal1612 and I agree it would be clearer if the caller switched to IO thread instead of this fun.

gino-m (dismissed):

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -61,25 +61,22 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
...
   private var taskWaitingForPhoto: String? = null
   private var capturedPhotoPath: String? = null
   private lateinit var capturedPhotoUri: Uri
 
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    capturePhotoLauncher =
+      registerForActivityResult(ActivityResultContracts.TakePicture()) { result: Boolean ->
+        externalScope.launch(Dispatchers.IO) { handleCaptureResult(result) }
```
shobhitagarwal1612: Can we inject `IoDispatcher` instead of using `Dispatchers.IO`? This would allow testing the feature using test dispatchers in unit tests.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -63,13 +63,13 @@ constructor(
...
 
     try {
-      val bitmap = bitmapUtil.fromUri(uri)
-      val file = userMediaRepository.savePhoto(bitmap, currentTask)
+      val file = userMediaRepository.savePhotoFromUri(uri, currentTask)
       userMediaRepository.addImageToGallery(file.absolutePath, file.name)
       val remoteFilename = FirebaseStorageManager.getRemoteMediaPath(surveyId, file.name)
-      setValue(PhotoTaskData(remoteFilename))
+
+      withContext(Dispatchers.Main) { setValue(PhotoTaskData(remoteFilename)) }
```
shobhitagarwal1612, line 70: As discussed, we can directly update the value from background thread using `postValue` without having to change the thread.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -97,21 +94,20 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
...
-    }
   }
 
   override fun onTaskViewAttached() {
     viewModel.surveyId = dataCollectionViewModel.surveyId
     viewModel.taskWaitingForPhoto = taskWaitingForPhoto
+
+    viewModel.capturedUri?.let { uri ->
+      if (!viewModel.hasLaunchedCamera) {
+        externalScope.launch {
```
shobhitagarwal1612: Should this be using ioDispatcher too?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
anandwana001: this setValue is not the LiveData, it is a part of AbstractTaskViewModel.

PR #3169, opened 2025-07-02
[DrawAreaTaskMapFragment] Hide add marker when point too close

### LINKED ISSUE #3168: [DrawArea] Disable Plus button while disabling Add Point button
Disable Plus button for draing area while disabling Add Point button

anandwana001:
Fixes #3168 

@shobhitagarwal1612  PTAL?

| Before  | After |
| --- | --- |
| Screenshot 2025-07-07 at 12 23 15 PM | Screenshot 2025-07-07 at 12 21 57 PM |

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskMapFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskMapFragment.kt
@@ -145,9 +145,8 @@ abstract class AbstractTaskMapFragment<TVM : AbstractTaskViewModel> :
       }
     }
 
-  fun updateCenterMarker(isTooClose: Boolean, isMarkedComplete: Boolean) {
-    val shouldShow = !isMarkedComplete && !isTooClose
-    binding.centerMarker.visibility = if (shouldShow) View.VISIBLE else View.GONE
+  fun setCenterMarkerVisibility(visible: Boolean) {
+    binding.centerMarker.visibility = if (visible) View.VISIBLE else View.GONE
   }
 
   @MustBeInvokedByOverriders
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskMapFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskMapFragment.kt
@@ -145,6 +145,11 @@ abstract class AbstractTaskMapFragment<TVM : AbstractTaskViewModel> :
       }
     }
 
+  fun updateCenterMarker(isTooClose: Boolean, isMarkedComplete: Boolean) {
+    val shouldShow = !isMarkedComplete && !isTooClose
+    binding.centerMarker.visibility = if (shouldShow) View.VISIBLE else View.GONE
+  }
+
   @MustBeInvokedByOverriders
   protected open fun onMapCameraMoved(position: CameraPosition) {
     if (getMapViewModel().locationLock.value.getOrDefault(false)) {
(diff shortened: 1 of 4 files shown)


shobhitagarwal1612:
Can you please add a before/after screen recording of the bug fix?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskMapFragment.kt
```diff
@@ -145,6 +145,11 @@ abstract class AbstractTaskMapFragment<TVM : AbstractTaskViewModel> :
       }
     }
 
+  fun updateCenterMarker(isTooClose: Boolean, isMarkedComplete: Boolean) {
+    val shouldShow = !isMarkedComplete && !isTooClose
+    binding.centerMarker.visibility = if (shouldShow) View.VISIBLE else View.GONE
+  }
```
shobhitagarwal1612: Can we change this method to:

```
fun setCenterMarkerVisibility(visible: Boolean) {
   binding.centerMarker.visibility = if (visible) View.VISIBLE else View.GONE
}
```

This would help simplify the structure of the class.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -50,4 +51,11 @@ class DrawAreaTaskMapFragment @Inject constructor() :
     taskViewModel.draftArea
       .map { feature: Feature? -> if (feature == null) setOf() else setOf(feature) }
       .asLiveData()
+
+  fun updateCenterMarker() {
+    super.updateCenterMarker(
+      isTooClose = taskViewModel.isTooClose,
+      isMarkedComplete = taskViewModel.isMarkedComplete(),
+    )
```
shobhitagarwal1612: Instead of calling super, this would be replaced with `setCenterMarkerVisibility`

shobhitagarwal1612 (changes requested):
Please update the title and description of the PR as well to specify why this change is necessary. This is useful when looking at the code history.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -32,6 +32,7 @@ class DrawAreaTaskMapFragment @Inject constructor() :
   override fun onMapCameraMoved(position: CameraPosition) {
     super.onMapCameraMoved(position)
     if (!taskViewModel.isMarkedComplete()) {
+      updateCenterMarker()
```
shobhitagarwal1612: I feel it is error prone to manually check if the marker should be updated. The better approach would be to trigger an event whenever any of the dependent values `isMarkedComplete()` or `isTooClose` is changed in the view model. This would be ensure that we don't accidentally update the marker ui before the values are mutated.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -94,6 +94,7 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
         viewModel.addLastVertex()
         viewModel.checkVertexIntersection()
         viewModel.triggerVibration()
+        drawAreaTaskMapFragment.updateCenterMarker()
```
shobhitagarwal1612: Same as previous comment

shobhitagarwal1612 (approved):
💯

PR #3173, opened 2025-07-07
[UserRepository] User can submit to Public as well as Unlisted survey

### LINKED ISSUE #3157: Error collecting data for a publicly listed survey from another account
**Describe the bug**
Unable to collect data for a public survey owned by someone else

**App version**
At head

**To Reproduce**
1. User A creates a new survey and makes it public
2. User B opens the app and activates it
3. See error in logcat

**Expected behavior**
User should be able to see the "+" button

**Actual behavior**
User doesn't see anything.

anandwana001:
Fixes #3157

We might need migration here as well
WDYT?

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/repository/UserRepository.kt
+++ b/app/src/main/java/org/groundplatform/android/repository/UserRepository.kt
@@ -83,6 +83,11 @@ constructor(
    * survey. If no survey is active at the moment, then it returns false.
    */
   suspend fun canUserSubmitData(): Boolean {
+    val isPublic = surveyRepository.activeSurvey?.generalAccess
+    if (isPublic?.ordinal == 3) {
+      return true
+    }
+
     val user = getAuthenticatedUser()
     return try {
       surveyRepository.activeSurvey?.getRole(user.email) != Role.VIEWER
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/groundplatform/android/repository/UserRepository.kt
```diff
@@ -83,6 +83,11 @@ constructor(
    * survey. If no survey is active at the moment, then it returns false.
    */
   suspend fun canUserSubmitData(): Boolean {
+    val isPublic = surveyRepository.activeSurvey?.generalAccess
+    if (isPublic?.ordinal == 3) {
```
gino-m: Can we use the enum here?

gino-m: We should also for unlisted surveys as well.

gino-m: surveyRepository.activeSurvey?.generalAccess == || ==..

anandwana001: done

shobhitagarwal1612 (changes requested):
Can we add some test coverage?

PR #3177, opened 2025-07-14
[MainActivity] Fix Deeplink for both cases, With/Without Activated Survey

### LINKED ISSUE #3176: [Open via link] QR/link doesn't work after removing survey
@rfontanarosa for repro details, @anandwana001 for fix.

anandwana001:
Fixes #3176 

@rfontanarosa @shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/MainViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/MainViewModel.kt
@@ -15,6 +15,7 @@
  */
 package org.groundplatform.android
 
+import android.net.Uri
 import androidx.core.view.WindowInsetsCompat
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -63,6 +69,10 @@ constructor(
   /** The window insets determined by the activity. */
   val windowInsets: MutableLiveData<WindowInsetsCompat> = MutableLiveData()
 
+  private val _deepLinkUri = MutableStateFlow<Uri?>(null)
+  val isDeepLinkAvailable =
```
shobhitagarwal1612: Is this unused?

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -78,6 +88,12 @@ constructor(
     }
   }
 
+  fun isDeepLinkAvailable(): Boolean = _deepLinkUri.value != null
```
shobhitagarwal1612: This can be declared as private or inlined.

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -126,6 +142,18 @@ constructor(
       userRepository.saveUserDetails(user)
       if (!isTosAccepted()) {
         MainUiState.TosNotAccepted
+      } else if (isDeepLinkAvailable()) {
+        val deepLinkUri = _deepLinkUri.value
+        val surveyId =
+          deepLinkUri
+            ?.takeIf { it.pathSegments.firstOrNull() == SURVEY_PATH_SEGMENT }
```
shobhitagarwal1612, line 144: Let's move the logic of parsing survey id from uri to a separate method for readability

app/src/main/java/org/groundplatform/android/MainViewModel.kt
```diff
@@ -126,6 +142,18 @@ constructor(
...
+            ?.lastPathSegment
+
+        if (!surveyId.isNullOrBlank()) {
+          MainUiState.ActiveSurveyById(surveyId)
+        } else {
+          MainUiState.NoActiveSurvey
```
shobhitagarwal1612, line 150: Should we also log an error with the uri?

anandwana001: not an error

PR #3178, opened 2025-07-14
Added loading while TermsOfService is loading

### LINKED ISSUE #3170: [Tos] "Agree" checkbox and blank screen shown while terms of service load
Instead we should show a spinner and no UI for the user to interact with.

anandwana001:
Fixes #3170

Screenshot 2025-07-14 at 7 57 41 AM

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/MainActivity.kt
+++ b/app/src/main/java/org/groundplatform/android/MainActivity.kt
@@ -28,7 +28,6 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.core.view.WindowInsetsCompat
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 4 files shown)


app/src/test/java/org/groundplatform/android/ui/tos/TermsOfServiceFragmentTest.kt
```diff
@@ -105,6 +105,7 @@ class TermsOfServiceFragmentTest : BaseHiltTest() {
 
   @Test
   fun agreeButton_default_shouldNotBeEnabled() {
+    whenever(networkManager.isNetworkConnected()).thenReturn(true)
```
shobhitagarwal1612: This line is common for almost all tests. Can we move it to `setup` instead?

PR #3179, opened 2025-07-15
Only show popup for read-only surveys, remove all zoom related states

### LINKED ISSUE #3171: "Zoom to start collecting data" still shown
This message should be removed, since we now always auto-zoom for the user (https://github.com/google/ground-android/issues/2130), and since you don't need to zoom in to collect data anymore (@shobhitagarwal1612). @scolsen Would you like to do the honors?

shobhitagarwal1612:
Fixes #3171

Removes info popups related to zoom states.

Verified the changes by testing the app locally. 

@gino-m  PTAL?


gino-m (approved):
👍

PR #3181, opened 2025-07-17
Fixing the Core issue for the Photo Crash

### LINKED ISSUE #3193: [Picasso] Replace Picasso with Android Image Loading

anandwana001:
Fixes #3193 

The core reason of the photo crash is due to onResume hit.

Each time a user capture the image, onResume callback hit in the MainActivity which restarts the navigation execution and due to collecting the only first emission, it never jumps to photo fragment and due to which it was jumping to home screen in the loop.

We have to make a new flow of checking auth status each time we resume the app.

https://github.com/user-attachments/assets/331c3906-f77d-4445-bcb3-20ff9c4ae5cb

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/schemas/org.groundplatform.android.persistence.local.room.LocalDatabase/122.json
+++ b/app/schemas/org.groundplatform.android.persistence.local.room.LocalDatabase/122.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 122,
-    "identityHash": "ec359a748f5139dc24205b7cbea2b4ed",
+    "identityHash": "5b8e3c49ff710bd4f80b1dae1be0fe09",
     "entities": [
       {
         "tableName": "draft_submission",
@@ -617,7 +617,7 @@
       },
       {
         "tableName": "survey",
-        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `title` TEXT, `description` TEXT, `acl` TEXT, `data_sharing_terms` BLOB, PRIMARY KEY(`id`))",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `title` TEXT, `description` TEXT, `acl` TEXT, `data_sharing_terms` BLOB, `general_access` INTEGER, PRIMARY KEY(`id`))",
         "fields": [
           {
             "fieldPath": "id",
@@ -647,6 +647,12 @@
             "fieldPath": "dataSharingTerms",
             "columnName": "data_sharing_terms",
             "affinity": "BLOB",
+            "notNull": false
+          },
+          {
+            "fieldPath": "generalAccess",
+            "columnName": "general_access",
+            "affinity": "INTEGER",
             "notNull": false
           }
         ],
@@ -1139,7 +1145,7 @@
     "views": [],
     "setupQueries": [
       "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
-      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, 'ec359a748f5139dc24205b7cbea2b4ed')"
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, '5b8e3c49ff710bd4f80b1dae1be0fe09')"
     ]
   }
 }
--- a/app/schemas/org.groundplatform.android.persistence.local.room.LocalDatabase/122.json
+++ b/app/schemas/org.groundplatform.android.persistence.local.room.LocalDatabase/122.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 122,
-    "identityHash": "5b8e3c49ff710bd4f80b1dae1be0fe09",
+    "identityHash": "ec359a748f5139dc24205b7cbea2b4ed",
     "entities": [
       {
         "tableName": "draft_submission",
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/groundplatform/android/persistence/local/room/LocalDatabase.kt
```diff
@@ -90,7 +90,12 @@ import org.groundplatform.android.persistence.local.room.fields.TileSetEntitySta
     ],
   version = Config.DB_VERSION,
   exportSchema = true,
-  autoMigrations = [AutoMigration(from = 120, to = 121), AutoMigration(from = 121, to = 122)],
+  autoMigrations =
```
shobhitagarwal1612: Should this be moved to a separate PR?

app/schemas/org.groundplatform.android.persistence.local.room.LocalDatabase/122.json
```diff
@@ -617,7 +617,7 @@
       },
       {
         "tableName": "survey",
-        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `title` TEXT, `description` TEXT, `acl` TEXT, `data_sharing_terms` BLOB, PRIMARY KEY(`id`))",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `title` TEXT, `description` TEXT, `acl` TEXT, `data_sharing_terms` BLOB, `general_access` INTEGER, PRIMARY KEY(`id`))",
```
shobhitagarwal1612: Why is this required?

anandwana001:
https://github.com/google/ground-android/issues/3182

gino-m:
@anandwana001 Please describe the issue more clearly and we can merge.

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
```diff
@@ -52,14 +55,25 @@ object BindingAdapters {
...
-        .load(uri)
-        .placeholder(R.drawable.ic_photo_grey_600_24dp)
-        .transform(RotateUsingExif(uri, view.context))
-        .into(view)
-    } catch (exception: IllegalStateException) {
-      // Needed for running unit tests
-      Timber.e(exception)
+      val context = view.context
+
+      val bitmap =
```
gino-m: @anandwana001 Is this necessary?

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
anandwana001: yes, for EXIF rotation. we need this.

gino-m: Then we should use https://developer.android.com/reference/android/media/ExifInterface to load the header without loading the whole file.

anandwana001: done

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
```diff
@@ -52,14 +52,11 @@ object BindingAdapters {
     if (view == null || uri == null) return
 
     try {
-      Picasso.get()
```
shobhitagarwal1612, line 55: Can we cleanup the gradle dependency too?

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
```diff
@@ -52,14 +52,11 @@ object BindingAdapters {
...
-        .load(uri)
-        .placeholder(R.drawable.ic_photo_grey_600_24dp)
```
shobhitagarwal1612, line 57: Is it possible to preserve the placeholder logic?

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
```diff
@@ -52,14 +52,11 @@ object BindingAdapters {
...
-        .transform(RotateUsingExif(uri, view.context))
-        .into(view)
-    } catch (exception: IllegalStateException) {
-      // Needed for running unit tests
-      Timber.e(exception)
+      val rotatedBitmap = loadBitmapWithCorrectOrientation(view.context, uri)
+      view.setImageBitmap(rotatedBitmap)
+    } catch (e: Exception) {
```
shobhitagarwal1612: It's generally not advisable to use a generic "catch-all" for exceptions. A more robust approach is to explicitly list and handle the specific exceptions you expect to occur. In cases where no particular exceptions are foreseen, it's often better to allow the program to fail gracefully rather than attempting to catch an unknown error.

app/src/main/java/org/groundplatform/android/ui/util/RotateUsingExif.kt
```diff
@@ -17,60 +17,51 @@ package org.groundplatform.android.ui.util
...
     }
-
-  /** Returns the EXIF orientation attribute of the JPEG image at the specified URI. */
-  private fun getOrientationFromExif(uri: Uri): Int {
-    val inputStream =
-      context.contentResolver.openInputStream(uri)
-        ?: throw IOException("Content resolver returned null for $uri")
-    val exif = ExifInterface(inputStream)
-    return exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
   }
```
shobhitagarwal1612: Can we preserve the existing methods `getRotationDegrees` and `getOrientationFromExif`? These lines haven't changed and would help reduce the diff count in the PR

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
anandwana001: other places are still using it

anandwana001: https://github.com/google/ground-android/issues/3193

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
anandwana001: not directly, but one approach is to set the placeholder meanwhile the bitmap is coming.

anandwana001: done

app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
shobhitagarwal1612: Are there any other usages of Picasso left? Can we do it in the same PR?

anandwana001: removed

shobhitagarwal1612 (approved):
LGTM! Thanks

Issue #3192, opened 2025-07-19
Create interfaces for repositories in `domain` layer and their implementation in `data` layer


shobhitagarwal1612:
- [ ] LocationOfInterestRepository
- [ ] SubmissionRepository
- [ ] SurveyRepository
- [ ] TermsOfServiceRepository
- [ ] UserRepository

Issue #3186, opened 2025-07-17
[Code health] Improve directory structure as per clean architechture


shobhitagarwal1612:
- [ ] Create a new `presentation` layer to contain all screen and ui related classes
- [ ] Move `usecases` to `domain` package
- [ ] Within each screen, use `components` subpackage for storing various composables

Final structure:
--------------
 - common ✅
 - data ✅
 -  - local ✅
 -  - remote ✅
 -  - repository
 -  di (for storing app level modules only)
 -  domain 
 - - model
 - - repository
 - - use_case
 - presentation
 - - home
 - - settings
 - - data collection 
 - - ....
 - - ui 
 - - - theme 
 - - - - Color 
 - - - - Theme 
 - - MainActivity
 - - MainViewModel
 - Ground Application ✅

--------------

View Models should delegate the business logic for user interactions (e.g. GetSurveys) to domain layer use_cases.

PR #3194, opened 2025-07-19
[Clean Architecture] Refactor 2: Create an interface for LocationOfInterestRepository

shobhitagarwal1612:
Towards #3192

This PR aims at creating an interface for the LocationOfInterestRepository as per clean architecture. This is a pure refactor and doesn't modify the class definition.

@gino-m  PTAL?

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -47,6 +47,7 @@
       <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
       <option name="IMPORT_LAYOUT_TABLE">
         <value>
+          <package name="" withSubpackages="true" static="false" module="true" />
           <package name="" withSubpackages="true" static="true" />
           <emptyLine />
           <package name="" withSubpackages="true" static="false" />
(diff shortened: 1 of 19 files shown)


gino-m (changes requested):
How does creating an interface for LocationOfInterestRepository help with clean architecture? LocationOfInterestRepository only has one implementation, and our mock libraries seem sufficient for testing.. unless I'm missing something?

shobhitagarwal1612:
The proposed final structure would look something like this: https://github.com/google/ground-android/issues/3186#issue-3240788248

I've found that leveraging interfaces for our domain layer's repositories significantly decouples our app's Clean Architecture. This approach provides easier testing by enabling data mocking, offers greater flexibility to swap data sources without altering core business logic, and ensures simpler maintenance as changes are isolated. Crucially, it also strictly adheres to Clean Architecture's dependency rule, making our application more robust and scalable.

gino-m:
> Clean Architecture's dependency rul

Since the repo's deps are injected, I don't think we actually break the dependency rule for repositories; the calling code doesn't need to know anything about the downstream implementation, so unidirectional dependencies are preserved.

I assume this helps testing because it makes it easier to implement fakes. Is the idea that application tests would use repository fakes instead of fake Firestore impls? That would provide lower fidelity - is that what we want?

shobhitagarwal1612:
You're right that dependency injection already handles decoupling for current repository classes, making interfaces seem redundant for that specific purpose. 

However, interfaces excel at providing a crucial layer of abstraction between high-level modules (like domain use cases) and low-level modules (such as repository implementations). They define a clean, easily understandable, and maintainable public API for your repository layer, regardless of underlying complexities.

Consider `SurveyRepository`, it's currently handling both loading/persisting surveys and managing the "active" survey id, the latter of which shouldn't be its responsibility. An interface would allow you to separate these concerns, clearly defining what the `SurveyRepository` should do, and exposing problematic design choices.

Issue #2353, opened 2024-03-14
[Survey list] List page blinks back and forth between sign out link and list on unstable connection


gino-m:
Survey list seems to be reactive to the network connection. Should check for network connection after page loads and show sign out link if no connection or no surveys.

anandwana001:
This will fix this issue - https://github.com/google/ground-android/issues/3047

gino-m:
@anandwana001 Thanks. Please be sure to add`Fixes #3047` to the relevant PR.

gino-m:
The page should not reload when going offline, only the section with non-offline surveys.

PR #3203, opened 2025-07-22
Add section listing in survey selector screen

### LINKED ISSUE #3047: [Survey list] Add section listing "public" surveys
A second query will need to be run to fetch surveys which are accessible to all users. 

Depends on https://github.com/google/ground-platform/issues/2121.

@lecrabe FYI

anandwana001:
Fixes #3047
Fixes https://github.com/google/ground-android/issues/2353

https://github.com/user-attachments/assets/aafa1c4a-cc90-4a98-9fba-e44d7355e6ca

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
shobhitagarwal1612, line 1: Can we rename this file to `SurveyListScreen`?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
```diff
@@ -0,0 +1,278 @@
...
+    R.string.section_shared_with_me to false,
+    R.string.section_public to false,
+  )
+}
+
+private fun Survey.GeneralAccess.iconRes(): Int =
+  when (ordinal) {
+    2 -> R.drawable.ic_unlisted
+    3 -> R.drawable.ic_public
+    else -> R.drawable.ic_restricted
```
shobhitagarwal1612: Please replace with the proto reference rather than using the field number value.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+private fun Survey.GeneralAccess.labelString(): Int =
+  when (ordinal) {
+    2 -> R.string.access_unlisted
+    3 -> R.string.access_public
+    else -> R.string.access_restricted
```
shobhitagarwal1612: Same as above

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
```diff
@@ -0,0 +1,278 @@
...
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
+          )
+        }
+    }
+  }
+}
+
+@Composable
+private fun HeaderRow(item: SurveyListItem, onPopUpClick: (String) -> Unit) {
```
shobhitagarwal1612: Can we improve the naming of the function `onPopUpClick`? It's not clear what popup is referring to here.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
```diff
@@ -0,0 +1,278 @@
...
+        modifier = Modifier.size(24.dp).clickable { onPopUpClick(item.id) }.padding(end = 4.dp),
+      )
+    }
+  }
+}
+
+private fun LazyListScope.SurveyItemExpandedList(
+  items: List<SurveyListItem>,
+  onActivate: (String) -> Unit,
+  onPopUpClick: (String) -> Unit,
```
shobhitagarwal1612: We don't mention `activate` keyword anywhere in the ui. It might be better to name the functions based on what triggers it. e.g. replace `onActivate` with `onCardClick`. Something similar needs to be done for other callback method.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListUI.kt
```diff
@@ -0,0 +1,278 @@
...
+      fontFamily = FontFamily(Font(R.font.text_500)),
+      lineHeight = 16.sp,
+      fontSize = 16.sp,
+      fontWeight = FontWeight(500),
+      color = MaterialTheme.colorScheme.onSurfaceVariant,
+    )
+  }
+}
+
+fun formatSectionTitle(title: String, count: Int): String = "$title ($count)"
```
shobhitagarwal1612: Can we add the previews for the composables for various states at the bottom of this class?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorFragment.kt
```diff
@@ -90,7 +83,7 @@ class SurveySelectorFragment : AbstractFragment(), BackPressListener {
       container.visibleIf(surveys.isNotEmpty())
       emptyContainer.visibleIf(surveys.isEmpty())
     }
-    adapter.updateData(surveys)
+    binding.composeView.setContent { AppTheme { SurveyList(surveys, viewModel) } }
```
shobhitagarwal1612, line 87: Can we add a TODO to move the above usages to composable as well?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -51,6 +53,20 @@ internal constructor(
...
+  private val _surveys = MutableStateFlow<List<SurveyListItem>>(emptyList())
+
+  private val _onDevice = MutableStateFlow<List<SurveyListItem>>(emptyList())
+  val onDevice: StateFlow<List<SurveyListItem>> = _onDevice
+
+  private val _sharedWith = MutableStateFlow<List<SurveyListItem>>(emptyList())
+  val sharedWith: StateFlow<List<SurveyListItem>> = _sharedWith
+
+  private val _publicList = MutableStateFlow<List<SurveyListItem>>(emptyList())
+  val publicList: StateFlow<List<SurveyListItem>> = _publicList
```
shobhitagarwal1612: Let's improve the naming of the variables to clearly indicate that these are surveys. Maybe something like `_offlineSurveys`. Similarly for others too.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -51,6 +53,20 @@ internal constructor(
...
+
+  val showDialog = MutableStateFlow(false)
```
shobhitagarwal1612: How about calling it `showDeleteDialog` to be more explicit about the scope of this state?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -107,11 +123,44 @@ internal constructor(
...
+
+    _sharedWith.value =
+      surveys.filter {
+        !it.availableOffline &&
+          (it.generalAccess == Survey.GeneralAccess.RESTRICTED ||
+            it.generalAccess == Survey.GeneralAccess.UNLISTED)
+      }
+
+    _publicList.value =
+      surveys.filter { !it.availableOffline && it.generalAccess == Survey.GeneralAccess.PUBLIC }
```
shobhitagarwal1612: Is it possible to have the same survey listed in multiple columns? Can you please clarify? If the survey in each section needs to be exclusive, then what's the priority order? In that case, we should also add some validation checks to ensure that the surveys are not duplicated across sections.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -107,11 +123,44 @@ internal constructor(
...
+    showDialog.value = true
+  }
+
+  fun closeDeleteDialog() {
+    showDialog.value = false
+    selectedSurveyId.value = null
+  }
+
+  fun confirmDelete() {
+    selectedSurveyId.value?.let { deleteSurvey(it) }
```
shobhitagarwal1612: Can we pass this survey id as a parameter rather than maintaining as a state in view model?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -51,6 +53,20 @@ internal constructor(
...
+  val onDeviceSurveys: StateFlow<List<SurveyListItem>> = _onDeviceSurveys
+
+  private val _sharedWithSurveys = MutableStateFlow<List<SurveyListItem>>(emptyList())
+  val sharedWithSurveys: StateFlow<List<SurveyListItem>> = _sharedWithSurveys
+
+  private val _publicListSurveys = MutableStateFlow<List<SurveyListItem>>(emptyList())
+  val publicListSurveys: StateFlow<List<SurveyListItem>> = _publicListSurveys
+
+  val showDeleteDialog = MutableStateFlow(false)
+  val selectedSurveyId = mutableStateOf<String?>(null)
```
shobhitagarwal1612, line 68: Can we cleanup usages of this as we are now passing the surey id as a parameter to `openDeleteDialog`?

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
```diff
@@ -107,11 +123,44 @@ internal constructor(
...
+
+    _sharedWithSurveys.value =
+      surveys.filter {
+        !it.availableOffline &&
+          (it.generalAccess == Survey.GeneralAccess.RESTRICTED ||
+            it.generalAccess == Survey.GeneralAccess.UNLISTED)
+      }
+
+    _publicListSurveys.value =
+      surveys.filter { !it.availableOffline && it.generalAccess == Survey.GeneralAccess.PUBLIC }
```
shobhitagarwal1612: We can use `groupBy` to split the surveys into all categories in a single pass. This will make the code cleaner and more efficient.

```
val grouped = surveys.groupBy {
    when {
        it.availableOffline -> "onDevice"
        it.generalAccess == Survey.GeneralAccess.PUBLIC -> "public"
        it.generalAccess == Survey.GeneralAccess.RESTRICTED ||
        it.generalAccess == Survey.GeneralAccess.UNLISTED -> "sharedWith"
        else -> "other"
    }
}
_onDeviceSurveys.value = grouped["onDevice"].orEmpty()
_sharedWithSurveys.value = grouped["sharedWith"].orEmpty()
_publicListSurveys.value = grouped["public"].orEmpty()
```

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListScreen.kt
```diff
@@ -0,0 +1,340 @@
...
+    R.string.section_shared_with_me to false,
+    R.string.section_public to false,
+  )
+}
+
+private fun Survey.GeneralAccess.iconRes(): Int =
+  when (ordinal) {
+    Survey.GeneralAccess.UNLISTED.ordinal -> R.drawable.ic_unlisted
+    Survey.GeneralAccess.PUBLIC.ordinal -> R.drawable.ic_public
+    else -> R.drawable.ic_restricted
```
shobhitagarwal1612: Instead of using `else` block for `RESTRICTED` type, let's explicitly handle for it and throw an exception in else block. This would prevent any unexpected errors if there is any changes with default values or some other new value is defined.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListScreen.kt
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+private fun Survey.GeneralAccess.labelString(): Int =
+  when (ordinal) {
+    Survey.GeneralAccess.UNLISTED.ordinal -> R.string.access_unlisted
+    Survey.GeneralAccess.PUBLIC.ordinal -> R.string.access_public
+    else -> R.string.access_restricted
```
shobhitagarwal1612: Same as above

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListScreen.kt
```diff
@@ -0,0 +1,340 @@
...
+  onCardClick: (String) -> Unit,
+  menuClick: (String) -> Unit,
+) {
+  this.items(items, key = { it.id }) { item ->
+    SurveyCardItem(item = item, onCardClick = { onCardClick(it) }, menuClick = { menuClick(it) })
+  }
+}
+
+@Composable
+private fun SectionHeader(
```
shobhitagarwal1612: Can we move this component to a new sub-package `components`? 

This will help with future refactoring of keeping all individual composables in `components` package and the main composable which binds everything together into this class (e.g. *Screen) which can live outside of this `components` package (like this file)

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListScreen.kt
```diff
@@ -0,0 +1,340 @@
...
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
+          )
+        }
+    }
+  }
+}
+
+@Composable
+private fun HeaderRow(item: SurveyListItem, menuClick: (String) -> Unit) {
```
shobhitagarwal1612: Same as the other comment. Let's move it to `components` sub-package.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveyListScreen.kt
```diff
@@ -0,0 +1,340 @@
...
+
+private fun Survey.GeneralAccess.labelString(): Int =
+  when (ordinal) {
+    Survey.GeneralAccess.UNLISTED.ordinal -> R.string.access_unlisted
+    Survey.GeneralAccess.PUBLIC.ordinal -> R.string.access_public
+    else -> R.string.access_restricted
+  }
+
+@Composable
+private fun SurveyCardItem(
```
shobhitagarwal1612: Same as the other comment, let's it to the separate package.

app/src/main/java/org/groundplatform/android/ui/surveyselector/SurveySelectorViewModel.kt
anandwana001: We are using this variable to pass and eep track of current id.

shobhitagarwal1612 (approved):

shobhitagarwal1612:
Some tests are still failing. Can you please check?

anandwana001:
Screenshot 2025-07-30 at 10 32 12 PM


failing at github action but all passing locally.

PR #3219, opened 2025-07-30
Update Condition.kt

GeorgCantor:
Using `.count { ... }` is more efficient
The `.count { predicate }` function internally iterates over the list, counts the number of elements matching the predicate, and returns that count as an integer, without creating any intermediate collections. This reduces memory usage and can be faster, especially for large lists.

@... PTAL?


touhidurrr:
why not just use some I wonder
oh ok, some does not tell you if there is exactly one or more than one

PR #3235, opened 2025-08-21
[Draw/walk area task] Prevent user from drawing self-intersecting polygons

### LINKED ISSUE #3234: Mis-shapen polygon can be drawn through "Complete"
Related to https://github.com/google/ground-android/issues/2604

Adding a point during polygon drawing that create a mis-shapen polygon correctly gives an error. However, if adding it as the last point ("Complete") the mis-shapen polygon goes through.

https://github.com/user-attachments/assets/687965e5-8f9a-4f54-be50-89f071c38b5e

Discovered by @lecrabe, who asked why we even prohibit such polygons?

@gino-m @anandwana001 @shobhitagarwal1612 @kenstershiro

anandwana001:
Fixes #3234 

This PR introduces a full system to detect and handle polygon self-intersections:

Implemented isSelfIntersecting(vertices) to check for crossing edges, with support for both open polylines and closed polygons.

Integrated validation into the workflow:
Add Point: checks the newly added edge, reverts the vertex, and shows a dialog if self-intersection is detected.
Complete Polygon: validates the would-be closed ring (vertices + first), blocking completion and showing a dialog if invalid.
Ensured the intersection flag resets when geometry is valid, preventing stale “intersecting” states.
Updated onFeatureUpdated to disable the Complete button when a polygon is invalid.

Added unit tests covering:
Open and closed polygons
Adjacent edges sharing vertices
True self-intersections (e.g., “X” shapes)
Edge cases like empty list, single point, or triangles

This provides reliable guardrails to prevent creating invalid polygons and ensures user feedback when intersections occur.

https://github.com/user-attachments/assets/25c32662-d3df-48ec-bcde-4b366db1e214

@gino-m @jo-spek  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -95,8 +95,12 @@ internal constructor(
   private val _showSelfIntersectionDialog = MutableSharedFlow<Unit>()
   val showSelfIntersectionDialog = _showSelfIntersectionDialog.asSharedFlow()
 
+  private var hasSelfIntersection: Boolean = false
```
shobhitagarwal1612: We can declare the setter as private and getter as public. Then we won't need a separate `hasSelfIntersection()` method

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +241,38 @@ internal constructor(
...
-  fun checkVertexIntersection() {
-    if (isSelfIntersecting(vertices)) {
+  fun checkVertexIntersection(): Boolean {
+    val hit = isSelfIntersecting(vertices)
+    hasSelfIntersection = hit
+    if (hit) {
       vertices = vertices.dropLast(1)
       onSelfIntersectionDetected()
+    } else {
+      hasSelfIntersection = false
```
shobhitagarwal1612: Isn't this condition implicit? Any reason for explicitly assigning it to false?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +241,38 @@ internal constructor(
...
+    }
+    return hit
+  }
+
+  fun validateBeforeComplete(): Boolean {
```
shobhitagarwal1612: Please consider using a better name for the function. Maybe something which clarifies the purpose of the method.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +241,38 @@ internal constructor(
...
+        vertices
+      }
+
+    val hit = isSelfIntersecting(ring)
+    hasSelfIntersection = hit
+    if (hit) {
+      onSelfIntersectionDetected()
+      return false
     }
+    hasSelfIntersection = false
```
shobhitagarwal1612: Isn't this statement also implicit?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +241,38 @@ internal constructor(
...
+
+  fun validateBeforeComplete(): Boolean {
+    val ring =
+      if (vertices.size >= 3 && vertices.first() != vertices.last()) {
+        vertices + vertices.first()
+      } else {
+        vertices
+      }
+
+    val hit = isSelfIntersecting(ring)
```
shobhitagarwal1612: Can we reuse `hasSelfIntersection` instead of defining another local variable?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -92,11 +92,16 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
...
-        viewModel.triggerVibration()
+        val intersected = viewModel.checkVertexIntersection()
+        if (!intersected) viewModel.triggerVibration()
       }
     completeButton =
-      addButton(ButtonAction.COMPLETE).setOnClickListener { viewModel.completePolygon() }
+      addButton(ButtonAction.COMPLETE).setOnClickListener {
+        if (viewModel.validateBeforeComplete()) {
+          viewModel.completePolygon()
+          viewModel.clearIntersectionFlag()
```
shobhitagarwal1612: Is it even possible to have this flag true when the validation is passing? I feel this line is adding confusion.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
   c.lat in minOf(a.lat, b.lat)..maxOf(a.lat, b.lat) &&
     c.lng in minOf(a.lng, b.lng)..maxOf(a.lng, b.lng)
 
+private data class Segment(val a: Coordinates, val b: Coordinates)
+
+private fun isClosed(vertices: List<Coordinates>): Boolean =
```
shobhitagarwal1612: We already have this method defined inside `Geometry` class. Can we reuse it instead?

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
...
+  vertices.size >= 2 && vertices.first() == vertices.last()
+
+private fun shareEndpoint(s1: Segment, s2: Segment): Boolean =
+  s1.a == s2.a || s1.a == s2.b || s1.b == s2.a || s1.b == s2.b
+
+private fun buildSegments(vertices: List<Coordinates>): List<Segment> {
```
shobhitagarwal1612: Please add ktdoc

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
...
+  val n = vertices.size
+  if (n < 2) return emptyList()
+  val closed = isClosed(vertices)
+
+  val segs = ArrayList<Segment>(if (closed) n else n - 1) // <- removed unnecessary parentheses
```
shobhitagarwal1612: Please cleanup unused comments.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
   c.lat in minOf(a.lat, b.lat)..maxOf(a.lat, b.lat) &&
     c.lng in minOf(a.lng, b.lng)..maxOf(a.lng, b.lng)
 
+private data class Segment(val a: Coordinates, val b: Coordinates)
+
+private fun isClosed(vertices: List<Coordinates>): Boolean =
+  vertices.size >= 2 && vertices.first() == vertices.last()
```
shobhitagarwal1612: Should this condition be `> 3`?

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
...
+private fun buildSegments(vertices: List<Coordinates>): List<Segment> {
+  val n = vertices.size
+  if (n < 2) return emptyList()
+  val closed = isClosed(vertices)
+
+  val segs = ArrayList<Segment>(if (closed) n else n - 1) // <- removed unnecessary parentheses
+  for (i in 0 until n - 1) {
+    segs += Segment(vertices[i], vertices[i + 1])
+  }
+  if (closed) segs += Segment(vertices[n - 1], vertices[0])
```
shobhitagarwal1612: Can you add some comments to help add context why this is needed?

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
...
+  return segs
+}
+
+private fun areAdjacent(i: Int, j: Int, size: Int, closed: Boolean): Boolean {
```
shobhitagarwal1612: Please add ktdoc and use better parameter names. `i` and `j` should not be used outside of a loop.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -85,19 +85,54 @@ private fun onSegment(a: Coordinates, b: Coordinates, c: Coordinates) =
   c.lat in minOf(a.lat, b.lat)..maxOf(a.lat, b.lat) &&
     c.lng in minOf(a.lng, b.lng)..maxOf(a.lng, b.lng)
 
+private data class Segment(val a: Coordinates, val b: Coordinates)
```
shobhitagarwal1612: Please add some ktdoc. What do we mean by a segment?

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
anandwana001: That is more specific to LineString object; here we only have a list of coordinates.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +240,34 @@ internal constructor(
     }
   }
 
-  fun checkVertexIntersection() {
-    if (isSelfIntersecting(vertices)) {
+  fun checkVertexIntersection(): Boolean {
+    val hit = isSelfIntersecting(vertices)
+    hasSelfIntersection = hit
+    if (hit) {
```
shobhitagarwal1612: Let's reuse the `hasSelfIntersection` variable here as well instead of using a local variable.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +240,34 @@ internal constructor(
...
+    return hit
+  }
+
+  fun validatePolygonCompletion(): Boolean {
+    val ring =
+      if (vertices.size >= 3 && vertices.first() != vertices.last()) {
+        vertices + vertices.first()
+      } else {
+        vertices
+      }
```
shobhitagarwal1612: We should return early if it's not a polygon.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -245,11 +240,34 @@ internal constructor(
...
+    if (hasSelfIntersection) {
+      onSelfIntersectionDetected()
+      return false
+    }
+    return true
+  }
+
+  fun clearIntersectionFlag() {
+    hasSelfIntersection = false
   }
```
shobhitagarwal1612: Please remove if unused.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
shobhitagarwal1612: The logic below is incorrect. A polygon must have 4 vertices. I'd still suggest somehow reusing the logic instead of duplicating the definition at multiple places.

anandwana001: done

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -81,23 +109,116 @@ private fun orientation(a: Coordinates, b: Coordinates, c: Coordinates): Int {
...
+  val edges = ArrayList<PolygonEdge>(if (closed) n else n - 1)
+
+  // Create edges between each consecutive pair of vertices.
+  for (i in 0 until n - 1) {
+    edges += PolygonEdge(vertices[i], vertices[i + 1])
+  }
+
+  // If the polygon is closed, add an edge from the last vertex back to the first
+  // to ensure the polygon forms a complete loop.
+  if (closed) edges += PolygonEdge(vertices[n - 1], vertices[0])
```
shobhitagarwal1612: If the polygon is closed, why do we need a edge whose vertices are same? Won't that be of zero length?

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -81,23 +109,116 @@ private fun orientation(a: Coordinates, b: Coordinates, c: Coordinates): Int {
...
+ * @param edgeCount Total number of edges.
+ * @param isClosedPolygon Whether the polygon is closed (first vertex equals last vertex).
+ */
+private fun areAdjacent(
+  firstEdgeIndex: Int,
+  secondEdgeIndex: Int,
+  edgeCount: Int,
+  isClosedPolygon: Boolean,
+): Boolean {
+  if (firstEdgeIndex == secondEdgeIndex) return true
```
shobhitagarwal1612: Does this condition represent a scenario where the edges being compared are same? If yes, we should prevent it from occurring in the first place itself rather than having to handle it separately.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
```diff
@@ -81,23 +109,116 @@ private fun orientation(a: Coordinates, b: Coordinates, c: Coordinates): Int {
...
+  if (abs(firstEdgeIndex - secondEdgeIndex) == 1) return true
+  if (isClosedPolygon) {
+    val last = edgeCount - 1
+    if (
+      (firstEdgeIndex == 0 && secondEdgeIndex == last) ||
+        (secondEdgeIndex == 0 && firstEdgeIndex == last)
+    ) {
+      return true
+    }
+  }
```
shobhitagarwal1612: We can simplify these method (this and the other ones) by specifying exactly which geometry we are dealing with (i.e. linear ring or polygon). Handling for generic vertices is causing readability issues as you need to check for `completeness` at every step. Either pass a linear ring which ensures that the first and last vertices are same or handle generically for a list of vertices without checking for `completeness`.

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
gino-m: +1. If the polygon is closed, I would expect the last edge to end at the first vertex.

app/src/main/java/org/groundplatform/android/ui/util/PolygonUtil.kt
gino-m: Aren't polygons represented by the coordinates of the vertices? In that case, the edges are inferred by looking at those coordinates, and there's no way to represent non-adjacent edges.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
```diff
@@ -157,10 +161,12 @@ class DrawAreaTaskFragment @Inject constructor() : AbstractTaskFragment<DrawArea
     val closed = geometry.isClosed()
     val markedComplete = viewModel.isMarkedComplete()
     val tooClose = viewModel.isTooClose.value
+    val selfIntersecting = viewModel.hasSelfIntersection
 
     addPointButton.showIfTrue(!closed)
     addPointButton.enableIfTrue(!closed && !tooClose)
     completeButton.showIfTrue(closed && !markedComplete)
+    completeButton.enableIfTrue(closed && !markedComplete && !selfIntersecting)
```
gino-m, line 169: How does the user know the "complete button" is disabled because the polygon is self-intersecting? Should we instead be disabling the "add point" button as the user pans the map?

gino-m:
@anandwana001 Note I updated the PR description to reflect the new behavior rather than a description of the code

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragment.kt
shobhitagarwal1612: +1, we should disable both `add` and `complete` buttons when the polygon is irregular

shobhitagarwal1612: Discussed offline, we are showing a dialog and reverting the last point when adding a point which intersects the polygon

shobhitagarwal1612 (approved):

PR #3236, opened 2025-08-21
[Draw area task] Update last edge and distance while dragging map

### LINKED ISSUE #3066: [Draw area task] Update last edge and distance while dragging map
These currently update when the map stops moving - instead they should update while dragging.

anandwana001:
Fixes #3066

https://github.com/user-attachments/assets/893a4bd6-5737-41e3-a148-4fe2fc6ddf13

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
@@ -42,6 +42,7 @@ import javax.inject.Inject
 import kotlin.math.min
 import kotlin.math.sqrt
 import kotlinx.collections.immutable.toImmutableList
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.launch
 import org.groundplatform.android.common.Constants
@@ -79,7 +80,11 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
   override val startDragEvents = MutableSharedFlow<Unit>()
 
   /** Camera move events. Emits items after the camera has stopped moving. */
-  override val cameraMovedEvents = MutableSharedFlow<CameraPosition>()
+  override val cameraMovedEvents =
+    MutableSharedFlow<CameraPosition>(
+      extraBufferCapacity = 64,
+      onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    )
 
   @Inject lateinit var featureManager: FeatureManager
   @Inject lateinit var remoteStorageManager: RemoteStorageManager
@@ -167,6 +172,7 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
 
     map.setOnCameraIdleListener(this::onCameraIdle)
     map.setOnCameraMoveStartedListener(this::onCameraMoveStarted)
+    map.setOnCameraMoveListener(this::onCameraMoving)
     map.setOnMapClickListener { onMapClick(it) }
 
     with(map.uiSettings) {
@@ -264,6 +270,24 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
   private fun onCameraMoveStarted(reason: Int) {
     if (reason == OnCameraMoveStartedListener.REASON_GESTURE) {
       viewLifecycleOwner.lifecycleScope.launch { startDragEvents.emit(Unit) }
+    }
+  }
+
+  private fun onCameraMoving() {
+    lifecycleScope.launch {
+      repeatOnLifecycle(Lifecycle.State.STARTED) {
+        map.setOnCameraMoveListener {
+          val cameraPosition = map.cameraPosition
+          val projection = map.projection
+          cameraMovedEvents.tryEmit(
+            CameraPosition(
+              cameraPosition.target.toCoordinates(),
+              cameraPosition.zoom,
+              projection.visibleRegion.latLngBounds.toModelObject(),
+            )
+          )
+        }
+      }
     }
   }
 


shobhitagarwal1612:
Can we do something about the black box?

gino-m:
Also, can we update the line rather than delete and re-add to reduce the flicker?

anandwana001:
Moving to draft, as after the latest sync from master, this solution stopped working.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -136,4 +184,45 @@ constructor(
   fun onCameraIdle() {
     clusterManager.onCameraIdle()
   }
+
+  /** Helper to detect whether a [Feature] is a draft line that should be updated in place. */
+  private fun isDraftLineString(feature: Feature): Boolean =
```
gino-m: Can we make this an extension function?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -90,17 +96,59 @@ constructor(
...
+   * Clustering:
+   * - Clustering is always refreshed after applying the diff. For draft updates we skip diffing
+   *   entirely and just update the geometry directly.
    */
   fun setFeatures(updatedFeatures: Collection<Feature>) {
-    // remove stale
-    val removedOrChanged = features - updatedFeatures.toSet()
+    val incomingDraft = updatedFeatures.firstOrNull { isDraftLineString(it) }
+
+    // Case 1: We already have a draft rendered and got another draft update → update in place.
```
gino-m: This class is intended to deal with rendering `Features` without any knowledge of the specific flows or CUJs they derive from. Making a special case for "draft" feature expands the scope / leaks the abstraction from the caller. Instead, we should be modifying `updatedFeatures` at the caller and passing in the new list each time the map moves.

anandwana001: Done

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManagerExt.kt
```diff
@@ -0,0 +1,27 @@
...
+package org.groundplatform.android.ui.map.gms.features
+
+import org.groundplatform.android.model.geometry.LineString
+import org.groundplatform.android.ui.map.Feature
+
+/**
+ * Returns true if this [Feature] represents a user-drawn "draft" line string (i.e. in-progress
+ * polygon drawing that should be updated in place).
+ */
+fun Feature.isDraftLineString(): Boolean =
```
gino-m: This doesn't need to be in its own file, it can be a private extension function in `FeatureManager`.

app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -241,7 +246,31 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
 
   override fun setFeatures(newFeatures: Set<Feature>) {
     Timber.v("setFeatures() called with ${newFeatures.size} features")
-    featureManager.setFeatures(newFeatures)
+
+    val draft = newFeatures.firstOrNull { it.isDraftLineString() }
```
gino-m: Sorry to keep pushing on this, but this isn't the right place either. This whole package implements the bridge between Maps SDK and our own geometry model and events. "Draft drawing state" has no meaning in this context; it just shows the features it's given, with the style it's given.

Since this behavior is specific to the "Draw area" feature, I would expect it to be implemented in the `DrawAreaTaskViewModel`, not here in the map adapter.

app/src/test/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragmentTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    val draft = fakeFeature(id = "draft1", isDraft = true)
+
+    fragment.setFeatures(setOf(draft))
+
+    verify(_featureManager).setFeatures(setOf(draft))
+    verify(_featureManager, never()).updateLineString(any(), any(), any(), any(), anyOrNull())
+  }
+
+  @Test
+  fun `subsequent draft updates go through updateLineString`() {
```
gino-m: Thanks for adding tests! Please move these to DrawAreaTaskViewModelTest when you move the logic over!

gino-m (changes requested):

app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
anandwana001: ok,
The reason to keep here was to reducce the add remove churn and multi object building, let me shift to viewmodel layer and see if we can ignore the memory issue.

gino-m: All items in the set except the one that changes are references to existing objects - the only churn will be for two objects, the immutable set itself, and the draft feature. 

I assume when that refactor is done this extension function will move into the vm as well.

Thanks, @anandwana001 !

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -286,24 +290,60 @@ internal constructor(
...
         }
-      )
+      }
     }
 
-  private suspend fun buildPolygonFeature() =
+  private suspend fun buildPolygonFeature(style: Feature.Style) =
     Feature(
-      id = uuidGenerator.generateUuid(),
+      id = uuidGenerator.generateUuid(), // suspend, called only once per draft
```
gino-m: What does this comment mean here?

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
```diff
@@ -85,6 +86,7 @@ class DrawAreaTaskFragmentTest :
     )
   }
 
+  @Ignore
```
gino-m: Did you mean to check the these in?

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
```diff
@@ -177,6 +180,7 @@ class DrawAreaTaskFragmentTest :
...
```
gino-m: Same here

gino-m (changes requested):
Nice progress..

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
anandwana001: planning to take this in next PR and take this pr in coming release.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -69,6 +69,10 @@ internal constructor(
   private val _draftArea: MutableStateFlow<Feature?> = MutableStateFlow(null)
   val draftArea: StateFlow<Feature?> = _draftArea.asStateFlow()
 
+  private var draftTag: Feature.Tag? = null
+  private val _draftUpdates = MutableSharedFlow<Feature>(extraBufferCapacity = 1)
+  val draftUpdates = _draftUpdates.asSharedFlow()
```
shobhitagarwal1612, line 96: Please add ktdoc

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -136,4 +137,33 @@ constructor(
...
+   * can use this without forcing a full add/remove diff.
+   */
+  fun updateLineString(
+    tag: Feature.Tag,
+    geometry: LineString,
+    style: Feature.Style,
+    selected: Boolean,
+    tooltipText: String?,
+  ) {
+    coroutineScope.launch {
```
shobhitagarwal1612: Do we need this here? We don't launch a couroutine when adding or removing a feature

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -69,6 +69,10 @@ internal constructor(
   private val _draftArea: MutableStateFlow<Feature?> = MutableStateFlow(null)
   val draftArea: StateFlow<Feature?> = _draftArea.asStateFlow()
 
+  private var draftTag: Feature.Tag? = null
+  private val _draftUpdates = MutableSharedFlow<Feature>(extraBufferCapacity = 1)
```
shobhitagarwal1612, line 88: Do we really need this new flow? Can't we reuse `_draftArea` here as well?

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -136,4 +137,33 @@ constructor(
...
+      mapsItemManager.updateLineString(tag, geometry, style, selected, tooltipText)
+      // keep indices consistent for future diffs
+      featuresByTag[tag]?.let { prev ->
+        val updated =
+          prev.copy(
+            geometry = geometry,
+            style = style,
+            selected = selected,
+            tooltipText = tooltipText ?: prev.tooltipText,
+          )
```
shobhitagarwal1612: What's the point of this? I thought we already reused the tag from previous feature. So, it seems redundant to me

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
anandwana001: We introduced _draftUpdates because _draftArea is used for draft creation/clearing, while this flow handles high-frequency in-place updates during drawing.

Reusing _draftArea would cause the entire feature set to re-render on every small vertex move, leading to flicker and extra allocations.

_draftUpdates uses a SharedFlow with a small buffer to emit lightweight updates via tryEmit, keeping map rendering smooth without unnecessary churn.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
anandwana001: This part keeps the in-memory indices (features, featuresByTag) consistent after updating the line on the map.

Since Feature is immutable, we need to replace the old instance with a new one reflecting the latest geometry and style — otherwise future setFeatures() diffs or lookups would operate on stale data and cause unnecessary add/remove churn or flicker.

The logic itself is correct; we could optionally add a small optimization to skip the replace if nothing actually changed, but functionally this approach is sound and keeps the caches in sync.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
anandwana001: The coroutine wrapper was added because mapsItemManager isn’t always initialized yet when the first draft update arrives — causing a crash. Launching inside coroutineScope (which uses the main dispatcher) defers the update just enough for the map to finish binding. We can make it safer by adding an explicit ::mapsItemManager.isInitialized check before calling updateLineString() to avoid silent race conditions.

gino-m: > defers the update just enough for the map to finish binding

This sounds a bit like "Hail Mary" concurrency. Praying that things happen in the right order because we intentionally slow one down enough isn't really a valid strategy. Why would we be trying to update the map before the map has initialized? Can we make it such that we can have a reference to this and related classes only _after_ the map is initialized?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
shobhitagarwal1612: Is this really needed now? I noticed that this is replicated below as well? Can you please check?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -46,11 +49,39 @@ class DrawAreaTaskMapFragment @Inject constructor() :
...
-        .collect { shouldShow -> setCenterMarkerVisibility(shouldShow) }
+
+        launch {
+          taskViewModel.draftUpdates.collect { feature ->
+            map.updateLineString(
+              tag = feature.tag,
+              geometry = feature.geometry as LineString,
+              style = feature.style,
+              selected = feature.selected,
+              tooltipText = feature.tooltipText,
```
shobhitagarwal1612, line 69: Instead of passing each value individually, let's pass the entire feature object from here (in alignment with existing setFeatures function).

shobhitagarwal1612: Also, instead of naming it as `updateLineString`, let's keep it generic `updateFeature` and only add handling for `LineString` in `FeatureManager`. We can throw an exception for remaining geometry types.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -136,4 +140,33 @@ constructor(
...
+    tag: Feature.Tag,
+    geometry: LineString,
+    style: Feature.Style,
+    selected: Boolean,
+    tooltipText: String?,
+  ) {
+    if (!mapReady) return
+
+    mapsItemManager.updateLineString(tag, geometry, style, selected, tooltipText)
+    // keep indices consistent for future diffs
```
shobhitagarwal1612: Please cleanup unclear and draft comments.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureManager.kt
```diff
@@ -136,4 +140,33 @@ constructor(
...
+    featuresByTag[tag]?.let { prev ->
+      val updated =
+        prev.copy(
+          geometry = geometry,
+          style = style,
+          selected = selected,
+          tooltipText = tooltipText ?: prev.tooltipText,
+        )
+      features.remove(prev)
+      features.add(updated)
```
shobhitagarwal1612, line 167: I ran your changes locally and it seemed to work fine even without these lines. I simply replaced it with `feature` passed to this method (please refer to previous comment)

```
featuresByTag[tag] = feature
```

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -87,4 +79,40 @@ constructor(
...
+
+  fun updateExistingPolyline(
+    polyline: Polyline,
+    geometry: LineString,
+    style: Feature.Style,
+    selected: Boolean,
+    map: GoogleMap,
+    tooltipText: String?,
+  ) {
+    // Update geometry every tick
```
shobhitagarwal1612: Please cleanup the comments. They seem implicit to me.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -87,4 +79,40 @@ constructor(
...
+    polyline.points = geometry.coordinates.toLatLngList()
+
+    // Only re-apply style if it changed (cheap guard)
```
shobhitagarwal1612: Same as previous comment

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -87,4 +79,40 @@ constructor(
...
+    if (
+      polyline.color != style.color ||
+        polyline.width != defaultStrokeWidth * if (selected) 2f else 1f
+    ) {
+      polyline.applyStyle(style, selected)
+    }
+
+    // Tooltip throttled
```
shobhitagarwal1612: Same as previous comment.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/LineStringRenderer.kt
```diff
@@ -87,4 +79,40 @@ constructor(
...
     tooltipMarkerRenderer.update(map, midpoint, tooltipText)
   }
+
+  private fun Polyline.applyStyle(style: Feature.Style, selected: Boolean) {
+    if (style.vertexStyle == Feature.VertexStyle.CIRCLE) {
+      startCap = circleCap
+      endCap = circleCap
+    }
+    val strokeScale = if (selected) 2f else 1f
+    width = defaultStrokeWidth * strokeScale
```
shobhitagarwal1612: Let's move this to a private method and use here and in `updateExistingPolyline` when checking for updated values.

app/src/main/java/org/groundplatform/android/ui/map/gms/features/MapsItemManager.kt
```diff
@@ -98,4 +98,34 @@ class MapsItemManager(
...
+    style: Feature.Style,
+    selected: Boolean,
+    tooltipText: String?,
+  ) {
+    // Find the map items currently associated with this feature tag.
+    val items = itemsByTag[tag] ?: return
+
+    // Only polylines are relevant here. For each one, update its points and style.
+    items.forEach { item ->
+      if (item is Polyline) {
```
shobhitagarwal1612, line 125: Let's throw an error for else cases instead of silently consuming it. It would come handy in future if someone misuses the API

app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -267,6 +287,19 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
...
+  private fun onCameraMoving() {
+    val cameraPosition = map.cameraPosition
+    val projection = map.projection
+    cameraMovedEvents.tryEmit(
+      CameraPosition(
+        cameraPosition.target.toCoordinates(),
+        cameraPosition.zoom,
+        projection.visibleRegion.latLngBounds.toModelObject(),
+      )
+    )
```
shobhitagarwal1612: As per the ktdoc of `cameraMovedEvents`, this should only be used when the camera has stopped moving. Please remove this change and only use `cameraDragEvents`

"Camera move events. Emits items after the camera has stopped moving."

app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -92,6 +93,8 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
 
   override val featureClicks = MutableSharedFlow<Set<Feature>>()
 
+  override val cameraDragPositions = MutableSharedFlow<Coordinates>(extraBufferCapacity = 1)
```
shobhitagarwal1612: Can we rename it to `cameraDragEvents` for consistency? Also let's group it together with other similar variable in the class

app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
```diff
@@ -236,8 +240,24 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
...
+    tooltipText: String?,
+  ) {
+    featureManager.updateLineString(tag, geometry, style, selected, tooltipText)
+  }
+
+  /**
+   * Returns true if this [Feature] represents a user-drawn "draft" line string (i.e. in-progress
+   * polygon drawing that should be updated in place).
+   */
+  fun Feature.isDraftLineString(): Boolean =
```
shobhitagarwal1612: Is this unused method? If so, please cleanup

app/src/main/java/org/groundplatform/android/ui/map/MapFragment.kt
```diff
@@ -53,6 +54,17 @@ interface MapFragment {
...
+  /**
+   * A flow that continuously emits the camera target coordinates while the user is dragging.
+   * - Published by [GoogleMapsFragment.onCameraMoving].
+   * - Consumed by task fragments (e.g., DrawAreaTaskMapFragment) to update draft geometries
+   *   (vertices) in real-time as the map moves.
+   *
+   * Unlike [cameraMovedEvents] which only emits when the camera stops (idle), this stream fires on
+   * every movement tick.
+   */
+  val cameraDragPositions: SharedFlow<Coordinates>
```
shobhitagarwal1612: Please rename this to `cameraDragEvents` for consistency. Also, please use short and crisp documentation / comments (see other variable's comments). This would make the code more readable by removing the obvious and unnecessary items from the comment

app/src/main/java/org/groundplatform/android/ui/map/MapFragment.kt
```diff
@@ -102,4 +114,27 @@ interface MapFragment {
...
+   *
+   * It is typically called during an active polygon or polyline drawing session, when the user
+   * drags or adds vertices and we need to reflect those changes smoothly on the map.
+   *
+   * @param tag Unique [Feature.Tag] identifying the feature to update.
+   * @param geometry New [LineString] geometry to render.
+   * @param style Visual [Feature.Style] for the updated feature (e.g., stroke color, vertex style).
+   * @param selected Whether this feature is currently selected or being edited.
+   * @param tooltipText Optional text shown as a tooltip or distance label.
+   */
```
shobhitagarwal1612: Similar to previous comment.

app/src/main/java/org/groundplatform/android/ui/map/MapFragment.kt
```diff
@@ -102,4 +114,27 @@ interface MapFragment {
...
+  fun updateLineString(
```
shobhitagarwal1612, line 111: Can we rename this to `updateFeature` to keep the API generic? Then we can simply pass the whole feature object instead of deconstructing to each indiividual variable.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
```diff
@@ -90,11 +90,11 @@ class DrawAreaTaskFragmentTest :
     setupTaskFragment<DrawAreaTaskFragment>(job, task.copy(isRequired = false))
 
     runner()
-      .assertButtonIsHidden(NEXT_POINT_BUTTON_TEXT)
+      .assertButtonIsDisabled(NEXT_POINT_BUTTON_TEXT)
       .assertButtonIsEnabled(SKIP_POINT_BUTTON_TEXT)
       .assertButtonIsDisabled(UNDO_POINT_BUTTON_TEXT, true)
       .assertButtonIsDisabled(REDO_POINT_BUTTON_TEXT, true)
-      .assertButtonIsEnabled(ADD_POINT_BUTTON_TEXT)
+      .assertButtonIsHidden(ADD_POINT_BUTTON_TEXT)
```
shobhitagarwal1612: Do we know the reason for this change? Is this an existing regression that your code is fixing? If not, do we really think this is safe given the Add button is being hidden which previously used to be enabled.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
```diff
@@ -103,7 +103,7 @@ class DrawAreaTaskFragmentTest :
     setupTaskFragment<DrawAreaTaskFragment>(job, task.copy(isRequired = true))
 
     runner()
-      .assertButtonIsHidden(NEXT_POINT_BUTTON_TEXT)
+      .assertButtonIsDisabled(NEXT_POINT_BUTTON_TEXT)
```
shobhitagarwal1612: same as previous comment.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
```diff
@@ -182,7 +182,7 @@ class DrawAreaTaskFragmentTest :
     setupTaskFragment<DrawAreaTaskFragment>(job, task.copy(isRequired = false))
     ShadowDialog.getLatestDialog().dismiss()
 
-    runner().assertButtonIsEnabled(ADD_POINT_BUTTON_TEXT)
+    runner().assertButtonIsHidden(ADD_POINT_BUTTON_TEXT)
```
shobhitagarwal1612: same as previous comment.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
anandwana001: fixed

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
anandwana001: fixed

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskFragmentTest.kt
anandwana001: fixed

shobhitagarwal1612:
After the latest set of changes, the map dragging doesn't work smoothly. Can you please check?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -34,23 +35,35 @@ import org.groundplatform.android.ui.map.gms.GmsExt.toBounds
...
+              taskViewModel.updateLastVertexAndMaybeCompletePolygon(coord) { c1, c2 ->
+                map.getDistanceInPixels(c1, c2)
+              }
+            }
+          }
+        }
+
+        launch {
+          taskViewModel.draftUpdates.collect { feature ->
+            launch { taskViewModel.draftUpdates.collect { feature -> map.updateFeature(feature) } }
```
shobhitagarwal1612: The lag is because of this line. Can you please remove the duplicate line?

anandwana001: fixed and revert heavy object passing which is causing lag

Issue #3213, opened 2025-07-30
[Draw polygon] Always use meters for draw polygon and area estimate


gino-m:
Let's just always use meters until we have a chance to add an option to the Settings screen.

gino-m:
@shobhitagarwal1612 Do you have a sec for this?

shobhitagarwal1612:
Should we add an entry to the settings screen as well instead of forcing it to be always SI unit?

shobhitagarwal1612:
Should we have the settings pref for "metric" vs "imperial"? Or "ft" vs "metre"?

PR #3239, opened 2025-08-24
Add a new setting for switching unit of length between "meter" and "feet"

shobhitagarwal1612:
Fixes https://github.com/google/ground-android/issues/3213

Adds a new option to switch unit of length between "m" and "ft". The default will be set to `Meters (m)`.

[distance_pref.webm](https://github.com/user-attachments/assets/2782c18e-5f12-47b1-95a1-07967a15a936)

@anandwana001 @gino-m  PTAL?


priyanshscpp (approved):

.idea/codeStyles/Project.xml
anandwana001, line 1: do we need this file?

app/src/main/java/org/groundplatform/android/ui/util/LocaleAwareMeasureFormatter.kt
```diff
@@ -25,37 +25,35 @@ import kotlin.math.floor
...
 /** Distances equal to or greater than this value are rounded down to the nearest integer. */
 private const val MIN_ROUNDED_DISTANCE = 10
+
 /** The number of decimal places shown for distances < MIN_ROUNDED_DISTANCE. */
 private const val SMALL_DISTANCE_DECIMAL_PLACES = 1
 
-class LocaleAwareMeasureFormatter @Inject constructor(private val locale: Locale) {
-  private val uLocale = ULocale.forLocale(locale)
+class LocaleAwareMeasureFormatter @Inject constructor() {
+  private val uLocale = ULocale.forLocale(Locale.getDefault())
```
anandwana001: aren't we using the dependency provider for Locale.getDefault()

anandwana001 (commented):
now that we are using "m" at multple places, can we have a single source?

.idea/codeStyles/Project.xml
shobhitagarwal1612: This change is automatically done by the IDE and appears every single time. Can we include just so that it doesn't appear again? Is it a no op change.

app/src/main/java/org/groundplatform/android/ui/util/LocaleAwareMeasureFormatter.kt
shobhitagarwal1612: Done

shobhitagarwal1612:
@anandwana001 Done. PTAL

PR #3244, opened 2025-08-29
Refactor: Moves map related data classes to model package which are shared application wide

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/3186

Moves the shared map model classes from `ui` to `model` package.

Some classes e.g. `Bounds`, `CameraPosition`, `MapType` are currently defined under `ui` package but are used outside of it as well. As per our directory structure, all shared model classes must be put under top-level `model` package. 

@anandwana001  PTAL?


gino-m:
@shobhitagarwal1612 Please pull and resolve conflicts so we can merge.

PR #3252, opened 2025-09-12
Deprecate usages of gitVersioner

### LINKED ISSUE #3251: [Code health] Deprecate gitVersioner plugin
This plugin currently auto-generates the version name and version code for debug builds, while production builds are already versioned manually. Given that the `gitVersioner` plugin is no longer supported with Kotlin 2.0 and the latest `Gradle` plugin, I propose we either replace it with a modern alternative or adopt a manual versioning process for all builds.

shobhitagarwal1612:
Fixes #3251

@gino-m  PTAL?


gino-m:
Thank you, @shobhitagarwal1612

Issue #3216, opened 2025-07-30
[Trending crash] TimeoutKt.TimeoutCancellationException


gino-m:
<img width="610" height="341" alt="Image" src="https://github.com/user-attachments/assets/222280c8-35ae-4bbf-83a4-9e9f9aec81ca" />

```
          Fatal Exception: kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 3000 ms
       at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:191)
       at kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:159)
       at kotlinx.coroutines.EventLoopImplBase$DelayedRunnableTask.run(EventLoop.common.kt:501)
       at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:280)
       at kotlinx.coroutines.DefaultExecutor.run(DefaultExecutor.kt:109)
       at java.lang.Thread.run(Thread.java:1119)        
```

```
Failed to get survey due to timeout kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 3000 ms at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:191) at kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:159) at kotlinx.coroutines.EventLoopImplBase$DelayedRunnableTask.run(EventLoop.common.kt:501) at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:280) at kotlinx.coroutines.DefaultExecutor.run(DefaultExecutor.kt:109) at java.lang.Thread.run(Thread.java:1119)
```

gino-m:
This was likely introduced by the fix to #2998, where we blocking wait with timeout for the state to be initialized. Suggested fix from https://github.com/google/ground-android/issues/2773#issuecomment-3164003248:

>I suspect this is also related to the race condition of Survey and LOIs not being loaded before the data collection fragment is rendered.

> Suggestion: Wrap the DataCollectionFragment in a container fragment which shows a loading spinner while survey and LOIs are loading.

kenstershiro:
@anandwana001 to add some documentation to the PR above in order for @shobhitagarwal1612  / @gino-m  to test

PR #3264, opened 2025-09-26
Added Datacollection Viewmode helper class

anandwana001:
towards #3216

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,448 @@
...
+
+  private fun getDeltas(): List<ValueDelta> {
+    val tasksDataMap = taskDataHandler.dataState.value
+    val validTasks = taskSequenceHandler.getValidTasks()
+    return tasksDataMap
+      .filterKeys { it in validTasks }
+      .map { (task, taskData) -> ValueDelta(task.id, task.type, taskData) }
+  }
+
+  private fun saveChanges(state: DataCollectionUiState.Ready, deltas: List<ValueDelta>) {
```
shobhitagarwal1612: These methods (save changes, move next/previous, submit, etc.) should still remain in the view model class.

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,448 @@
...
+    shouldLoadFromDraft = savedStateHandle[TASK_SHOULD_LOAD_FROM_DRAFT] ?: false
+
+    scope.launch { initUiState() }
+  }
+
+  private suspend fun initUiState() {
+    _uiState.value = computeInitialUiState()
+  }
+
+  private suspend fun computeInitialUiState(): DataCollectionUiState {
```
shobhitagarwal1612: In my opinion, only this method should be moved to this class. Also, let's rename the class to be more precise (e.g. InitialStateProvider, etc.)

anandwana001: done

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,186 @@
...
+    shouldLoadFromDraft = savedStateHandle[TASK_SHOULD_LOAD_FROM_DRAFT] ?: false
+
+    scope.launch { initUiState() }
+  }
+
+  private suspend fun initUiState() {
+    _uiState.value = initialStateProvider()
+  }
+
+  private suspend fun initialStateProvider(): DataCollectionUiState {
```
shobhitagarwal1612: Can we make this method public and get rid of the stateFlow from this class?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,186 @@
...
+
+  private fun resolveInitialTaskId(validIds: List<String>, saved: String?): String =
+    saved?.takeIf { it.isNotBlank() && it in validIds } ?: validIds.first()
+
+  private suspend fun computeLoiName(surveyId: String, loiId: String?): String? =
+    if (loiId == null) {
+      // user is adding a new LOI; pull whatever user typed (may be null/empty)
+      savedStateHandle[TASK_LOI_NAME_KEY]
+    } else {
+      withContext(ioDispatcher) {
```
shobhitagarwal1612: Instead of using the ioDispatcher here, let's move it to the calling place so that the entire method `initialStateProvider` is run under this dispatcher.

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,186 @@
...
+
+  companion object {
+    private const val TASK_JOB_ID_KEY = "jobId"
+    private const val TASK_LOI_ID_KEY = "locationOfInterestId"
+    private const val TASK_LOI_NAME_KEY = "locationOfInterestName"
+    private const val TASK_POSITION_ID = "currentTaskId"
+    private const val TASK_SHOULD_LOAD_FROM_DRAFT = "shouldLoadFromDraft"
+    private const val SURVEY_LOAD_TIMEOUT_MILLIS = 3_000L
+
+    fun getViewModelClass(taskType: Task.Type): Class<out AbstractTaskViewModel> =
```
shobhitagarwal1612: Can we remove this method if it is unused?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionUiState.kt
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.ui.datacollection
+
+import org.groundplatform.android.model.job.Job
+import org.groundplatform.android.model.task.Task
+
+sealed interface DataCollectionUiState {
```
shobhitagarwal1612, line 33: Can you please add ktdoc to each method?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    val tasks = pickTasks(job, loiId)
+    if (!initSequenceOrReturnError(tasks)) {
+      return DataCollectionUiState.Error("No valid tasks for this job.")
+    }
+
+    val initialTaskId = computeInitialTaskId()
+    savedStateHandle[TASK_POSITION_ID] = initialTaskId
+
+    val loiName = computeLoiName(survey.id, loiId)
```
shobhitagarwal1612: I would suggest adding a try-catch block for the entire logic and return `Error` state with the thrown error. Then each of the methods can throw error with appropriate message internally. Also, instead of return error string message, let's return error_code or enum which can then be translated to proper UI strings in the ui layer.

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
anandwana001: done

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,201 @@
...
+import org.groundplatform.android.ui.common.LocationOfInterestHelper
+
+/**
+ * Extracted “engine” that contains all mutable logic previously living in DataCollectionViewModel.
+ * - Keeps API surface similar so VM can delegate gradually.
+ * - Owns flows, sequencing, validation, and draft persistence.
+ *
+ * Usage: helper.initialize(savedStateHandle, viewModelScope) expose helper.uiState, jobName,
+ * loiName from VM delegate calls: onNextClicked(), onPreviousClicked(), saveCurrentState(), etc.
+ */
```
shobhitagarwal1612: Can you please update this comment?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,201 @@
...
+constructor(
+  private val locationOfInterestHelper: LocationOfInterestHelper,
+  private val locationOfInterestRepository: LocationOfInterestRepository,
+  private val surveyRepository: SurveyRepository,
+) {
+
+  private lateinit var savedStateHandle: SavedStateHandle
+
+  private lateinit var jobId: String
+  private var loiId: String? = null
```
shobhitagarwal1612: Can this be declared as `lateinit` as well?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Initializes this helper with persisted navigation/context state.
+   *
+   * @throws DataCollectionException.InvalidJobId if the required jobId is absent.
+   */
+  fun initialize(savedStateHandle: SavedStateHandle) {
+    this.savedStateHandle = savedStateHandle
+    jobId = savedStateHandle[TASK_JOB_ID_KEY] ?: throw DataCollectionException.InvalidJobId
+    loiId = savedStateHandle[TASK_LOI_ID_KEY]
+  }
```
shobhitagarwal1612: Can these be moved as method params for `initialStateProvider` method?

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionHelper.kt
```diff
@@ -0,0 +1,201 @@
...
+   * Cancellation:
+   * - Propagates [CancellationException] to respect coroutine structured concurrency.
+   */
+  suspend fun initialStateProvider(): DataCollectionUiState =
+    try {
+      val survey = loadSurveyOrThrow()
+      val job = resolveJobOrThrow(survey)
+
+      val tasks = pickTasks(job, loiId)
+      initSequenceOrThrow(tasks)
```
shobhitagarwal1612: Is it possible to remove sequence state from this class?

shobhitagarwal1612 (approved):
💯

PR #3292, opened 2025-10-02
Migrate from kapt to KSP

### LINKED ISSUE #3291: Migrate from kapt to KSP
kapt is causing issues when upgrading the gradle to latest version. So, as per recommendation, we should first migrate to ksp to reduce the scope of gradle plugin updation PR.

https://developer.android.com/build/migrate-to-ksp

https://kotlinlang.org/docs/ksp-why-ksp.html

shobhitagarwal1612:
Fixes #3291

This is a non-user facing change. Verified for regressions by running the app locally. 

@anandwana001 @gino-m  PTAL?


shobhitagarwal1612:
@anandwana001 Can you PTAL?

PR #3337, opened 2025-10-26
Improve UX of sync status screen

### LINKED ISSUE #3112: [Data sync status]  Improve sync status UX
We spoke about this but I can't find an FR or notes.

| Before    | After |
| -------- | ------- |
| Upload pending | Pending |
| Failed | Pending retry |
| Syncing | Uploading form data |
| Data synced\n(media pending) | Photos pending |
| Data synced\n(media pending retry) | Photos pending retry |
| Media syncing | Uploading photos |
| Synced | Uploaded |

We should also precede the site name with the marker icon and move it above the user's name data, since it's the primary entity to which the submissions are attached.

Current screen:

Image

We should also update `Mutation.SyncStatus` to reflect these labels.

@gino-m keeping this issue for the UX part and https://github.com/google/ground-android/issues/2866 is for only label text.

anandwana001:
Fixes #3112

Figma
https://www.figma.com/design/r2uGtyOiWoDa29za7DNuUc/Ground-UX-Designs?node-id=8024-6220&t=yzbeYivA4674OTG7-0

Screenshot 2025-10-31 at 10 29 29 AM

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncListItem.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncListItem.kt
@@ -65,7 +65,7 @@ fun SyncListItem(modifier: Modifier, detail: SyncStatusDetail) {
         )
         Spacer(Modifier.height(4.dp))
         Text(
-          text = "${detail.label} • ${detail.user}",
+          text = "${detail.label} • ${detail.subtitle}",
           style =
             TextStyle(
               fontSize = 16.sp,
@@ -81,7 +81,7 @@ fun SyncListItem(modifier: Modifier, detail: SyncStatusDetail) {
             fontWeight = FontWeight(400),
             color = MaterialTheme.colorScheme.onSurfaceVariant,
           )
-        Text(text = detail.subtitle, style = textStyle)
+        Text(text = detail.description, style = textStyle)
         Spacer(Modifier.height(4.dp))
         Text(text = stringResource(id = detail.status.toLabel()), fontSize = 11.sp)
       }
@@ -166,6 +166,7 @@ fun PreviewSyncListItem(
       label = "Map the farms",
       subtitle = "IDX21311",
       status = Mutation.SyncStatus.PENDING,
+      description = "Lacuna Fund Cocoa Mapping",
     )
 ) {
   AppTheme { SyncListItem(Modifier, detail) }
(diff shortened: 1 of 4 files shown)


shobhitagarwal1612:
Can you please attach figma mocks to the bug or PR description?

app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncListItem.kt
```diff
@@ -61,8 +63,9 @@ fun SyncListItem(modifier: Modifier, detail: SyncStatusDetail) {
           fontWeight = FontWeight(500),
           letterSpacing = 0.1.sp,
         )
+        Spacer(Modifier.height(4.dp))
         Text(
-          text = detail.user,
+          text = "${detail.label} • ${detail.user}",
```
shobhitagarwal1612: This looks incorrect when compared with the ux mocks. User's name isn't shown in the card title.

app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncListItem.kt
```diff
@@ -129,7 +132,7 @@ private fun Mutation.SyncStatus.toIcon(): Int =
     Mutation.SyncStatus.MEDIA_UPLOAD_IN_PROGRESS,
     Mutation.SyncStatus.IN_PROGRESS -> R.drawable.ic_sync
```
shobhitagarwal1612, line 151: As per the mocks, the in progress icon needs to be animated.

app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncListItem.kt
anandwana001: that idx001 is the user name, right?

shobhitagarwal1612: That must be the custom id of the LOI

anandwana001: from group - "  It's the ID and/or name of the site, if present. There's already a function to build the string from the ID and/or user-defined name"

app/src/test/java/org/groundplatform/android/ui/syncstatus/SyncStatusFragmentTest.kt
```diff
@@ -84,9 +84,8 @@ class SyncStatusFragmentTest : BaseHiltTest() {
     setupFragment()
 
     composeTestRule.onNodeWithTag("sync list").assertIsDisplayed()
-    composeTestRule.onNodeWithText("User").assertIsDisplayed() // User's display name
+    composeTestRule.onNodeWithText("Job • User").assertIsDisplayed() // User's display name
```
shobhitagarwal1612: This needs to be updated, right?

anandwana001: updated

PR #3342, opened 2025-10-30
Fix Conditional Task Evaluation Bug in Large Surveys by Properly Invalidating Cache

### LINKED ISSUE #3248: Conditional tasks not working properly
see here: https://openforis.support/questions/4317/ground-app-conditional-questions-not-working-after-latest-update#gsc.tab=0

> Since the latest Ground app update, conditional questions in forms are no longer working properly. All questions now display regardless of previous answers, ignoring the configured conditions. This was working fine before the update.

anandwana001:
Fixes #3248 

Summary

Fixes a major issue where conditional tasks were not showing or hiding correctly in large surveys, particularly when loading from drafts or restoring saved states.

Root Cause

The TaskSequenceHandler maintained a cached list of valid tasks, but this cache was not being invalidated when:

Tasks were initialized with draft data (getTaskViewModel())

The current state was saved (saveCurrentState())

As a result, conditional logic was using stale data, leading to incorrect task visibility and navigation.

The Fix

Added explicit cache invalidation in two key locations:

In getTaskViewModel()

After initializing task data with taskDataHandler.setData()

Ensures cache is refreshed when loading from drafts or initializing tasks

In saveCurrentState()

After syncing task data before saving

Ensures cache remains accurate after saving or restoring state

These updates guarantee that the task sequence always reflects the latest conditional logic.

Impact

Conditional tasks now appear/disappear correctly when conditions are met or unmet
Draft loading restores conditional tasks properly
Navigation works correctly across large, complex surveys
No performance regressions — cache invalidation is O(1)

Testing

Added test: Load tasks from draft with conditional task data shows conditional task

All existing tests continue to pass, confirming no regressions

Files Modified

DataCollectionViewModel.kt

DataCollectionFragmentTest.kt

Verification Steps

Create a survey with conditional tasks

Trigger a conditional task via input and save as draft

Reopen draft — the conditional task should appear correctly

Change the answer — the conditional task should hide as expected

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -209,6 +210,7 @@ internal constructor(
       val taskData = if (shouldLoadFromDraft) getValueFromDraft(state.job, task) else null
       created.initialize(state.job, task, taskData)
       taskDataHandler.setData(task, taskData)
+      taskSequenceHandler.invalidateCache()
```
shobhitagarwal1612: Can you move these lines to a new method and reuse that method across all instances in this class?

anandwana001: done

shobhitagarwal1612: I meant moving both lines since the issue is originating from using setData without invalidating. 

```
fun updateDataAndInvalidateTasks(task, taskData) {
taskDataHandler.setData(task, taskData)
taskSequenceHandler.invalidateCache()
}
``

anandwana001: done

Issue #3352, opened 2025-11-10
[TRENDING] Fatal Exception: java.lang.IllegalArgumentException Unknown GeneralAccess: GENERAL_ACCESS_UNSPECIFIED


gino-m:
```
Fatal Exception: java.lang.IllegalArgumentException: Unknown GeneralAccess: GENERAL_ACCESS_UNSPECIFIED
       at org.groundplatform.android.ui.surveyselector.components.HeaderRowKt.iconRes(HeaderRow.kt:88)
       at org.groundplatform.android.ui.surveyselector.components.HeaderRowKt.HeaderRow(HeaderRow.kt:47)
       at org.groundplatform.android.ui.surveyselector.components.SurveyCardItemKt$SurveyCardItem$2.invoke(SurveyCardItem.kt:53)
       at org.groundplatform.android.ui.surveyselector.components.SurveyCardItemKt$SurveyCardItem$2.invoke(SurveyCardItem.kt:51)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:130)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
       at androidx.compose.material3.CardKt$Card$1.invoke(Card.kt:888)
       at androidx.compose.material3.CardKt$Card$1.invoke(Card.kt:96)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
       at androidx.compose.material3.SurfaceKt$Surface$1.invoke(Surface.kt:126)
       at androidx.compose.material3.SurfaceKt$Surface$1.invoke(Surface.kt:108)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
       at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
       at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:364)
       at androidx.compose.material3.SurfaceKt.Surface-T9BRK9s(Surface.kt:105)
       at androidx.compose.material3.CardKt.Card(Card.kt:89)
       at org.groundplatform.android.ui.surveyselector.components.SurveyCardItemKt.SurveyCardItem(SurveyCardItem.kt:46)
[...]

Issue #3326, opened 2025-10-22
Old surveys are missing from the list


rfontanarosa:
This depends on unspecified general access value.

Old surveys don't have that value, so the application should consider missing value or GENERAL_ACCESS_UNSPECIFIED = 0 as restricted.

<img width="508" height="327" alt="Image" src="https://github.com/user-attachments/assets/b53bc885-e9c2-49bb-ab8e-7742fbda2872" />

andreia-ferreira:
the fix for #3352 ended up fixing this issue as well 🎉 closing this

PR #3356, opened 2025-11-12
Fix general access unspecified crash

andreia-ferreira:
Fixes [#3352](https://github.com/google/ground-android/issues/3352) and [#3326 ](https://github.com/google/ground-android/issues/3326)

Currently the app is crashing when switching surveys if a user is part of a survey that didn't have the generalAccess property.
In these cases, if a survey doesn't have the parameter, it adds the default value of `GENERAL_ACCESS_UNSPECIFIED` and they should be shown in the restricted surveys section. This was crashing the app since `HeaderRow` was throwing an exception for the value.
This PR removes the logic throwing the crash, displaying the survey instead in the 'Shared with me' section, and removes the nullability from the general access property.

- [x] Remove logic throwing an exception in `HeaderRow`  for specific generalAccess values
- [x] Make the `GENERAL_ACCESS_UNSPECIFIED` surveys show up in the restricted section of the survey selection
- [x] Refactor  `Survey`, `SurveyEntity` and `SurveyListItem` making generalAccess non-null
- [x] Handle db migration to make the parameter non-null and add tests for it 

**How to reproduce**
1. Remove general access (8) from a survey in the firestore
2. Open the app of a user that has access to that survey
3. Go to the survey selection and click 'Switch survey'
4. The app doesn't crash anymore and the user can access the survey correctly

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/surveyselector/components/HeaderRow.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/surveyselector/components/HeaderRow.kt
@@ -49,7 +49,7 @@ fun HeaderRow(item: SurveyListItem, menuClick: (String) -> Unit) {
   Row(modifier = Modifier.fillMaxWidth(), verticalAlignment = Alignment.CenterVertically) {
     Icon(
       painter = painterResource(iconRes),
-      contentDescription = stringResource(R.string.offline_icon_description),
+      contentDescription = stringResource(labelRes),
       modifier = Modifier.size(24.dp).padding(end = 4.dp),
     )
 
--- a/app/src/test/java/org/groundplatform/android/ui/surveyselector/components/SurveyCardItemTest.kt
+++ b/app/src/test/java/org/groundplatform/android/ui/surveyselector/components/SurveyCardItemTest.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2025 Google LLC
+ *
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/groundplatform/android/data/local/room/entity/SurveyEntity.kt
```diff
@@ -27,6 +27,6 @@ data class SurveyEntity(
   @ColumnInfo(name = "description") val description: String?,
   @ColumnInfo(name = "acl") val acl: JSONObject?,
   @ColumnInfo(name = "data_sharing_terms") val dataSharingTerms: ByteArray?,
-  @ColumnInfo(name = "general_access") val generalAccess: Int?,
+  @ColumnInfo(name = "general_access") val generalAccess: Int,
```
shobhitagarwal1612, line 30: Would adding a default value here let us skip writing the manual migration? I'm just wondering if there's a benefit to doing it manually versus letting Room handle it automatically.

andreia-ferreira: unfortunately Room doesn't seem to handle this migration automatically and for existing users, the app would crash without the manual migration: `java.lang.IllegalStateException: Migration didn't properly handle: survey(org.groundplatform.android.data.local.room.entity.SurveyEntity).`

shobhitagarwal1612 (approved):
LGTM 💯

PR #3367, opened 2025-11-24
Implement button to recenter location

### LINKED ISSUE #3255: [Add point/area] Replace "location lock" with "Recenter" feature
Follow model used by Google Maps in "navigation mode" where the default state is with location lock enabled, show "Recenter" button when disabled to reenable.

andreia-ferreira:
Fixes #3255 

In order to make it clearer to the user that the map is not centered on their current location, a button to recenter will show up either on the general map view or when adding new points/areas

**Summary**
- [x] Add translations
- [x] Add recenter button and its logic to show/hide
- [x] Add unit tests

When the map is not centered in the user's location, a new re-center button should show up. Clicking on it locks the location back to the user:

https://github.com/user-attachments/assets/020a5b79-429b-472a-9c35-00f17007a318

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
@@ -236,16 +236,27 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
   /** Adds the action buttons to the UI. */
   private fun renderButtons() {
     taskView.actionButtonsContainer.composeView.setComposableContent {
-      Column(modifier = Modifier.fillMaxWidth()) {
-        HeaderCard()
-
-        Spacer(Modifier.height(12.dp))
-        Row(modifier = Modifier.fillMaxWidth(), horizontalArrangement = Arrangement.SpaceBetween) {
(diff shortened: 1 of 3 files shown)


app/src/main/res/values-es/strings.xml
```diff
@@ -235,4 +235,6 @@
   <string name="delete_site_dialog_title">¿Eliminar sitio?</string>
   <string name="delete_site_dialog_message">Este sitio y todos los datos asociados se eliminarán permanentemente. Esta acción no se puede deshacer.</string>
   <string name="delete">Eliminar</string>
+
+  <string name="recenter">Recentrar</string>
```
shobhitagarwal1612, line 242: Do we need to update this as well? (The en locale string contains a hyphen)

Same comment for the rest of the string locales too.

shobhitagarwal1612:
UX nit: the alignment of the new button looks off in the draw area fragment.

app/src/main/res/values-es/strings.xml
andreia-ferreira: In PT I can confirm it’s written without a hyphen. And from what I can see from a quick search, the other languages don’t seem to use one either

shobhitagarwal1612: Thanks for confirming!

PR #3341, opened 2025-10-30
Consolidate location capture tasks and add location lock support

anandwana001:
Fixes #3148 and #3149

This PR consolidates the CaptureLocationTask functionality into the existing DropPinTask and adds configurable location lock behavior for both point and area drawing tasks. This reduces code duplication and provides more flexible location capture options based on task configuration.

### Key Changes
1. Merged CaptureLocationTask into DropPinTask 
Removed 3 separate CaptureLocationTask files 260 lines of code
Extended DropPinTaskFragment, DropPinTaskViewModel, and DropPinTaskMapFragment to support both modes:
Free placement mode: User can pan/zoom map to drop pin anywhere
GPS-only mode: User can only capture their current device location
Mode is determined by the new allowMovingPoint property on the Task model
Appropriate UI buttons shown based on mode DROP_PIN vs CAPTURE_LOCATION
Different instruction dialogs shown for each mode

2. Added Location Lock Support for DrawAreaTask
Added allowManualOverride property to Task model for area drawing tasks
When allowManualOverride is false, location lock cannot be disengaged and map gestures are disabled
When allowManualOverride is true default, location lock starts enabled but can be turned off by panning
Added location permission dialog flow for enforced location lock scenarios

3. Updated Task Model & Conversion
Added two new boolean properties to Task:
allowMovingPoint: Controls whether user can freely place points applies to DROP_PIN tasks
allowManualOverride: Controls whether location lock can be overridden applies to DRAW_AREA tasks
Updated TaskConverter to read these properties from Firebase proto schema
Updated ValueJsonConverter for proper JSON serialization of new properties

4. Data Submission Updates
Modified SubmitDataUseCase to handle both DROP_PIN and legacy CAPTURE_LOCATION task types
Both task types now submit point geometry data appropriately

5. Removed Unused Resources
Cleaned up unused string resources across all localization files es, fr, pt, vi
Removed CaptureLocationTaskViewModel from ViewModelModule

### Impact
Code reduction: Net reduction of 150 lines of code 456 insertions, 311 deletions
Improved maintainability: Single implementation for point-based location capture
Enhanced flexibility: Task behavior now configurable via proto schema
Backward compatibility: Legacy CAPTURE_LOCATION tasks still supported in data submission

### Testing
Updated CaptureLocationTaskFragmentTest to work with consolidated implementation
Updated FakeData to include new task properties

@shobhitagarwal1612  PTAL?

Issue #3148, opened 2025-05-23
Combine "capture location" and "drop pin" tasks


gino-m:

anandwana001:
@gino-m  
Can we add more details to this?

I think this also needs a parallel web ticket? Do we need to update proto too?

gino-m:
> [@gino-m[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/gmiceli@google.com)](https://github.com/gino-m) Can we add more details to this?
> 
> I think this also needs a parallel web ticket? Do we need to update proto too?

Added details to PR description. Please let me know if you have any questions!

gino-m:
@vittorino and I spoke 🥳

gino-m:
Notes on discussion:

* @jo-spek reports that the GPS lock behavior is not generally understood by data collectors. @gino-m suggests a text label / CTA to make this clearer. @amegantz @vittorino can you help here. @kenstershiro to create a FR.
* @jo-spek simplifying and allowing multiple geometries per submission would make things easier to understand, ie if there's no special magic "add site" task.
* @jo-spek says ideally we would always capture location in the background as metadata. @gino-m I suggest we automatically capture location when data is submitted. @jo-spek suggests we only capture location automatically on geometry and photo tasks.

jo-spek:
Overall, people are already successfully geolocating, so the below suggestions are not as pressing as may seem. Regardless, the "Capture location" task keeps getting people confused and oftentimes is considered the same as "Drop a pin" and it would be good for completeness to find a better implementation long-term, potentially post GA.

On the GPS lock: I only recognized that this is a thing in regular Google Maps for the first time today. Ground is quite a different case because it is used for actively mapping plots and not just reading maps. Having a button, e.g., with an actual lock + satellite symbology might go a much longer way for users to understand it.

Here is my idea: 

In survey creation, the survey organizer can select between "Drop Pin" and "Draw polygon". Next to it, there is a slider "allow free placement". If that one is ticked, the data collector will be able to freely draw, as is currently the case. 

<img width="800" height="185" alt="Image" src="https://github.com/user-attachments/assets/349f06d0-db4d-44a2-8d65-ef4584c8459d" />

If it is not ticked, the data collector will have to go exactly where the pin shall be located, or walk around the field, registering each corner, respectively.
[...]

PR #3374, opened 2025-12-01
Introduce unify_capture_location_task and a task bit

anandwana001:
Towards #3148 
PR 1 as per https://github.com/google/ground-android/pull/3341#issuecomment-3584142573

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/model/task/Task.kt
+++ b/app/src/main/java/org/groundplatform/android/model/task/Task.kt
@@ -33,7 +33,6 @@ constructor(
   val multipleChoice: MultipleChoice? = null,
   val isAddLoiTask: Boolean = false,
   val condition: Condition? = null,
-  val isUnified: Boolean = false,
 ) {
 
   // TODO: Define these in data layer!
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/groundplatform/android/model/task/Task.kt
```diff
@@ -33,6 +33,7 @@ constructor(
   val multipleChoice: MultipleChoice? = null,
   val isAddLoiTask: Boolean = false,
   val condition: Condition? = null,
+  val isUnified: Boolean = false,
```
shobhitagarwal1612: Let's not add it here just yet.

app/src/main/java/org/groundplatform/android/Flags.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android
+
+object Flags {
+  const val UNIFY_CAPTURE_LOCATION_TASK = false
```
shobhitagarwal1612: Please add some ktdoc mentioning the purpose of this flag

app/src/main/java/org/groundplatform/android/model/task/Task.kt
anandwana001: as per https://github.com/google/ground-android/pull/3341#issuecomment-3584142573, we have to add here only, right?

PR #3376, opened 2025-12-02
Merged capture and drop fragment

anandwana001:
Towards #3148 
PR 2 as per https://github.com/google/ground-android/pull/3341#issuecomment-3584142573

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
@@ -22,41 +22,20 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import org.groundplatform.android.model.geometry.Point
 import org.groundplatform.android.model.submission.CaptureLocationTaskData
-import org.groundplatform.android.ui.common.BaseMapViewModel
 import org.groundplatform.android.ui.datacollection.tasks.AbstractTaskViewModel
+import org.groundplatform.android.ui.datacollection.tasks.LocationLockEnabledState
 import org.groundplatform.android.ui.map.gms.getAccuracyOrNull
 import org.groundplatform.android.ui.map.gms.getAltitudeOrNull
 import org.groundplatform.android.ui.map.gms.toCoordinates
-
-/** Location lock states relevant for attempting to enable it or not. */
-enum class LocationLockEnabledState {
-  /** The default, unknown state. */
-  UNKNOWN,
-
-  /** The location lock was already enabled, or an attempt was made. */
-  ALREADY_ENABLED,
-
-  /** The location lock was not already enabled. */
-  NEEDS_ENABLE,
-
-  /** Trigger to enable the location lock. */
-  ENABLE,
-}
 
 class CaptureLocationTaskViewModel @Inject constructor() : AbstractTaskViewModel() {
 
   private val _lastLocation = MutableStateFlow<Location?>(null)
   val lastLocation = _lastLocation.asStateFlow()
-  /** Allows control for triggering the location lock programmatically. */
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskFragment.kt
```diff
@@ -65,7 +65,9 @@ class DropPinTaskFragment @Inject constructor() : AbstractTaskFragment<DropPinTa
   }
 
   override fun onTaskResume() {
-    if (isVisible && !viewModel.instructionsDialogShown) {
+    if (
+      isVisible && !viewModel.instructionsDialogShown && viewModel.captureLocation.value != true
```
shobhitagarwal1612, line 69: Can you please abstract the logic for displaying dialog in the viewmodel?

```
if (isVisible && viewModel.shouldShowInstructionsDialog()) {
  showInstructionsDialog()
}
```

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,32 @@ class DropPinTaskMapFragment @Inject constructor() :
...
+    }
+
     // Disable pan/zoom gestures if a marker has been placed on the map.
     taskViewModel.features.observe(this) {
       if (it.isEmpty()) {
-        map.enableGestures()
+        if (taskViewModel.captureLocation.value == true) {
+          map.disableGestures()
+        } else {
+          map.enableGestures()
```
shobhitagarwal1612, line 45: Can the logic be included in the previous if block to prevent nesting?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,32 @@ class DropPinTaskMapFragment @Inject constructor() :
...
     }
+
+    taskViewModel.captureLocation.observe(this) { captureLocation ->
+      if (taskViewModel.features.value.isNullOrEmpty()) {
+        if (captureLocation) {
+          map.disableGestures()
+        } else {
+          map.enableGestures()
+        }
+      }
```
shobhitagarwal1612, line 59: Can we merge this with the previous logic?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
shobhitagarwal1612, line 1: Shouldn't these lines not be deleted in this PR as we are planning to preserve existing logic too?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,32 @@ class DropPinTaskMapFragment @Inject constructor() :
   override fun onMapReady(map: MapFragment) {
     super.onMapReady(map)
 
+    viewLifecycleOwner.lifecycleScope.launch {
+      taskViewModel.initLocationUpdates(getMapViewModel())
+    }
+
```
shobhitagarwal1612, line 38: Please guard these changes behind the added flag

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskViewModel.kt
anandwana001: we are shifting all this to AbstractTaskViewModel, and incase keeping this will not work as "'enableLocationLockFlow' hides member of supertype 'AbstractTaskViewModel' and needs an 'override' modifier" will throw such error.

PR #3381, opened 2025-12-02
Merged capture and drop fragment

anandwana001:
Towards #3148 
PR 2 as per https://github.com/google/ground-android/pull/3341#issuecomment-3584142573

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/repository/SurveyRepository.kt
+++ b/app/src/main/java/org/groundplatform/android/repository/SurveyRepository.kt
@@ -31,7 +31,6 @@ import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.withTimeout
-import kotlinx.coroutines.withTimeoutOrNull
 import org.groundplatform.android.FirebaseCrashLogger
 import org.groundplatform.android.coroutines.ApplicationScope
 import org.groundplatform.android.data.local.LocalValueStore
@@ -43,7 +42,7 @@ import org.groundplatform.android.model.User
 import timber.log.Timber
 
 private const val ACTIVATE_SURVEY_TIMEOUT_MILLS: Long = 3 * 1000
-private const val LOAD_REMOTE_SURVEY_TIMEOUT_MILLS: Long = 15 * 1000
+private const val LOAD_REMOTE_SURVEY_TIMEOUT_MILLS: Long = 30 * 1000
 
 /** Maintains the state of currently active survey. */
 @OptIn(ExperimentalCoroutinesApi::class)
@@ -71,7 +70,7 @@ constructor(
   suspend fun saveSurvey(survey: Survey) = localSurveyStore.insertOrUpdateSurvey(survey)
 
   suspend fun getRemoteSurvey(surveyId: String): Survey? =
-    withTimeoutOrNull(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(surveyId) }
+    withTimeout(LOAD_REMOTE_SURVEY_TIMEOUT_MILLS) { remoteDataStore.loadSurvey(surveyId) }
 
   fun getRemoteSurveys(user: User): Flow<List<SurveyListItem>> =
     combine(remoteDataStore.getRestrictedSurveyList(user), remoteDataStore.getPublicSurveyList()) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,26 @@ class DropPinTaskMapFragment @Inject constructor() :
...
     super.onMapReady(map)
 
-    // Disable pan/zoom gestures if a marker has been placed on the map.
-    taskViewModel.features.observe(this) {
-      if (it.isEmpty()) {
-        map.enableGestures()
-      } else {
-        map.disableGestures()
+    viewLifecycleOwner.lifecycleScope.launch {
+      if (taskViewModel.unifyCaptureLocationTask) {
```
shobhitagarwal1612: Can we move this if condition outside?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,26 @@ class DropPinTaskMapFragment @Inject constructor() :
...
+        taskViewModel.initLocationUpdates(getMapViewModel())
       }
     }
+
+    // Disable pan/zoom gestures if a marker has been placed on the map.
+    taskViewModel.features.observe(this) { updateGestures() }
+
+    taskViewModel.captureLocation.observe(this) { updateGestures() }
```
shobhitagarwal1612: Can these coroutines be merged into a single one?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +32,26 @@ class DropPinTaskMapFragment @Inject constructor() :
...
+  }
+
+  private fun updateGestures() {
+    if (
+      !taskViewModel.features.value.isNullOrEmpty() || taskViewModel.captureLocation.value == true
```
shobhitagarwal1612: This logic or true or false should come from the previously mentioned stream and not outside.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskViewModel.kt
```diff
@@ -53,10 +56,17 @@ constructor(
...
 
+  private val _captureLocation = MutableLiveData<Boolean>(false)
+  val captureLocation: LiveData<Boolean> = _captureLocation
+
   override fun initialize(job: Job, task: Task, taskData: TaskData?) {
     super.initialize(job, task, taskData)
     pinColor = job.getDefaultColor()
 
+    if (unifyCaptureLocationTask) {
+      _captureLocation.value = task.type == Task.Type.CAPTURE_LOCATION
```
shobhitagarwal1612: Why is this a liveData? Can this not be a simple boolean?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskViewModel.kt
shobhitagarwal1612, line 1: Can we create a new abstract view model for map based tasks `AbstractMapTaskViewModel` which extends `AbstractTaskViewModel`? The location specific logic should live in that.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
anandwana001: initLocationUpdates is a suspend function

shobhitagarwal1612: I was thinking of something like

```
if (taskViewModel.unifyCaptureLocationTask) {
     taskViewModel.initLocationUpdates(getMapViewModel())
  }
}
```

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -30,14 +33,30 @@ class DropPinTaskMapFragment @Inject constructor() :
...
+  }
+
+  // Disable pan/zoom gestures if a marker has been placed on the map.
+  private fun observeGestures() {
+    lifecycleScope.launch {
+      taskViewModel.features.asFlow().collect { features ->
+        updateGestures(features, taskViewModel.captureLocation)
+      }
+    }
+  }
```
shobhitagarwal1612: Let's inline this in onMapReady function. The method name `observeGestures` doesn't seem accurate.

shobhitagarwal1612 (approved):

Issue #3294, opened 2025-10-03
Cascading lists not resetting when parent value changes


sfyelemou:
Hello,
In my Ground form I use a cascading logic Region → Province → Commune. The issue is that when I change the region, the provinces/communes from the previous region remain active and are not cleared automatically (unlike in KoboToolbox). Is there an option or recommended way to fix this?

kenstershiro:
Duplicate of #3248

kenstershiro:
@jo-spek can we close this now that #3248 has been fixed?

jo-spek:
Yes

PR #3383, opened 2025-12-02
Fix chained conditional task logic

### LINKED ISSUE #3248: Conditional tasks not working properly

anandwana001:
Fixes #3248  #3293 #3294 

Conditional tasks were not updating correctly when their dependencies were part of a chain. Specifically, if Task C depended on Task B, and Task B depended on Task A:

User selects value in A -> B becomes visible.
User selects value in B -> C becomes visible.
User changes A -> B becomes hidden.
Issue: Task C remained visible because it was still evaluating its condition against Task B's stale data, ignoring the fact that Task B was now hidden/invalid.

Updated TaskSequenceHandler.generateValidTasksList() to iteratively build the list of valid tasks.

When evaluating if a task is valid, the system now only considers data from tasks that have already been determined to be valid in the current sequence.
This ensures that hidden tasks are treated as having no data, correctly propagating the "hidden" state down the dependency chain.

Verified with a local reproduction test case (TaskSequenceHandlerReproTest) simulating the A->B->C dependency chain.

@rfontanarosa @shobhitagarwal1612  PTAL?


anandwana001:
Hi @rfontanarosa If you could help us here in testing this with the exact bug you were facing, it will be quicky.

anandwana001:
Looks like the problem keeps happening on a single survey, we have tested this PR with 100 tasks in a survey whcich is waht we consider a heavy and all test case passing, working fine too.

anandwana001:
@rfontanarosa did you tried this?

andreia-ferreira:
This does fix issues with conditional chains :tada: but it doesn’t fully address all conditional-task issues.
Testing with the survey that @rfontanarosa mentioned, it seems that there is still an edge case that can make a conditional question be shown regardless if its condition is met or not. This is due to the fact that when the tasks are stored in the Room database, they don't follow any specific order, which means that a conditional task can be inserted before its parent task is in the database, causing Room to skip inserting the associated `ExpressionEntity` because the required foreign key doesn’t exist yet.

If, for example we update this in the `RoomSurveyStore`:
```
  private suspend fun insertOrUpdateTasks(jobId: String, tasks: Collection<Task>) {
    tasks
      .sortedBy { it.index }
      .forEach { insertOrUpdateTask(jobId, it) }
    taskDao.deleteNotIn(jobId, tasks.map { it.id })
  }
```
the conditional tasks seem to work since they are inserted by index order. However, this doesn’t correct cases where surveys were already stored incorrectly on user devices, but that can be solved with a Room migration.

Let me know what you think, and if you’d like help debugging this further.

anandwana001:
@rfontanarosa can you check now?

rfontanarosa:
it works! :)

app/src/main/java/org/groundplatform/android/data/local/room/migration/Migration_125_126.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.groundplatform.android.data.local.room.migration
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+/**
+ * Clean up the "task" table to force a refetch of tasks from the remote server. This fixes an issue
+ * where conditional tasks were inserted out of order, leading to broken foreign key constraints.
```
shobhitagarwal1612: Are we sure of this? Also, would this cause any data loss of local submissions for the users?

app/src/main/java/org/groundplatform/android/data/local/room/migration/Migration_125_126.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+val Migration_125_126 =
+  object : Migration(125, 126) {
+    override fun migrate(db: SupportSQLiteDatabase) {
+      db.execSQL("DELETE FROM task")
```
andreia-ferreira: apparently this doesn't force a re-fetch, this will clear any tasks from the downloaded surveys and causes a bug where no green action button is shown. Looking at the code, tasks only seem to be inserted when a survey is added/updated.

Maybe we could delete only the surveys whose tasks contain Conditions that don’t have matching entries in the Expression table. That should trigger a re-fetch only for the surveys that are actually affected.

andreia-ferreira: after discussing it a bit with @shobhitagarwal1612 we agreed that the best first step is to fix the issue for new users. We can handle the Room migration and the fix for existing users in a separate follow-up issue. Could you update this PR accordingly?

anandwana001: Updated

anandwana001:
@andreia-ferreira PTAL

andreia-ferreira (approved):

Issue #3149, opened 2025-05-23
[Draw area] Enforce manual override option on “Draw area” tasks


gino-m:
Adapting details from #3148 to "Draw area":

This introduces a new option set by the organizer called "allow manual override" on "Walk/draw area tasks":

Details:

* The location lock will be on by default when the task is first opened.
* If `true`, the user can pan the map to disengage location location (current behavior of "Draw area").
* If `false`, the location lock will be disabled, and the cannot pan/zoom the map (current behavior of "Capture location").
* The tooltip will be the same as it currently is for "Draw area", but when location lock is on will also show info shown in "Capture location" task.

anandwana001:
Can we add more details to this?

gino-m:
> Can we add more details to this?

Added more details above.

kenstershiro:
Check if we already have a corresponding task for Drop pin

jo-spek:
> Check if we already have a corresponding task for Drop pin

If this one is for polygon drawing, then the previous one is for "Drop Pin" I think --> https://github.com/google/ground-android/issues/3148

CMIIW please @gino-m

PR #3391, opened 2025-12-04
Toggle the flag to true for QA/Verification.

### LINKED ISSUE #3148: Combine "capture location" and "drop pin" tasks

anandwana001:
Fixes #3148
PR 4 as per https://github.com/google/ground-android/pull/3341#issuecomment-3584142573

@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
Can you please attach a screencast of a survey with this flag set to false and then to true? Let's ensure there are no regressions before enabling it. 

Please add location based tasks which are affected by this experiment.

anandwana001:
@shobhitagarwal1612 


https://github.com/user-attachments/assets/034a5853-1bf7-4c19-94b7-2d57634da412


https://github.com/user-attachments/assets/7ee9bedd-7f24-40f9-8791-ab62cd933f38

shobhitagarwal1612:
As per https://github.com/google/ground-android/issues/3148#issuecomment-3211104658, if the "allow free placement" slider in form creation is `true`, then "drop a pin" and "draw polygon" tasks would allow moving the map marker in the app (as it does today). But, if the option is `false` then the user must walk the perimeter manually for "drop a pin" or "draw polygon" task and click on the capture or add point once the user is physically on that location. The GPS button would be visible but disabled in this case as we don't want the user to be able to pan or zoom the map.

In the above attached screen recording (when the flag is true), we are still able to zoom/pan the map which shouldn't have been possible as per the requirements. Can you please confirm if my understanding is correct? If yes, then we can't flip the bit just yet as the feature is not yet fully functional.

Also, it's not clear to me how to client should behave if the form contains a "capture location" task as the ux mentioned in https://github.com/google/ground-android/issues/3148#issuecomment-3211104658 only considers the "drop a pin" and "draw polygon". Should the task be dropped? Or should it be shown regardless to respect the legacy surveys?

cc @gino-m @vittorino

anandwana001:
Following ticket with this PR next - https://github.com/google/ground-android/issues/3149

gino-m:
The requirements and behavior here are not clear. Let's revisit execution steps in #3148.

Issue #3351, opened 2025-11-07
[App area] Consistent units


jo-spek:
Upon completion of drawing a polygon, the size of the polygon is briefly shown at the bottom of the screen for a second in square meters m², no matter the system's default unit settings. With an imperial unit system locale, you get the lengths of polygon sides in **feet** but upon completion of the polygon the size is shown in **square meters** (see below).

<img width="536" height="1170" alt="Image" src="https://github.com/user-attachments/assets/bb16a6b2-1983-43f6-949c-856fcf9e599a" />

**Feature request:**
- Please show the polygon size in the typical unit for larger fields, i.e. hectares in metric units, or acres in imperial units.
- Please either make all units depend on the system locale or add a setting in Ground to switch between metric and imperial units (see suggestion below).

<img width="540" height="1170" alt="Image" src="https://github.com/user-attachments/assets/84d31b37-6f7a-4659-90d2-9ff8a2f0ddec" />

This FR relates partially to https://github.com/google/ground-android/issues/3350

PR #3392, opened 2025-12-04
Refactor settings shared preferences access

andreia-ferreira:
Towards #3351 

The `LocalValueStore` was being accessed directly in the `SettingsFragment` which tightly couples the presentation and data layer. In order to solve that, a new use case was introduced `GetUserSettingsUseCase` which provides the current preferences to `SettingsViewModel` which stores the state presented by the view.

@... PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/GroundApplication.kt
+++ b/app/src/main/java/org/groundplatform/android/GroundApplication.kt
@@ -27,7 +27,7 @@ import com.google.firebase.remoteconfig.FirebaseRemoteConfig
 import dagger.hilt.android.HiltAndroidApp
 import javax.inject.Inject
 import org.groundplatform.android.common.Constants.isReleaseBuild
-import org.groundplatform.android.usecases.user.GetUserSettingsUseCase
+import org.groundplatform.android.data.local.LocalValueStore
 import timber.log.Timber
 
 @HiltAndroidApp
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/groundplatform/android/repository/UserRepository.kt
```diff
@@ -120,4 +121,17 @@ constructor(
...
+    val measurementUnits = localValueStore.selectedLengthUnit
+    val shouldUploadPhotosOnWifiOnly =
+      localValueStore.shouldUploadMediaOverUnmeteredConnectionOnly()
+
+    return UserSettings(
+      language = language,
+      measurementUnits = measurementUnits,
+      shouldUploadPhotosOnWifiOnly = shouldUploadPhotosOnWifiOnly,
+    )
+  }
```
shobhitagarwal1612: We can simplify this as,

```
with(localValueStore) {
      UserSettings(
        language = selectedLanguage,
        measurementUnits = selectedLengthUnit,
        shouldUploadPhotosOnWifiOnly = shouldUploadMediaOverUnmeteredConnectionOnly()
      )
    }
```

andreia-ferreira: Applied your suggestion, thanks!

app/src/main/java/org/groundplatform/android/ui/settings/SettingsFragment.kt
```diff
@@ -101,8 +124,6 @@ class SettingsFragment : PreferenceFragmentCompat(), Preference.OnPreferenceClic
   }
 
   private fun updateLocaleAndRestart(languageCode: String) {
```
shobhitagarwal1612: Should we rename this function to `restartApp` since the `updateLocale` is already taken care of in `refreshUserPreferences` call?

andreia-ferreira: good point. The method still does apply the locale to the app though, with `AppCompatDelegate.setApplicationLocales(appLocale)`, so I've updated the name to `applyLocaleAndRestart`, hopefully it reflects the behavior a bit better

PR #3405, opened 2025-12-13
Refactor MOG providers to use dependency injection

shobhitagarwal1612:
This PR aims at using dependency injection for providing remote tile providers instead of manually building the classes at multiple locations.

Verified for regressions by running the app locally and testing offline areas 

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/test/java/org/groundplatform/android/ui/map/gms/mog/MogSourceProviderTest.kt
+++ b/app/src/test/java/org/groundplatform/android/ui/map/gms/mog/MogSourceProviderTest.kt
@@ -17,10 +17,7 @@ package org.groundplatform.android.ui.map.gms.mog
 
 import com.google.common.truth.Truth.assertThat
 import org.junit.Test
-import org.junit.runner.RunWith
-import org.robolectric.RobolectricTestRunner
 
-@RunWith(RobolectricTestRunner::class)
 class MogSourceProviderTest {
 
   @Test


app/src/main/java/org/groundplatform/android/ui/map/gms/mog/MogClient.kt
```diff
@@ -237,11 +236,3 @@ class MogClient(
       }
   }
 }
-
-private inline fun <T> nullIfError(fn: () -> T) =
```
andreia-ferreira, line 241: won't removing this helper method and its usage increase the probability of having the app crash in cases of network errors?

app/src/main/java/org/groundplatform/android/ui/offlineareas/selector/OfflineAreaSelectorFragment.kt
```diff
@@ -133,7 +134,7 @@ class OfflineAreaSelectorFragment : AbstractMapContainerFragment() {
     viewLifecycleOwner.lifecycleScope.launch {
       mapContainerViewModel.mapLoiFeatures.collect { map.setFeatures(it) }
     }
-    map.addTileOverlay(viewModel.remoteTileSource)
+    map.addTileOverlay(RemoteMogTileSource)
```
andreia-ferreira: nit: this bypasses the ViewModel role to determine what data should be represented. Even though it's just a simple data object I believe keeping the variable in the VM as it was keeps the code more testable and easier to switch/mock sources if needed

app/src/test/java/org/groundplatform/android/ui/map/gms/mog/MogSourceProviderTest.kt
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.groundplatform.android.ui.map.gms.mog
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
andreia-ferreira: if I understand correctly, these unit tests don't use anything from the Android framework so they don't need to run with the Robolectric. Using the default JUnit runner would make the tests faster and platform agnostic in this case

app/src/main/java/org/groundplatform/android/ui/map/gms/mog/MogClient.kt
shobhitagarwal1612: This helper method is being used with objects which return non-nullable instance. So, I feel this logic is redundant and was added for preventive reasons rather than based on actual error handling. So, if we do see any errors, we should handle them on a case-by-case basic rather than consuming all errors at once, which potentially means consuming critical errors too. Happy to discuss further!

app/src/main/java/org/groundplatform/android/ui/offlineareas/selector/OfflineAreaSelectorFragment.kt
shobhitagarwal1612: Done

app/src/test/java/org/groundplatform/android/ui/map/gms/mog/MogSourceProviderTest.kt
shobhitagarwal1612: Done. Thanks!

app/src/main/java/org/groundplatform/android/ui/map/gms/mog/MogClient.kt
andreia-ferreira: I see. I do agree that errors should be handled case-by-case instead of a 'catch all' kind of way, thanks for explaining

Issue #3365, opened 2025-11-21
[Code health] Setup CI/CD for e2e testing and automated deployment


andreia-ferreira:
In order to streamline the release process it would be ideal to implement a CI/CD workflow that triggers a new release and publishes the app on the Google Play Console automatically. However, in order to achieve that, a E2E test flow must be in place and run before each release.

PR #3406, opened 2025-12-15
Cleanup and optimize dependencies

andreia-ferreira:
Towards #3365 

In order to update the current E2E test the code needs to take into account Compose elements that were added in the meantime. However, adding the dependency and the composeTestRule in the `e2eTest` module would cause a `NoSuchMethodError` crash. 

This PR fixes that by replacing `androidx.fragment:fragment-testing` with `fragment-testing-manifest`. This ensures the tests continue to have the necessary components while avoiding dependency conflicts and runtime crashes.

Additionally a couple of improvements were done:
- Use BOM for compose, coroutines, and serialization so that all related libraries are aligned to compatible versions, reducing the risk of dependency conflicts and simplifying version management
- Remove unused dependencies 

In order to make sure that no issues were introduced by this, a general sanity check of the app was performed.

@shobhitagarwal1612  PTAL?

### DIFF
--- a/e2eTest/src/main/java/org/groundplatform/android/e2etest/SurveyRunnerTest.kt
+++ b/e2eTest/src/main/java/org/groundplatform/android/e2etest/SurveyRunnerTest.kt
@@ -20,14 +20,11 @@ import android.widget.Button
 import android.widget.CheckBox
 import android.widget.EditText
 import android.widget.RadioButton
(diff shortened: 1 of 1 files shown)


shobhitagarwal1612:
Can you please update the description to specify if any regression / sanity testing was performed?

PR #3408, opened 2025-12-16
Fix possible crash in areas that don't have offline imagery

andreia-ferreira:
When trying to download offline imagery for an area that doesn't have this data, the app crashed with [StorageException](https://console.firebase.google.com/u/0/project/ground-dev-sig/crashlytics/app/android:org.groundplatform.android/issues/e3304389425320ba8191ff4be41862b5?time=24h&types=crash&sessionEventKey=6941329201A300011EC210EF697E71DD_2162844563414382503). In order to fix that, error handling was added to `MogPathOrUrl.toUrl()`, which was removed in the [previous PR](https://github.com/google/ground-android/pull/3405).

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/map/gms/mog/MogClient.kt
```diff
@@ -216,10 +217,15 @@ class MogClient(
 
   private suspend fun MogPathOrUrl.toUrl(): MogUrl? =
     if (startsWith("/")) {
-      remoteStorageManager.getDownloadUrl(this).toString()
+      try {
+        remoteStorageManager.getDownloadUrl(this).toString()
+      } catch (_: StorageException) {
+        Timber.w("File not found for path: $this")
```
shobhitagarwal1612: Can we log the exception stacktrace along with the error?

andreia-ferreira: sure, done :+1:

shobhitagarwal1612 (approved):

PR #3414, opened 2025-12-19
Add E2E test to fill in a survey with all task types

andreia-ferreira:
Towards #3365 

This PR focuses on implementing a E2E test that covers the full flow of filling a survey, including all supported task types.
The previous test implementation was relying too much on UIAutomator, which can make the tests slow and flaky. To address this, this PR introduces a new test architecture centered around a TestDriver abstraction and the Robot pattern

The following was implemented:
- [x] Introduced a `TestDriver` interface that defines the supported user interactions which the implementation then define how to perform. In this case, for the Android tests it will prioritize Espresso and ComposeTestRule, using UIAutomator only when necessary (system interactions, maps etc)
- [x] Added the `Robot` base class and its implementations to encapsulate the interactions specific for each page
- [x] Updated `SurveyRunnerTest` with the new pattern
- [x] Cover the full survey flow with tests

**How to test**
You can run the test locally by using the `localDebug` build variant while running the Firebase emulator on the ground-platform project. The current emulator data doesn't have the survey used in this test, you need to use the one which will be added in this PR: https://github.com/google/ground-platform/pull/2327

[Screen_recording_20251222_182518.webm](https://github.com/user-attachments/assets/afea46b0-526d-4ddb-ad97-e49b3f0f6bab)

@shobhitagarwal1612 @gino-m PTAL?

### DIFF
--- a/e2eTest/src/main/java/org/groundplatform/android/e2etest/TestConfig.kt
+++ b/e2eTest/src/main/java/org/groundplatform/android/e2etest/TestConfig.kt
@@ -18,8 +18,7 @@ package org.groundplatform.android.e2etest
 import org.groundplatform.android.model.task.Task
 
 object TestConfig {
-  const val LONG_TIMEOUT = 30000L
-  const val SHORT_TIMEOUT = 10000L
+  const val DEFAULT_TIMEOUT = 10000L
   const val SURVEY_NAME = "Ground app E2E test"
   const val TEST_JOB_ALL_TASK_TYPES_EXCEPT_DRAW_AREA = "Test all task types except draw area"
   val TEST_LIST_ALL_TASK_TYPES_EXCEPT_DRAW_AREA =
--- a/e2eTest/src/main/java/org/groundplatform/android/e2etest/drivers/AndroidTestDriver.kt
+++ b/e2eTest/src/main/java/org/groundplatform/android/e2etest/drivers/AndroidTestDriver.kt
@@ -22,6 +22,7 @@ import androidx.compose.ui.test.ExperimentalTestApi
 import androidx.compose.ui.test.hasContentDescription
 import androidx.compose.ui.test.hasTestTag
 import androidx.compose.ui.test.hasText
+import androidx.compose.ui.test.isEnabled
 import androidx.compose.ui.test.junit4.AndroidComposeTestRule
 import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performScrollTo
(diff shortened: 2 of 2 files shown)


gino-m:
This is excellent! How would this flow actually get triggered from GitHub Actions? I assume it would work against a local emulated db populated by the web test interactions?

andreia-ferreira:
@gino-m I haven't updated the GitHub actions in this MR yet, I will do it in a follow up.

After adding these tests, I was thinking on updating and reusing the existing `submit-test` action in the Android repo to run them. As part of that, instead of relying on data produced by the web e2e tests, the action would run against Firebase emulator data pre-populated in [this PR](https://github.com/google/ground-platform/pull/2327). 

This removes a dependency on the web e2e, which I wasn’t able to run locally, and makes the Android tests more deterministic and easier to extend. Using pre-populated emulator data lets us test against a more varied and representative survey configuration (including all task types) without having to modify or re-run web e2e flows that currently would only create a single-task survey.

Let me know what you think, happy to discuss further

shobhitagarwal1612 (approved):
🤖 🧪

gino-m:
> @gino-m I haven't updated the GitHub actions in this MR yet, I will do it in a follow up.
> 
> After adding these tests, I was thinking on updating and reusing the existing `submit-test` action in the Android repo to run them. As part of that, instead of relying on data produced by the web e2e tests, the action would run against Firebase emulator data pre-populated in [this PR](https://github.com/google/ground-platform/pull/2327).
> 
> This removes a dependency on the web e2e, which I wasn’t able to run locally, and makes the Android tests more deterministic and easier to extend. Using pre-populated emulator data lets us test against a more varied and representative survey configuration (including all task types) without having to modify or re-run web e2e flows that currently would only create a single-task survey.
> 
> Let me know what you think, happy to discuss further

Thanks for the context! @sufyanAbbasi and I had originally considered this, but decided against it since using data produced by the web app itself would help us identify regressions due to drift between the web and Android app persisted data models. This is a significant source of risk, since there are no automated tests to check for alignment between the two clients. Our proto-based pseudo-schema mitigates this someone, but it is purely declarative, and also doesn't test alignment of the proto (de)serializers themselves.

I think ideally we need to run both web and Android tests when either changes to ensure changes to neither breaks the other. If this takes too long to run on PR, maybe we can only run them on Merge to master?

andreia-ferreira:
Running these tests on push to master is indeed the best approach since the Android emulator can be quite heavy.

I agree that having the web app generate and verify the data on each run would be the ideal way to catch drift. The [data which would be used](https://github.com/google/ground-platform/pull/2327) for these android tests was originally produced by running the current web app against the Firebase emulator, creating a survey (manually), and exporting the emulator data. Even though it's static for now, it represents the real data structure and lets us verify that the android app handles it correctly while also catching potential regressions.

Maybe we could start with with this approach and then, if there's capacity on the web side, evolving the setup by updating the web E2E tests so that it correctly creates a complex survey as test data instead of using the static one. What do you think @gino-m ?

PR #3417, opened 2025-12-22
Migrate SignInScreen to Jetpack Compose

shobhitagarwal1612:
Towards #1795

Migrates the XML based layout of sign-in screen to Jetpack Compose. This is in preparation for KMP migration.

image

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
@@ -121,6 +121,6 @@ private fun GoogleSignInButton(modifier: Modifier = Modifier, onClick: () -> Uni
 @Preview(showBackground = true)
 @Composable
 @ExcludeFromJacocoGeneratedReport
-private fun SignInScreenPreview() {
+fun SignInScreenPreview() {
   SignInScreen(onSignInClick = {})
 }


app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
```diff
@@ -0,0 +1,123 @@
...
+@Composable
+private fun GoogleSignInButton(onClick: () -> Unit) {
+  AndroidView(
+    modifier = Modifier.wrapContentSize(),
+    factory = { context -> SignInButton(context).apply { setSize(SignInButton.SIZE_WIDE) } },
+    update = { button -> button.setOnClickListener { onClick() } },
+  )
+}
+
+@Preview(showBackground = true)
```
anandwana001, line 121: do we add ignore preview annotations?

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
```diff
@@ -0,0 +1,123 @@
...
+      verticalArrangement = Arrangement.SpaceAround,
+    ) {
+      LogoAndTitle()
+      GoogleSignInButton { onSignInClick() }
+    }
+  }
+}
+
+@Composable
+private fun BackgroundOverlay() {
```
andreia-ferreira: nit: it’s a good practice to include an optional `Modifier` parameter in composables to promote flexibility. It would be nice to have it added to the components in this file

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
```diff
@@ -0,0 +1,123 @@
...
+  AndroidView(
+    modifier = Modifier.wrapContentSize(),
+    factory = { context -> SignInButton(context).apply { setSize(SignInButton.SIZE_WIDE) } },
+    update = { button -> button.setOnClickListener { onClick() } },
+  )
+}
+
+@Preview(showBackground = true)
+@Composable
+fun SignInScreenPreview() {
```
andreia-ferreira: nit: it might be good to keep previews private to reduce the risks of accidentally using these components

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
shobhitagarwal1612: Done. Thanks!

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
shobhitagarwal1612: `detekt` plugin failed with `unused method` error when making it private. We'll have to config the plugin and then make this change across all compose components.

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
andreia-ferreira: good point. I have added a new issue for that, thanks

PR #3420, opened 2025-12-23
Update detekt to allow private previews

### LINKED ISSUE #3418: [Code health] Update detekt and set Compose previews as private
Compose previews are important helpers for development purposes only. Keeping them private would make this more explicit and help prevent their accidental use as real components.

Right now, if we make Previews private, the `detekt` task fails with `UnusedPrivateMember`, so the plugin needs to be updated before making them private across the codebase.

andreia-ferreira:
Fixes #3418 

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (approved):
Thanks for the quick fix!

PR #3429, opened 2026-01-05
Update github actions for E2E test

andreia-ferreira:
Towards #3365 

This PR updates the existing GitHub action files for running the Android tests. 
The workflow defined in `test-e2e.yml` will run on every push to master, but it can also be triggered manually if needed. This workflow triggers the `start-emulator` action of `ground-platform` in order to start the Firebase emulator and then the `submit-test` on `ground-android` in order to start the Android emulator and run the tests in the `e2eTest` module.

While testing this workflow some changes had to be implemented in order to make the tests less flaky:
- Replacing the logic to open the phone camera with Espresso intents. A new image was added to the assets in order to mock the flow
- Remove the step to complete a 'draw area job'. The current implementation to drag the map and add points seems to be flaky so this was left out for now and can be improved and re-introduced later. At the moment all task types are being tested except for this one.

You can see how the workflow run would look like here: https://github.com/andreia-ferreira/ground-android/actions/runs/21182506715

PTAL @shobhitagarwal1612 @gino-m

### DIFF
--- a/.github/workflows/todo-to-issue.yml
+++ b/.github/workflows/todo-to-issue.yml
@@ -38,15 +38,21 @@ jobs:
           INSERT_ISSUE_URLS: "true"
           CLOSE_ISSUES: "false"
       - name: Set Git user
+        # Use env to pass values from workflow expressions into the shell safely.
+        env:
+          ACTOR: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


gino-m (approved):
Excellent progress! 🚀

PR #3430, opened 2026-01-07
Fix cluster sites not rendering

### LINKED ISSUE #3387: [Data collection] Some Predefined sample sites don't render
Created this survey with 25 pre-defined sites: https://ground.openforis.org/en/survey/au4T5os5gRtyOEMGCYnX
Sites: https://drive.google.com/file/d/1pPd3T5FXUjVAqKdAFGquuT2yY0TfUUzh/view?usp=sharing (no sensitive data).

In Build 0.3.1 - 2025-12-01, zooming in from the green grouping dots, some of the polygons are properly shown, but for some, only the grouping dot stays and doesn't change and the desired polygon does not appear.

https://github.com/user-attachments/assets/ab8f42ce-75bb-498f-a9dc-f0670dea851f

andreia-ferreira:
Fixes #3387 

There was an issue when zooming in on map clusters, where the individual items wouldn't render. This happened when the clusters themselves didn't change, but the zoom crossed the threshold where individual items should be displayed.

In order to fix that, a new override was added `shouldRender` which forces the rerender in this particular case, when the clusters don't change but the rendering type did.

**Before fix**

[Screen_recording_20260107_175849.webm](https://github.com/user-attachments/assets/7fef23c2-e6e0-4cad-b25e-3a565a3d3812)

**With fix**

[Screen_recording_20260107_175947.webm](https://github.com/user-attachments/assets/16c3e814-53ac-4934-bed1-b7605e253d06)

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureClusterRenderer.kt
```diff
@@ -101,8 +108,33 @@ class FeatureClusterRenderer(
...
+    oldClusters: Set<Cluster<FeatureClusterItem?>?>,
+    newClusters: Set<Cluster<FeatureClusterItem?>?>,
+  ): Boolean =
+    if (hasRenderingModeChanged()) {
+      true
+    } else {
+      super.shouldRender(oldClusters, newClusters)
+    }
+
+  private fun hasRenderingModeChanged(): Boolean = isClusterMode(oldZoom) != isClusterMode(zoom)
```
shobhitagarwal1612, line 137: Should we update the value of `oldZoom` in this method after it has returned true?

andreia-ferreira: that value is updated by `setZoom`, which is called by each `onCameraIdle`. Then rendering related methods, such as this one, only consume that state to decide whether a re-render is needed

shobhitagarwal1612: I was just wondering if there's a way to optimize things by re-rendering the clusters even if the zoom hasn't changed.

andreia-ferreira: zoom crossing the threshold is the only case where a re-render without cluster changes is forced, because if zoom hasn't changed and clusters are the same, re-rendering wouldn't provide a benefit and would be less optimal. If there's something I'm missing let me now and I can revisit this :+1:

Issue #3416, opened 2025-12-21
Add support for iOS devices using Kotlin Multiplatform (KMP)


shobhitagarwal1612:
### Summary

This issue proposes expanding the "Ground" Android application to iOS using Kotlin Multiplatform (KMP). This will allow the team to preserve a majority of existing business logic while meeting the high-performance requirements for geospatial rendering.

### Key Highlights

- Room Persistence: Room now offers KMP support, allowing for the reuse of existing database schemas and DAO interfaces across platforms.
- Dependency Injection: A migration from Hilt to Koin is necessary, as Hilt’s tight coupling with Android components renders it incompatible with KMP.
- UI Framework: To enable cross-platform UI sharing, all remaining XML layouts must be converted to Jetpack Compose.
- Setup XCode project and iOS infrastructure.

Detailed tasks are captured in [this doc](https://docs.google.com/document/d/18ix0KVdrywUbWN_9r0m2OusvSTusAm2ishaqKwRRysQ/r/0-nVoy4XTsCLmh_4qAt2zV5w/edit?tab=t.0#heading=h.ydot0uxxpdpx).

PR #3461, opened 2026-01-18
Migrate remaining logic from StartupFragment to compose and view models

shobhitagarwal1612:

This PR aims to shift all non-Android logic, including screen generation, over to Jetpack Compose. This change is a step towards enabling Kotlin Multiplatform (KMP) support (as detailed in issue #3416) by removing the reliance on UI Views for iOS. While the long-term plan is to replace Fragments with Jetpack Navigation Compose, this specific PR focuses solely on the logic migration to Compose. The full switch to Navigation Compose will occur in a single update once every screen has been successfully migrated.

Verified that the functionality doesn't change by running the app locally. Also added test coverage for newly added compose screen and view models.

We've replaced the native progress dialog with a Jetpack Compose version, resulting in a slight visual change. Other screens will be updated to use this new dialog over time for consistency.

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/map/gms/GoogleMapsFragment.kt
@@ -179,8 +179,8 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
       isIndoorLevelPickerEnabled = false
     }
 
-    if (isAdded && view != null) {
-      viewLifecycleOwner.lifecycleScope.launch {
+    viewLifecycleOwner.lifecycleScope.launch {
+      viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
         featureManager.markerClicks.collect { featureClicks.emit(setOf(it)) }
       }
     }
@@ -249,7 +249,7 @@ class GoogleMapsFragment : SupportMapFragment(), MapFragment {
     val cameraPosition = map.cameraPosition
     val projection = map.projection
 
-    featureManager.zoom = map.cameraPosition.zoom
+    featureManager.setZoom(map.cameraPosition.zoom)
     featureManager.onCameraIdle()
 
     lifecycleScope.launch {
--- a/app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureClusterRenderer.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/map/gms/features/FeatureClusterRenderer.kt
@@ -41,7 +41,7 @@ class FeatureClusterRenderer(
   context: Context,
   map: GoogleMap,
   clusterManager: ClusterManager<FeatureClusterItem>,
-  var zoom: Float,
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/groundplatform/android/ui/components/LoadingDialog.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Dialog
+import org.groundplatform.android.R
+import org.groundplatform.android.ui.common.ExcludeFromJacocoGeneratedReport
+import org.groundplatform.android.ui.theme.AppTheme
+
+@Composable
+fun LoadingDialog(messageId: Int) {
```
andreia-ferreira: nit: as a general best practice composables should have a default modifier in the params

app/src/main/java/org/groundplatform/android/ui/components/LoadingDialog.kt
```diff
@@ -0,0 +1,64 @@
...
+  Dialog(onDismissRequest = {}) {
+    Surface(
+      shape = MaterialTheme.shapes.medium,
+      color = MaterialTheme.colorScheme.surface,
+      tonalElevation = 4.dp,
+    ) {
+      Row(modifier = Modifier.padding(24.dp), verticalAlignment = Alignment.CenterVertically) {
+        CircularProgressIndicator(
+          modifier = Modifier.size(36.dp),
```
andreia-ferreira: maybe we can add these properties (`size` and `strokeWidth`) to `Size.kt`? I believe there are other loading components across the app so having these in a place where they can be reused could be handy

app/src/main/java/org/groundplatform/android/ui/startup/StartupViewModel.kt
```diff
@@ -15,23 +15,58 @@
...
+        _state.value = StartupState.Error(t)
+      }
+    }
+  }
+
+  /** Displays an error popup if the error is related to Google Play Services. */
+  private fun maybeDisplayError(throwable: Throwable) {
+    if (throwable is GooglePlayServicesNotAvailableException) {
+      popups.ErrorPopup().show(R.string.google_api_install_failed)
+    }
```
andreia-ferreira: the ViewModels should emit events, not display UI. I believe this logic fits better in the Fragment where `onLoadFailed` is handled.

app/src/main/java/org/groundplatform/android/ui/components/LoadingDialog.kt
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/startup/StartupViewModel.kt
shobhitagarwal1612: Done

andreia-ferreira (approved):

PR #3462, opened 2026-01-19
Move logic from SignInFragment to SignInScreen

shobhitagarwal1612:

Similar to https://github.com/google/ground-android/pull/3461, this PR aims at moving remaining business logic from the fragment to it's compose screen. 

This would allow us to later switch to Jetpack Compose Navigation to allow for screens to be reused for iOS without having to maintain separate view screens for iOS.

[Screen_recording_20260121_105250.webm](https://github.com/user-attachments/assets/a7b367eb-2c43-42a5-81a1-67ee3aa0f019)

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/test/java/org/groundplatform/android/ui/main/MainActivityTest.kt
+++ b/app/src/test/java/org/groundplatform/android/ui/main/MainActivityTest.kt
@@ -34,7 +34,6 @@ import org.groundplatform.android.system.auth.FakeAuthenticationManager
 import org.groundplatform.android.system.auth.SignInState
 import org.junit.Test
 import org.junit.runner.RunWith
-import org.mockito.Mockito
 import org.robolectric.Robolectric
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.Shadows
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
```diff
@@ -110,17 +154,54 @@ private fun LogoAndTitle(modifier: Modifier = Modifier) {
...
+) {
   AndroidView(
-    modifier = modifier.wrapContentSize().testTag(BUTTON_TEST_TAG),
-    factory = { context -> SignInButton(context).apply { setSize(SignInButton.SIZE_WIDE) } },
-    update = { button -> button.setOnClickListener { onClick() } },
+    modifier =
+      modifier.wrapContentSize().testTag(BUTTON_TEST_TAG).semantics {
+        // Allows UI tests written with the Compose testing framework to accurately check the
+        // button's state using semantic matchers like assertIsDisabled(). Otherwise, testing
+        // framework cannot check the state of a View embedded within AndroidView.
```
andreia-ferreira: suggestion: maybe we can just convert the whole button to compose to simplify testing?

shobhitagarwal1612: Thanks for the suggestion! Done.

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
```diff
@@ -110,17 +155,59 @@ private fun LogoAndTitle(modifier: Modifier = Modifier) {
...
-    factory = { context -> SignInButton(context).apply { setSize(SignInButton.SIZE_WIDE) } },
-    update = { button -> button.setOnClickListener { onClick() } },
-  )
+    colors =
+      ButtonDefaults.buttonColors(
+        containerColor = Color.White,
+        contentColor = Color.DarkGray,
+        disabledContainerColor = Color.LightGray,
+        disabledContentColor = Color.Gray,
+      ),
```
anandwana001, line 173: most of our button themes are going to be same, so we can keep such themes at a single place and make it usable by everyone.

app/src/main/res/values/strings.xml
```diff
@@ -59,6 +59,7 @@
   <string name="no_basemaps_downloaded">No map imagery downloaded for offline use</string>
   <string name="connect_to_download_message">You are offline. Connect to download offline map imager</string>
   <string name="signing_in">Signing in</string>
+  <string name="sign_in_with_google">Sign in with Google</string>
```
andreia-ferreira, line 62: could you also add the strings for the other languages? We can reuse the same copy that was shown by the previous button in the different languages

shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/signin/SignInScreen.kt
shobhitagarwal1612: I checked other buttons in the applications and each button had it's own color palette. For e.g. this one doesn't use primary colors for the icon tint and uses white color for background instead of lighter shade of green. So, we can't reuse this particular combination for other buttons.

app/src/main/res/values/strings.xml
andreia-ferreira: thanks!

PR #3473, opened 2026-01-23
Move sign-in error handling to SignInScreen

shobhitagarwal1612:

Migrates the logic related to handling sign-in errors from `MainViewModel` to `SignInScreen`.

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/main/MainViewModel.kt
```diff
@@ -85,37 +82,19 @@ constructor(
...
-
   private fun onUserSignedOut(): MainUiState {
     // Scope of subscription is until view model is cleared. Dispose it manually otherwise, firebase
     // attempts to maintain a connection even after user has logged out and throws an error.
     viewModelScope.launch { withContext(ioDispatcher) { clearUserSessionUseCase() } }
     return MainUiState.OnUserSignedOut
   }
 
-  private suspend fun onUserSignedIn(user: User): MainUiState =
+  private suspend fun onUserSignedIn(user: User): MainUiState? =
```
andreia-ferreira: nit: this should still be non-nullable in this method right?

app/src/main/java/org/groundplatform/android/ui/signin/SignInViewModel.kt
```diff
@@ -56,4 +56,8 @@ internal constructor(networkManager: NetworkManager, private val userRepository:
   fun onSignInButtonClick() {
     userRepository.signIn()
   }
+
+  fun onSignOutButtonClick() {
+    userRepository.signOut()
+  }
```
andreia-ferreira, line 62: perhaps a good moment to add unit tests for this VM

app/src/main/java/org/groundplatform/android/ui/main/MainViewModel.kt
shobhitagarwal1612: Good catch! Fixed.

app/src/main/java/org/groundplatform/android/ui/signin/SignInViewModel.kt
shobhitagarwal1612: Done

PR #3478, opened 2026-01-26
Fix 'Other' text field rendering issue

### LINKED ISSUE #3428: Writing in "Other" rendering problem
Build 0.3.3: 

When a single or multiple choice question has the option "Other" with possibility to type, on some phones the rendering isn't working properly, see video for better understanding:

https://github.com/user-attachments/assets/e21f3bab-20df-4e15-a0ca-879f41f36cdb

This happened on multiple devices, my Samsung Galaxy S24 included.

andreia-ferreira:
Fixes #3428 

This PR addresses an issue where the screen would bounce up and down when the user typed in the “Other” input field of multiple-choice tasks on small screens.

The screen is a mixed XML + Compose layout. The logic to render the buttons in the bottom (`TaskButton`) is complex and consists on dynamically adding and removing buttons based on a task state. Since the height of the parent layout was `wrap_content`, the fact that writing in the text field changed the 'done' button state,  probably caused the container height to fluctuate and the bouncy behavior to occur.

This PR adds:
- fixed height for the layout that wraps the task buttons in `ActionButtonsRow`
- adjustments on the modifiers used in `MultipleChoiceItemView` since there was some mixup between the parent and children modifiers - tests were added for this composable to make sure no regressions were introduced

As a follow up, the overall setup of the task buttons is going to be refactored in https://github.com/google/ground-android/issues/3437 and some content will be refactored fully to compose to provide a cleaner fix in the long term

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
@@ -250,7 +250,10 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
 
   @Composable
   private fun ActionButtonsRow() {
-    Row(modifier = Modifier.fillMaxWidth(), horizontalArrangement = Arrangement.SpaceBetween) {
+    Row(
+      modifier = Modifier.fillMaxWidth().height(48.dp),
+      horizontalArrangement = Arrangement.SpaceBetween,
+    ) {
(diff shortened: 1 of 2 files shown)


shobhitagarwal1612:
One test is failing.

```
CaptureLocationTaskFragmentTest > capture button disabled when accuracy is poor FAILED
    java.lang.AssertionError at CaptureLocationTaskFragmentTest.kt:166
```

andreia-ferreira:
setting the height in `task_frag_action_buttons` seemed to interfere with the capture location accuracy warning popup so I moved the height parameter to the `ActionButtonsRow` composable and everything seems ok

PR #3486, opened 2026-01-28
Add google-services.json for debug configuration

### LINKED ISSUE #3485: [Code health] Check in `google-services.json` for dev server
Firebase can only be accessed by builds by devs whose encryption key is set in Firebase. Checking this file in would simply dev setup.

From https://groups.google.com/g/firebase-talk/c/bamCgTDajkw/m/uVEJXjtiBwAJ :

> ...the google-services.json is safe to check in to your repo and is something that should be shared among engineers on your team.  The JSON file does not contain any super-sensitive information (like a server API key).  It does contain some information like your database URL, Android API key, and storage bucket.  These are not secrets, but if your security rules are not set up correctly an attacker could use them against you.  However since these values are also compiled into your APK as resources, the theoretical attacker would not need your JSON file to get these values anyway.

gino-m:
Fixes #3485

### DIFF
--- a/app/src/debug/dev/google-services.json
+++ b/app/src/debug/dev/google-services.json
@@ -1,0 +1,127 @@
+{
+  "project_info": {
+    "project_number": "429183886479",
+    "project_id": "ground-dev-sig",
+    "storage_bucket": "ground-dev-sig.appspot.com"
+  },
+  "client": [
+    {
+      "client_info": {
+        "mobilesdk_app_id": "1:429183886479:android:17703fcaf0c28800293cfa",
+        "android_client_info": {
+          "package_name": "org.groundplatform.android"
+        }
+      },
+      "oauth_client": [
+        {
+          "client_id": "429183886479-3q7cs6dj8fv9shc9krad227d4is0j0uj.apps.googleusercontent.com",
+          "client_type": 1,
+          "android_info": {
+            "package_name": "org.groundplatform.android",
+            "certificate_hash": "af19263d66b6227d8a72338e0e40995355172b94"
+          }
+        },
+        {
+          "client_id": "429183886479-68js6ddvgob6eihpqihd4ms4hncl4j3m.apps.googleusercontent.com",
+          "client_type": 1,
+          "android_info": {
+            "package_name": "org.groundplatform.android",
+            "certificate_hash": "38c1bbbdb050f57f20f854c8f0abef0ea2080082"
+          }
+        },
+        {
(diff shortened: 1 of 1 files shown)


andreia-ferreira:
I believe this would be fine for a private or internal repo, but is it appropriate to add it to an open-source one?

The message you linked in the issue also mentions:
> For a library or open-source sample we do not include the JSON file because the intention is that users insert their own to point the code to their own backend

Which aligns with this project's contributing guidelines regarding [firebase](https://github.com/google/ground-android/blob/master/CONTRIBUTING.md#set-up-firebase) as well

gino-m:
> I believe this would be fine for a private or internal repo, but is it appropriate to add it to an open-source one?
> 
> The message you linked in the issue also mentions:
> 
> > For a library or open-source sample we do not include the JSON file because the intention is that users insert their own to point the code to their own backend
> 
> Which aligns with this project's contributing guidelines regarding [firebase](https://github.com/google/ground-android/blob/master/CONTRIBUTING.md#set-up-firebase) as well

I had replied but my comment was either lost or added to the wrong PR (?). 

I was saying that these values are shared in the APK and web js unencrypted, so they're essentially public. The reason they're secure is not because we keep them secret, but because only apps signed with the current encryption key can access the backend with these keys. 

Most of the Ground developers are internal and have access to the staging backend, so it seemed to make sense to reduce friction for them. That said, if you feel strongly we can abandon this PR!

andreia-ferreira:
Thanks for the context, I understand the friction concern. I don't feel strongly enough to block this, but wanted to understand the workflow better. 
IIUC devs still need to add their SHA fingerprint in Firebase to get a working config. If they're already going through that step, how much friction does committing the file actually save? Won't the flow still have to be:
Go to Firebase -> add SHA -> re-download google-services.json to get the updated version with their SHA

gino-m:
> Thanks for the context, I understand the friction concern. I don't feel strongly enough to block this, but wanted to understand the workflow better. IIUC devs still need to add their SHA fingerprint in Firebase to get a working config. If they're already going through that step, how much friction does committing the file actually save? Won't the flow still have to be: Go to Firebase -> add SHA -> re-download google-services.json to get the updated version with their SHA

This file doesn't change - adding their SHA fingerprint allows them to built a binary capable of actually using this file. 

That said, Gemini has a point here:

> No, it is not standard or recommended practice to check google-services.json (or GoogleService-Info.plist for iOS) into a public repository, even for staging projects.
> 
> While the data inside these files is technically "public" (since it ends up compiled into your app anyway), exposing it in a repository significantly lowers the barrier for abuse.
> 
> Here is a breakdown of why this is risky and how to handle it securely for a staging environment.

I'll close this.

PR #3501, opened 2026-01-30
Add new components and state management for task action buttons

andreia-ferreira:
Part 1/2 towards #3437 

This PR prepares for the refactor of TaskButton to a composable with correct state management. This refactor moves button state logic from the Fragment to the ViewModel and introduces new components which will be replacing the current implementation in a follow up PR.

Main changes:
- Added `taskActionButtonStates` to `AbstractTaskViewModel` to reactively expose the list of buttons and their states based on task data
- Created `TaskFooter` and `TaskButton` composables to render the button list declaratively, replacing the current imperative  implementation
- Centralized button click logic in `AbstractTaskViewModel.onButtonClick`, with overrides in child ViewModels for task-specific actions
- Added comprehensive unit tests for button states and click handling across all task ViewModels

Overall this PR only introduces the new components and its tests. They are only going to be implemented in a next PR 
image

@shobhitagarwal1612 PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskPositionInterface.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskPositionInterface.kt
@@ -25,7 +25,7 @@ import org.groundplatform.android.model.submission.TaskData
  * buttons) without requiring direct access to the full task list or sequencing logic.
  */
 interface TaskPositionInterface {
-  /** Returns true if the task is the first in the sequence */
+  /** Returns true if the task is the first in the sequence. */
   fun isFirst(): Boolean
 
   /**


app/src/main/java/org/groundplatform/android/ui/datacollection/components/refactor/TaskButton.kt
```diff
@@ -0,0 +1,109 @@
...
+  AppTheme {
+    Column(modifier = Modifier.padding(16.dp), verticalArrangement = Arrangement.spacedBy(8.dp)) {
+      ButtonAction.entries.forEach { action ->
+        Row(horizontalArrangement = Arrangement.spacedBy(8.dp)) {
+          TaskButton(state = ButtonActionState(action), onClick = {})
+          TaskButton(state = ButtonActionState(action), onClick = {})
+        }
+      }
+    }
+  }
```
shobhitagarwal1612, line 109: Instead of adding each button twice, let's have a single column for better visualization.

```
AppTheme {
    Column(
      modifier = Modifier.padding(16.dp),
      verticalArrangement = Arrangement.spacedBy(8.dp),
      horizontalAlignment = Alignment.CenterHorizontally,
    ) {
      ButtonAction.entries.forEach { action ->
        TaskButton(state = ButtonActionState(action), onClick = {})
      }
    }
  }
```

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/instruction/InstructionTaskViewModel.kt
```diff
@@ -16,7 +16,12 @@
...
+import org.groundplatform.android.model.submission.TaskData
+import org.groundplatform.android.ui.datacollection.components.refactor.ButtonActionState
 import org.groundplatform.android.ui.datacollection.tasks.AbstractTaskViewModel
 
-@Suppress("EmptyClassBlock")
-class InstructionTaskViewModel @Inject constructor() : AbstractTaskViewModel() {}
+class InstructionTaskViewModel @Inject constructor() : AbstractTaskViewModel() {
+
+  override fun getButtonStates(taskData: TaskData?): List<ButtonActionState> =
+    listOf(getPreviousButton(), getNextButton(taskData))
```
shobhitagarwal1612: Comparing this logic with the current `InstructionTaskFragment`, the previous buttons were PREVIOUS (from AbstractTaskFragment), SKIP and NEXT. Can you please confirm if the current change is intentional?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskPositionInterface.kt
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.ui.datacollection.tasks
+
+import org.groundplatform.android.model.submission.TaskData
+
+interface TaskPositionInterface {
```
shobhitagarwal1612, line 27: Please add ktdoc

app/src/main/java/org/groundplatform/android/ui/datacollection/components/refactor/TaskButton.kt
andreia-ferreira: good point, done

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/instruction/InstructionTaskViewModel.kt
andreia-ferreira: The current `InstructionTaskFragment` only displays PREVIOUS and NEXT. The `AbstractTaskFragment` adds PREVIOUS to all tasks by default and then some also override `onCreateActionButtons` to add only specific buttons except that one, as is the case for this fragment:
```
  override fun onCreateActionButtons() {
    addButton(ButtonAction.NEXT).setOnClickListener { moveToNext() }
  }
```

I did notice that the NEXT button should always be enabled for the Instructions task, which I hadn't taken into account, so I updated that

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskPositionInterface.kt
andreia-ferreira: sure! Done

gino-m:
@andreia-ferreira Can you paste in a screen recording of these elements for reference?

shobhitagarwal1612 (approved):

PR #3508, opened 2026-02-01
Migrate MapType bottom sheet dialog to Jetpack Compose 

### LINKED ISSUE #3507: MapType Dialog

shobhitagarwal1612:
Fixes #3507

- Creates compose screen for the xml layout
- Migrate logic from fragment to compose screen
- Remove nav_graph binding and directly show the bottom sheet dialog without using fragment

[Before](https://github.com/user-attachments/assets/c335a069-42fd-41f7-ace1-e69a24102899)

[After](https://github.com/user-attachments/assets/9ae7bcf0-79c4-4835-84cc-0ab69f7f89c6)

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
@@ -236,7 +236,7 @@ private fun MapType.labelResId(): Int =
 @Preview(showBackground = true)
 @Composable
 @ExcludeFromJacocoGeneratedReport
-fun BasemapSelectorScreenPreview() {
+private fun BasemapSelectorScreenPreview() {
   AppTheme {
     BasemapSelectorContent(
       mapTypes = listOf(MapType.ROAD, MapType.TERRAIN, MapType.SATELLITE),


app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
```diff
@@ -0,0 +1,249 @@
...
+  Text(
+    text = stringResource(R.string.base_map),
+    style = MaterialTheme.typography.titleMedium,
+    color = MaterialTheme.colorScheme.onSurface,
+    modifier = Modifier.padding(top = 27.dp),
+  )
+
+  LazyRow(
+    modifier = Modifier.fillMaxWidth().padding(top = 11.dp, bottom = 14.dp),
+    horizontalArrangement = Arrangement.spacedBy(16.dp),
```
andreia-ferreira: comparing with the old layout, it looks like the spacing between the items should be bigger. This is more noticeable in medium/big screens 

since we're just displaying 3 items here, maybe we can use a Row and set its children with a Modifier.weight(1f) in order to have the items occupy the entire available space?

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
```diff
@@ -0,0 +1,249 @@
...
+    Column(modifier = Modifier.weight(1f)) {
+      Text(
+        text = stringResource(R.string.offline_map_imagery),
+        style = MaterialTheme.typography.titleMedium,
+        color = MaterialTheme.colorScheme.onSurface,
+      )
+      Text(
+        text = stringResource(R.string.offline_map_imagery_pref_description),
+        style = MaterialTheme.typography.bodySmall,
+        color = MaterialTheme.colorScheme.secondary,
```
andreia-ferreira, line 176: this text was previously the same color as the rest of the content, is this change of color intended?

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
```diff
@@ -0,0 +1,249 @@
...
+  when (this) {
+    MapType.ROAD -> R.string.road_map
+    MapType.TERRAIN -> R.string.terrain
+    MapType.SATELLITE -> R.string.satellite
+  }
+
+@Preview(showBackground = true)
+@Composable
+@ExcludeFromJacocoGeneratedReport
+fun BasemapSelectorScreenPreview() {
```
andreia-ferreira: preview can be private

app/src/test/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreenTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@HiltAndroidTest
+@RunWith(RobolectricTestRunner::class)
+@Config(qualifiers = "w600dp-h1024dp")
+class BasemapSelectorScreenTest : BaseHiltTest() {
+
+  @Inject lateinit var mapStateRepository: MapStateRepository
+
+  private lateinit var viewModel: BasemapSelectorViewModel
```
andreia-ferreira, line 45: suggestion: currently these tests mix UI and ViewModel assertions. We can consider extracting ViewModel tests to a dedicated BasemapSelectorViewModelTest. But since that VM is pretty small feel free to resolve this comment if you feel like the extra test file is not worth it

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
shobhitagarwal1612: Used `SpaceBetween` to distribute the items as it was happening in the original xml layout.

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
shobhitagarwal1612: If you look at the xml layout, the color scheme was set to `secondary` there as well. I think the kotlin theme class and the xml themes have diverged causing these discrepencies. Given rest of the app is already using the kotlin theme, would you be ok with keeping this UX change? If not, we can fix it too.

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
shobhitagarwal1612: Done

app/src/test/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreenTest.kt
shobhitagarwal1612: I tried moving this to a separate class but it duplicates some tests between the two classes. Resolving the comment

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
andreia-ferreira: nice one :+1:

app/src/main/java/org/groundplatform/android/ui/basemapselector/BasemapSelectorScreen.kt
andreia-ferreira: ah i see... I think it's not very noticeable so it's ok, if there are complaints it's a quick fix anyway :smile:

PR #3530, opened 2026-02-05
Fix socket timeout crash

### LINKED ISSUE #3527: [Trending crash]  InstrURLConnectionBase.getResponseCode SocketTimeoutException
Only seems to happen since release 0.3.6: https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/c5b240dce1598f829a1fa56e661ba8de?time=90d&sessionEventKey=698331E800AE00016C6750D05E0EB70F_2181421495330704690

This can be reproduced by trying to download a offline area with a bad network connection

andreia-ferreira:
Fixes #3527 

This fixes the trending crash in the latest release where, if trying to move the camera around while finding an area to download imagery with bad internet, the app would crash on timeout.

The network exceptions are now handled in `OfflineAreaRepository` and the `OfflineAreaSelectorViewModel` shows an error message for those cases. A missing PT translation was added for this.
This logic in the viewmodel is now covered with unit tests

@shobhitagarwal1612 @hassan-nsubuga PTAL?


app/src/main/java/org/groundplatform/android/ui/offlineareas/selector/OfflineAreaSelectorViewModel.kt
shobhitagarwal1612, line 164: Would it be better if we wrap `updateDownloadSize(bounds)` call with a `catching` block and handle the errors here?

```
 runCatching { updateDownloadSize(bounds) }.onFailure { onUnavailableAreaSelected() }
```

Then we won't have to worry about catching those errors and converting to `Result` in the repository or within `updateDownloadSize`. Thoughts?

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/offlineareas/selector/OfflineAreaSelectorViewModel.kt
andreia-ferreira: that was my first though too, but I think it's more the repository's job to catch the errors thrown by the client and translate them into Result types. This way the VM doesn't need to know about IOExceptions or other data layer implementation details, keeping boundaries cleaner

PR #3538, opened 2026-02-06
Update Gradle to version 9.0.0 

shobhitagarwal1612:
Bump `gradleVersion` from 8.13.2 to 9.0.0 in `libs.versions.toml`.

These changes were proposed by AGP (gradle plugin) assistant on android studio.

Changes for migrating to built-in kotlin support will be taken up separately https://github.com/google/ground-android/issues/3539

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/test/java/org/groundplatform/android/HiltExt.kt
+++ b/app/src/test/java/org/groundplatform/android/HiltExt.kt
@@ -20,7 +20,6 @@ import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
 import androidx.annotation.IdRes
-import androidx.annotation.StyleRes
 import androidx.core.os.bundleOf
 import androidx.core.util.Preconditions
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -27,4 +27,14 @@ org.gradle.jvmargs=-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError
...
+android.defaults.buildfeatures.resvalues=true
+android.sdk.defaultTargetSdkToCompileSdkIfUnset=false
+android.enableAppCompileTimeRClass=false
+android.usesSdkInManifest.disallowed=false
+android.uniquePackageNames=false
+android.dependency.useConstraints=true
+android.r8.strictFullModeForKeepRules=false
+android.r8.optimizedResourceShrinking=false
+android.builtInKotlin=false
+android.newDsl=false
```
andreia-ferreira: are both of these covered by #3539 ?

gradle.properties
```diff
@@ -27,4 +27,14 @@ org.gradle.jvmargs=-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError
...
\ No newline at end of file
+android.nonFinalResIds=false
+android.defaults.buildfeatures.resvalues=true
+android.sdk.defaultTargetSdkToCompileSdkIfUnset=false
+android.enableAppCompileTimeRClass=false
+android.usesSdkInManifest.disallowed=false
+android.uniquePackageNames=false
+android.dependency.useConstraints=true
+android.r8.strictFullModeForKeepRules=false
+android.r8.optimizedResourceShrinking=false
```
andreia-ferreira: most of these flags seem to be deprecated and show up as deprecation warnings when building the project

These were auto-added by the AGP assistant, but we may not actually need all of them if the project is already compliant. Thinking for example if we need `uniquePackageNames=false` or `defaultTargetSdkToCompileSdkIfUnset=false`

I suggest to try removing these flags and if any of them cause the project to stop compiling, we can restore that specific flag and create a follow-up issue to address it properly. What do you think?

PR #3540, opened 2026-02-11
Remove resources from OfflineAreaSelectorViewModel

andreia-ferreira:
`OfflineAreaSelectorViewModel` depended on `Resources` to resolve string resources directly, creating a tight coupling with the Android framework

This PR removes that dependency by introducing a new sealed class `BottomTextState` so that the VM emits semantic UI state and the Fragment handles string resolution. During this, the databinding was also removed from the xml.

This should help decouple the VM from the Android framework and improve testability.

@shobhitagarwal1612 PTAL?


shobhitagarwal1612 (approved):
🧹 ✨

PR #3548, opened 2026-02-13
Remove Glide library

shobhitagarwal1612:
All usages of image loading have been migrated from Glide to Coil. This PR removes remaining usages and gradle dep.


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -74,13 +76,12 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
...
-    taskBinding.fragment = this
-    taskBinding.viewModel = viewModel
+  override fun onCreateTaskBody(inflater: LayoutInflater): View = createComposeView {
+    val isPhotoPresent by viewModel.isPhotoPresent.observeAsState(false)
+    val uri by viewModel.uri.observeAsState()
     homeScreenViewModel = getViewModel(HomeScreenViewModel::class.java)
-    return taskBinding.root
+
+    PhotoTaskScreen(isPhotoPresent = isPhotoPresent, uri = uri, onTakePhoto = { onTakePhoto() })
   }
```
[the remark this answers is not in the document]
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/datacollection/components/UriImage.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.core.net.toUri
+import coil.compose.AsyncImage
+import coil.request.ImageRequest
+import coil.size.Scale
+import org.groundplatform.android.R
+import org.groundplatform.android.ui.common.ExcludeFromJacocoGeneratedReport
+import org.groundplatform.android.ui.theme.AppTheme
+
+const val MAX_IMAGE_SIZE = 2048
```
andreia-ferreira: nit: can be private or `@VisibleForTesting`

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragment.kt
```diff
@@ -75,12 +77,13 @@ class PhotoTaskFragment : AbstractTaskFragment<PhotoTaskViewModel>() {
     TaskViewFactory.createWithHeader(inflater)
 
   override fun onCreateTaskBody(inflater: LayoutInflater): View {
-    val taskBinding = PhotoTaskFragBinding.inflate(inflater)
-    taskBinding.lifecycleOwner = this
-    taskBinding.fragment = this
-    taskBinding.viewModel = viewModel
     homeScreenViewModel = getViewModel(HomeScreenViewModel::class.java)
-    return taskBinding.root
+    return createComposeView {
```
andreia-ferreira, line 81: it looks like this is missing setting the ViewCompositionStrategy

e2eTest/src/main/java/org/groundplatform/android/e2etest/drivers/AndroidTestDriver.kt
```diff
@@ -148,7 +148,7 @@ class AndroidTestDriver(
       Instrumentation.ActivityResult(Activity.RESULT_OK, Intent())
     }
 
-    click(TestDriver.Target.ViewId(R.id.btn_camera))
+    device.findObject(By.text("Camera")).click()
```
andreia-ferreira: we should prefer the composeRule for the app interactions since it's faster and more reliable, and use UI automator matching for system-level interactions or elements that are hard to access

shobhitagarwal1612:
I'm not able to figure out why this change is causing MainActivity test to fail. Will try splitting the changes into smaller parts.

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskFragmentTest.kt
```diff
@@ -137,7 +135,7 @@ class PhotoTaskFragmentTest : BaseTaskFragmentTest<PhotoTaskFragment, PhotoTaskV
   fun `taking photo sends intent`() {
     setupTaskFragment<PhotoTaskFragment>(job, task)
 
-    onView(withId(R.id.btn_camera)).perform(click())
+    composeTestRule.onNodeWithText("Camera").performClick()
```
andreia-ferreira, line 138: I think the problem affecting the pipeline might be here. This test file should have this override :
`@get:Rule override val composeTestRule = createAndroidComposeRule<ComponentActivity>()`

otherwise it will use the default one from `BaseHiltTest` (createComposeRule) that doesn't launch a real activity so any tests that work with navigation like this one need to override that. I'm now doubting if that rule should even be in that base test class, will investigate it a bit more

shobhitagarwal1612: Resolving as the original code has been submitted without any issues, so I suspect the issue is associated with glide cleanup

PR #3552, opened 2026-02-14
[Refactor] Convert date task fragment to compose

shobhitagarwal1612:
Towards #3549

Before:

[date task - compose - before.webm](https://github.com/user-attachments/assets/120d29e1-a1ca-4d6e-8926-3764fce1c32c)

After: 

[date task - compose - after.webm](https://github.com/user-attachments/assets/07986f42-652e-4e5e-a5a2-aa14e038326d)

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
@@ -72,7 +72,7 @@ fun DateTaskScreen(
 @Preview(showBackground = true)
 @Composable
 @ExcludeFromJacocoGeneratedReport
-fun DateTaskScreenPreview() {
+private fun DateTaskScreenPreview() {
   AppTheme {
     Column(
       modifier = Modifier.padding(16.dp),


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
```diff
@@ -0,0 +1,85 @@
...
+    // TODO: Replace with simple text field.
+    OutlinedTextField(
+      value = dateText,
+      onValueChange = {},
+      readOnly = true,
+      placeholder = { Text(hintText) },
+      modifier = Modifier.width(200.dp).testTag("dateInputText"),
+      interactionSource = interactionSource,
+    )
+  }
```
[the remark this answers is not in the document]
shobhitagarwal1612: This is a wrong suggestion.

In `Jetpack Compose`, `OutlinedTextField` does not have a native `onClick` parameter because it handles its own touch events for focus and text selection. If you add `.clickable` directly to the text field, the modifier will consume the touch event. The `InteractionSource` method allows interactions to pass through to the underlying component while still notifying you.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskFragment.kt
```diff
@@ -20,66 +20,50 @@ import android.content.DialogInterface
...
-          if (taskData != null) {
-            val calendar = Calendar.getInstance()
-            calendar.timeInMillis = taskData.timeInMillis
-            DateFormat.getDateFormat(requireContext()).format(calendar.time)
-          } else {
-            ""
-          }
-        }
-        .asLiveData()
+  override fun onCreateTaskBody(inflater: LayoutInflater): View = createComposeView {
```
andreia-ferreira, line 47: the compose view is missing the view composition strategy

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
```diff
@@ -0,0 +1,87 @@
...
+      modifier = Modifier.width(200.dp).testTag(DATE_TEXT_TEST_TAG),
+      interactionSource = interactionSource,
+    )
+  }
+}
+
+@Preview(showBackground = true)
+@Composable
+@ExcludeFromJacocoGeneratedReport
+fun DateTaskScreenPreview() {
```
andreia-ferreira: nit: can be private

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskFragment.kt
```diff
@@ -20,66 +20,50 @@ import android.content.DialogInterface
...
+        (taskData as? DateTimeTaskData)?.let {
+          DateFormat.getDateFormat(context).format(Date(it.timeInMillis))
+        } ?: ""
       }
-  }
 
-  override fun onCreateTaskView(inflater: LayoutInflater): TaskView =
-    TaskViewFactory.createWithHeader(inflater)
+    val hintText = remember {
+      (DateFormat.getDateFormat(context) as SimpleDateFormat).toPattern().uppercase()
```
andreia-ferreira, line 59: it would be safer to use `as?` here

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
```diff
@@ -0,0 +1,87 @@
...
+  LaunchedEffect(interactionSource) {
+    interactionSource.interactions.collect { interaction ->
+      if (interaction is PressInteraction.Release) {
+        onDateClick()
+      }
+    }
+  }
+
+  Column(modifier = modifier) {
+    // TODO: Replace with simple text field.
```
andreia-ferreira, line 60: is there any blocker on implementing a Text instead of OutlinedTextField in this PR? It could simplify the component a bit and let us use `Modifier.clickable` instead of the interactionSource workaround

shobhitagarwal1612: I added this TODO based on figma mocks. I wanted to make sure that we don't introduce any visual changes along with the compose migration.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskFragment.kt
shobhitagarwal1612: We don't expect `DateFormat.getDateFormat(context) ` to be null.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/date/DateTaskScreen.kt
andreia-ferreira: ah I see. I believe that a Box + Text wouldn't change the visuals, but I might be missing something. If there's a risk that it does differ feel free to resolve the thread, I'm fine keeping this as is :+1:

shobhitagarwal1612: I'll leave it as is and send a separate change to udpate the UX. Thanks for brining it up!

PR #3572, opened 2026-02-20
Cleanup test setup

andreia-ferreira:
`BaseHiltTest` currently holds a `composeTestRule`, which adds unnecessary overhead to tests that don't need to test anything related to the UI. That may have a slight impact on runtime for non-UI tests but also it may make it harder for future modularization.

This PR refactors this setup by:
- removing the `composeTestRule` from the base test class and moved it only to tests that actually need it
- adding `TestResourcesHelper` to provide resources without requiring `createAndroidComposeRule`
- cleaned up `BaseHiltTest` usage to decouple it from tests that don't need it (eg. pure compose tests)

@shobhitagarwal1612 PTAL?

### DIFF
--- a/.github/workflows/todo-to-issue.yml
+++ b/.github/workflows/todo-to-issue.yml
@@ -44,7 +44,7 @@ jobs:
           HEAD_REF: ${{ github.head_ref }}
         run: |
           git config --global user.name "$ACTOR"
-          git config --global user.email "$(git log -n 1 --pretty=format:%ae remotes/origin/$HEAD_REF)"
+          git config --global user.email "$(git log -n 1 --pretty=format:%ae -- "remotes/origin/$HEAD_REF")"
       - name: Commit and Push Changes
         env:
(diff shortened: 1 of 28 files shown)


andreia-ferreira:
Thanks! On local tests runs I only noticed a slight decrease in the average test runtime (around 10s or so), so nothing dramatic :) 

However, I'm marking this PR as a draft for now because I noticed the changes might introduce some flakiness in certain tests which I want to investigate better

andreia-ferreira:
after https://github.com/google/ground-android/pull/3580 this PR is now safe to merge. Since those changes removed the static dispatcher from `TestCoroutineDispatchersModule` , leftover tasks or uncaught exceptions from previous tests no longer persist and there's less risk of cross-test state leaks

The flakiness only surfaced after this decoupling refactor, which likely changed enough about the test setup to make the pre-existing issue visible. After the dispatchers fix that flakiness doesn't occur anymore

PR #3574, opened 2026-02-20
Refactor settings screen to compose

### LINKED ISSUE #3573: Convert Settings Activity and Fragment to Compose

shobhitagarwal1612:
Fixes #3573

Before:

[settings-activity-compose-before.webm](https://github.com/user-attachments/assets/46ddd985-dde1-43d4-9124-d01b9154990f)

After:

[Screen_recording_20260227_102932.webm](https://github.com/user-attachments/assets/6b07b658-7184-4950-b865-ebf10f77c605)

@andreia-ferreira  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/settings/SettingsViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/settings/SettingsViewModel.kt
@@ -33,7 +33,7 @@ internal constructor(
   private val updateUserSettingsUseCase: UpdateUserSettingsUseCase,
 ) : AbstractViewModel() {
 
-  private var _uiState: MutableStateFlow<UserSettings> = MutableStateFlow(getUserSettingsUseCase())
+  private val _uiState: MutableStateFlow<UserSettings> = MutableStateFlow(getUserSettingsUseCase())
   val uiState: StateFlow<UserSettings> = _uiState
 
   private fun updateState(newUiState: UserSettings) {


app/src/main/java/org/groundplatform/android/ui/settings/SettingsViewModel.kt
```diff
@@ -15,31 +15,42 @@
...
 import org.groundplatform.android.model.settings.UserSettings
+import org.groundplatform.android.repository.UserRepository
 import org.groundplatform.android.ui.common.AbstractViewModel
-import org.groundplatform.android.usecases.user.GetUserSettingsUseCase
 
+@HiltViewModel
 class SettingsViewModel
 @Inject
-internal constructor(private val getUserSettingsUseCase: GetUserSettingsUseCase) :
+internal constructor(private val localValueStore: LocalValueStore, userRepository: UserRepository) :
```
andreia-ferreira: this simplifies getting/setting preferences, but it tightly couples the VM to `LocalValueStore`, a data-layer class. 

Previously we only used `GetUserSettingsUseCase` for reading and the xml preferences did its magic to write them. Now that we need to write from the VM, I'd suggest creating a `UpdateUserSettingsUseCase` domain class. Not only that brings us closer to clean architecture but also keeps the VM agnostic of Android-specific storage implementations. What do you think?

app/src/test/java/org/groundplatform/android/ui/settings/SettingsScreenTest.kt
```diff
@@ -0,0 +1,201 @@
...
+          onUploadMediaOverUnmeteredConnectionOnlyChange = {},
+          onLanguageChange = {},
+          onMeasurementUnitsChange = {},
+          onVisitWebsiteClick = {},
+          onBack = {},
+        )
+      }
+    }
+
+    composeTestRule.onNodeWithText(context.getString(R.string.general_title)).assertIsDisplayed()
```
andreia-ferreira: suggestion: you can use the extension getString (from `TestResourcesHelper`) to fetch string resources without explicitly use context

app/src/main/java/org/groundplatform/android/ui/settings/components/SettingsSelectItem.kt
```diff
@@ -0,0 +1,84 @@
...
+
+  val allOptions =
+    remember(configuration) {
+      val labels = resources.getStringArray(entriesResId)
+      val values = resources.getStringArray(entryValues)
+      labels.zip(values) { label, value -> Option(label, value) }
+    }
+
+  val selectedOption = allOptions.find { it.value == currentValue } ?: allOptions.firstOrNull()
+  var showUnitDialog by remember { mutableStateOf(false) }
```
andreia-ferreira: this now shows the dialog for both units and language preferences (which was previously just showing a dropdown list). Is it intended? If so maybe we can keep this naming more generic

app/src/main/java/org/groundplatform/android/ui/settings/SettingsViewModel.kt
shobhitagarwal1612: That's a great suggestion. Done.

app/src/test/java/org/groundplatform/android/ui/settings/SettingsScreenTest.kt
shobhitagarwal1612: Done

app/src/main/java/org/groundplatform/android/ui/settings/components/SettingsSelectItem.kt
shobhitagarwal1612: Replaced the dialog with dropdown.

shobhitagarwal1612:
@andreia-ferreira PTAL? I've updated the screencast in the description as well.

app/src/main/java/org/groundplatform/android/ui/settings/SettingsViewModel.kt
```diff
@@ -15,31 +15,41 @@
...
-internal constructor(private val getUserSettingsUseCase: GetUserSettingsUseCase) :
-  AbstractViewModel() {
+internal constructor(
+  getUserSettingsUseCase: GetUserSettingsUseCase,
+  private val updateUserSettingsUseCase: UpdateUserSettingsUseCase,
+) : AbstractViewModel() {
 
-  private val _uiState: MutableStateFlow<UserSettings?> = MutableStateFlow(null)
-  val uiState: StateFlow<UserSettings?> = _uiState
+  private var _uiState: MutableStateFlow<UserSettings> = MutableStateFlow(getUserSettingsUseCase())
```
andreia-ferreira: this can be a `val` I believe
```suggestion
  private val _uiState: MutableStateFlow<UserSettings> = MutableStateFlow(getUserSettingsUseCase())
```

andreia-ferreira (approved):
awesome stuff 🚀 only a minor suggestion

PR #3579, opened 2026-02-23
Test/3481/add conditional tests

### LINKED ISSUE #3481: [Code health] Add conditionals to e2e test
At the moment the `CompleteAllTaskTypesTest` doesn't cover conditional logic. It would be great to add some conditional questions to the test survey and add tests for them in order to prevent regressions

hassan-nsubuga:
Fixes #3481 

The CompleteAllTaskTypesTest didn't cover conditional logic. Add a conditional question to the test survey and add tests for it in order to prevent regressions

[x] Add conditional question to the test survey
[x] Refactor AndroidTestDriver add conditional test

@... PTAL?


e2eTest/src/main/java/org/groundplatform/android/e2etest/drivers/AndroidTestDriver.kt
```diff
@@ -88,15 +95,32 @@ class AndroidTestDriver(
...
+        composeRule.onTarget(target, index).performClick()
+        composeRule.onNodeWithText(NEXT_BUTTON_TEXT).assertExists().performClick()
+        composeRule.onNodeWithText(ARABICA_TEXT).assertDoesNotExist()
+        composeRule.onNodeWithText(COVER_CROPPING_TEXT).assertExists().performClick()
+        composeRule.onNodeWithText(PREVIOUS_BUTTON_TEXT).assertExists().performClick()
+      } else {
+        composeRule.onTarget(target, index).performClick()
+      }
     }
   }
```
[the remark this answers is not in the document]
andreia-ferreira: agree with this suggestion. This could be better placed in `DataCollectionRobot`, inside the `multipleChoiceTask`, which would trigger the sequence of events if it verifies it's a conditional TestTask.

This would require creating a new method in `TestDriver` (eg. assertVisibility) to verify if something is shown or not

e2eTest/src/main/java/org/groundplatform/android/e2etest/robots/DataCollectionRobot.kt
```diff
@@ -109,17 +115,26 @@ class DataCollectionRobot(override val testDriver: TestDriver) : Robot<DataColle
...
-  private fun multipleChoiceTask(selectIndexes: List<Int>) {
-    if (selectIndexes.size == 1) {
-      testDriver.selectFromList(
-        TestDriver.Target.TestTag(SELECT_MULTIPLE_RADIO_TEST_TAG),
-        selectIndexes[0],
-      )
-    } else {
-      selectIndexes.forEach {
-        testDriver.selectFromList(TestDriver.Target.TestTag(SELECT_MULTIPLE_CHECKBOX_TEST_TAG), it)
+  private fun multipleChoiceTask(selectIndexes: List<Int>, isConditional: Boolean = false) {
```
andreia-ferreira: to test the conditional questions, we don’t actually use the `selectIndexes` parameter, the logic follows a different path. Because of that, the current function signature can be a bit misleading. I'd suggest to add class that represents the kind of multiple choice question being handled, and pass that as a parameter instead. Something like::
```
sealed class MultipleChoiceType {
  data class Regular(val selectIndexes: List<Int>): MultipleChoiceType()
  data object Conditional: MultipleChoiceType()
}
```
then the function would have a MultipleChoiceType in the arguments instead of selectIndexes + isConditional

e2eTest/src/main/java/org/groundplatform/android/e2etest/robots/DataCollectionRobot.kt
```diff
@@ -138,4 +153,20 @@ class DataCollectionRobot(override val testDriver: TestDriver) : Robot<DataColle
   private fun captureLocationTask() {
     testDriver.click(TestDriver.Target.Text(testDriver.getStringResource(R.string.capture)))
   }
+
+  private fun conditionalTask(target: TestDriver.Target) {
+    if (target is TestDriver.Target.TestTag && target.tag == SELECT_MULTIPLE_RADIO_TEST_TAG) {
+
+      testDriver.assertVisible(ARABICA_TEXT)
+      testDriver.click(TestDriver.Target.Text(COFFEE_TEXT))
```
andreia-ferreira: the constants used in the test could have more descriptive names to better understand why the sequence of actions goes the way it does. 

For example, for the text that needs to be selected to trigger the conditional question, maybe instead of `COFFEE_TEXT` we could name it `CONDITIONAL_TRIGGER_OPTION`. Then, for the option that becomes visible within that conditional question, instead of `ARABICA_TEXT` we can name it `EXPECTED_CONDITIONAL_OPTION` or similar. Same for other specific names bellow, what do you think?

Issue #3600, opened 2026-03-09
Refactor code into pure KMP modules


andreia-ferreira:
Right now most of the code is in the `app` module. By refactoring it into sepparate KMP modules (eg. `core:domain`, `:core:ui`, `:core:data`), we can make the code platform independent and ready to be shared by both Android and iOS.

PR #3585, opened 2026-02-27
Implement sharing geometry with QR code

andreia-ferreira:
Fixes: #3575 

This PR implements a QR component shown at the end of data collection and also later on by clicking an LOI and clicking 'Share'. This component contains the LOI's geometry GeoJSON which can be scanned and read by a QR scanner.

- Implemented `GetLoiReportUseCase` to compute the data for LoiReport, including optimizing the generated geometry by limiting coordinates to a maximum of 6 decimal
- Add `ShareLocationModal` and refactor `DataSubmissionConfirmationScreen` to display the QR code
- Updated `GroundQrCode` with optimizations so that a logo won't be shown if the content is too large

[Screen_recording_20260403_171443.webm](https://github.com/user-attachments/assets/79716893-b5f9-4abb-957c-349026fd1618)

@shobhitagarwal1612 PTAL?

PR #3601, opened 2026-03-09
Move geometry package to core:domain

andreia-ferreira:
Towards #3585 ; #3600 

This PR moves the geometry package to `:core:domain`. These classes will be needed for the new feature to share geometry by QR code, which will have its usecase in the domain module.

Most changes listed are import updates, the only 'real' changes are:
- Move `area` out of `Geometry` and to `GmsExt`, since it relies on google maps logic which is an android dependency at the moment
- Update `centerOrError`  method inside `Geometry` to only use Kotlin logic to achieve the same that was done by the maps-specific logic `LatLngBounds.center()`
- Extract `UiFormatUtil` from `PolygonUtil` (which was moved to domain). `PolygonUtil` is then focused on calculations, while `UiFormatUtil` handles formatting area values to be presented in the UI

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/util/UiFormatUtil.kt
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2025 Google LLC
```
shobhitagarwal1612: nit: 2025 -> 2026

PR #3603, opened 2026-03-10
[Compose] Refactor task body rendering to use Compose

shobhitagarwal1612:
Towards #3602

This is a pure refactor which aims to convert parts of TaskView to composable functions.

Verified for regressions by running the tests locally and collecting data on a local device.

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -100,8 +101,8 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
   /** Creates the view for common task template with/without header. */
   abstract fun onCreateTaskView(inflater: LayoutInflater): TaskView
 
-  /** Creates the view for body of the task. */
-  abstract fun onCreateTaskBody(inflater: LayoutInflater): View
+  /** Renders the body of the task. */
+  @Composable abstract fun RenderTaskBody()
```
andreia-ferreira: nit: composables are usually named as nouns so maybe something like `TaskBody` might sound more natural

Issue #3604, opened 2026-03-10
[Code health] Replace java.util.Date usage in models


andreia-ferreira:
In order to make core domain modules compatible with KMP, usages of the JVM-specific `java.util.Date` should be replaced with a Kotlin alternative (eg. using the `kotlinx.datetime` library).

Issue #3575, opened 2026-02-20
Share geometry across devices and apps


kenstershiro:
Share geometry across devices and apps => QR code containing geometry as GeoJSON
See #3373 for more details

andreia-ferreira:
Figma designs: https://www.figma.com/design/IyR6NuBzH3lsKiaQW9YUso/branch/ZecTJJe3YlejVhu2WW34eo/Ground-Mobile?node-id=12207-8692&p=f&t=BFf7v1oOxYFerD8c-0 specifically in the 'Map' and 'Completion' sections

kenstershiro:
@andreia-ferreira as per the new reqs discussed on Friday this will likely have a dependency on google/ground-platform#2485 now.

kenstershiro:
New requirements we discussed:
- No need for now to have interoperability with the receiving app WeMo (being developed by EFI - European Forestry Institution) , as the app is not live yet. A direct integration could be a future follow up if they are not able to consume the geojson directly
- Making the location available in geosjon format is useful for EUDR, where geojson format is required
- Agreed to provide both formats:
1. A QR code which provides the geojson of the polygon, in text format as the raw geojson, and including the geoid.
2. A link URL to the geoid of the polygon, which can be shared through a share button ([example link](https://data.apps.fao.org/geospatial/v2/api/catalog/features/catalogs/geoid/collections/test_coll/items/893cf758ab2c4ffdafb42994d308835b74c9149b31f633dbc95542726b1b75ec)) (at a minimum a copy action to easily copy and paste it into any app like WhatsApp)
Both options require the device using Ground to have online connectivity (to synch with the server and generate the GeoID - requires  implementation of google/ground-platform#2485 )
The person receiving the share could be offline (but for option 2 they will need to go online to open the URL).
@andreia-ferreira shout if I missed anything. @jo-spek FYI

kenstershiro:
From Gino: dangerous to assume we will always have online connectivity to complete the submission - currently Ground is built around the assumption you don't on the submission screen (you do for example after the synch status screen)
Next steps:
- WHIMO app team need to specify what format and app intent they are expecting:
- geojson, any specific structure?
- universal mime type? Or specific app deeplink? 
- can they just build a QR scanner to their app to read it?
- @jo-spek to set up a call with WHIMO app team to discuss the details

jo-spek:
Suggestion for a QR code holding a GeoJSON: 

[...]

PR #3605, opened 2026-03-10
Add usecase to get Loi reports

andreia-ferreira:
Towards #3575 

This PR adds the use case to fetch a `LoiReport`, which will be initially used to hold the geoJson which needs to be displayed later on in the UI by QR code. This doesn't change any functionality, the use case will be implemented in a follow up 

**Note**: `LocationOfInterest` model should be in the domain module. However, it was not moved at this stage since it still has some dependencies that need a significant refactor (eg. usage of `java.util.Date` - a dedicated issue has been created for it, since it affects quite a lot of files #3604 )

@shobhitagarwal1612  PTAL?

### DIFF
--- a/core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/LoiDataProviderInterface.kt
+++ b/core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/LoiDataProviderInterface.kt
@@ -18,11 +18,13 @@ package org.groundplatform.domain.usecases
 import org.groundplatform.domain.model.geometry.Geometry
 import org.groundplatform.domain.model.locationofinterest.LoiProperties
 
-// TODO: Extract LocationOfInterestRepositoryInterface to domain module
-// This is a temporary abstraction while the repository interface is still not fully decoupled from
-// Android-specific types
+/**
+ * Provides access to the data of a location of interest.
+ *
+ * This is a temporary abstraction while the repository interface is still not fully decoupled from
+ * Android-specific types.
+ */
 interface LoiDataProviderInterface {
-  /** Returns the Geometry for an existing LOI, or null if not found. */
   suspend fun get(surveyId: String, loiId: String): LoiData?
 
   data class LoiData(val geometry: Geometry, val properties: LoiProperties)


core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/GetLoiReportUseCase.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.groundplatform.domain.model.geometry.Geometry
+import org.groundplatform.domain.model.geometry.LineString
+import org.groundplatform.domain.model.geometry.LinearRing
+import org.groundplatform.domain.model.geometry.MultiPolygon
+import org.groundplatform.domain.model.geometry.Point
+import org.groundplatform.domain.model.geometry.Polygon
+import org.groundplatform.domain.model.locationofinterest.LoiProperties
+import org.groundplatform.domain.model.locationofinterest.LoiReport
+
+class GetLoiReportUseCase(private val loiGeometryProvider: LoiDataProviderInterface) {
```
shobhitagarwal1612, line 37: Please add ktdoc.

core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/GetLoiReportUseCase.kt
```diff
@@ -0,0 +1,78 @@
...
+        is LinearRing -> geoJsonObject("LineString", coordinatesToPositions(coordinates))
+        is Polygon -> geoJsonObject("Polygon", polygonToCoordinates(this))
+        is MultiPolygon ->
+          geoJsonObject("MultiPolygon", JsonArray(polygons.map { polygonToCoordinates(it) }))
+      }
+    return JsonObject(
+      mapOf(
+        "type" to JsonPrimitive("Feature"),
+        "properties" to JsonObject(loiProperties.mapValues { JsonPrimitive(it.value.toString()) }),
+        "geometry" to geometryJson,
```
shobhitagarwal1612: Can we declare constants for these keys?
Also, what are other possible values for "type"?

core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/GetLoiReportUseCase.kt
andreia-ferreira: sure, done!

core/domain/src/commonMain/kotlin/org/groundplatform/domain/usecases/GetLoiReportUseCase.kt
andreia-ferreira: Good point. Added constants for all the keys/values. 
Regarding the GeoJSON type, the possible values are `Feature` or `FeatureCollection`. Since this use case deals with a single LOI with one geometry, it only produces a `Feature`

PR #3617, opened 2026-03-17
Migrate TaskView to compose

shobhitagarwal1612:
Fixes https://github.com/google/ground-android/issues/3602

Converts the remaining xml layouts in task view to compose. 

Before: 

[data collection - before.webm](https://github.com/user-attachments/assets/c6f87858-eaa3-437b-9af0-db1982aa9bc3)

After:

[data collection - after.webm](https://github.com/user-attachments/assets/01ebbb80-9c58-405b-9141-92d0c2c07358)

@andreia-ferreira  PTAL?

### DIFF
--- a/.github/actions/submit-test/action.yml
+++ b/.github/actions/submit-test/action.yml
@@ -55,7 +55,7 @@ runs:
           exit 1
         fi
         echo "MAPS_API_KEY is set"
-        echo "MAPS_API_KEY=${{ secrets.MAPS_API_KEY }}" > secrets.properties
+        echo "MAPS_API_KEY=${MAPS_API_KEY}" > secrets.properties
 
     - name: Enable KVM group perms
       shell: bash
@@ -106,11 +106,22 @@ runs:
         if-no-files-found: ignore
 
     - name: Export Firebase emulator data
-      if: always()
       shell: bash
+      if: always()
+      continue-on-error: true
       run: |
         echo "Exporting emulator data..."
         npx -y -p firebase-tools firebase emulators:export ${{ github.workspace }}/test-android-result --project demo-local --force
+
+    - name: Upload Firebase emulator log
+      if: always() && inputs.upload-artifacts == 'true'
+      continue-on-error: true
+      uses: actions/upload-artifact@v6
+      with:
+        name: firebase-emulator-log
+        path: /tmp/firebase-emulator.log
+        retention-days: 7
+        if-no-files-found: warn
 
     - name: Upload Firebase emulator data
       if: always() && inputs.upload-artifacts == 'true'
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -33,15 +33,17 @@ jobs:
 
     # See docs/e2e-testing-doc.md for details on how this is setup
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
-
       - name: Start Firebase emulator
         uses: google/ground-platform/.github/actions/start-emulator@master
+        with:
+          platform-repository: google/ground-platform
+
+      - name: Restore workflow repository
+        uses: actions/checkout@v6
 
       - name: Run Android tests on e2eTest module
-        uses: google/ground-android/.github/actions/submit-test@master
+        uses: ./.github/actions/submit-test
         with:
-          android-repository: ${{ github.repository }}
+          android-repository: google/ground-android
           upload-artifacts: true
           google-maps-key: ${{ secrets.MAPS_API_KEY }}
--- a/app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/common/BindingAdapters.kt
@@ -15,44 +15,14 @@
  */
 package org.groundplatform.android.ui.common
 
-import android.graphics.BitmapFactory
 import android.view.View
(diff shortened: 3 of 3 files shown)


app/src/main/java/org/groundplatform/android/ui/datacollection/components/InstructionsDialog.kt
```diff
@@ -15,50 +15,56 @@
...
-      dismissButton = {
-        OutlinedButton(onClick = { showDialog.value = false }) {
-          Text(text = stringResource(R.string.close))
-        }
-      },
-    )
-  }
+fun InstructionsDialog(
+  @DrawableRes iconId: Int,
+  @StringRes stringId: Int,
```
andreia-ferreira: nit: we could pass `InstructionData` here instead of separate params. It simplifies the call sites slightly

app/src/main/java/org/groundplatform/android/ui/datacollection/components/TaskViewLayout.kt
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+/**
+ * Data class representing the header section of a task view.
+ *
+ * @property label The text to be displayed in the header.
+ * @property iconResId The optional resource ID of the icon to be displayed alongside the label.
+ */
+data class Header(val label: String, val iconResId: Int? = null)
```
andreia-ferreira: nit: the name of the class is a bit ambiguous. Something more specific like `TaskHeader` could help avoid confusion

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
```diff
@@ -107,6 +93,20 @@ class CaptureLocationTaskFragment @Inject constructor() :
...
+    // Ensure that the location lock is enabled, if it hasn't been.
+    if (isVisible) {
+      viewModel.enableLocationLock()
+      lifecycleScope.launch {
+        viewModel.enableLocationLockFlow.collect {
+          if (it == LocationLockEnabledState.NEEDS_ENABLE) {
+            viewModel.showPermissionDeniedDialog.value = true
+          }
+        }
+      }
```
andreia-ferreira, line 110:
```suggestion
      lifecycleScope.launch {
        lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
          viewModel.enableLocationLockFlow.collect {
            if (it == LocationLockEnabledState.NEEDS_ENABLE) {
              viewModel.showPermissionDeniedDialog.value = true
            }
          }
        }
      }
```
improving the flow collection here could avoid potential leaks

app/src/main/java/org/groundplatform/android/ui/datacollection/components/InstructionsDialog.kt
shobhitagarwal1612: done

app/src/main/java/org/groundplatform/android/ui/datacollection/components/TaskViewLayout.kt
shobhitagarwal1612: Done

shobhitagarwal1612:
@andreia-ferreira I've addressed all the comments. PTAL? Also the instrumentation tests are failing during emulator setup phase. How can I run these locally?

andreia-ferreira:
to run them locally, you can follow [these instructions](https://github.com/google/ground-android/blob/master/docs/e2e-testing-doc.md#running-tests-locally), but you need to make sure to use [this branch](https://www.github.com/google/ground-platform/pull/2458) of the ground-platform which contains the fix for the current build failures. Let me know if you face any issues

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskFragment.kt
```diff
@@ -107,6 +95,22 @@ class CaptureLocationTaskFragment @Inject constructor() :
     }
   }
 
+  override fun onTaskResume() {
+    // Ensure that the location lock is enabled, if it hasn't been.
+    if (isVisible) {
+      viewModel.enableLocationLock()
+      lifecycleScope.launch {
+        lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
```
andreia-ferreira: my bad, in the previous suggestion I didn't realize that this was inside `onTaskResume`, so this should be
```suggestion
        lifecycle.repeatOnLifecycle(Lifecycle.State.RESUMED) {
```
otherwise with STARTED the dialog would still show after returning from settings

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -74,6 +74,9 @@ internal constructor(
   private val dataCollectionInitializer: DataCollectionInitializer,
 ) : AbstractViewModel() {
 
+  /** The current vertical position of the task view footer. */
+  var footerVerticalPosition = MutableStateFlow(0.0f)
```
andreia-ferreira: This can be a `val` instead. Also, since this is only mutated through `updateFooterPosition`, consider exposing a `StateFlow` and having a private ` MutableStateFlow` to protect from unintended updates

PR #3629, opened 2026-03-23
Remove proto usage from Survey

### LINKED ISSUE #3622: Refactor Survey model for KMP compatibility
`Survey` currently depends on ProtoBuf generated types.. Not only should Proto types be part of the data layer, the `com.google.protobuf` lib is JVM only, blocking KMP compatibility. 
Replacing these Proto types with plain Kotlin equivalents allows `Survey` to be a proper domain model which can be moved to `:core:domain` module

andreia-ferreira:
Fixes #3622 

This PR removes the Survey direct dependency on protobuf-generated classes by introducing domain types for GeneralAccess, DataVisibility and DataSharingTerms. This is required in order to migrate the domain model to the KMP `:core:domain` module, where protobuf classes (Android/JVM-specific) cannot be referenced.

The main changes are focused on the mappers `ConverterExt` and `SurveyConverter`, where Room and Firebase data is converted to `Survey` and vice versa. A new mapper was created `SurveyProtoToSurveyModel` to help mapping the 3 new domain types. This should not cause any behavior change in the app.

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/test/java/org/groundplatform/android/data/remote/firebase/protobuf/SurveyProtoToSurveyModelTest.kt
+++ b/app/src/test/java/org/groundplatform/android/data/remote/firebase/protobuf/SurveyProtoToSurveyModelTest.kt
@@ -1,0 +1,219 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.groundplatform.android.data.remote.firebase.protobuf
+
+import com.google.common.truth.Truth.assertThat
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/groundplatform/android/data/local/room/converter/ConverterExt.kt
```diff
@@ -407,16 +410,16 @@ fun SurveyEntityAndRelations.toModelObject(): Survey {
...
     surveyEntity.generalAccess.toGeneralAccess(),
   )
 }
 
-fun Int.toGeneralAccess(): org.groundplatform.android.proto.Survey.GeneralAccess =
-  org.groundplatform.android.proto.Survey.GeneralAccess.entries.find { it.number == this }
-    ?: org.groundplatform.android.proto.Survey.GeneralAccess.UNRECOGNIZED
+fun Int.toGeneralAccess(): Survey.GeneralAccess =
+  SurveyProto.GeneralAccess.entries.find { it.number == this }?.toModel()
+    ?: Survey.GeneralAccess.UNRECOGNIZED
```
shobhitagarwal1612, line 422: Should this be `SurveyProto.GeneralAccess.UNRCOGNIZED`?

andreia-ferreira: this method maps the Int stored in Room to the domain model's `Survey.GeneralAccess`, not to the proto type. It first finds the matching `SurveyProto.GeneralAccess` entry by the proto number, then calls `.toModel()` to convert it to the domain enum. If no proto entry matches, it returns the domain fallback `Survey.GeneralAccess.UNRECOGNIZED`

PR #3654, opened 2026-04-03
Migrate PhotoTaskFragment to use TaskScreen Component

shobhitagarwal1612:

- Introduce `PhotoTaskScreen` Composable to handle the UI.
- Update `PhotoTaskFragment` to use `createComposeView` and render the new `PhotoTaskScreen`.
- Move all state logic to `PhotoTaskScreen`
- Replace `PhotoTaskFragmentTest` with `PhotoTaskScreenTest` and add more coverage
- Also removed one extra call to copy the image data from URI.
- Also adds new message strings for unhandled error states.

[Screen_recording_20260403_161428.webm](https://github.com/user-attachments/assets/0770e13f-b051-44c2-988b-671bf1841100)

@andreia-ferreira  PTAL?


shobhitagarwal1612:
@andreia-ferreira Can you PTAL?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskScreen.kt
```diff
@@ -0,0 +1,184 @@
...
+        PhotoTaskError.CAMERA_LAUNCH_FAILED ->
+          Pair(R.string.camera_launch_failed_title, R.string.camera_launch_failed_desc)
+        PhotoTaskError.PHOTO_SAVE_FAILED ->
+          Pair(R.string.photo_save_failed_title, R.string.photo_save_failed_desc)
+      }
+    ConfirmationDialog(
+      title = titleResId,
+      description = messageResId,
+      confirmButtonText = R.string.ok,
+      onConfirmClicked = onDismissError,
```
andreia-ferreira, line 137: this dialog shows ok/cancel but both options have the same outcome (onDismissError). Unless there would be a retry mechanism in place, maybe just the 'ok' option is enough in this case

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -15,46 +15,91 @@
...
-      userMediaRepository.getDownloadUrl(taskData.remoteFilename)
-    } else {
-      Uri.EMPTY
+  fun onTakePhoto() {
+    if (_isAwaitingPhotoCapture.value) return
+    _isAwaitingPhotoCapture.value = true
+    viewModelScope.launch { obtainCapturePhotoPermissions { launchPhotoCapture() } }
+  }
+
+  suspend fun obtainCapturePhotoPermissions(onPermissionsGranted: suspend () -> Unit = {}) {
```
andreia-ferreira: this can be private

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskViewModel.kt
```diff
@@ -66,38 +111,28 @@ class PhotoTaskViewModel @Inject constructor(private val userMediaRepository: Us
...
+    tempPhotoFilePath = null // Clear to avoid reusing stale path
+    viewModelScope.launch {
+      if (result && filePath != null) {
+        finalizePhotoCapture(File(filePath))
+      } else {
+        _events.send(PhotoTaskEvent.ShowError(PhotoTaskError.PHOTO_SAVE_FAILED))
+      }
     }
-    hasLaunchedCamera = false
+    _isAwaitingPhotoCapture.value = false
```
andreia-ferreira: since this is outside the coroutine, it will execute immediately while the result is still being processed. In order to prevent edge cases of inconsistent ui state, it would be safer to move this inside the coroutine

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/photo/PhotoTaskScreen.kt
```diff
@@ -0,0 +1,184 @@
...
+  viewModel: PhotoTaskViewModel,
+  onFooterPositionUpdated: (Float) -> Unit,
+  onAction: (TaskScreenAction) -> Unit,
+  onAwaitingPhotoCapture: (Boolean) -> Unit,
+) {
+  val taskActionButtonsStates by viewModel.taskActionButtonStates.collectAsStateWithLifecycle()
+  val uri by viewModel.uri.collectAsStateWithLifecycle(Uri.EMPTY)
+  val isAwaiting by viewModel.isAwaitingPhotoCapture.collectAsStateWithLifecycle()
+
+  var activeError by remember { mutableStateOf<PhotoTaskEvent.ShowError?>(null) }
```
andreia-ferreira: right now the error dialog will not survive screen rotation. An alternative might be to use `rememberSaveable` and store the enum for the error type instead, since they are saveable by default

andreia-ferreira (commented):
https://github.com/user-attachments/assets/6e4f092b-b3df-4363-a7c3-9e7700a02bab

cancelling the camera intent (by clicking back for example) should not be treated as an error I believe. Also both the error state and taking a picture seem to cause the progress bar to have a wrong position bellow the footer buttons.

Tested on a pixel 9 pro fold, API 36.1

andreia-ferreira (approved):

shobhitagarwal1612:
Not sure why the test is failing on CI? 
@andreia-ferreira any ideas? I'm unable to run these tests locally.

andreia-ferreira:
the tests passed locally for me, so probably the failure was because of a corrupted Gradle cache in the CI. The cache wasn't reused in this latest run (possibly due to invalidation), and the tests are now passing. But if it happens again we can try to delete that cache under actions -> caches

PR #3660, opened 2026-04-08
Fix offline areas crash

### LINKED ISSUE #3407: Crash when downloading offline imagery
There seems to be a crash that happens when downloading offline map imagery. I managed to reproduce it consistently by downloading a first area and then downloading a second one in a row, there may be more ways to reproduce it.
This has happened for some time, although only to a few users: [Crashlytics](https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/92da1cdaec625437f378e41980626258?time=90d&sessionEventKey=691DA4C10285000150B11D9E0D5F2664_2152884184169411536)

https://github.com/user-attachments/assets/2bc3e3a9-cf51-4a38-835f-1ce9d3cdcd0e

andreia-ferreira:
Fixes #3407 ; #3032 ; #3038

When downloading 2 offline areas in a row, the app would [crash](https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/92da1cdaec625437f378e41980626258?time=90d&sessionEventKey=6996C13C005800017252FE2191717BD4_2186933461265149854). This is because the `OfflineAreaSelectorViewModel` was annotated with `@SharedViewModel`, which meant that the `navigate` SharedFlow outlived the fragment and, when a new `OfflineAreaSelectorFragment` was created, it reuses the previous VM. Then, on emiting a `navigate` event, it was collected twice: the current fragment and the one in the backstack - the one which crashed, because the navigation was already on the home screen.

This PR removes the `@SharedViewModel` usage from that VM, since it's only used by a single fragment anyway. Additionally, the flow collectors in the fragment were refactored to prevent memory leaks (using the viewlifecycle owner to launch coroutines and `repeatOnLifecycle(Lifecycle.State.STARTED)`)

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (changes requested):
Can we combine all separate mutable states to a single `UiState`?

andreia-ferreira:
@shobhitagarwal1612 fully agree! I've refactored the VM to unify the overall state under `OfflineAreaSelectorState` and one time events with `OfflineAreaSelectorEvent`. Ready for another look

shobhitagarwal1612 (approved):

PR #3665, opened 2026-04-10
Migrate CaptureLocationTaskFragment to use TaskScreen Component

shobhitagarwal1612:

- Introduce `CaptureLocationTaskScreen` Composable to handle the UI.
- Update `CaptureLocationTaskFragment` to use `createComposeView` and render the new `CaptureLocationTaskScreen`.
- Move all state logic to `CaptureLocationTaskScreen`

Verified for regressions by running the app on emulator and verifying the following scenarios:
 * Opened a capture location task with GPS disabled. Confirmed that the permission dialog is shown. 
 * Opened a capture location task with bad location. Confirmed that the accuracy card is shown. 
 * Opened a capture location task with good location. Confirmed that the accuracy card is not shown. 

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+  onAction: (TaskScreenAction) -> Unit,
+  mapContent: @Composable () -> Unit,
+) {
+  TaskScreen(
+    taskHeader = TaskHeader(taskLabel, R.drawable.outline_pin_drop),
+    taskActionButtonsStates = taskActionButtonsStates,
+    onFooterPositionUpdated = onFooterPositionUpdated,
+    onAction = onAction,
+    shouldShowHeader = true,
+    headerCard = {
```
andreia-ferreira: this is the only task that shows an header card right? In that case I think we could simplify a bit and rely on whether `headerCard` is null or not to determine visibility, we don't need the `shouldShowHeader` variable.

Also small naming suggestion: headerCard sounds like something that would be at the top of the screen. Maybe something like footerHeader would make the purpose clearer

andreia-ferreira (approved):

PR #3667, opened 2026-04-13
Update saveFooterPosition to only update position if the current task is active

andreia-ferreira:
The progress bar was misaligned in some devices, probably due to the fact that, since view pagers preload neighbor fragments, they might also trigger position updates for the footer position. This PR adds a extra check in `saveFooterPosition`, to make sure that only the current active task's footer drives the guideline

Before:

[Screen_recording_20260413_152950.webm](https://github.com/user-attachments/assets/facd0759-fce4-4736-a7f4-05af721dca10)

After:

[Screen_recording_20260413_152459.webm](https://github.com/user-attachments/assets/b3d91438-d209-4e71-9d21-39dd6846f7ea)

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (approved):
🐛 🔨

PR #3670, opened 2026-04-14
Migrate DropPinTaskFragment to use TaskScreen Component

shobhitagarwal1612:

- Introduce `DropPinTaskScreen` Composable to handle the UI.
- Update `DropPinTaskFragment` to use `createComposeView` and render the new `DropPinTaskScreen`.
- Move all state logic to `DropPinTaskScreen`

Verified for regressions by running the app on emulator and verifying the following scenarios:
 * Opens drop pin task for first time, the instructions dialog is shown. Click dismiss.
 * Opens drop pin task for second time, the instructions dialog is now shown.
 * Place marker on any position, click Next. LoiName dialog is shown.
 * LoiName is persisted across config changes and screen navigation

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskMapFragment.kt
```diff
@@ -53,7 +51,7 @@ class DropPinTaskMapFragment @Inject constructor() :
     taskViewModel.updateCameraPosition(position)
   }
 
-  override fun renderFeatures(): LiveData<Set<Feature>> = taskViewModel.features
+  override fun renderFeatures(): LiveData<Set<Feature>> = taskViewModel.features.asLiveData()
```
andreia-ferreira: can this also be refactored to return a flow instead of LiveData? As far as I see, it's used in `onMapReady` on `AbstractTaskMapFragment` which needs to convert it to a flow anyway

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskViewModelTest.kt
```diff
@@ -1,147 +0,0 @@
...
-import org.groundplatform.domain.model.geometry.Point
-import org.groundplatform.domain.model.submission.DropPinTaskData
-import org.groundplatform.domain.model.submission.TaskData
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.robolectric.RobolectricTestRunner
-
-@HiltAndroidTest
-@RunWith(RobolectricTestRunner::class)
-class DropPinTaskViewModelTest : BaseHiltTest() {
```
andreia-ferreira, line 44: was this test file removed because the test cases are covered elsewhere? It's worth considering keeping this and adding tests for what isn't covered yet like the initialize logic or the logic triggered by dismissing the instructions dialog

andreia-ferreira (approved):

PR #3672, opened 2026-04-15
Refactor SurveyRepository

andreia-ferreira:
Towards #3634 

This PR extracts a `SurveyRepositoryInterface` to :core:domain. Other changes include:

- Moved `SurveyListItem` to :core:domain since one of the repo methods returns this type
- Moved `SyncSurveyUseCase` since all of its dependencies are already there it can now safely move to :core:domain
- Added [Kermit](https://kermit.touchlab.co/) to replace Timber usage in `SyncSurveyUseCase`, since it's a KMP logging tool which allows Crashlytics integrations later in the migration
- Created `FakeData` for tests to replace mocking frameworks, enabling simple call tracking and configurable behavior in KMP-compatible tests

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (approved):
Wohoo!!! 🥳

Issue #3679, opened 2026-04-19
[Bug] DrawAreaTask saves duplicate points and allows degenerate polygons


shobhitagarwal1612:
### Description
During a refactor of the `DrawAreaTask` state management to adhere to Unidirectional Data Flow (UDF) principles (https://github.com/google/ground-android/pull/3678), several issues were identified in the polygon drawing logic that could lead to invalid geometry being saved and a confusing user experience.

### Detected Issues

1. **Duplicate Points in Storage:** Incomplete polygons were saving duplicate points (specifically duplicating the last vertex), resulting in 4 points instead of 3 for a triangle.
1. **Incorrect Undo Behavior:** The Undo action would remove a committed vertex even when there was a tentative vertex actively being drawn on the screen.
1. **Missing Validation for Degenerate Shapes:** The proximity check (`isTooClose`) failed to trigger when closing a triangle with only 2 committed points, allowing degenerate lines to be formed.
1. **Premature Button Visibility:** The `COMPLETE` button was displayed for closed shapes that didn't have enough distinct points to form a valid polygon (requires 3 distinct points).

### Proposed Fix
The implementation has been refactored in the current branch to:

- Separate committed vertices from the transient tentative vertex in the ViewModel state.
- Clear the tentative vertex upon commit.
- Enforce a minimum of 3 distinct points for closure validation.
- Update proximity checks to cover all edge cases when closing shapes.
- Introduce `PolygonDrawingSession` class to encapsulate list of committed vertices, transient tentative vertex, redo stack, operations like adding a vertex, undoing, and redoing. 

### Steps to Reproduce

1. Start drawing an area task.
2. Place 3 points.
3. Observe that the saved incomplete geometry contains 4 points instead of 3.

PR #3680, opened 2026-04-19
Refactor Polygon Drawing State into PolygonDrawingSession

shobhitagarwal1612:
Towards #3679

This PR refactors the state management and business logic for the polygon drawing task (`DrawAreaTask`). It extracts the state and associated operations from `DrawAreaTaskViewModel` into a dedicated `PolygonDrawingSession` component. This improves code maintainability, separation of concerns, and testability, adhering better to Unidirectional Data Flow (UDF) principles.

The refactor was completed in two phases:

1. **State Extraction:** Moved the list of vertices and redo stack to a new class.
2. **Logic Encapsulation:** Extracted a public interface, moved snapping logic, self-intersection checks, and completion validation into the session implementation.

The primary goal is to isolate the drawing session's business logic for better testability. Current ViewModel and Fragment tests have become overly complex; this extraction allows for more focused unit testing. Note that this is a pure refactor—bug fixes will be handled in a subsequent PR.

@andreia-ferreira PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSession.kt
```diff
@@ -0,0 +1,108 @@
...
+   */
+  fun commitTentativeVertex(currentCameraTarget: Coordinates?): List<Coordinates>?
+
+  /**
+   * Checks if the current polygon geometry self-intersects. If it does, it may remove the last
+   * vertex to restore validity.
+   *
+   * @return true if self-intersection was detected, false otherwise.
+   */
+  fun checkVertexIntersection(): Boolean
```
andreia-ferreira, line 66: nit: `hasVertexIntersection` might be clearer for a boolean return

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSession.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Validates if the polygon can be completed.
+   *
+   * @return true if valid for completion, false otherwise.
+   */
+  fun validatePolygonCompletion(): Boolean
```
andreia-ferreira: nit: similar to the nit above, `isValidPolygon` might make things clearer here

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSession.kt
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Sets the vertices to a new list.
+   *
+   * @param newVertices The new list of vertices.
+   */
+  fun setVertices(newVertices: List<Coordinates>)
+
+  companion object {
+    /** Min. distance in dp between two points for them be considered as overlapping. */
+    const val DISTANCE_THRESHOLD_DP = 24
```
andreia-ferreira, line 99: this variable is still duplicated in the `DrawAreaTaskViewModel`, can it be removed from there?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSessionImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+    get() = _redoVertexStack
+
+  private var _isTooClose: Boolean = false
+  override val isTooClose: Boolean
+    get() = _isTooClose
+
+  override val hasSelfIntersection: Boolean
+    get() = isSelfIntersecting(_vertices)
+
+  override fun addVertex(vertex: Coordinates, shouldOverwriteLastVertex: Boolean) {
```
andreia-ferreira: this is only used in this class, so it should be private and not part of the interface methods

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -290,79 +258,46 @@ internal constructor(
   @VisibleForTesting
   fun addLastVertex() {
     check(!isMarkedComplete.value) { "Attempted to add last vertex after completing the drawing" }
```
andreia-ferreira: shouldn't `isMarkedComplete` also be part of `PolygonDrawingSession`, just like `isTooClose`?

also, if both variables are in `PolygonDrawingSession`, we can avoid manually synching across every VM method (eg. having `_isTooClose.value = session.isTooClose`, `_isMarkedComplete.value = session.isMarkedComplete`). We can instead create a new data class there to hold both variables and then just add a `syncSessionState()` call after each mutation in the VM.

something like:
```
interface PolygonDrawingSession {
  data class State(val isTooClose: Boolean = false, val isMarkedComplete: Boolean = false)

  val state: State // The impl would build this from its private backing fields (_isTooClose, _isMarkedComplete), so those individual properties no longer need to be exposed on the interface
  ...
}
```
and in the `DrawAreaTaskViewModel`:
```
  private val _sessionState: MutableStateFlow<PolygonDrawingSession.State> = MutableStateFlow(PolygonDrawingSession.State())
  val sessionState: StateFlow<PolygonDrawingSession.State> = _sessionState.asStateFlow()
(...)

  fun syncSessionState() {
    _sessionState.value = session.state
  }

```

what do you think?

shobhitagarwal1612: That's a great suggestion! Done.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSessionImpl.kt
```diff
@@ -0,0 +1,133 @@
...
+
+  private var _vertices: List<Coordinates> = listOf()
+
+  override val vertices: List<Coordinates>
+    get() = _vertices
+
+  private val _redoVertexStack = mutableListOf<Coordinates>()
+
+  internal val redoVertexStack: List<Coordinates>
+    get() = _redoVertexStack
```
andreia-ferreira: as far as I see this is only used for tests. Maybe this can be removed and `_redoVertexStack` can be annotated with `@VisibleForTesting` instead?

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModelTest.kt
```diff
@@ -192,18 +118,6 @@ class DrawAreaTaskViewModelTest : BaseHiltTest() {
...
   }
 
-  @Test
-  fun `Cannot complete polygon when polygon is not complete`() {
-    setupViewModel()
-    updateLastVertexAndAdd(COORDINATE_1)
-    updateLastVertexAndAdd(COORDINATE_2)
-    updateLastVertex(COORDINATE_3, false)
-
-    assertThrows("Polygon is not complete", IllegalStateException::class.java) {
```
andreia-ferreira, line 202: this case doesn't seem to be covered anywhere else so the test should be kept I believe.

but in general it may be worth reconsidering keeping most of the tests in `DrawAreaTaskViewModelTest`  as they assert the task data and geometry output through the VM's public methods. This is important to cover even if the outputs of PolygonDrawingSession are already well tested

andreia-ferreira (approved):

PR #3696, opened 2026-04-23
Migrate DrawAreaTaskFragment to use TaskScreen Component

shobhitagarwal1612:

- Introduce `DrawAreaTaskScreen` Composable to handle the UI.
- Update `DrawAreaTaskFragment` to use `createComposeView` and render the new `DrawAreaTaskScreen`.
- Move all state logic to `DrawAreaTaskScreen`

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModel.kt
```diff
@@ -149,6 +149,12 @@ internal constructor(
       .stateIn(viewModelScope, WhileSubscribed(5_000), emptyList())
   }
 
+  init {
+    if (!instructionsDialogShown) {
+      showInstructions()
+    }
+  }
```
andreia-ferreira: this can cause a bug. Since the data collection flow uses a viewpager with offscreenPageLimit = 1, the adjacent task VMs are initialized, so, in the (probably rare) case where the draw area task wouldn't be the first task,  this init block fires the instructions dialog on the previous task rather than in the draw area screen.

I suggest to move this out of `init` and into a separate method that the `DrawAreaTaskScreen` can trigger only when the task is actually shown, something like:

```
  val lifecycleOwner = LocalLifecycleOwner.current
  LaunchedEffect(lifecycleOwner) {
    lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
      viewModel.checkDrawAreaInstructionsShown()
    }
  }
```

andreia-ferreira: screen recording of the issue:

[Screen_recording_20260424_103539.webm](https://github.com/user-attachments/assets/e24b1872-f44a-46fb-92d1-652d83ea8ce3)

shobhitagarwal1612: Thanks for catching that, Andreia! You're right—the ViewPager pre-loading would definitely have triggered those instructions too early.

I've applied the improvements as suggested and tested them locally using a similar survey structure. Also, do you think we need screenshot tests in the espresso test suite for catching such kinds of UI regressions?

andreia-ferreira: nice, thanks for also fixing the drop pin task! 

Regarding the screenshot tests, I think they'd be great to have, but they probably wouldn't catch this specific issue. The bug wasn't just visual but more about viewpager+lifecycle. Maybe we can add a test in `DataCollectionFragmentTest` though to assert instructions dialogs are not displayed in adjacent tasks?

shobhitagarwal1612: Added test coverage. Also ensured that the test fails if the fixes are reverted.

andreia-ferreira (approved):

PR #3703, opened 2026-04-25
Refactor terms of service screen

shobhitagarwal1612:
Towards #1795

This PR aims to migrate the shared UI logic in the TOS fragment to a compose screen and only keep the remaining android specific elements in the fragment class (to make it KMP compatible). Other changes include applying the best practices (UDF pattern, ui states, event channel, etc.) to the view model.

Verified changes via unit tests and manually running the app.

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/tos/TermsOfServiceScreen.kt
```diff
@@ -0,0 +1,175 @@
...
+    viewModel.events.collect { event ->
+      when (event) {
+        is TosEvent.NavigateToSurveySelector -> onNavigateToSurveySelector()
+        is TosEvent.LoadError -> onLoadError()
+      }
+    }
+  }
+
+  TermsOfServiceContent(
+    isViewOnly = isViewOnly,
```
andreia-ferreira: it would make sense for this to be part of the uiState, I believe. The VM can read the argument from the SavedStateHandle and expose it through `uiState`

shobhitagarwal1612: Done

andreia-ferreira (approved):

PR #3705, opened 2026-04-27
Remove legacy UI code in AbstractTaskFragment after Compose migration

shobhitagarwal1612:
Towards #3630

leans up `AbstractTaskFragment` by deleting the former UI rendering code. This is now dead code since all task implementations use the `TaskScreen` Composable. This change helps in decoupling from Android-specific UI frameworks, supporting the KMP initiative.

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskFragment.kt
```diff
@@ -44,7 +44,10 @@ class DropPinTaskFragment @Inject constructor() : AbstractTaskFragment<DropPinTa
       onFooterPositionUpdated = { saveFooterPosition(it) },
       shouldShowLoiNameDialog = shouldShowLoiNameDialog,
       loiName = loiName,
-      onAction = { action -> handleTaskScreenAction(action) },
+      onButtonClicked = { handleButtonClick(it) },
+      onLoiNameConfirm = { onLoiNameConfirm(it) },
+      onLoiNameDismiss = { onLoiNameDismiss() },
+      onLoiNameChanged = { onLoiNameChanged(it) },
```
andreia-ferreira: suggestion: it looks cleaner to have a dedicated `onButtonClicked` and loi name actions separate 👍 but maybe these can be refactored into a sealed class so they can be aggregated in a single lambda eg. `onLoiNameAction`, similar to how the button actions are handled with `onButtonClicked`

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskScreen.kt
```diff
@@ -45,7 +47,11 @@ fun TaskScreen(
   showInstructionsDialog: Boolean = false,
   loiName: String = "",
   onFooterPositionUpdated: (Float) -> Unit,
-  onAction: (TaskScreenAction) -> Unit,
+  onButtonClicked: (ButtonAction) -> Unit = {},
```
andreia-ferreira, line 51: suggestion: not related to this PR exactly, but it might be a good moment to rename `ButtonAction` to better clarify that it refers to actions that happen during data collection

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/AbstractTaskFragment.kt
```diff
@@ -219,52 +85,20 @@ abstract class AbstractTaskFragment<T : AbstractTaskViewModel> : AbstractFragmen
...
-        data = instructionData,
-        onDismissed = {
-          viewModel.dismissInstructions()
-          onInstructionDialogDismissed()
-        },
-      )
-    }
+  /** Handles changes to the LOI name draft. */
+  protected fun onLoiNameChanged(name: String) {
+    dataCollectionViewModel.setLoiNameDraft(name)
```
andreia-ferreira: on the previous version, there was a safeguard `if (loiName != "")`, maybe we can add a similar one here? Or a `if (loiName.isNotBlank())` would be even safer to prevent empty names

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/TaskScreen.kt
shobhitagarwal1612: I'd suggest doing this rename in a separate PR as it is adding a lot of noise to the changes.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskScreen.kt
```diff
@@ -73,18 +72,25 @@ fun DrawAreaTaskScreen(
...
+/**
+ * The stateless content of the draw area task screen.
+ *
+ * @param taskLabel The label of the task.
+ * @param taskActionButtonsStates The states of the action buttons.
+ * @param showInstructionsDialog Whether to show the instructions' dialog.
+ * @param showSelfIntersectionDialog Whether to show the self-intersection warning dialog.
+ * @param onFooterPositionUpdated Callback when the footer position changes.
+ * @param onDismissSelfIntersectionDialog Callback when the self-intersection dialog is dismissed.
+ * @param mapContent Composable for rendering the map.
```
andreia-ferreira, line 97: can you also add the `onButtonClicked`; `onLoiNameAction` and `onInstructionsDismiss`

andreia-ferreira (approved):

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskScreen.kt
shobhitagarwal1612: Done

Issue #3709, opened 2026-04-29
[KMP] Migrate ViewPager2 in DataCollectionFragment with a Compose HorizontalPager


shobhitagarwal1612:
We can now eliminate `AbstractTaskFragment` and its 10 subclasses by replacing the `ViewPager2` with a Compose-based `HorizontalPager`. This is feasible because the current fragments are thin wrappers that simply delegate logic to ViewModels and Compose screens.

This will be done in following phases:

1. Consolidation into a single Generic Fragment 
  - Eliminate all 11 fragment classes (AbstractTaskFragment + 10 subclasses) and replace them with a single DataCollectionTaskFragment.
  - Update the DataCollectionViewPagerAdapter to use only this fragment.
  
2. Compose Host Migration
  - Migrate the host DataCollectionFragment to Jetpack Compose.
  - Convert the XML layout (data_collection_frag.xml) to a Compose screen. 
  - Wrap the existing ViewPager2 inside an AndroidView within this new Compose host.

3. Pure Compose Paging (KMP compatible)
  - Remove the ViewPager2 and the FragmentStateAdapter. 
  - Switch to HorizontalPager in the shared UI layer. 
  - The map components will now be hosted directly in slots via AndroidView (on Android).

PR #3710, opened 2026-04-29
Consolidation task fragments and minor refactor

shobhitagarwal1612:
Towards #3709

This PR completes Phase 1 of the Task UI consolidation, significantly simplifying the data collection flow and preparing the codebase for KMP expansion. This is now possible as we have already migrated all ui logic from the task fragments to compose screens, also proving that our previous refactor (`fragment -> compose`) is stable.

### Key Changes

**1. Fragment Consolidation**

- Introduced `DataCollectionTaskFragment`, a single generic fragment that hosts the Compose screens for all task types.
- Deleted 11 task specific fragment classes.

**2. Reactive Shared Event Bus Architecture**

- ViewModels now emit abstract `DataCollectionEvent` signals instead of returning values or calling parent methods directly.
- `DataCollectionViewModel` now hosts a shared event bus that reactively captures and executes UI and navigation intents.
- Task ViewModels are now completely unaware of the underlying flow implementation, using a simple functional eventReporter callback.

**3. Architectural Refinements**

- Moved state-modifying logic (like SKIP action handling) from the orchestrator into the respective task ViewModels, keeping it closer to the data it affects.
- KMP Readiness: The new "Effect Bridge" pattern is platform-agnostic, allowing the shared ViewModel to drive UI flows on both Android and iOS without native fragment dependencies.

Demo:

[Screen_recording_20260429_102104.webm](https://github.com/user-attachments/assets/970f2a2b-a865-46fc-a9f8-c3a09a4c6d5c)

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/multiplechoice/MultipleChoiceTaskScreen.kt
```diff
@@ -38,6 +38,8 @@ import org.groundplatform.domain.model.task.Option
 import org.groundplatform.ui.theme.AppTheme
 import org.groundplatform.ui.theme.sizes
 
+const val MULTIPLE_CHOICE_LIST_TEST_TAG = "multiple choice items test tag"
```
andreia-ferreira, line 43: this could be annotated with `@VisibleForTesting`

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/DataCollectionEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+/** Represents high-level UI events triggered by task ViewModels during data collection. */
+sealed interface DataCollectionEvent {
+  /** Navigate to the previous task. */
+  data object NavigatePrevious : DataCollectionEvent
+
+  /** Navigate to the next task (includes validation and submission check). */
+  data object NavigateNext : DataCollectionEvent
+
+  /** Show the LOI Name Dialog before proceeding. */
+  data object ConfirmLoiName : DataCollectionEvent
```
andreia-ferreira: suggestion: maybe `ShowLoiDialog` or `SetupLoiName` would be clearer to show it's a trigger for a UI dialog and not for a direct confirmation (confirming only happens after filling in the name)

andreia-ferreira (approved):

PR #3711, opened 2026-04-29
Implement QR code scanner to join surveys

### LINKED ISSUE #3639: Survey QR code reader right in the App
Reading the QR code to join a survey currently requires a different app (usually camera app, but on a lot of older devices, people still have to install a separate app, which is often some bloaty trash app).

Occasionally, it also creates problems, because Ground is then started within that other app.

Ideally, we would have a "Join a survey" button in the survey overview, from where the survey QR code can be read right from Ground itself. 

Could look something like this:

Image

andreia-ferreira:
Fixes #3639 

This PR introduces a QR code reader in the survey selector screen so the user can join unlisted surveys more easily: 

https://github.com/user-attachments/assets/babb4b76-3a55-4997-b12a-00a754786348

ML Kit was used even though technically we could re-use the ZXing library for reading QR codes. However, that would require a custom implementation, while ML Kit ships its own scanner UI as a Play Services module, so we don't have to build/maintain a camera Activity or handle camera permissions ourselves at runtime

The error handling in the survey selector was polished: errors from survey listing, activation, and QR scanning now route through a single `ShowError` with a `ErrorType` with the different variants, mapped consistently in the fragment to the right popup string

@shobhitagarwal1612  PTAL?


shobhitagarwal1612:
Should this show a confirmation dialog with survey details before activating the survey?

andreia-ferreira:
@shobhitagarwal1612 great point! :100: Discussed that with @jo-spek and implemented a confirmation dialog to be shown before opening the scanned survey. Updated the code and the recording of the whole flow, ready for another review

core/domain/src/commonMain/kotlin/org/groundplatform/domain/util/SurveyQrCodeParser.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param deepLinkHost The host of the deep link URL (e.g. "groundplatform.org")
+ * @param deepLinkPath The path of the deep link URL (e.g. "/android/survey/")
+ */
+class SurveyQrCodeParser(private val deepLinkHost: String, private val deepLinkPath: String) {
+  operator fun invoke(payload: String): String? {
+    val regex =
+      Regex(
+        """^https?://${Regex.escape(deepLinkHost)}${Regex.escape(deepLinkPath)}([A-Za-z0-9_-]+)/?$"""
+      )
```
shobhitagarwal1612: The regex is quite strict. If a scanned url contains any query params, the match will fail because of the $ anchor. Can we confirm that this behavior is ok?

shobhitagarwal1612 (approved):

core/domain/src/commonMain/kotlin/org/groundplatform/domain/util/SurveyQrCodeParser.kt
andreia-ferreira: good catch! I replaced the regex with Uri-based parsing that matches the deeplink intent filter, also unifying with the existing deeplink flow

PR #3717, opened 2026-05-01
Migrate data collection fragment to compose and improve UDF

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/3709

This PR contains a series of refactorings and improvements to the Data Collection flow, moving it toward a more pure Compose architecture and better Unidirectional Data Flow (UDF) practices.

Key Changes:
 * Converted the xml layout to compose except for the view pager which will be taken up separately.
 * Replaced remaining instances of `mutableState` with `StateFlow`
 * Replaced legacy window insets with compose APIs.
 * Removed obsolete `TaskUpdated` state.
 

[Screen_recording_20260501_164001.webm](https://github.com/user-attachments/assets/fc6b272a-740f-4506-87f0-d7506440407f)

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+      title = R.string.data_collection_cancellation_title,
+      description = R.string.data_collection_cancellation_description,
+      confirmButtonText = R.string.data_collection_cancellation_confirm_button,
+      onConfirmClicked = { viewModel.confirmExit() },
+      onDismiss = { viewModel.dismissExitWarning() },
+    )
+  }
+}
+
+object DataCollectionScreenTestTags {
```
andreia-ferreira, line 87: this could be restricted to `@VisibleForTesting`

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+private fun LoadingContent(modifier: Modifier = Modifier) {
+  CircularProgressIndicator(
+    modifier = modifier.testTag(DataCollectionScreenTestTags.LOADING_INDICATOR)
+  )
+}
+
+@Composable
+private fun ErrorContent(code: DataCollectionErrorCode, modifier: Modifier = Modifier) {
+  Text(
+    text = "Error: $code",
```
andreia-ferreira: suggestion: maybe we can display an existing generic error message to the user in this case. The error code is already being logged in Timber so we can check that if needed

app/src/main/java/org/groundplatform/android/ui/datacollection/DataSubmissionConfirmationScreen.kt
```diff
@@ -57,14 +57,16 @@ import org.groundplatform.domain.model.locationofinterest.LoiReport
...
+  loiReport: LoiReport? = null,
+  onDismissed: () -> Unit,
+) {
   val baseModifier =
-    Modifier.fillMaxSize()
+    modifier
+      .fillMaxSize()
       .background(MaterialTheme.colorScheme.surface)
-      .padding(top = DEFAULT_TOOLBAR_HEIGHT)
       .systemBarsPadding()
```
andreia-ferreira: this should be removed, since the parent screen is now migrated to Compose and uses `Scaffold`, the insets are already handled. This would duplicate the bottom insets which makes the layout in smaller devices look like this: 

Image

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionScreen.kt
```diff
@@ -0,0 +1,240 @@
...
+
+  LinearProgressIndicator(
+    progress = { progress },
+    modifier =
+      modifier
+        .fillMaxWidth()
+        .offset { IntOffset(0, progressPositionY.toInt()) }
+        .testTag(DataCollectionScreenTestTags.PROGRESS_BAR),
+  )
+}
```
andreia-ferreira: now that this screen is in compose, would it be possible to move the progress bar directly into `TaskFooter`? This would simplify things quite a bit since we wouldn't have to rely on manually positioning it, we can make it part of that composable and pass on the `TaskPosition` to fill it correctly. What do you think?

shobhitagarwal1612: That's a great idea! Done.

app/src/test/java/org/groundplatform/android/ui/datacollection/components/TaskFooterTest.kt
```diff
@@ -69,4 +71,35 @@ class TaskFooterTest {
...
+
+    composeTestRule.setContent {
+      TaskFooter(
+        buttonActionStates = states,
+        onButtonClicked = {},
+        taskPosition = taskPosition,
+      )
+    }
+
+    composeTestRule.onNodeWithTag("progress_bar").assertIsDisplayed()
```
andreia-ferreira: suggestion: instead of hardcoding the tag, it could be exposed in the composable through a variable visible for testing

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionScreen.kt
```diff
@@ -0,0 +1,186 @@
...
+private fun DataCollectionUiState.getTitle(): String =
+  when (this) {
+    is DataCollectionUiState.TaskSubmitted -> stringResource(R.string.data_collection_complete)
+    is DataCollectionUiState.Ready -> job.name ?: ""
+    else -> ""
+  }
+
+@Composable
+private fun DataCollectionUiState.getSubtitle(): String? =
+  if (this is DataCollectionUiState.Ready) loiName else null
```
andreia-ferreira, line 190: worth adding a check if `loiName` is not empty too before returning it. Otherwise, before the LOI is defined, `getSubtitle()` returns an empty string instead of null, which causes the empty Text to still take vertical space and slightly misalign the toolbar title:

Image

andreia-ferreira (commented):
just a couple of small things and it's good to go! Great progress 🚀 🚀

andreia-ferreira (approved):

PR #3735, opened 2026-05-13
Replace ViewPager2 with Compose-based TaskScreenContainer for Data Collection

### LINKED ISSUE #3709: [KMP] Migrate ViewPager2 in DataCollectionFragment with a Compose HorizontalPager

shobhitagarwal1612:
Fixes https://github.com/google/ground-android/issues/3709

This PR overhauls the data collection flow by migrating it to a Compose-based architecture. It removes the legacy `ViewPager2` and fragment-based approach (`DataCollectionTaskFragment`, `DataCollectionViewPagerAdapter`) in favor of a new `TaskScreenContainer`. It also refactors event handling to be more deeply integrated with ViewModels.

**Key Changes:**
- [x] Replace `ViewPager2` and task fragments with Compose-based `TaskScreenContainer`.
- [x] Replace custom `TaskMapFragmentContainer` with the official Compose `AndroidFragment`.
- [x] Move map fragment instantiation into individual Compose task screens.
- [x] Simplify event handling in `DataCollectionViewModel` by removing `MutableSharedFlow` and using UI effects.
- [x] Remove `onButtonClicked` parameter from task screens and handle clicks via individual ViewModels.
- [x] Refactor `OpenSettings` and `SetAwaitingPhotoCapture` to use `DataCollectionViewModel` UI effects.
- [x] Fix failing tests and clean up unused imports/parameters.

**Steps to verify:**
1. Open the app and start a new data collection submission.
2. Verify that you can navigate smoothly between tasks (Next/Previous).
3. Check map-based tasks (Drop Pin, Draw Area, Capture Location) to ensure the map renders correctly and interactions work as expected.
4. Test permission dialogs and photo capture to ensure the new UI effects handle them properly.

@andreia-ferreira PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/location/CaptureLocationTaskScreen.kt
```diff
@@ -82,9 +83,7 @@ fun CaptureLocationTaskScreen(
  * @param showPermissionDeniedDialog Whether to show the permission denied dialog.
  * @param onDismissAccuracyCard Callback when the accuracy card is dismissed.
  * @param onAllowLocationClicked Callback when the allow location button is clicked in the dialog.
- * @param onOpenSettings Callback to open app settings.
  * @param onButtonClicked Callback when a button is clicked.
- * @param mapContent Composable for rendering the map.
```
andreia-ferreira, line 87: nit: this parameter is still in the arguments so it should be part of the kdoc

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskScreen.kt
```diff
@@ -41,17 +44,20 @@ import org.groundplatform.ui.theme.AppTheme
  * routing.
  *
  * @param viewModel The view model for this task.
- * @param mapContent Composable for rendering the map.
```
andreia-ferreira, line 44: nit: same comment as above. The params here in general don't match the ones in the composable

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskScreen.kt
```diff
@@ -95,7 +102,6 @@ fun DrawAreaTaskScreen(
  * @param onLoiNameAction Callback when user interacts with the LOI name dialog.
  * @param onInstructionsDismiss Callback when user dismisses the instructions dialog.
  * @param onDismissSelfIntersectionDialog Callback when the self-intersection dialog is dismissed.
- * @param mapContent Composable for rendering the map.
```
andreia-ferreira, line 98: nit: same comment as above

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionViewModel.kt
```diff
@@ -332,7 +319,22 @@ internal constructor(
...
+              when (event) {
+                DataCollectionEvent.NavigatePrevious -> onPreviousClicked(taskId)
+                DataCollectionEvent.NavigateNext -> onNextClicked(taskId)
+                DataCollectionEvent.ShowLoiDialog -> openLoiNameDialog()
+                DataCollectionEvent.OpenSettings ->
+                  viewModelScope.launch { _uiEffects.send(DataCollectionUiEffect.OpenSettings) }
+                is DataCollectionEvent.SetAwaitingPhotoCapture ->
+                  viewModelScope.launch {
+                    _uiEffects.send(DataCollectionUiEffect.SetAwaitingPhotoCapture(event.awaiting))
+                  }
```
andreia-ferreira: suggestion: the logic for each of these events could be extracted to a separate method like the events above

andreia-ferreira (commented):
The only issue I found was when restoring a draft where the current task was conditional. Clicking on any of the previous/next buttons makes the app crash. Tested this behavior on master to confirm it's not an existing issue and couldn't reproduce

https://github.com/user-attachments/assets/53fa927f-848a-496e-bfa7-64b4cd67b560

tested on API 24 and 36 and happens on both

shobhitagarwal1612:
@andreia-ferreira Thanks for catching the bug! I think it was happening because of using `relativeIndex` instead of `absoluteIndex` for rendering tasks. But I was unable to reproduce it locally. Could you please try again with your local setup and check if this resolves the issue?

andreia-ferreira:
thanks for looking into it, I tested it again but it seems to still happen unfortunately. You can try to join [this survey link](https://ground-dev-sig.web.app/android/survey/YyCAsc39zt6Lzgis6wF1) which I used in the recording below, to see if you can reproduce it too


[Screen_recording_20260514_105749.webm](https://github.com/user-attachments/assets/07ba628b-67a1-48d4-aa8e-6670ea727591)

shobhitagarwal1612:
There is an issue with creating new conditional surveys with the web client. Could you please make the above linked survey public or add me as a collaborator?

andreia-ferreira:
@shobhitagarwal1612 added you to the survey's data collectors. The 'Conditionals test' survey should show up on your device now

shobhitagarwal1612:
Upon deeper investigation, I found that the issue is happening due to the complete migration from fragment to compose based views. Earlier during draft resume flow, all task fragments were auto-loaded by the view pager. This used to instantiate their view model and add their values to `taskDataHandler`. 

With the new architecture, only the current active task and it's view model are loaded. Therefore, unvisited previous task responses are missing from the collected `ValueDelta`, causing the crash.

The proposed solution is to pre-populate `taskDataHandler` with all draft values immediately after data collection state becomes `Ready`.

andreia-ferreira (approved):
tested again and it works smoothly, thanks for fixing the issue with the drafts! This was quite a big refactor and the data collection flow looks simpler now, great work! 🎉

PR #3750, opened 2026-05-20
[Code health] Don't allow saving a duplicate vertex during polygon drawing session

shobhitagarwal1612:
Towards #3679

Separate committed and tentative vertices in `PolygonDrawingSession`. This allows the persistence layer to safely store only the committed vertices without having to assume that the last vertex is user committed vertex or not.

Changes verified using unit tests and testing the flow on a physical device.

@andreia-ferreira  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSession.kt
```diff
@@ -28,6 +28,12 @@ interface PolygonDrawingSession {
   /** The list of committed vertices in the current drawing session. */
   val vertices: List<Coordinates>
 
+  /** The tentative vertex that tracks the uncommitted active position (e.g. camera target). */
+  val tentativeVertex: Coordinates?
```
andreia-ferreira: this is only used in the implementation, does it need to be exposed in the interface?

app/src/test/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskViewModelTest.kt
```diff
@@ -199,11 +199,11 @@ class DrawAreaTaskViewModelTest : BaseHiltTest() {
 
     viewModel.removeLastVertex()
 
-    assertThat(featureTestObserver.value()?.tooltipText).isEqualTo("1,565,109 m")
+    assertThat(featureTestObserver.value()?.tooltipText).isNull()
```
andreia-ferreira, line 202: this changes the behavior of the test, since the tooltip text is not being checked anymore, is this intended?

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/PolygonDrawingSessionImpl.kt
```diff
@@ -34,56 +35,60 @@ class PolygonDrawingSessionImpl : PolygonDrawingSession {
...
+    get() = _tentativeVertex?.let { _vertices + it } ?: _vertices
 
-    updatedVertices.add(vertex)
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal val redoVertexStack = mutableListOf<Coordinates>()
 
-    _vertices = updatedVertices.toImmutableList()
-  }
+  override val hasSelfIntersection: Boolean
+    get() = isSelfIntersecting(displayVertices)
```
andreia-ferreira, line 47: I can't seem to close the polygons. When debugging i noticed that `isSelfIntersecting` now returns true when getting the state of the complete button. It could be because now this method checks `displayVertices` which includes the `tentativeVertex`, and that is equal to the first vertex

https://github.com/user-attachments/assets/ebcc2906-b89e-42d5-93c0-41fbd9f51889

shobhitagarwal1612: I'm unable to reproduce this on my device. Are you able to consistently reproduce it or is it flaky?

andreia-ferreira: it's consistently reproducible. But I've found out it's because I was clicking twice to drop the first point because clicking the first time doesn't make the white circle marker to show up until the camera is dragged. So I believe the problem is that dropping two points in the starting vertex is allowed, but then leads to an invalid polygon

shobhitagarwal1612: Apologies for the delay. Fixed now

andreia-ferreira (commented):
Tested and works well 👍 Just a UI/UX question, the first vertex doesn't display a circle on it anymore, only after moving the map, is it intended? 

https://github.com/user-attachments/assets/f253a6a8-52a6-4889-9218-c24f805af1cd

shobhitagarwal1612:
Let me take another pass.

Issue #3715, opened 2026-04-30
Instant offline PDF report


jo-spek:
As discussed in issue https://github.com/google/ground-android/issues/3373, an immediate result to share or store on the phone has been requested by multiple actors in the field.

The QR code holding the GeoJSON at the end of a survey (https://github.com/google/ground-android/issues/3575) was highly appreciated by participants in workshops in Cameroon in April 2026 and even makes the app more accessible to farmers themselves (no need to go through survey organizers and Ground platform anymore for a direct result with a tangible value for a farmer.)

The possibility to locally download the PDF report which holds all the survey questions and survey answers in a tabular overview, the GeoJSON QR code, and the images would go a long way in making Ground more accessible and useful to the target audience.

jo-spek:
This design suggestion from Victor accidentally landed in the Ground platform issues:

https://github.com/google/ground-platform/issues/2242#issuecomment-3902800948

<img width="309" height="662" alt="Image" src="https://github.com/user-attachments/assets/aaabbaaf-bd3b-46b6-9a00-a077445c6595" />

kenstershiro:
- Edge case to handle: need a submissions list to handle which submission the PDF should be for (if there is more than one)
- @gino-m  to follow up with  @vittorino for this edge case
- suggest separte pdfs per submission

kenstershiro:
Additional request to add an image of the polygon will be in a follow up ticket

kenstershiro:
- Use case of multiple submissions associated to an LOI: agreed to display the latest submission only for the first release
- Currently the PDF is only generated locally (for offline use), in future the plan is to generate it server-side from the web platform. At this point we could also include displaying a list of all submissions for the LOI
- Request from Cameroon to put the QR code with the location on a separate page (ideally the first page), ideally on a different page size (A5) if feasible - can be done as a follow up

PR #3758, opened 2026-05-28
Introduce PdfExportService and platform interfaces

andreia-ferreira:
Towards #3715

Introduces the scaffolding for PDF export of submission data. This is platform-independent and has the `PdfExportService` as entry point. The provider/renderer/launcher interfaces will be implemented on the platform level on the next PR. This new code is part of a new KMP module `:feature:pdf`

Additionally, as a small cleanup, removed the jvm from the `core:ui` gradle config because otherwise all expect/actual logic would have to be also implemented on jvmMain. Also added androidLibrary target to `core:testing` so that the fake testing classes are possible to be used by modules that contain that target

This introduces no functionality changes.

@shobhitagarwal1612 PTAL?


feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/PdfExportService.kt
```diff
@@ -0,0 +1,71 @@
...
+  suspend fun export(request: Request, action: Action) {
+    val outputPath = mutex.withLock {
+      withContext(coroutineDispatcher) {
+        outputProvider.pruneOldFiles()
+        val path = outputProvider.newFilePath(request.fileName)
+        if (!outputProvider.exists(request.fileName)) {
+          val images = imageProvider.load(request.qrContent, request.document.photoFilenames())
+          try {
+            renderer.render(request.document, images, path)
+          } finally {
```
shobhitagarwal1612, line 53: If `renderer.render()` throws an exception (e.g. out of memory, disk full, etc.), a partially written or corrupted 0-byte file might be left on the disk. To prevent opening/sharing broken PDFs, please consider catching exceptions during rendering and delete the corrupted file. Then re-throw the exception so it can be handled upstream

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/mapper/LoiReportMapper.kt
```diff
@@ -0,0 +1,108 @@
...
+            loiName = loiReport.loiName,
+            rows = rows,
+          ),
+      )
+    val timestamp =
+      "${dateFormatter.formatDate(details.dateMillis)}_${dateFormatter.formatTime(details.dateMillis)}"
+    val fileName =
+      listOf(details.surveyName, loiReport.loiName, details.userName, timestamp)
+        .map { it.filter(::isSafeFileChar) }
+        .filter { it.isNotBlank() }
```
shobhitagarwal1612, line 63: Will this work for other locales (e.g. japanese or arabic characters)?

shobhitagarwal1612 (changes requested):

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/PdfExportService.kt
andreia-ferreira: thanks for catching this. Added the exception handling :+1:

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/mapper/LoiReportMapper.kt
andreia-ferreira: good point, this would not work. I updated it so now characters from all languages are preserved

shobhitagarwal1612 (approved):

PR #3780, opened 2026-06-10
Add jacoco plugin to android multiplatform libs

andreia-ferreira:
Only `:app` and `:core:domain` were reporting test coverage at the moment. At the moment, the jacoco setup supported only kmp modules with a jvm target. This PR adds the code to the jacoco gradle script, so that android multiplatform libs correctly report their coverage and applies the plugin to `:core:ui` and `:feature:pdf`.

@shobhitagarwal1612 PTAL?


shobhitagarwal1612 (approved):
💯

PR #3774, opened 2026-06-09
Add Android rendering logic for PDF reports

andreia-ferreira:
Towards #3715

This PR adds the specific Android implementations for rendering PDF files. 

Main changes:
- New helpers: 
    - `PdfCursor` to keep track of the current position within a page and its reserved spaces
    - `PdfPageController` to keep track of page counts and their lifecycles
- Android implementations of the previously introduced interfaces as well as platform specific helpers `PdfWriter`, `PdfCanvas`, `PdfTextPaints` and `DocumentPdfCanvas` in order to draw the layout

There are no functionality changes yet. The full flow will be implemented in a final follow up PR. For reference, the current layout will generate the following document structure:
Screenshot From 2026-06-10 14-25-22

@shobhitagarwal1612 PTAL?

PR #3782, opened 2026-06-10
Add common layout components for PDF

andreia-ferreira:
Towards https://github.com/google/ground-android/issues/3715

(Split from https://github.com/google/ground-android/pull/3774)
This PR focuses on adding the building blocks for the PDF document: `PageHeaderLayout` (to display the survey name, job and date+time at the top), `PageFooterLayout` (to display the data collector information and the page number at the bottom), `QrBlockLayout` (to display the QR code embedding the geometry), `TableLayout` (with all the submission data) and their unit tests.

It also adds the following helpers:
- `PdfConfig` for dimensions and margins
- `PdfGeometry` for utilities and classes which handle calculations and measurements

@shobhitagarwal1612  PTAL?

### DIFF
--- a/feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
+++ b/feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
@@ -44,7 +44,7 @@ internal object PdfConfig {
   const val LINE_SPACING = 4f
 
   /** Target size of the QR code block. */
-  const val QR_SIZE = 200
+  const val QR_SIZE = 200f
 
   /** Horizontal gap between the two columns of the page header. */
   const val HEADER_COLUMN_GAP = 16
--- a/feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
+++ b/feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
@@ -43,60 +43,9 @@ internal object PdfConfig {
   /** Vertical spacing added between lines of text. */
   const val LINE_SPACING = 4f
 
-  /** Target size of the QR code block. */
-  const val QR_SIZE = 200f
-
-  /** Horizontal gap between the two columns of the page header. */
-  const val HEADER_COLUMN_GAP = 16
-
-  /** Inner padding between a table cell's border and its text. */
-  const val CELL_PADDING = 6
-
-  /** Stroke width used for table and cell borders. */
-  const val BORDER_WIDTH = 0.5f
-
-  /** Vertical gap below the header before body content begins. */
-  const val HEADER_BOTTOM_GAP = 28f
-
-  /** Vertical gap above the footer separating it from body content. */
-  const val FOOTER_TOP_GAP = 28f
-
-  /** Maximum number of lines rendered for each header value. */
-  const val MAX_HEADER_VALUE_LINES = 1
-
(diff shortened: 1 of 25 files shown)


shobhitagarwal1612 (commented):
Regarding the package structure, right now `PageHeaderLayout`, `TableLayout`, etc live in `components` package. Since `components` usually implies actual UI rendering code (like `@Composable` functions), and these classes are purely math models, would it make more sense to move these classes one level up (or maybe `org.groundplatform.feature.pdf.render.layout` or `model`)?

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/layout/TableLayout.kt
```diff
@@ -0,0 +1,113 @@
...
+    leftTextHeight: Float,
+    rightTextHeight: Float,
+    rightImageSize: PdfItemSize?,
+  ): Float {
+    val imageHeight = rightImageSize?.height ?: 0f
+    val imageSpacing = if (imageHeight > 0f && rightTextHeight > 0f) LINE_SPACING else 0f
+    return maxOf(leftTextHeight, rightTextHeight + imageHeight + imageSpacing) + 2 * CELL_PADDING
+  }
+
+  fun getRow(
```
shobhitagarwal1612, line 107: Since this includes both the top and bottom borders for every row, would it cause rendering of a line twice? Would it be safer to only draw the bottom border per row or does the rendering would take care of deduplication?

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.groundplatform.feature.pdf.render
+
+/**
+ * Dimensional and type-scale constants shared by the Android and iOS PDF renderers. Keeping these
+ * in commonMain prevents the two platforms from drifting on page size, margins, or type scale.
+ *
+ * All measurements are in PDF points (1/72 inch).
+ */
+internal object PdfConfig {
```
shobhitagarwal1612, line 24: Please consider adding ktdoc for these constants. Also, there is a mix of Int and Float in here. Would converting them to Float be easier down the line?

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/layout/TableLayout.kt
andreia-ferreira: good catch, this logic would cause rendering of inner lines twice. We do need the top border to be drawn on the first row of each page though, but I refactored this layout to make it optional so that it will only be drawn in those cases

feature/pdf/src/commonMain/kotlin/org/groundplatform/feature/pdf/render/PdfConfig.kt
andreia-ferreira: added ktdoc to these constants, and moved component-specific ones into their layout files. This file only holds the more generic, shared dimensions now.

Regarding converting to Float, I checked and converted QR_SIZE to Float which removed some `.toFloat()` operations, but other variables are not worth converting since they are mostly used as Int

andreia-ferreira:
@shobhitagarwal1612 thanks for the review :pray: I have applied the suggestions, including renaming the layout components package, it is now `org.groundplatform.feature.pdf.render.layout`. Ready for another look!

shobhitagarwal1612 (approved):
🚢

PR #3845, opened 2026-07-17
Fix missing placeholder in string

andreia-ferreira:
The translation for the `area_message` in Khmer was missing the placeholder. This PR adds it and a lint rule to prevent this cases in the future.

@shobhitagarwal1612  PTAL?


shobhitagarwal1612 (approved):
Nice!

PR #3853, opened 2026-07-22
Fix navigation crash

### LINKED ISSUE #3846: [Trending crash] IllegalArgumentException: action_home_screen_fragment_to_dataCollectionFragment cannot be found from the current destination
[Crashlytics url](https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/7cd23302820dbab2f698b23b4a74b8c8?time=90d&types=crash&sessionEventKey=6A5DC96600DB000160F8791EBE788EDE_2242957220210455315)

Logs show crashes happening since at least 0.3.8 with a bigger spike in 0.4.0

andreia-ferreira:
Fixes #3846

The crash could be reproduced with the following steps:
1 - Enable 'Don't keep activities' on Developer Settings
2 - Turn off network connection
3 - Start data collection and complete a few tasks
4 - Put the app on the background 
5 - Go back to Ground 
6 - Crash

After the app was killed in the background, `MainViewModel` was recreated and immediately received the signed-in user again. When in offline mode, the MainVM would be recreated and get the user right away since there was no connection for `updateRemoteUserInfo`. Offline this happened almost instantly, because `updateRemoteUserInfo` returns early when there is no connection instead, navigation then proceeded to the Home screen, popping the restored back stack. So when the data collection screen tried to fetch its VM from the back stack it crashed.

Now the MainActivity will only proceed with this rerouting if the user is still on startup or signIn screen so a restored screen is left alone.
There is also a small refactor changing `MainUiState` to `MainUiEffect` since it mostly works with one time events rather than a screen state.

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/main/MainActivity.kt
```diff
@@ -105,23 +118,20 @@ class MainActivity : AbstractActivity() {
     )
   }
 
-  private fun updateUi(uiState: MainUiState) {
-    when (uiState) {
-      MainUiState.OnUserSignedOut -> {
-        navigateTo(SignInFragmentDirections.showSignInScreen())
-      }
-      MainUiState.TosNotAccepted -> {
+  private fun updateUi(destination: MainUiEffect.StartDestination) {
```
shobhitagarwal1612, line 121: Currently this function calls `navigateTo()` inside every branch of `when`. We can lift it out of the when to make the code more concise.

```
private fun updateUi(destination: MainUiEffect.StartDestination) {
  val direction = when (destination) {
    MainUiEffect.StartDestination.TermsOfService -> SignInFragmentDirections.showTermsOfService(false)
    MainUiEffect.StartDestination.SurveySelector -> SurveySelectorFragmentDirections.showSurveySelectorScreen(true)
    MainUiEffect.StartDestination.Home -> HomeScreenFragmentDirections.showHomeScreen()
    is MainUiEffect.StartDestination.ActiveSurvey -> 
      SurveySelectorFragmentDirections.showSurveySelectorScreen(false).apply {
        surveyId = destination.surveyId
     }
   }
  navigateTo(direction)
}

```

app/src/main/java/org/groundplatform/android/ui/main/MainViewModel.kt
```diff
@@ -78,45 +76,47 @@ constructor(
...
-        MainUiState.TosNotAccepted
-      } else if (isDeepLinkAvailable()) {
-        val surveyId = _deepLinkUri.value?.let { surveyDeepLinkParser.parse(it) }
-        if (surveyId != null) {
-          MainUiState.ActiveSurveyById(surveyId)
+  private suspend fun onUserSignedIn(user: User) {
+    val destination =
+      try {
+        userRepository.saveUserDetails(user)
+        if (!isTosAccepted()) {
```
shobhitagarwal1612: Optional: We can replace `if / else if / else` structure with `when` block 

```
when {
  !isTosAccepted() -> MainUiEffect.StartDestination.TermsOfService
  isDeepLinkAvailable() -> {
    // ...
  }
  else -> {
    // ...
  }
}
```

app/src/main/java/org/groundplatform/android/ui/main/MainViewModel.kt
```diff
@@ -78,45 +76,47 @@ constructor(
...
+          MainUiEffect.StartDestination.TermsOfService
+        } else if (isDeepLinkAvailable()) {
+          val surveyId = _deepLinkUri.value?.let { surveyDeepLinkParser.parse(it) }
+          if (surveyId != null) MainUiEffect.StartDestination.ActiveSurvey(surveyId)
+          else MainUiEffect.StartDestination.SurveySelector
```
shobhitagarwal1612: We can replace nested if-else with ?: operator.

```
val surveyId = _deepLinkUri.value?.let { surveyDeepLinkParser.parse(it) }
surveyId?.let { MainUiEffect.StartDestination.ActiveSurvey(it) } 
  ?: MainUiEffect.StartDestination.SurveySelector
```

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/main/MainActivity.kt
andreia-ferreira: good call, done!

shobhitagarwal1612 (approved):

PR #3858, opened 2026-07-25
fix: show add LOI button when zoomed out

### LINKED ISSUE #3553: [Home screen] "+" fab only shown when zoomed in
The "+" is not shown when zoomed out.

This is likely a relic of the old card behavior when cards were only shown when zoomed in. 

@andreia-ferreira @hassan-nsubuga

arpitagarwal1301:
Fixes #3553

- Keep the Add LOI FAB available regardless of map zoom level.
- Preserve zoom-dependent LOI viewport filtering and map clustering behavior.
- Add focused coverage for the below-clustering-threshold case.

## Verification

- `./gradlew checkCode`
- `./gradlew :app:ktfmtCheck :app:testLocalDebugUnitTest --tests 'org.groundplatform.android.ui.home.mapcontainer.HomeScreenMapContainerViewModelTest'`
- `git diff --check`

## Limitations

- The behavior was not visually exercised on a device; the affected ViewModel state is covered by Robolectric.

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
@@ -130,10 +130,7 @@ internal constructor(
   /** [Feature] clicked by the user. */
   val featureClicked: MutableStateFlow<Feature?> = MutableStateFlow(null)
 
-  /**
-   * List of [Job]s which allow LOIs to be added during field collection, populated only when zoomed
-   * in far enough.
-   */
+  /** List of [Job]s which allow LOIs to be added during field collection. */
   private val adHocLoiJobs: Flow<List<Job>>
 
   private val showJobSelectionModal = MutableStateFlow(false)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/groundplatform/android/ui/home/mapcontainer/HomeScreenMapContainerViewModel.kt
```diff
@@ -179,8 +176,8 @@ internal constructor(
         .stateIn(viewModelScope, SharingStarted.Lazily, listOf())
 
     adHocLoiJobs =
-      activeSurvey.combine(isZoomedInFlow) { survey, isZoomedIn ->
-        if (survey == null || !isZoomedIn) listOf()
+      activeSurvey.map { survey ->
+        if (survey == null) listOf()
         else survey.jobs.filter { it.canDataCollectorsAddLois && it.getAddLoiTask() != null }
```
andreia-ferreira, line 181: optional: can be slightly simplified with

```suggestion
        survey?.jobs?.filter { it.canDataCollectorsAddLois && it.getAddLoiTask() != null } ?: listOf()
```

andreia-ferreira (approved):

PR #3865, opened 2026-07-30
Show existing LOIs on map while creating a new LOI

cuongnt:
## Summary

When a user draws a new polygon or drops a new pin, already-saved LOIs are now rendered as background features on the map for spatial context.

### `DrawAreaTaskViewModel`
- Inject `LocationOfInterestRepositoryInterface` and `SurveyRepositoryInterface`
- Expose `existingLoiFeatures: StateFlow<Set<Feature>>` derived from `activeSurveyFlow → getValidLois()`

### `DrawAreaTaskMapFragment`
- `renderFeatures()` now combines `draftArea` with `existingLoiFeatures`

### `DropPinTaskViewModel`
- Same injection pattern and `existingLoiFeatures` flow

### `DropPinTaskMapFragment`
- `renderFeatures()` now combines pin features with `existingLoiFeatures`

## Behaviour

Existing LOIs are rendered with `Feature.Type.LOCATION_OF_INTEREST`, `clusterable = false`, `selected = false` — visually consistent with the home map but non-interactive during data collection.

## Testing

Existing Hilt test infrastructure injects real repositories; no test file changes required.


cuongnt:
[ground-android-master.pdf](https://github.com/user-attachments/files/30539384/ground-android-master.pdf)

cuongnt:
[ground-android-master.pdf](https://github.com/user-attachments/files/30539394/ground-android-master.pdf)
Unfortunately, I didn't record a screen video. I've summarized the calibration process to display the drawn polygons on the new batch drawing module in the PDF file above.

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskViewModel.kt
```diff
@@ -32,22 +37,48 @@ import org.groundplatform.android.ui.map.Feature
...
+  private fun LocationOfInterest.toExistingFeature(): Feature =
+    Feature(
+      id = id,
+      type = Feature.Type.LOCATION_OF_INTEREST,
+      geometry = geometry,
+      style = Feature.Style(job.getDefaultColor()),
+      clusterable = false,
+      selected = false,
+    )
+
```
andreia-ferreira, line 81: instead of duplicating this both in `DropPinTaskViewModel` and `DrawAreaTaskViewModel`, this can be moved to `BaseMapViewModel`, which both fragments can later access with `getMapViewModel()`

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -66,8 +67,8 @@ class DrawAreaTaskMapFragment @Inject constructor() :
   }
 
   override fun renderFeatures(): Flow<Set<Feature>> =
-    taskViewModel.draftArea.map { feature: Feature? ->
-      if (feature == null) setOf() else setOf(feature)
+    combine(taskViewModel.draftArea, taskViewModel.existingLoiFeatures) { draft, existing ->
+      (if (draft == null) emptySet() else setOf(draft)) + existing
```
andreia-ferreira: can be simplified with
```suggestion
      setOfNotNull(draft) + existing
```

andreia-ferreira (changes requested):
thanks for taking this up! Left some comments to improve the implementation, also could you please add some unit tests for the new logic? 🙏 

the pipeline seems to be failing because the CLA hasn't been signed, you can do that here: https://cla.developers.google.com/

app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/point/DropPinTaskViewModel.kt
cuongnt: This is also a good option.

andreia-ferreira:
this feature has become a high priority, and we're looking to include it in an upcoming release. Since the CLA is still pending, I'm going to close this PR and move forward with: https://github.com/google/ground-android/pull/3871

Thanks again for your contribution and for taking the time to work on this! 🙏

Issue #3866, opened 2026-08-03
Large surveys are unusable in the app


andreia-ferreira:
It was reported by Hai Pham that, when a survey is too large (aprox 2000-5000 LOIs), the app gets stuck in the loading screen and the survey never opens

<img width="360" height="800" alt="Image" src="https://github.com/user-attachments/assets/aef82d52-a5a6-4707-8dbc-d83d2360b929" />

andreia-ferreira:
Updated the title to make this issue cover the 2 main problems:
- The initial survey download 
- Performance of the map with large sets of data

PR #3870, opened 2026-08-04
Improve performance of the map for large surveys

andreia-ferreira:
Towards #3866

The performance of large surveys while in the home screen map was bad and could lead to OOM crashes. This was because every LOI outline was handed to the google maps sdk as soon as the survey loaded, just marked as invisible.

This PR changes that so clusterable items are added to the cluster manager but not to the map. Their outlines are created only when the clusterer decides to draw them individually, and released again once they're absorbed back into a cluster.
Additionally, the algorithm of the cluster manager was set as `NonHierarchicalViewBasedAlgorithm` so that only on-screen features are drawn. This is important especially past the `CLUSTERING_ZOOM_THRESHOLD`, where all polygons would uncluster and render regardless if they were visible or not

Before:

https://github.com/user-attachments/assets/128d2c4c-e895-44b7-97d8-3490bee5099e

After:

[Screen_recording_20260805_190515.webm](https://github.com/user-attachments/assets/e7fd8b6b-13b2-4ef4-a5b6-2571660df3d3)

@shobhitagarwal1612 PTAL?


shobhitagarwal1612 (approved):
🚀🚀🚀

PR #3871, opened 2026-08-05
Show existing LOIs in draw area and drop pin tasks

### LINKED ISSUE #3482: [Map] Show sites already added during geometry-making task
Extension of https://github.com/google/ground-android/issues/3068

Currently, if allowed for the survey, data collected by other surveyors can be seen on the "starting screen" (?) on the map. However, while creating a new geometry, it disappears. 

Would be good to **still see neighboring geometries during creation of new geometries**, because the functionality is partially intended to prevent overlapping boundaries. See video for better understanding.

https://github.com/user-attachments/assets/e373b77f-67db-4c04-a6d7-20cc2321655f

andreia-ferreira:
Fixes #3482 

[Screen_recording_20260805_193830.webm](https://github.com/user-attachments/assets/1a3d680b-d4f2-4dbf-befb-279ca751ff36)

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/tasks/polygon/DrawAreaTaskMapFragment.kt
```diff
@@ -66,8 +67,10 @@ class DrawAreaTaskMapFragment @Inject constructor() :
   }
 
   override fun renderFeatures(): Flow<Set<Feature>> =
-    taskViewModel.draftArea.map { feature: Feature? ->
-      if (feature == null) setOf() else setOf(feature)
+    combine(getMapViewModel().existingLoiFeatures, taskViewModel.draftArea) {
```
shobhitagarwal1612, line 70: Should we give some toggle button to the users instead of forcefully enabling it for everyone?
Was this not considered due to some reason?

andreia-ferreira: except for own submissions, polygon visibility is already controlled by the survey's data visibility setting (whether data is visible to all participants or not). An in-app toggle wasn't added since the immediate goal was to prevent data collectors from drawing overlapping geometries because they can't see what's already on the map. But if needed we can always add a toggle later :+1:

shobhitagarwal1612 (approved):

Issue #3867, opened 2026-08-04
App doing too many unecessary reads from Firebase


andreia-ferreira:
- [ ] `SyncSurveyUseCase` syncs all LOIs of the entire survey. Even if a user makes just 1 submission , a sync event for the whole survey data will be triggered for all users that have that survey https://github.com/google/ground-android/pull/3918
- [x] `SurveySyncWorker` retries surveys indefinitely https://github.com/google/ground-android/pull/3922
- [x]  There are no `unsubscribeFromTopic` calls anywhere, which means that even if a user eventually deletes a survey from the device, they will silently still get its updates  #3876 ; #3888
- [x] User LOIs are downloaded twice per sync when `dataVisibility == ALL_SURVEY_PARTICIPANTS` #3889

PR #3876, opened 2026-08-10
Unsubscribe from survey updates when removing them from the device

andreia-ferreira:
Towards #3867 

In order to reduce the amount of unnecessary reads from Firebase, this PR adds the logic to unsubscribe from survey updates when a user deletes surveys from the device. Currently, even if a user deletes a survey, if this survey has any update from other user, the phone will still receive and react to the FCM and the survey will be silently re-downloaded.

The fix includes:
- Refactoring `RemoveOfflineSurveyUseCase` moving it to `core:domain`. Methods were added to the following interfaces:`SurveyRepositoryInterface`, `RemoteDataStore` and `MapStateRepositoryInterface` in order to allow decoupling the domain logic from the data 
- `SurveySyncWorker` now skips syncing surveys that are no longer available offline

@shobhitagarwal1612 PTAL?


app/src/main/java/org/groundplatform/android/data/remote/RemoteDataStore.kt
```diff
@@ -73,6 +73,8 @@ interface RemoteDataStore {
    */
   suspend fun subscribeToSurveyUpdates(surveyId: String)
 
+  suspend fun unsubscribeFromSurveyUpdates(surveyId: String)
```
shobhitagarwal1612, line 80: nit: Please add ktdoc

shobhitagarwal1612 (approved):

PR #3884, opened 2026-08-14
Add cache for map icons

andreia-ferreira:
Towards #3866

Every icon was rebuilt from scratch (cluster icons and regular marker icons) on panning and zooming the map. On large surveys this could cause the map performance to drop significantly and also lead to OOM crashes due to memory pressure.

This PR introduces a cache for both clusters and marker icons.

The impact of these changes can be seen in the heap dump comparison of before/after.
Before:
Screenshot From 2026-08-14 10-55-40

After:
image

@shobhitagarwal1612 PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/ui/IconFactory.kt
+++ b/app/src/main/java/org/groundplatform/android/ui/IconFactory.kt
@@ -39,17 +39,21 @@ import org.groundplatform.android.ui.util.obtainTextPaintFromStyle
 class IconFactory @Inject constructor(@ApplicationContext private val context: Context) {
 
   private val markerIcons =
-    // One icon per job colour at each of two scales, so a handful per survey.
+    // Bitmap cache keyed by job color and scale. Capacity of 16 covers most combinations and avoids
+    // recreating icons.
     object : LruCache<Pair<Int, Float>, BitmapDescriptor>(16) {
       override fun create(key: Pair<Int, Float>): BitmapDescriptor =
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/groundplatform/android/ui/IconFactory.kt
```diff
@@ -36,6 +37,24 @@ import org.groundplatform.android.ui.util.obtainTextPaintFromStyle
...
+  private val markerIcons =
+    // Bitmap cache keyed by job color and scale. Capacity of 16 covers most combinations and avoids
+    // recreating icons.
+    object : LruCache<Pair<Int, Float>, BitmapDescriptor>(16) {
+      override fun create(key: Pair<Int, Float>): BitmapDescriptor =
+        BitmapDescriptorFactory.fromBitmap(getMarkerBitmap(key.first, key.second))
+    }
+
+  private val clusterIcons =
+    object : LruCache<String, BitmapDescriptor>(1) {
```
shobhitagarwal1612: nit: Can we increase the default to 16 as well?

shobhitagarwal1612 (approved):
Nice improvements! 👏 👏 👏 

Left a minor comment for your consideration. Please feel free to ignore if you disagree with the suggestion.

app/src/main/java/org/groundplatform/android/ui/IconFactory.kt
andreia-ferreira: done

PR #3885, opened 2026-08-14
Fix crash on SubmitDataUseCase

### LINKED ISSUE #3882: [Trending crash] IllegalStateException: AddLoi task response missing
https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/fd8668b0538b169795d0e7718ccc998f?time=7d&types=crash&sessionEventKey=6A7D4531010900017A72BABCA4719A6F_2251824018727133004

```
Fatal Exception: java.lang.IllegalStateException: AddLoi task response missing
       at org.groundplatform.domain.usecases.submission.SubmitDataUseCase.addLocationOfInterest(SubmitDataUseCase.kt:68)
       at org.groundplatform.domain.usecases.submission.SubmitDataUseCase.invoke(SubmitDataUseCase.kt:50)
       at org.groundplatform.android.ui.datacollection.DataCollectionViewModel.saveChanges(DataCollectionViewModel.kt:395)
       at org.groundplatform.android.ui.datacollection.DataCollectionViewModel.access$saveChanges(DataCollectionViewModel.kt:79)
       at org.groundplatform.android.ui.datacollection.DataCollectionViewModel$onNextClicked$1$1$1.invokeSuspend(DataCollectionViewModel.kt:295)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
       at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
       at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:798)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

andreia-ferreira:
Fixes #3882

The crash could be reproduced by doing the following:
1. Start collecting data and past a draw area/drop pin task
2. Put the app on background
3. Simulate destroying the background process by executing either `adb shell am kill org.groundplatform.android` or having developer options 'no background processes' 
4. Go back to the app and complete data collection
5. The app would crash when reaching the end of data collection

This was because, when the OS kills the app in the background, the data collection screen is recreated at the task the user was in, but with none of the data collected recovered, which caused the crash mentioning the AddLoi task was missing, and previous answers were also silently missing. The only draft recovery path was when the activity would be recreated and the home screen carried the data in nav arguments.

This PR does the following changes:
- `DataCollectionInitializer` now restores draft data so that the screen recovers correctly regardless of how it was opened
- `TaskSequenceHandler` makes it impossible for a task sequence to resume past a mandatory task that the user hasn't answered to avoid cases where data is silently missing 

Before:

[Screen_recording_20260814_160426.webm](https://github.com/user-attachments/assets/25282432-5a42-44ca-99fb-0e0037098f0c)

After:

[Screen_recording_20260814_160539.webm](https://github.com/user-attachments/assets/1f37d404-7ab9-400b-a052-6f788dbe4d3b)

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionInitializer.kt
```diff
@@ -122,14 +137,32 @@ constructor(
...
+  /**
+   * Returns the draft of an interrupted data collection session for the given [survey], [jobId] and
+   * [loiId], or `null` when there is none, or the stored one belongs to a different session.
+   */
+  private suspend fun findDraft(survey: Survey, jobId: String, loiId: String?): DraftSubmission? {
+    val draftId = submissionRepository.getDraftSubmissionsId()
+    val draft =
+      if (draftId.isEmpty()) null else submissionRepository.getDraftSubmission(draftId, survey)
+    return draft?.takeIf { it.surveyId == survey.id && it.jobId == jobId && it.loiId == loiId }
+  }
```
shobhitagarwal1612: Should we move this to SubmissionRepository? Or move this logic within `getDraftSubmission`

shobhitagarwal1612 (changes requested):

app/src/main/java/org/groundplatform/android/ui/datacollection/DataCollectionInitializer.kt
andreia-ferreira: yes, good point. I have added a new method to the repo `getDraftSubmissionForSession` to fetch a draft for a specific survey, job and LOI. Kept the `getDraftSubmission` to just fetch a draft in general, since the home screen needs this to know if it should open the data collection on startup

shobhitagarwal1612 (approved):

PR #3902, opened 2026-08-26
Move all remaining usecases to :core:domain

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/3633

Migrates all survey use cases from `:app` to `:core:domain` (KMP) and abstracts network connectivity behind an interface.

Key changes:
- **Domain Use Cases**: Moved `ActivateSurveyUseCase`, `MakeSurveyAvailableOfflineUseCase`, `ReactivateLastSurveyUseCase`, and `ListAvailableSurveysUseCase` to `:core:domain`.
- **Network Abstraction**: Introduced `NetworkManagerInterface` in `:core:domain` and `FakeNetworkManager` in `:core:testing`, decoupling use cases and repositories from Android connectivity APIs.
- **Repository Decoupling**: Exposed `lastActiveSurveyId` and `isSurveyActive()` on `SurveyRepositoryInterface` to remove use case dependencies on Android's `LocalValueStore`.
- **Tests**: Replaced slow Robolectric use case tests with fast KMP unit tests in `:core:domain:commonTest`, and modernized ViewModel tests to use `FakeNetworkManager`.

@andreia-ferreira PTAL?


app/src/main/java/org/groundplatform/android/system/NetworkManager.kt
```diff
@@ -27,17 +27,16 @@ import javax.inject.Singleton
...
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
+import org.groundplatform.domain.system.NetworkManagerInterface
 import timber.log.Timber
 
-enum class NetworkStatus {
-  AVAILABLE,
-  UNAVAILABLE,
-}
+typealias NetworkStatus = org.groundplatform.domain.system.NetworkStatus
```
andreia-ferreira: nit: this can just be imported instead of creating a typealias

core/testing/src/commonMain/kotlin/org/groundplatform/testing/FakeSurveyRepository.kt
```diff
@@ -31,9 +30,23 @@ class FakeSurveyRepository : SurveyRepositoryInterface {
   override val activeSurvey: Survey?
     get() = _activeSurveyFlow.value
 
-  var offlineSurveys: List<Survey> = emptyList()
+  override var lastActiveSurveyId: String = ""
```
andreia-ferreira, line 33: the value for this is set by `activateSurvey` but never cleared. `clearActiveSurvey` should do that

andreia-ferreira (approved):

PR #3903, opened 2026-08-26
Move MapConfig and related constants to :core:domain

shobhitagarwal1612:
Towards https://github.com/google/ground-android/issues/3633

Migrates base map configuration and common domain constants from `:app` to `:core:domain` as part of the ongoing KMP migration:
- Moved `MapConfig` from `org.groundplatform.android.ui.common` to `org.groundplatform.domain.model.map.MapConfig`.
- Added property-level KtDocs and default values to `MapConfig`.
- Relocated domain/map constants (`DEFAULT_LOI_ZOOM_LEVEL`, `CLUSTERING_ZOOM_THRESHOLD`, `TEXT_DATA_CHAR_LIMIT`, `DEFAULT_MAP_TYPE`, `ACCURACY_THRESHOLD_IN_M`) from `app/common/Constants.kt` to `org.groundplatform.domain.util.Constants`.
- Updated all imports and call sites across fragments, view models, and unit tests in `:app`.
- Added `MapConfigTest` in `:core:domain:commonTest` covering default values, custom overrides, and copy mutations.

@andreia-ferreira PTAL?


core/domain/src/commonMain/kotlin/org/groundplatform/domain/model/map/MapConfig.kt
```diff
@@ -13,13 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.groundplatform.android.ui.common
+package org.groundplatform.domain.model.map
```
andreia-ferreira: wondering if this class would be better placed in `:core:ui` instead? It's only used between fragments (eventually between compose screens), so the map behavior it defines feels like a view layer concern. What do you think?

shobhitagarwal1612: Good point! MapConfig is purely configuring UI interactions and rendering overrides without any business logic or persistence in the domain layer. Moving it to `:core:ui` makes sense and keeps `:core:domain` clean.

andreia-ferreira (approved):

PR #3918, opened 2026-09-02
Implement incremental LOI sync

### LINKED ISSUE #3867: App doing too many unecessary reads from Firebase

andreia-ferreira:
<!-- NOTE: The comments can be left as is as they don't end up in the final preview. -->

<!-- Add one or more issues below if already present. Otherwise, create one. -->
Fixes #3867 

FCMs now carry `deleted` and `loiId`, which we apply straight as a local delete rather than discovering it through a full read. That's only a shortcut, so before each incremental sync we compare the server's LOI count against the local one and fall back to a full read if needed.

A full read still happens in the following cases:
- the survey has never been synced                                                                                                                                                                                                                
- the server holds fewer LOIs than we expect, which means a delete happened and a FCM was lost or collapsed                                                                                                                                                                           
- the survey's data visibility changed, since that changes which LOIs we should hold                                                                                                                     
- the last full read was over 7 days ago, as a catch-all

The count operation used to detect delete cases bills 1 read per 1K LOIs so it is lightweight, but soft deletion on the server side should be a future improvement for this implementation.

@shobhitagarwal1612 PTAL?


shobhitagarwal1612 (changes requested):
Can we break this PR into 2 parts? First introduces the necessary room tables, daos, etc. Second one with the rest of the changes?

andreia-ferreira:
done! This PR is ready for review now that most of the Room changes have been merged

shobhitagarwal1612 (approved):

PR #3922, opened 2026-09-08
Fix looping survey syncs

andreia-ferreira:
Towards #3867

At the moment survey syncs can retry indefinitely. This can cause issues, for example in older app versions that don't have optimizations to load large surveys, the sync can timeout after 10 min and then retry later. And keep on retrying.

This PR attempts to fix that with the following:
-  `SurveySyncService` now uses `KEEP` instead of `APPEND`, with a short initial delay, so a burst of FCM collapses into one run
- Added `MAX_SYNC_ATTEMPTS` along with a timeout, and stopped cancelling work manager failures the same way, since it has its own retry mechanism

@shobhitagarwal1612  PTAL?


app/src/main/java/org/groundplatform/android/data/sync/SurveySyncWorker.kt
```diff
@@ -54,19 +58,42 @@ constructor(
...
       }
       success()
+    } catch (t: TimeoutCancellationException) {
+      // Sync timed out, count as a failed attempt.
+      retryOrGiveUp(t)
+    } catch (t: CancellationException) {
+      // WorkManager stopped the execution, it will handle rescheduling itself.
+      throw t
     } catch (t: Throwable) {
-      Timber.e(t, "Failed to sync survey $surveyId, retrying")
```
shobhitagarwal1612, line 69: Do we see these logs on firebase which confirms the infinite retry loop bug?

andreia-ferreira: yes, it's one of the most frequent non-fatals at the moment : https://console.firebase.google.com/u/0/project/of-ground/crashlytics/app/android:org.openforis.ground/issues/d677c91f7b5a867cdcbbcd5c7d26f844?time=7d&types=error&sessionEventKey=6A9FEC6D00DF00012E8BF2FDA96FC2AA_2261565131165010012

shobhitagarwal1612 (approved):
⏲️ ✅

PR #3923, opened 2026-09-08
Migrate Sync Status Screen to Compose

shobhitagarwal1612:

Migrates `SyncStatusFragment` from the legacy XML layout to pure Jetpack Compose.
- Replace `sync_status_frag.xml` with `SyncStatusScreen` composable
- Convert `SyncStatusViewModel` from `LiveData` to `StateFlow<SyncStatusState>`
- Convert `SyncStatusFragment` to a thin `ComposeView` wrapper
- Remove unused `sync_status_frag.xml` and `ic_arrow_back.xml`
- Add unit and Compose tests for ViewModel, Screen, and Fragment

| Before | After |
|------|------|
| Screenshot_20260908_170219 |  Screenshot_20260908_171237 |

@andreia-ferreira PTAL?


app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncStatusScreen.kt
```diff
@@ -0,0 +1,133 @@
...
+    modifier = modifier.fillMaxSize(),
+    topBar = {
+      Toolbar(
+        stringRes = R.string.data_sync_status,
+        showNavigationIcon = true,
+        titleCentered = true,
+        iconClick = onNavigateUp,
+      )
+    },
+    containerColor = MaterialTheme.colorScheme.surfaceContainer,
```
andreia-ferreira: nit: this gives the screen a bit of a purple tint, maybe we should use `MaterialTheme.colorScheme.background` instead to keep the white background?

andreia-ferreira (approved):

app/src/main/java/org/groundplatform/android/ui/syncstatus/SyncStatusScreen.kt
shobhitagarwal1612: Done.

image

PR #3936, opened 2026-09-11
Improve force update to also suspend firebase requests

andreia-ferreira:
When a force update is enabled, the outdated apps still continue requests to firebase. This PR makes the requests also stop before the app is updated. All pending mutations stay queued locally and should upload after the update. 

This could be useful in case a specific version ships with a bug that may flood firebase with requests or bad data (since work manager can keep making requests in the background regardless if the app is in use or not), so that all traffic can be shut off right away instead of waiting for all users to update. 

In order to do this, the logic that fetched the remote config and verified if an update was needed was refactored to a repository and a usecase so it could be better reused.

@shobhitagarwal1612  PTAL?

### DIFF
--- a/app/src/main/java/org/groundplatform/android/data/remote/DataStoreException.kt
+++ b/app/src/main/java/org/groundplatform/android/data/remote/DataStoreException.kt
@@ -41,3 +41,5 @@ open class DataStoreException(message: String?) : RuntimeException(message) {
     }
   }
 }
+
+class UpdateRequiredException : DataStoreException("App update required")
--- a/app/src/main/java/org/groundplatform/android/di/UseCaseModule.kt
+++ b/app/src/main/java/org/groundplatform/android/di/UseCaseModule.kt
@@ -19,6 +19,8 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import org.groundplatform.android.BuildConfig
+import org.groundplatform.domain.repository.AppConfigRepositoryInterface
 import org.groundplatform.domain.repository.LocationOfInterestRepositoryInterface
 import org.groundplatform.domain.repository.MapStateRepositoryInterface
 import org.groundplatform.domain.repository.OfflineAreaRepositoryInterface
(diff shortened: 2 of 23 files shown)


app/src/main/java/org/groundplatform/android/data/remote/DataStoreException.kt
```diff
@@ -41,3 +41,5 @@ open class DataStoreException(message: String?) : RuntimeException(message) {
     }
   }
 }
+
+class UpdateRequiredException : DataStoreException("App update required")
```
shobhitagarwal1612, line 45: Can we rename to `AppUpdateRequiredException`?

app/src/main/java/org/groundplatform/android/di/UseCaseModule.kt
```diff
@@ -125,4 +127,11 @@ object UseCaseModule {
     surveyRepository: SurveyRepositoryInterface,
     userRepository: UserRepositoryInterface,
   ) = ListAvailableSurveysUseCase(networkManager, surveyRepository, userRepository)
+
+  @Provides
+  fun providesShouldForceUpdateUseCase(appConfigRepositoryInterface: AppConfigRepositoryInterface) =
+    org.groundplatform.domain.usecases.ShouldForceUpdateUseCase(
```
shobhitagarwal1612, line 133: nit: Can we import the use case instead of using fully qualified imports?

app/src/main/java/org/groundplatform/android/ui/main/MainViewModel.kt
```diff
@@ -114,6 +114,9 @@ constructor(
             }
           }
         }
+      } catch (_: UpdateRequiredException) {
+        // Popup prompting the user to update is displayed, so we don't need to do anything here.
```
shobhitagarwal1612, line 118: Instead of silently consuming the exception, can we instead create a new UiEvent which the ui layer can then listen to for showing the dialog?

app/src/main/java/org/groundplatform/android/ui/main/MainActivity.kt
```diff
@@ -161,7 +161,7 @@ class MainActivity : AbstractActivity() {
 
   override fun onResume() {
     super.onResume()
-    if (viewModel.isAppUpdateAvailable()) {
+    if (viewModel.isAppUpdateRequired()) {
       showForceUpdateDialog()
```
shobhitagarwal1612, line 165: Please see the other comment.

shobhitagarwal1612 (changes requested):

PR #3941, opened 2026-09-15
Harden JaCoCo configuration and limit coverage excludes to generated code

shobhitagarwal1612:
Two independent changes to the JaCoCo setup, one per commit.
- [x] **Fail loudly on missing coverage data.** `JacocoReport` *skips* rather than fails
      when execution data is missing, publishing an empty report on a green build. Adds a
      `verifyJacocoExecutionData` task, separate because a check inside the report task
      would never run once it's skipped.
- [x] **Resolve `.exec` from either known path.** AGP redirects coverage to
      `outputs/unit_test_code_coverage` when `jacoco` is applied before it, otherwise the
      Gradle plugin's `build/jacoco` default wins. A `fileTree` matches whichever exists.
- [x] **Rename `jacocoLocalDebugUnitTestReport` → `jacocoTestReport`.** It's registered
      for KMP modules too, where it runs `jvmTest`. CI updated.
- [x] **Move the JaCoCo version into the version catalog.**
- [x] **Limit excludes to generated code.** Six patterns matched only hand-written Kotlin
      (`migration/*`, the `firebase/*` ones, `LocationSharedFlowCallback*`) and were
      hiding real gaps — dropped. `**/*Module*` matched 159 classes on substring alone and
      would hide any hand-written `...Module...`; replaced with `**/di/**`. Completed the
      generated side, which was leaking Room's auto-migrations.

Excludes were decided by measuring a report built with excludes disabled, and by mapping
all 424 generated sources under `app/build/generated` against the report — not by name.
### Codecov will show a drop — expected
`:app` **78.46% → 77.44%**. No test or production code changed; 215 previously hidden
lines are now measured (+55 covered).
| module | before | after |
|---|---|---|
| app | 78.46% | 77.44% |
| core/domain, core/ui, feature/pdf | unchanged | unchanged |
| **total** | **75.22%** | **74.44%** |

Newly visible: `FirestoreDataStore` (0/82),  `FirebaseStorageManager` (7/17), `firebase/schema/*Reference` (50/110).
### Verification
```
./gradlew jacocoTestReport checkCode
./gradlew testLocalDebugUnitTest jvmTest testAndroidHostTest
```

@gino-m @andreia-ferreira  PTAL?


shobhitagarwal1612:
@gino-m PTAL?

PR #3946, opened 2026-09-16
Reschedule media uploads when the Wi-Fi-only preference changes

### LINKED ISSUE #3945: Media upload pending despite switching off "WiFi-upload only"
If in settings "Upload photos over Wi-Fi only" is checked and data have been collected, unchecking it does not trigger the upload over mobile data.

This is a big problem at the moment for a geolocation campaign in Nigeria, where several data collectors had "Upload photos over Wi-Fi only" checked. After unchecking it, the photos keep pending in *Data sync status* and are never uploaded. (Once connecting to a WiFi, the photos are uploaded immediately. The Nigerian campaign is trying a workaround, connecting to hotspots, but this needs urgent fixing, therefore P0.)

Tried on older version, this seems to be an old undetected bug.

@andreia-ferreira @rfontanarosa @gino-m @shobhitagarwal1612 @kenstershiro 

Andreia is on leave. Are you @shobhitagarwal1612 or somebody else available to fix this?

https://github.com/user-attachments/assets/398d541c-7964-4432-b76b-b4453a104fd3

shobhitagarwal1612:
Fixes #3945

WorkManager constraints are immutable once a request is enqueued, so unchecking "Upload photos over Wi-Fi only" never applied to media uploads which had already been scheduled. Appending a new request left it blocked behind work still awaiting an unmetered connection, so pending photos were only uploaded once the device rejoined Wi-Fi.

Replace pending media upload work when the preference changes so that the newly selected network type takes effect, and cancel it when local data is cleared, so that work left behind by a signed out user can't block uploads enqueued by the next one.

Also adds missing test coverage for the worker.

[Screen_recording_20260916_222109.webm](https://github.com/user-attachments/assets/4d783577-0889-4185-9f14-456b38521b3a)

@jo-spek @gino-m  PTAL?


jo-spek:
Awesome! Thanks for the hyper-quick fix :)

gino-m (approved):

jo-spek (approved):
All good from my end

